PR #7, opened 2019-02-27
Feature/add storybook

azangru:
Included in this fairly large PR are:

- [x] Configured Storybook
- [x] npm script for running a Storybook instance
- [x] npm script for building Storybook and deploying the build to [Github Pages](https://ensembl.github.io/ensembl-client)
- [x] Documentation (ADR for Storybook and updated README)
- [x] Several design primitives (a list of colors and typography) added to Storybook
- [x] A number of components for Species Selector as well as shared components (not at their final stage; the work on components will continue; but they can serve as example of using Storybook for developing components)


imran-ebi:
Doesn't seem to break anything. Only suggestion would be to change `SFC` in `assets.d.ts` to `FunctionComponent` as it has been deprecated.

PR #9, opened 2019-03-06
Add google analytics tracking

msmanoj:
Added google analytics tracking using the react library ReactGA([react-ga](https://github.com/react-ga/react-ga)).

- A middleware has been created on the root level to track all the actions that are dispatched
- Every action dispatched needs to include the google analytics meta data (meta.ga) parameter. A warning will be displayed on the console if not. 

**Note**:
Tracking is enforced only to the actions dispatched to the root store at the moment.  
We will have to manually use ReactGA within the components to track the events that doesn't affect the state using the following code:
`ReactGA.event({
  category: 'category',
  action: 'action'
});`

Please visit [react-ga](https://github.com/react-ga/react-ga) for more information regarding what all can be tracked.


src/ensembl/package.json
```diff
@@ -44,6 +44,7 @@
     "react": "16.8.1",
     "react-cookie": "^3.0.8",
     "react-dom": "16.8.1",
+    "react-ga": "^2.5.7",
```
azangru: could we install exact versions of packages (not with tildes)?
`npm install something --save-exact` writes the exact version to package.json

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -5,62 +5,124 @@ import { TrackType } from './track-panel/trackPanelConfig';
...
+      });
   }
 );
 
 export const toggleDrawer = createAction('browser/toggle-drawer', (resolve) => {
-  return (drawerOpened?: boolean) => resolve(drawerOpened);
+  return (drawerOpened?: boolean) =>
+    resolve(drawerOpened, {
+      ga: {
+        category1: 'Drawer',
```
azangru: is `category1` intentional or is it a typo?

src/ensembl/package.json
msmanoj: I've now updated it to use an exact value.

src/ensembl/src/store.ts
```diff
@@ -4,6 +4,41 @@ import { composeWithDevTools } from 'redux-devtools-extension';
...
 
-export default createStore(reducers, composeEnhancers(applyMiddleware(thunk)));
+const googleAnalyticsMiddleWare = (store: any) => (next: any) => (
+  action: any
+) => {
+  try {
+    // If we have the google anlytics meta data passed in, and
+    // if disable_ga_tracking flag is not set to true,
+    // We need to track this event
+    if (!action.meta.ga.disable_tracking && action.meta.ga.category.length) {
```
azangru: Will action always have the `meta` field? Will there always be the `ga` object on the `meta` field?

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: Well spotted! It was intentional to test the warning message but I missed to revert it. I've reverted it now.

src/ensembl/src/store.ts
msmanoj: This is the way i'm trying to enforce the developer to include the meta information every time they implement a new action. If the meta or meta.ga is missing, a warning message will be displayed on the console.

azangru: How will it work with e.g. thunks? Wouldn't all of them cause errors?

msmanoj: It will cause error as long as the meta is not defined. I couldn't think of any workarounds other than passing the 'disable_tracking' flag.

Also, we will be able to track how long an AJAX request took to get the response, for which we will need to pass some meta along with the event dispatched from the thunk. This will be helpful to filter the requests that takes long time.

azangru: Could we perhaps disable this check then and rely on each other during code review to make sure we are adding analytics metainfo when required?

azangru: Could we perhaps remove the automatic check then and rely on each other during the code review process to point out that an analytics event is required (when it is)?

ens-st3: Why do you suggest removing the automatic check ?That's more robust than code review...

msmanoj: > Could we perhaps remove the automatic check then and rely on each other during the code review process to point out that an analytics event is required (when it is)?

I feel it will be a good way to remind the developers to include the meta info as we still in the initial stages of development. It will pretty much reduce the number of checks we need to during the code review process.

azangru: > Why do you suggest removing the automatic check ?

The check assumes that all our actions will be objects with a `meta` field on them. Thunks are actions that are not objects; they are functions; so I would expect the check to break. Moreover, if we are going to use a more sophisticated redux middleware for side effects (such as `redux-saga` or `redux-observable`), that will result in more actions that will not necessarily have the `meta` field on them, nor should they.

Showing console errors in all these cases will be distracting and once false-positive errors appear, the developer will never pay attention to them.

azangru: Redux actions is an internal messaging mechanism within the app. They do not correspond 1:1 to user's interactions with the UI.

msmanoj: >The check assumes that all our actions will be objects with a meta field on them. Thunks, on the other hand, are actions that are not objects; they are functions; so I would expect the check to break.

As far as I know, all the actions must be of type object which is explained in the links:
- https://redux.js.org/basics/actions#actions
- https://github.com/redux-utilities/flux-standard-action#actions

If am not wrong, the way thunk works is that it will initially return a function that eventually dispatches an action, which is again of type object. The function returned by thunk will be handled by thunk itself and not the middlewares. 

If the above statement is right, it should all be the same for thunk, saga and observable where we will only be handling the final action, which will be an object.

In the second link I have included above, it states that all the actions MUST include the 'type' property and MAY have the meta property, which is what I am trying to make use of to implement this tracking.

> Showing console errors in all these cases will be distracting, and once false-positive errors appear, the developer will no longer pay attention to them.

I agree that false-positive errors have other side effects as you have mentioned, but I've used console.warn which instead gets displayed as a warning message in amber. 

> Redux actions are an internal messaging mechanism within the app. They do not correspond 1:1 to user's interactions with the UI.

This is true, but again, this mechanism already has pretty much everything we needed for google analytics tracking, using which we can avoid building and running another similar mechanism in parallel.

azangru: > As far as I know, all the actions must be of type object which is explained in the links:

That’s the general rule, yes, but thunks break this contract — see their implementation here:
https://github.com/reduxjs/redux-thunk/blob/master/src/index.js#L1-L9

Notice also that (if I am reading the source code correctly) these functions will get propagated through the chain of middlewares (thunk middleware calls `next(action)` and passes this action function into `next`).

Thunks do not observe the shape of a Flux standard action. And I am sure we will be using them. It is also not a far stretch to imagine that other redux middleware we might add will not need the `meta` field on the action object.

Speaking of Flux standard actions, raising an error (even though its manifestation is a warning text in the console) when the `meta` field is absent is essentially breaking with the FSA standard, because, as you rightly quoted, the `meta` field is optional.

> this mechanism already has pretty much everything we needed for google analytics tracking

The mechanism is fine; let’s just not produce errors when the action lacks the `meta` field. This puts us in the position when we either have to ensure that all actions, even for internal workings of the app, have an analytics object attached to them (which may not even be desirable), or have to accept false-positives (and thus lose the usefulness of the notification). Let’s have the freedom of adding analytics objects only to the actions that are relevant for analytics.

msmanoj: > (thunk middleware calls next(action) and passes this action function into next).

I think the next(action) will be called only when the action is not of type function. If it is of type function, it dispatches another action which again gets handled by Thunk. When this Thunk gets resolved, the final dispatch of action will be an object, which gets passed on to our custom analytics middleware. This is where we will actually be looking for the `meta` and all the previous actions does not require it.

> Speaking of Flux standard actions, raising an error (even though its manifestation is a warning text in the console) when the meta field is absent is essentially breaking with the FSA standard, because, as you rightly quoted, the meta field is optional, and the error implies that it's obligatory.

It does break the FSA standard, but I was thinking about having a standard of our own in order to solve the requirement of tracking almost all the events. 


> The mechanism is fine; let’s just not produce errors when the action lacks the meta field. This puts us in the position when we either have to ensure that all actions, even for internal workings of the app, have an analytics object attached to them (which may not even be desirable), or have to accept false-positives (and thus lose the usefulness of the notification).

At the moment, this automatic enforcement is only applied to all the actions that affects the root state, which I think it would all be worthy enough to be tracked. The idea behind displaying the warning on the console is just to make sure that its addressed during the development phase.
I believe that the actions involved in internal workings would mostly not affect the global state like the accordion toggle for example, which we don't have to worry about adding the `meta` info.


> Let’s have the freedom of adding analytics objects only to the actions that are relevant for analytics.
 
If we are okay to make a list of all the actions that needs to be tracked and also rely on manually ensuring that the tracking meta data is correctly added to the new actions during the code review stage, I am happy to remove the warnings. I would suggest to keep it for a while and then we can always scrap it when we get to a point where it's gets very less helpful.

azangru: > I think the next(action) will be called only when the action is not of type function.

Oh, right :smile:

imran-ebi: > If we are okay to make a list of all the actions that needs to be tracked and also rely on manually ensuring that the tracking meta data is correctly added to the new actions during the code review stage, I am happy to remove the warnings.

How about using a prefix for all action types that need to be tracked? We can then do this check for those specific actions.

msmanoj: We can do that but it wouldn't give us a way enforce the tracking data when it's necessary. Let's say after a year, we will not be very much aware that certain actions needs to be tracked while developing and we could end up adding an entire feature without adding any tracking data to it.

src/ensembl/src/store.ts
```diff
@@ -1,9 +1,48 @@
 import { applyMiddleware, createStore } from 'redux';
 import thunk from 'redux-thunk';
 import { composeWithDevTools } from 'redux-devtools-extension';
-
+import { createBrowserHistory } from 'history';
+import { routerMiddleware } from 'connected-react-router';
 import reducers from './rootReducer';
 
+import GoogleAnalyticsTracking from './services/AnalyticsService';
```
azangru: We didn't discuss this, but I think the convention is that PascalCase, if it's used at all, is used only for naming of React component files. A service wouldn't be a component file, so I would advice the name `analytics-service`. @imran-ebi, wonder what you think.

imran-ebi: Yes, that would make it easier to differentiate the service files from the React components.

msmanoj: Cool, I've renamed it.

src/ensembl/src/store.ts
```diff
@@ -1,9 +1,48 @@
...
+  }
+
+  next(action);
+};
+
+export default function configureStore(preloadedState?: any) {
+  const store = createStore(
+    reducers(history), // root reducer with router state
+    preloadedState,
+    composeEnhancers(
```
azangru: I am a bit confused here: why do we need `composeEnhancers`?

imran-ebi: We need it for Redux dev tools

src/ensembl/src/Root.tsx
```diff
@@ -53,7 +53,7 @@ const Root: FunctionComponent<RootProps> = (props: RootProps) => {
   }, [cookies]);
 
   return (
-    <BrowserRouter>
+    <React.Fragment>
```
azangru: Two questions:
- why do we need Fragment here (there is only one child, not an array of children)
- should we write this as `React.Fragment` or using the shorthand `<> ... </>` notation? I am a bit partial to the shorthand notation (once I tried it, I never wanted to go back), but would like to know what you think. @imran-ebi ?

src/ensembl/src/store.ts
azangru: Ah!

src/ensembl/src/store.ts
```diff
@@ -1,9 +1,48 @@
...
-export default createStore(reducers, composeEnhancers(applyMiddleware(thunk)));
+const googleAnalyticsMiddleWare = (store: any) => (next: any) => (
+  action: any
+) => {
+  // If we have the google anlytics meta data passed in
+  // We need to track this event
+  if (action.meta && action.meta.ga && action.meta.ga.category) {
+    // The action and category fields are mandatory
+    GoogleAnalyticsTracking.trackEvent(action);
+  } else if (action.type.indexOf('LOCATION_CHANGE') != -1) {
```
azangru: Not as a string, please. Let's import the `LOCATION_CHANGE` constant from connected-react-router and use it here.

src/ensembl/src/Root.tsx
imran-ebi: That's probably because I was too lazy to remove it, after the `div` tag was added. We can definitely remove it. As for fragments, I prefer the `Fragment` without the `React`.

I don't have an opinion about the shorthand, since I've never used it. Won't it look hacky compared to using a name tag?

azangru: > Won't it look hacky compared to using a name tag?
I don't think it will — it's a standard React tag after all. I used Fragment without React for a long time too; then I tried the shorthand, and now I'm a strong believer in it.

msmanoj: I too would prefer <Fragment> as it will be a little more readable.

azangru: Not that I want to push either of you one way or the other; I just want to offer myself as an example of someone who, conservatively, thought that `Fragment` is more readable and that the empty angle brackets look a bit weird and unsatisfying — and then after the majority of our team decided on the shorthand, tried it, and found it very liberating. No more expressly declared imports from `react` or using the `React` namespace; it was like having to import something from `react` to write a `<div>` in JSX, and then being relieved from that effort.

It took about 10 minutes to get used to empty angle brackets :smile:

src/ensembl/src/store.ts
msmanoj: That's done.

src/ensembl/src/store.ts
```diff
@@ -19,21 +19,22 @@ const googleAnalyticsMiddleWare = (store: any) => (next: any) => (
...
     GoogleAnalyticsTracking.trackEvent(action);
-  } else if (action.type.indexOf('LOCATION_CHANGE') != -1) {
+  } else if (action.type === LOCATION_CHANGE) {
     // If the location history has been changed, track it as a pageview
     GoogleAnalyticsTracking.trackPageView(
       action.payload.location.pathname +
         action.payload.location.search +
         action.payload.location.hash
     );
+    return false;
```
azangru: Why the `return false`? Don’t we want to call `next` to let this action reach the reducer eventually?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -107,10 +107,13 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
-      .replace(':species', params.species)
-      .replace(':objSymbol', params.objSymbol)
-      .replace(':location', newChrLocationStr);
+    const newUrl =
+      path
+        .substring(0, path.length - 1)
+        .replace(':species', params.species)
+        .replace(':objSymbol', params.objSymbol) +
+      '?region=' +
+      newChrLocationStr;
```
azangru: can't we do newUrl = `/app/browser/${params.species}/${params.objSymbol}?region=${newChrLocationStr}`?

src/ensembl/src/store.ts
```diff
@@ -15,7 +15,10 @@ const googleAnalyticsMiddleWare = (store: any) => (next: any) => (
   if (action.meta && action.meta.ga && action.meta.ga.category) {
     // The action and category fields are mandatory
     GoogleAnalyticsTracking.trackEvent(action);
-  } else if (action.type === LOCATION_CHANGE) {
+  } else if (
+    action.type === LOCATION_CHANGE &&
+    action.payload.action !== 'REPLACE'
```
azangru: no, we shouldn't filter out replaces; replaces are fine.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -107,10 +107,13 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
 
     props.history.replace(newUrl);
```
azangru: with `ConnectedReactRouter` this should be done via the `replace` method imported from `connected-react-router`

src/ensembl/src/store.ts
msmanoj: That's removed now. I added it while investigating the double rendering issue.

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: That's done.

src/ensembl/src/store.ts
msmanoj: Cool, I've updated it to check if the path has changed.

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: That's done.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -105,16 +107,13 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
-        .replace(':objSymbol', params.objSymbol) +
-      '?region=' +
-      newChrLocationStr;
-
-    props.history.replace(newUrl);
+    const newUrl = `/app/browser/${params.species}/${
+      params.objSymbol
+    }?region=${newChrLocationStr}`;
+
+    replace(newUrl);
```
azangru: I suspect `replace` is an action that has to be `dispatch`-ed. See example with `push` in connecter-react-router repo:

https://github.com/supasate/connected-react-router/blob/66215b121ebee82f36baec74a552cd3149c6e33e/FAQ.md#with-storedispatch

src/ensembl/src/content/home/Home.tsx
```diff
@@ -22,12 +22,12 @@ const Home: FunctionComponent<HomeProps> = (props: HomeProps) => {
         <h2>Previously viewed</h2>
         <dl>
           <dd>
-            <Link to="/app/browser/human/BRCA2/13:32315474-32400266">
+            <Link to="/app/browser/human/BRCA2/?region=13:32315474-32400266">
```
azangru: _(here and below)_ Perhaps remove the slash after the gene name?

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: Calling the replace function from connected-react-route seems to be dispatching the action we need. Would that cause any issues?

src/ensembl/src/content/home/Home.tsx
msmanoj: That's done.

src/ensembl/src/content/app/browser/Browser.tsx
azangru: Ok, I now see that in the latest commit you put `replace` in `mapDispatchToProps`, and are using `props.replace`, which means that the replace action is now properly dispatched. Good.

src/ensembl/src/content/app/App.tsx
```diff
@@ -66,7 +66,8 @@ export const App: FunctionComponent<AppProps> = (props: AppProps) => {
           <Route path={`${url}/global-search`} component={GlobalSearch} />
           <Route path={`${url}/species-selector`} component={SpeciesSelector} />
           <Route
-            path={`${url}/browser/:species/:objSymbol/:location`}
+            // Using '*' here as query string is not supported in the Route
```
azangru: This comment is no longer needed, right?

msmanoj: Yes, it's not. I've removed it now.

PR #10, opened 2019-03-07
Feature/species search match

azangru:
- [x] Added lodash (because we are loading it as a sub-dependency anyway, and it's bound to make our life easier)
- [x] Added the SpeciesSearchMatch element whose responsibility it is to markup the matched substrings in the search results (see design mockup below):

image

- [x] Added tests for SpeciesSearchMatch

### DIFF
--- a/src/ensembl/src/content/app/species-selector/components/species-autosuggestion-panel/SpeciesAutosuggestionPanel.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/species-autosuggestion-panel/SpeciesAutosuggestionPanel.tsx
@@ -4,15 +4,24 @@ import SpeciesSearchMatch from '../species-search-match/SpeciesSearchMatch';
 
 import styles from './SpeciesAutosuggestionPanel.scss';
 
-import { SearchMatches } from 'src/content/app/species-selector/types/species-search';
+import { SearchMatch } from 'src/content/app/species-selector/types/species-search';
 
 type Props = {
-  matches: SearchMatches;
+  matches: SearchMatch[];
+  onMatchSelected: (match: SearchMatch) => void;
 };
 
 const SpeciesAutosuggestionPanel = (props: Props) => {
+  const onMatchSelected = (match: SearchMatch) => () => {
+    props.onMatchSelected(match);
+  };
+
   const matches = props.matches.map((match) => (
-    <SpeciesSearchMatch match={match} key={match.description} />
+    <SpeciesSearchMatch
+      match={match}
+      onClick={onMatchSelected(match)}
+      key={match.description}
+    />
   ));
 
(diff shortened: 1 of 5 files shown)


ens-ap5:
FWIW, it should be "Homo sapiens" not "Homo Sapiens" - only the genus is capitalised (as per the examples from bacteria).

azangru:
Well, the data will be coming from the api anyway; I hope it will be formatted appropriately :-)

In the mock data I used for developing this component, [sapiens is lower-cased](https://github.com/Ensembl/ensembl-client/blob/master/src/ensembl/tests/data/species-selector/species-search.ts#L3) :-)

src/ensembl/src/content/app/species-selector/components/species-autosuggestion-panel/SpeciesAutosuggestionPanel.tsx
```diff
@@ -4,15 +4,24 @@ import SpeciesSearchMatch from '../species-search-match/SpeciesSearchMatch';
...
 const SpeciesAutosuggestionPanel = (props: Props) => {
+  const onMatchSelected = (match: SearchMatch) => () => {
+    props.onMatchSelected(match);
+  };
+
   const matches = props.matches.map((match) => (
-    <SpeciesSearchMatch match={match} key={match.description} />
+    <SpeciesSearchMatch
+      match={match}
+      onClick={onMatchSelected(match)}
```
imran-ebi: A bit curious about this one. Does using a wrapper function prevent re-renders that are usually caused by using inline functions?

azangru: No, it doesn't :-) It was written solely to trick the linter, which at the moment is complaining when you write functions inline. This is a perfect example of why this linting rule is stupid, because performance-wise both ways of writing the handler are exactly the same (and the performance cost of generating a new function each time the component re-renders is minuscle anyway).

Here's my favourite Ryan Florence's tweet about unhealthy obsession with functions created during render:

image

I suggest we get rid of that linting rule :-)

imran-ebi: There is an interesting discussion about it here: https://twitter.com/dan_abramov/status/1005566478089965574. Some are leaning towards not using them at all. Having said that, functional components will re-render the function anyway 😄

azangru: > In practice I’d suggest ignoring this rule until you have real perf problems, and then maybe following it where it matters

Wise words! :smile:

PR #11, opened 2019-03-07
Use environment variables to configure the app

azangru:
- [x] Use the `.env` file to keep (sensitive) configuration variables 
- [x] Copy `.env` file locally from `.env.example` if the `.env` file is missing 
- [x] Use the `dotenv` package to read the `.env` file and make the values of environment variables available to the client-side js using `webpack.DefinePlugin`

### DIFF
--- a/src/ensembl/package.json
+++ b/src/ensembl/package.json
@@ -13,8 +13,9 @@
   "author": "EMBL-EBI",
   "license": "MIT",
   "scripts": {
-    "start": "npm run serve:dev",
-    "serve:dev": "webpack-dev-server --config ./webpack/webpack.config.dev.js",
+    "copy-dotenv": "test ! -f .env && (cp .env.example .env; echo '.env file created') || true",
+    "start": "npm install --no-save && npm run serve:dev",
+    "serve:dev": "npm run copy-dotenv && webpack-dev-server --config ./webpack/webpack.config.dev.js",
     "serve:prod": "node ./server.js",
     "build": "rimraf ./dist && webpack --config ./webpack/webpack.config.prod.js",
     "deploy": "node deploy",
@@ -44,6 +45,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
+    "dotenv": "6.2.0",
     "foundation-sites": "6.5.3",
     "react": "16.8.1",
     "react-cookie": "^3.0.8",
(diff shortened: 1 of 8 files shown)


src/ensembl/package.json
```diff
@@ -13,8 +13,9 @@
   "author": "EMBL-EBI",
   "license": "MIT",
   "scripts": {
-    "start": "npm run serve:dev",
-    "serve:dev": "webpack-dev-server --config ./webpack/webpack.config.dev.js",
+    "copy-dotenv": "test ! -f .env && (cp .env.example .env; echo '.env file created') || true",
+    "start": "npm install --no-save && npm run serve:dev",
```
azangru: Using `npm install` will slightly (because the packages will be checked against the cache) increase the startup time, but will ensure that the developer always has proper packages.
The --no-save option makes sure that `npm install` does not modify `package-lock.json` (which we'll need to return to the repo, hopefully soon).

src/ensembl/package.json
```diff
@@ -13,8 +13,9 @@
...
+    "serve:dev": "npm run copy-dotenv && webpack-dev-server --config ./webpack/webpack.config.dev.js",
```
azangru: Also, not sure whether `npm run copy-dotenv` command belongs here or in `npm start`

azangru: Ideally, prod environment will have environment variables provided to it, which will be accessible to Node via the `process.env` field at build time.

Right now, if the variables are not provided, the defaults from `.env.example` will be used. If they are exist in the environment, they will override the defaults.

_(For example, NODE_ENV should already be changing to production — I guess via some webpack magic. At least it does so on my machine)._

msmanoj: Thanks, @azangru. 

@imran-ebi, if you could have a quick look on this PR, we can get this merged.

ens-ds23: It would be great to get this on master as soon as practical as we're already losing quite a lot of time to mucking up commits due to different environments. If it's done in the next couple of days I'll cherry-pick it onto vizbi too. If not, it can wait till we merge/PR/squash-commit the vizbi branch.

ens-ds23: Just seen this has already been done! Sorry for the noise, :-) .

PR #12, opened 2019-03-12
Add track panel eye toggle icon to storybook

msmanoj:
Created the eye toggle icon as a shared component and crated a storybook entry for it.


src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import eyeOnIcon from './eye-on.svg';
+import eyeOffIcon from './eye-off.svg';
```
azangru: We have been using 2 svg's (as images) everywhere, but I think this should be done by

- first, importing the svg as a React component
- secondly, by applying various classes to the svg to change its fill property.

_(see the CloseButton in the Storybook as an example)._

The only drawback to this approach is that the svg will become part of the corresponding javascript file and make it a bit larger.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import styles from './EyeToggleIcon.scss';
+
+type Props = {
+  iconStatus: string;
+  description: string;
+  callbackFunction: () => void;
+};
+
+const EyeToggleIcon = (props: Props) => {
```
azangru: I also have a naming concern. Is this component an "icon", or is it a "button" (since it has an `onClick` behavior)? @imran-ebi @ens-ap5 what would you say?

We are talking about this thingy, right?

image

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import eyeOnIcon from './eye-on.svg';
+import eyeOffIcon from './eye-off.svg';
+
+import styles from './EyeToggleIcon.scss';
+
+type Props = {
+  iconStatus: string;
```
azangru: Is it an 'on/off' string? Would it be better represented by a boolean (e.g. `disabled: boolean`)?

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
```diff
@@ -0,0 +1,37 @@
...
+  description: string;
+  callbackFunction: () => void;
```
azangru: `onClick` instead of `callbackFunction `?

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
ens-ap5: I would say that if it has behaviour, it's a button. An icon is an image used to represent a concept (rather than for its decorative effect).

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
imran-ebi: This is something I intend to look at for this sprint, which ends only by end of next week. Most of these icons will get converted into React components using `svgr` and make use of classes to change to the different icon states.

src/ensembl/stories/shared-components/eye-toggle-icon/EyeToggleIcon.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+        callbackFunction={toggleEye}
+      />
+    </div>
+  );
+};
+
+storiesOf('Components|Shared Components/EyeToggleIcon', module)
+  .add('On', () => {
+    return <EyeToggleIconParent eyeStatus={'on'} />;
+  })
```
azangru: Would just one story with a clickable icon/button be more representative of the button's behavior?

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
imran-ebi: Since the component replaces the whole button instead of only the icon, it looks more like a "button" component to me.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
imran-ebi: The icon can have at least three statuses. The three I've come across are: on, off and disabled.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
imran-ebi: Or maybe `clickHandler` to differentiate between the prop and the React `onClick` event?

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: Oh, a 'disabled' button is visually different from an 'off' button, huh? 🤔 Then the button should encounter for all three of these statuses.

Would an `enum` be better than a `string` then?

imran-ebi: Agree, it would be better than a `string`.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: There was a `handleClick` function in the component's implementation, which makes `clickHandler` potentially as confusing as `onClick` :-) I am happy to go with either, but think `onClick` is a slightly more natural name.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
msmanoj: It was implemented as a string in the track panel so I didn't change it. It's updated to boolean now.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
msmanoj: To me it's more like asking "which external function needs to be called when this button is clicked?" and also it would differentiate it from the React onClick like @imran-ebi mentioned.

How about using `onClickHandler`?

src/ensembl/stories/shared-components/eye-toggle-icon/EyeToggleIcon.stories.tsx
msmanoj: Yes, that's a good point. I've updated it now.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: Well, if it's a boolean, then the prop should be renamed to `disabled` or `isDisabled` or something like that (`iconStatus` is not something that you would expect to be true or false). But as @imran-ebi pointed out, there are more than just 2 states of that button; so probably we should go with `enum`.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: I would say, too verbose.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: Again continuing on the subject of naming, wouldn't you agree that the name `VisibilityButton` better describes the purpose of this button than `EyeToggleButton`?

imran-ebi: I am not sure what the proper name should be. At the time of naming it, the purpose of it was not very clear to me as it did keep changing slightly now and then. But now it is surely more refined and we can confirm with Andrea what it is exactly for.

msmanoj: I also noticed that we have the following buttons with similar functionality:

Screenshot 2019-03-13 at 09 25 54
Screenshot 2019-03-13 at 09 25 34

I guess it could all be implemented with just one component.

I feel `VisibilityButton` describes the purpose of the button but not the button itself.

What if we call it something like `ImageToggleButton` which would mean that it is an image that's used like a button and it can be toggled between on, off and disabled states?

imran-ebi: The track panel bar (vertical bar) buttons have a similar functionality. However, the launchbar icons belong to links so I am not sure if we can put them in the same category.

imran-ebi: The header buttons though can go into the same category.

image

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
msmanoj: Should we have `isDisabled` and `isActive` so that we can use it like:
<Button isDisabled />, <Button isActive />, <Button isDisabled isActive={false} />

imran-ebi: I'd still prefer it to be an `enum` as in the future we could have more than three states for this button.

msmanoj: Cool, `enum` is fine. I think `disabled` needs to be separated out as we will not be calling the onClick function if it is.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
azangru: It might be worth having a single component if it consistently behaves the same (will it? I miss Andrea!). Only what will its interface (props) be?

imran-ebi: Won't we need to rely on a function for the link to do the redirection then? The link should be separate from the button for semantic reasons too in my opinion.

azangru: If there is going to be a single button, I think we should start by prototyping its interface: what it requires from the outside world to function.

Something like:

```
OnClickProps = {
  onClick: () => void
} & CommonProps

LinkProps = {
  path: string
} & CommonProps

CommonProps = {
  ??? (is it going to be one `icon` or the `children` component)?
}
```

with

```
const VisibilityButton = () => {

  return (
    <ToggleImageButton onClick={blah} status={blah}>
      <EyeIcon/>
    </ToggleImageButton>
  );

}
```

or something?

I can understand why the wrapper component (like ToggleImageButton) may be useful, but on the other hand am not sure it will make our lives simpler.

imran-ebi: This looks interesting. Also, this made me wonder what if we have a generic icon component that has states. And we pass the icon information to it to render the eye icon rather than having a specific icon for it? A very abstract example would be:

`
const GenericStateIcon = ({ iconDetails }) => {
  return 
}

const ToggleImageButton = () => {
  return <button><GenericStateIcon iconDetails={...} /></button>
}
`

Or the eye icon component can use this generic component instead?

`
const EyeIcon = () => {
  const iconDetails = {...}

  return <GenericStateIcon {...props} />
}

const ToggleImageButton = () => {
  return <button><EyeIcon /></button>
}
`

azangru: > `return `

`svgr` though!

Also, now that you told me that a button can have more than 2 states, and that these 2 states can be visually represented by different (as opposed to just differently styled) svg's, I am a bit uncertain as to the practicality of this approach in general.

Of course, different svg's can still be put into 1 svg (as an svg sprite), and shown/hidden/styled using CSS.

Something we need to brainstorm...

imran-ebi: I get what you mean. This approach won't be suitable for SVG components, however can't we make these components configurable the same way?

src/ensembl/package.json
```diff
@@ -59,6 +60,7 @@
     "redux": "4.0.1",
     "redux-devtools-extension": "2.13.8",
     "redux-thunk": "2.3.0",
+    "svg-react-loader": "0.4.6",
```
azangru: Regarding @svgr/core and svg-react-loader: what functionality do they add compared to @svgr/webpack that we've already got?

(Also, if we were to add them, they would have gone to dev dependencies (`npm install --save-dev --save-exact blah`)

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -115,10 +122,11 @@ const TrackPanelListItem: FunctionComponent<TrackPanelListItemProps> = (
         <button onClick={changeDrawerViewHandler}>
           <img src={ellipsis.icon.on} alt={`Go to ${track.label}`} />
         </button>
-        <EyeToggleButton
-          iconStatus={trackStatus}
+        <ToggleImageButton
+          buttonStatus={trackStatus}
           description={'enable/disable track'}
           onClick={toggleTrack}
+          imageFile={Eye}
```
azangru: it isn't a file anymore; it has been transformed into a ReactComponent. `image` perhaps? or `icon`?

src/ensembl/package.json
msmanoj: I have actually removed them in the latest commit.

src/ensembl/src/shared/eye-toggle-icon/EyeToggleIcon.tsx
imran-ebi: In the comment I actually meant to ask if we can have a generic wrapper component that will take in props on behalf of the SVG component, in this case the `EyeIcon`.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
msmanoj: Changed it to `image`.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -113,12 +122,12 @@ const TrackPanelListItem: FunctionComponent<TrackPanelListItemProps> = (
...
-            src={trackStatus === 'on' ? eye.icon.on : eye.icon.off}
-            alt={eye.description}
-          />
-        </button>
+        <ToggleImageButton
+          buttonStatus={trackStatus}
+          description={'enable/disable track'}
+          onClick={toggleTrack}
+          image={Eye}
+        />
```
azangru: I am not sure what TrackPanelListItem does (and now that `master` has hardcoded `http://127.0.0.1:4000` as api host I can't check it in my browser), but wouldn't you say that an intermediary component would be more suitable here?

For example, let's assume that this intermediary component is called `VisibiliryButton`. It, in turn, calls `ToggleImageButton`. This makes `VisibiliryButton` responsible for the proper icon (TrackPanelListItem shouldn't know that it's an eye) and for supplying `ToggleImageButton` with custom classes (if required) that can override default styling.

src/ensembl/src/shared/toggle-image-button/ImageHolder.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import { ImageButtonStatus } from 'src/shared/toggle-image-button/ToggleImageButton';
+
+type Props = {
+  buttonStatus: ImageButtonStatus;
+  image: any;
```
azangru: I think we know what the type of the image is.

If we are passing an svg transformed to React component, that will probably be `ReactNode`, and if we are allowing passing either inline svg or path to image, that will be `ReactNode | string`.

src/ensembl/src/shared/toggle-image-button/ImageHolder.tsx
```diff
@@ -0,0 +1,30 @@
...
+  imageStyles: any;
```
azangru: This should also be possible to type more precisely:

```
classNames: { [key: keyof ImageButtonStatus ]: string }
```
?

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export enum ImageButtonStatus {
+  ACTIVE = 'active',
+  INACTIVE = 'inactive',
+  DISABLED = 'disabled'
+}
+
+type Props = {
+  buttonStatus: ImageButtonStatus;
+  description?: string;
```
azangru: If it's an optional prop, it shouldn't be defined in defaultProps. Alternatively, if it is defined in defaultProps, then it's not an optional prop.

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import dafaultStyles from './ToggleImageButton.scss';
```
azangru: dafault -> default

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+  buttonStatus: ImageButtonStatus;
+  description?: string;
+  image: any;
+  imageStyles?: any;
+  onClick: () => void;
+};
+
+const ToggleImageButton = (props: Props) => {
+  return (
+    <div
```
azangru: Should it be a div or a button? @imran-ebi ?

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+      onClick={
+        props.buttonStatus === ImageButtonStatus.DISABLED
+          ? () => {
+              return;
+            }
+          : () => props.onClick()
```
azangru: Perhaps instead of writing this in jsx itself, we could have something like:

```
const buttonProps = props.buttonStatus === ImageButtonStatus.DISABLED ? {} : { onClick: props.onClick }

<div {...buttonProps}>
  blah
</div>
```

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
imran-ebi: I think it should be a button

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+        description={props.description}
+        image={props.image}
+        imageStyles={props.imageStyles ? props.imageStyles : dafaultStyles}
+      />
+    </div>
+  );
+};
+
+ToggleImageButton.defaultProps = {
+  buttonStatus: ImageButtonStatus.ACTIVE,
```
azangru: Do we want a default buttonStatus for a button, or should it just always be controlled by the parent?

src/ensembl/src/shared/toggle-image-button/ImageHolder.tsx
msmanoj: Should it be `ComponentClass | string` or `ReactNode | string` ?

I might be doing something wrong but I'm getting the following error when I use `ReactNode`:

> JSX element type 'props.image' does not have any construct or call signatures.

On the line:
`{typeof props.image === 'string' ?  : <props.image/>}`

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
msmanoj: I guess we wouldn't be creating a button without defining it's initial state will.

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
msmanoj: That's done.

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
msmanoj: That's done.

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
msmanoj: That's done.

src/ensembl/src/shared/toggle-image-button/ToggleImageButton.tsx
msmanoj: That's done.

src/ensembl/src/shared/toggle-image-button/ImageHolder.tsx
msmanoj: Not sure if this still applies as we will be passing multiple classNames?

src/ensembl/src/shared/image-button/ImageButton.scss
```diff
@@ -1,29 +1,59 @@
...
 
 .default {
+  position: relative;
   svg {
     width: 100%;
     height: 100%;
     fill: $ens-blue;
     cursor: pointer;
+    transition: opacity 0.4s;
+    -webkit-transition: opacity 0.4s;
```
azangru: Here and elsewhere: won’t `postcss-preset-env` that we have in our dependencies add all this stuff automatically?

imran-ebi: Yes it should

msmanoj: Cool, I wasn't aware of that. I've updated it now.

src/ensembl/src/shared/image-button/ImageButton.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+      expect(wrapper.props().image).toEqual('');
+    });
+
+    it('respects the image prop', () => {
+      const wrapper = mount(<ImageButton image={'foo.svg'} />);
+
+      expect(wrapper.props().image).toEqual('foo.svg');
+    });
+  });
```
azangru: For the three `describe` blocks above, what is the intended behaviour for one or the other prop? It's not terribly interesting to know that there is a particular default prop that gets replaced with a prop that is passed to a component — it is more interesting to know why we are passing the props, i.e. how these props are used by the component.

src/ensembl/src/shared/image-button/ImageButton.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    it('respects the classNames prop', () => {
+      const wrapper = mount(<ImageButton classNames={'foo'} />);
+
+      expect(wrapper.props().classNames).toEqual('foo');
+    });
+  });
+
+  describe('prop onClick', () => {
+    const onClick = jest.fn();
+    it('respects the onClick prop', () => {
```
azangru: `calls the onClick prop when clicked`?

src/ensembl/src/shared/image-button/ImageButton.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    it('respects the image prop', () => {
+      const wrapper = mount(<ImageButton image={'foo.svg'} />);
+
+      expect(wrapper.props().image).toEqual('foo.svg');
+    });
+  });
+
+  describe('prop classNames', () => {
+    it('respects the classNames prop', () => {
```
azangru: Why wouldn't it respect the classNames prop? There is nothing competing for this prop (like a default prop could); and accepting a prop is just a standard React behaviour.

src/ensembl/src/shared/image-button/ImageButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  DISABLED = 'disabled',
+  DEFAULT = 'default',
+  HIGHLIGHTED = 'highlighted'
+}
+
+type Props = {
+  buttonStatus: ImageButtonStatus;
+  description: string;
+  image: React.FunctionComponent<React.SVGProps<SVGSVGElement>> | string;
+  classNames?: any;
```
azangru: Well, we know the type of this thing, don't we?

([example](https://www.typescriptlang.org/play/#src=export%20enum%20ImageButtonStatus%20%7B%0D%0A%20%20ACTIVE%20%3D%20'active'%2C%0D%0A%20%20INACTIVE%20%3D%20'inactive'%2C%0D%0A%20%20DISABLED%20%3D%20'disabled'%2C%0D%0A%20%20DEFAULT%20%3D%20'default'%2C%0D%0A%20%20HIGHLIGHTED%20%3D%20'highlighted'%0D%0A%7D%0D%0A%0D%0A%0D%0A%0D%0Aconst%20classNames%3A%20%7B%20%5Bkey%20in%20ImageButtonStatus%5D%3F%3A%20string%20%7D%20%3D%20%7B%0D%0A%20%20active%3A%20'foo-active'%2C%0D%0A%7D%0D%0A%0D%0Aconst%20invalidClassNames%3A%20%7B%20%5Bkey%20in%20ImageButtonStatus%5D%3F%3A%20string%20%7D%20%3D%20%7B%0D%0A%20%20active%3A%20'foo-active'%2C%0D%0A%20%20passive%3A%20'foo-passive'%0D%0A%7D))

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
```diff
@@ -39,19 +41,22 @@ const TrackPanelBarIcon: FunctionComponent<TrackPanelBarIconProps> = memo(
...
       <dt className={styles.barIcon}>
-        <button onClick={toggleModalView}>
-          <img src={getViewIcon()} alt={props.iconConfig.description} />
-        </button>
+        <ImageButton
+          buttonStatus={getViewIconStatus()}
+          description={props.iconConfig.description}
+          onClick={toggleModalView}
+          image={props.iconConfig.icon}
+        />
```
azangru: Here's what I am currently seeing on the production site:

Screenshot 2019-03-28 at 15 27 04

And here's what I am seeing in this branch:

Screenshot 2019-03-28 at 15 27 20

The styles are off.

azangru:
image

Typescript is unhappy for various reasons (apart from its complaints about SpeciesAutosuggestionPanel)

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
msmanoj: Seems to be working fine for me. Can you clear the cache and try again, please?

src/ensembl/src/shared/image-button/ImageButton.tsx
msmanoj: I've update it.

src/ensembl/src/shared/image-button/ImageButton.test.tsx
msmanoj: That's done

src/ensembl/src/shared/image-button/ImageButton.test.tsx
msmanoj: The tests are now updated.

src/ensembl/src/shared/image-button/ImageButton.test.tsx
```diff
@@ -13,62 +14,88 @@ describe('<ImageButton />', () => {
...
+      expect(wrapper.prop('image')).toEqual('');
     });
 
-    it('respects the image prop', () => {
-      const wrapper = mount(<ImageButton image={'foo.svg'} />);
+    it('renders and img tag if a path to an image file is passed', () => {
+      const wrapper = mount(<ImageButton image={'foo.png'} />);
 
-      expect(wrapper.props().image).toEqual('foo.svg');
+      expect(wrapper.find(ImageHolder).find('img')).toHaveLength(1);
```
azangru: And make sure that  has the expected source (string passed in the image prop).

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
azangru: Yep, looks fine now.

src/ensembl/src/shared/image-button/ImageButton.test.tsx
msmanoj: That's done

PR #14, opened 2019-03-20
ADR for react animation library

msmanoj:
I have compared the top three animations libraries available for React and the gap between the top two seems to be very less.
Turns out that we could also use both React Transition Group and React Spring, but that's something we need to discuss and decide.

### DIFF
--- a/src/ensembl/doc/architecture/decisions/0013-animation-library.md
+++ b/src/ensembl/doc/architecture/decisions/0013-animation-library.md
@@ -20,6 +20,8 @@ Example animations:
 The following are top three libraries available:
 
 **1) React Transition Group** - (https://github.com/reactjs/react-transition-group)
+ Works by simply changing classes on an element depending on lifecycle of a transition (entering transition, being in transition, exiting transition). Since all it does is just changing classes, the transition itself has to be written in CSS. On the plus side, this means these transitions are the most performant. On the minus side, CSS gets decoupled from JS. Also, the duration of animation is imperatively controlled in two places: by time passed to the component and by duration specified in CSS.
+
 - Minified size: 16.9 KB
 - 67 issues open and 260 closed
 - Gives more flexibility and control with increase in complexity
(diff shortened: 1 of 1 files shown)


azangru:
Some important considerations missing from the ADR:

- `react-transition-group` works by simply changing classes on an element depending on lifecycle of a transition (entering transition, being in transition, exiting transition). Since all it does is just changing classes, the transition has to be written in CSS. On the plus side, this means these transitions are the most performant. On the minus side, CSS gets decoupled from JS. Also, the duration of animation is imperatively controlled by time passed to the component.

- `react-spring` is a spiritual inheritor of `react-motion`, sharing the general approach that `react-motion` introduced. It works entirely in javascript, by interpolating values (any values) over time using interpolation functions ("springs"). What this means is that duration of animation is controlled by parameters such as "tension" or "friction" passed to the spring function (so instead of thinking in terms of time you have to think in terms of spring metaphor: how rigid or elastic your spring is and how heavy of a weight it is trying to move). This, arguably, makes the animation more "physical" (and therefore "realistic"). Also, the animation itself is done by getting interpolated variables constantly changing and constantly re-rendering a part of your component passing new inline styles to it. In terms of performance, it is almost inevitably worse than by having a CSS animation. The performance hit, however, is usually tolerable.

- `pose` is an adaptation of the `popmotion` library for react. People who have experience with it like it for its declarative interface.

There is an interesting project comparing various animation libraries for creating a particular (reasonably complicated) animation: https://github.com/aholachek/react-animation-comparison

msmanoj:
Thanks, Andrey.

Would there be any additional benefits if we make use of both react-transition-group as well as react-spring?

azangru:
There's both a benefit and a drawback to doing so.

**Benefit:** simple animations can then be done with CSS alone (which is native to the browser and thus more performant).

**Drawback:** we will ship more code (2 libraries) to the client.

msmanoj:
@azangru would it be an overkill to implement simple animations using `react-spring`?

azangru:
Good question. We can probably give it a try and see how it goes.

imran-ebi:
Can we merge this PR into `dev`?

PR #16, opened 2019-03-20
Add Button component

azangru:
Talked to Andrea about button names and their purpose.

It seemed like the consensus naming convention was to have a `PrimaryButton` and a `SecondaryButton`, which are added in this PR.

Going forward, we will also likely have a ButtonLink component (implemented as an `a` tag) and a Tab or a TabButton component; but that's for future PRs.


src/ensembl/src/shared/button/Button.tsx
```diff
@@ -0,0 +1,46 @@
...
+  children: ReactNode;
+};
+
+export const PrimaryButton = (props: Props) => {
+  const className = classNames(
+    styles.primaryButton,
+    { [styles.primaryButtonDisabled]: props.isDisabled },
+    props.className
+  );
+  return <Button {...props} className={className} />;
```
msmanoj: Should we separate out `className` from props before passing it down to Button?

azangru: It should overwrite the `className` in props, it that's what you are asking.

msmanoj: I was wondering if it would be better to do something like `const {className, ...rest} = props` so that there is no overwriting?

Also, would `<Button  className={className} {...props} />` still work the same way?

azangru: > would <Button className={className} {...props} /> still work the same way?

It shouldn't. [Later attributes override previous ones.](https://gist.github.com/sebmarkbage/07bbe37bc42b6d4aef81#spread-attributes).

I don't know whether explicitly destructuring `className` from the `props` will be better. It lacks explicitness, but loses terseness, and for a developer familiar with JSX it's apparent what's going on anyway. I can re-write it this way, if that's what you would prefer.

msmanoj: Although it's apparent, I was just wondering if all the developers who will be working on updating this component in the future will be aware that this ordering cannot be changed.

azangru: I would say that, since it's standard JSX behaviour, they will (or should, or eventually will) be aware of that. But as a safeguard, there already is a test in place that checks that classes are correctly assigned. It fails if you change the order of props:

image

src/ensembl/src/shared/button/Buttons.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  test('assigns the primaryButton class to the button', () => {
+    const renderedButton = wrapper.find('button');
+    expect(renderedButton.hasClass('primaryButton')).toBe(true);
+  });
+
+  test('extends own class with external classname', () => {
+    const externalClass = faker.lorem.word();
+    const buttonChildren = faker.lorem.words();
```
msmanoj: Do we need to redeclare `buttonChildren` here?

src/ensembl/stories/shared-components/button/Button.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      </>
+    ),
+    { notes: primaryButtonNotes }
+  );
+
+storiesOf('Components|Shared Components/Button/SecondaryButton', module).add(
+  'default',
+  () => (
+    <div style={{ padding: '40px' }}>
```
msmanoj: `${styles.wrapper}` instead?

src/ensembl/src/shared/button/Buttons.test.tsx
azangru: We don't. Fixed.

src/ensembl/stories/shared-components/button/Button.stories.tsx
azangru: 👍 fixed

src/ensembl/src/shared/button/Button.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  test('assigns the secondaryButton class to the button', () => {
+    const renderedButton = wrapper.find('button');
+    expect(renderedButton.hasClass('secondaryButton')).toBe(true);
+  });
+
+  test('extends own class with external classname', () => {
+    const externalClass = faker.lorem.word();
+    const buttonChildren = faker.lorem.words();
```
msmanoj: Might not be required here as well.

src/ensembl/src/shared/button/Button.tsx
msmanoj: That's cool, we have something to safeguard.

src/ensembl/src/shared/button/Button.test.tsx
azangru: Removed :-)

PR #17, opened 2019-03-20
Accordion shared component

msmanoj:
Added the shared accordion component that's based on [react-accessible-accordion ](https://github.com/springload/react-accessible-accordion)

I have made very less modifications compared to the original repo, such as fixing the linter errors and using sass instead of css.


src/ensembl/src/shared/accordion/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+type AccordionProps = React.HTMLAttributes<HTMLDivElement>;
+
+const Accordion = ({ ...rest }: AccordionProps): JSX.Element => {
```
imran-ebi: I am curious why you didn't use a param for this rather than rest operator? You could write this as (props: AccordionProps) and use it as a spread operator in the `div` element as you've already done with ...rest.

src/ensembl/src/shared/accordion/Accordion/Accordion.wrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type AccordionWrapperProps = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  'onChange'
+> & {
+  accordion?: boolean;
+  onChange(args: UUID | UUID[]): void;
+};
+
+export default class AccordionWrapper extends React.Component<
```
imran-ebi: Any reason why you didn't like using a functional component with hooks? Just curious 😄

msmanoj: This is something I wanted to have a discussion about. At the moment, I just used the code from the package `react-accessible-accordian` which is based on classes. Before I start to re-write it to use functional components, I would like to know if we are not planning to use class components for any particular reason?

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
+import { mount, ReactWrapper } from 'enzyme';
+import * as React from 'react';
```
azangru: Due to our typescript settings, `import React from 'react'` works just fine; no need to import star.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+import {
+  Item,
+  Provider,
+  ProviderState
+} from '../AccordionContainer/AccordionContainer';
+import { default as AccordionItem } from '../AccordionItem/AccordionItem.wrapper';
```
azangru: `import AccordionItem from ...`

src/ensembl/src/shared/accordion/Accordion/Accordion.tsx
msmanoj: Well spotted! I couldn't think of any reason why `...rest` will be required here instead of props. I've updated it now.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+import { default as AccordionItemTitle } from '../AccordionItemTitle/AccordionItemTitle.wrapper';
+import { default as Accordion } from './Accordion.wrapper';
+
+describe('Accordion', () => {
+  it('different className', () => {
```
azangru: let's write test descriptions in a grammatically sensible way? `describe` is a verb that takes a noun (`describe('Accordion')), but `it` is a noun that expects a predicate to make a clause (`it('accepts a class name')`, or something).

Instead of `describe('<Accordion accordion="true" />'`, why not describe what this `accordion=true` prop is representing (I don't know what it is, so can't offer any suggestions; perhaps you could explain?)

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+      const instance = wrapper.find(Provider).instance() as Provider;
+
+      expect(
+        instance.state.items.filter((item: Item) => item.expanded === true)
+          .length
+      ).toEqual(0);
+    });
+  });
+
+  it('does not expanded disabled items on click', () => {
```
azangru: does not expanded -> does not expand

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
azangru: Does the `accordion=true` prop allow opening only a single panel at a time?

Foundation's accordion calls this property "multi" (when it's on, multiple panels can be opened). Isn't it a better name?

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
msmanoj: > Does the accordion=true prop allow opening only a single panel at a time?

Yes, that's right. I feel `multi' is still a bit vague. How about `allowMultiple`?

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+      instance.state.items.filter((item: Item) => item.expanded).length
+    ).toEqual(1);
+    expect(
+      wrapper.findWhere((item: ReactWrapper) =>
+        item.hasClass(hideBodyClassName)
+      ).length
+    ).toEqual(1);
+  });
+
+  it('pre expanded accordion when accordion is false', () => {
```
azangru: it('expands multiple accordion items when...')

src/ensembl/src/shared/accordion/Accordion/Accordion.wrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+  AccordionWrapperProps
+> {
+  public static defaultProps: AccordionWrapperProps = {
+    accordion: true,
+    className: accordionStyles.accordion,
+    onChange: (): void => {
```
azangru: () => void?

src/ensembl/src/shared/accordion/AccordionContainer/AccordionContainer.tsx
```diff
@@ -0,0 +1,165 @@
+// tslint:disable:max-classes-per-file
```
azangru: do we need this?

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+    wrapper
+      .find('.foo')
+      .first()
+      .simulate('click');
+    expect(
+      instance.state.items.filter((item: Item) => item.expanded === true).length
+    ).toEqual(0);
+  });
+
+  it('pre expanded accordion', () => {
```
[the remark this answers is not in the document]
imran-ebi: Is this about an item being expanded by default? If yes, how about `it('expands by default')`?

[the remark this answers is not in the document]
msmanoj: I've updated it to `pre expands the item that has expanded === true`. That sounds okay?

src/ensembl/src/shared/accordion/AccordionContainer/AccordionContainer.tsx
msmanoj: We need to include this line as the linter is not happy with having two class defined in a single file. We could alternatively disable this rule in the tslint config if that's okay?

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+    );
+
+    const instance = wrapper.find(Provider).instance() as Provider;
+
+    expect(
+      instance.state.items.filter((item: Item) => item.expanded === true).length
+    ).toEqual(1);
+  });
+
+  it('works with multiple pre expanded accordion. Extra expands are just ignored.', () => {
```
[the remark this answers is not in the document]
msmanoj: That's done.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/accordion/AccordionContainer/AccordionContainer.tsx
azangru: I think disabling the rule should be okay; I can't think of a reason why we need the arbitrary limitation of no more than 1 class per file. @imran-ebi ?

imran-ebi: Yes we can do that. Anyway, I doubt we will be having more than one class when we switch most of these components to functional components.

src/ensembl/src/shared/accordion/Accordion/Accordion.test.tsx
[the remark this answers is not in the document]
azangru: `pre-expands the item that is marked as expanded` ?

(to avoid the implementation detail of `expanded=true`, and to avoid detachment of the prefix from the word it belongs to 😄 )

src/ensembl/src/shared/accordion/components/Accordion.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+      const wrapper = mount(<Accordion data-testid={UUIDS.FOO} />);
+      expect(wrapper.find('div').props().className).toEqual('accordion');
+    });
+
+    it('can be overridden', () => {
+      const wrapper = mount(
+        <Accordion className="foo" data-testid={UUIDS.FOO} />
+      );
+      expect(wrapper.find('div').props().className).toEqual('foo');
+    });
```
azangru: Do we want to overwrite the className entirely, or to extend it with a classname received from the parent (so that a portion of styles could be reused)?

compare:  `expect(wrapper.find('div').props().className).toEqual('foo')` vs `expect(wrapper.find('div').props().className).toEqual('accordion foo')` (or `.toContain('foo')`?

src/ensembl/src/shared/accordion/components/Accordion.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { UUID } from './ItemContext';
+import styles from '../css/Accordion.scss';
+
+type AccordionProps = Pick<
+  DivAttributes,
+  Exclude<keyof DivAttributes, 'onChange'>
+> & {
+  preExpanded?: UUID[];
+  allowMultipleExpanded?: boolean;
+  allowZeroExpanded?: boolean;
```
azangru: If allowMultipleExpanded and allowZeroExpanded are among default props, then they are always present, which means there is no need in the question mark.

src/ensembl/src/shared/accordion/components/AccordionItem.test.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
```
azangru: import React from 'react'

src/ensembl/src/shared/accordion/components/AccordionItem.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { cleanup, render } from 'react-testing-library';
```
azangru: But we aren't using `react-testing-library`.
Either enzyme for all unit tests, or react-testing-library for all unit tests, but not both.

msmanoj: Yes, I'm currently re-writing all the tests.

src/ensembl/src/shared/accordion/components/AccordionItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { nextUuid } from '../helpers/uuid';
+import { Provider as ItemProvider, UUID } from './ItemContext';
+import styles from '../css/Accordion.scss';
+
+type Props = DivAttributes & {
+  uuid?: UUID;
+};
+
+const defaultProps = {
+  className: styles.accordion__item
```
azangru: Seems we are not using this style of class naming.

src/ensembl/src/shared/accordion/components/AccordionItemButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Consumer as ItemConsumer, ItemContext } from './ItemContext';
+
+import styles from '../css/Accordion.scss';
+
+type Props = DivAttributes & {
+  toggleExpanded(): void;
+};
+
+const defaultProps = {
+  className: styles.accordion__button
```
azangru: naming style

src/ensembl/src/shared/accordion/components/AccordionItemPanel.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import DisplayName from '../helpers/DisplayName';
+import { DivAttributes } from '../helpers/types';
+import { Consumer as ItemConsumer, ItemContext } from './ItemContext';
+
+type Props = DivAttributes;
+
+const defaultProps = {
+  className: 'accordion__panel'
```
azangru: name style

src/ensembl/src/shared/accordion/components/ItemContext.tsx
```diff
@@ -0,0 +1,100 @@
+// tslint:disable:max-classes-per-file
+
+import * as React from 'react';
```
azangru: import React from 'react'

src/ensembl/src/shared/accordion/helpers/types.ts
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
```
azangru: import React from 'react'

src/ensembl/src/shared/accordion1/AccordionItem/AccordionItem.wrapper.tsx
```diff
@@ -36,13 +35,15 @@ export default class AccordionItemWrapper extends React.Component<
...
-    uuid: undefined
+    hideBodyClassName: ''
   };
 
   public id: number = nextUuid();
 
-  public render(): JSX.Element {
+  public renderAccordionItem = (): JSX.Element => {
+    console.log(this.context);
+    console.log(this.props);
```
azangru: don't forget to remove

src/ensembl/src/shared/accordion/components/Accordion.test.tsx
msmanoj: Ideally I would like to extend it with the classname received, but it will make it hard to update the default styling later on as any changes we make will be applied to all the accordions.

src/ensembl/src/shared/accordion/components/Accordion.tsx
msmanoj: That's done

src/ensembl/src/shared/accordion/components/AccordionItem.tsx
msmanoj: That's now updated

src/ensembl/src/shared/accordion/components/AccordionContext.test.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
```
azangru: import React from 'react' :-)

src/ensembl/src/shared/accordion/components/AccordionItem.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    it('can be overridden', () => {
+      const wrapper = mount(
+        <Accordion>
+          <AccordionItem uuid={'FOO'} className="foo" />
+        </Accordion>
+      );
+      expect(wrapper.find(AccordionItem).props()['className']).toEqual('foo');
+    });
```
azangru: @imran-ebi, here's where we had a difference of opinion.

One way of styling accordion components is, like in the test, by replacing default class with a provided class. The `accordionItem` class gets replaced by `foo`.

Another way of styling would be to extend the default class with a custom class. The result would be `accordionItem foo`.

An argument against extending is that if the default class gets updated, it will (or might) cause changes in all extended classes. I.e. if you take accordion and put it in a specific context overwriting some of its styles, then after update to the default class, the accordion in that specific context will also get updated (which, I would argue, is something that we might actually want).

An argument for extending is that wherever you do want to apply custom classes, you do not have to define all style rules, but can only update the relevant ones.

Which would you prefer?

src/ensembl/src/shared/accordion/components/AccordionItemButton.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </AccordionItemHeading>
+          </AccordionItem>
+        </Accordion>
+      );
+
+      expect(
+        wrapper
+          .find(AccordionItemButton)
+          .find('div')
+          .props()['className']
```
azangru: perhaps `.props('className')`?

src/ensembl/src/shared/accordion/components/AccordionItemHeading.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+              <AccordionItemHeading>
+                <AccordionItemButton>Hello World</AccordionItemButton>
+              </AccordionItemHeading>
+            </AccordionItem>
+          </Accordion>
+        )
+      ).not.toThrowError(SPEC_ERROR);
+    });
+
+    it('permits a single AccordionItemButton as a child - variation #2', () => {
```
azangru: erm. variation?

It seems the difference is that in the second test the single item is inside of an array. Should we put this in the test description?

(Also, what happens if there are more than one item in the array? Tests don't say).

src/ensembl/src/shared/accordion/components/AccordionItemHeading.tsx
```diff
@@ -14,8 +14,6 @@ const defaultProps = {
 
 export const SPEC_ERROR = `AccordionItemButton may contain only one child element, which must be an instance of AccordionItemButton.
 
-From the WAI-ARIA spec (https://www.w3.org/TR/wai-aria-practices-1.1/#accordion):
-
```
azangru: Why delete this line? Without it, we are not showing where the quoted paragraph on the next line originates from.

src/ensembl/src/shared/accordion/components/AccordionContext.test.tsx
msmanoj: That's done

src/ensembl/src/shared/accordion/components/AccordionItemButton.test.tsx
msmanoj: That's done

src/ensembl/src/shared/accordion/components/AccordionItemHeading.test.tsx
msmanoj: I've updated the tests and also added one more to test to check more items in the array.

src/ensembl/src/shared/accordion/components/AccordionItemHeading.tsx
msmanoj: I guess I removed it by mistake while removing the line breaks. I've put it back now.

imran-ebi:
I notice there is a conflict now 😕

msmanoj:
I’m about to make some more changes to this component now. Once that’s done, I’ll resolve the conflict.

src/ensembl/src/shared/accordion/Accordion/Accordion.wrapper.tsx
imran-ebi: We might use class components when it is difficult to write a component with functions. However, I think we will mainly stick with functional components with hooks.

src/ensembl/src/shared/accordion/components/AccordionItemPermanentBlock.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+// This block has to be please either before or after AccordionItemPanel
```
azangru: erm... `please` -> `placed`?

src/ensembl/src/shared/accordion/components/AccordionItem.test.tsx
imran-ebi: As we discussed yesterday, I would prefer extending the styling. Having said that, we can change the implementation of this to overriding the default styling if the accordions don't look similar to each other.

src/ensembl/src/shared/accordion/components/Accordion.test.tsx
```diff
@@ -19,15 +19,29 @@ describe('Accordion', () => {
...
+        'accordionDefault foo'
+      );
+    });
+
+    it('can also be overridden by using extendDefaultStyles === false', () => {
+      const wrapper = mount(
+        <Accordion
+          className="foo"
+          extendDefaultStyles={false}
+          data-testid={UUIDS.FOO}
```
azangru: is this test id used anywhere?

src/ensembl/src/shared/accordion/components/Accordion.tsx
```diff
@@ -12,28 +13,40 @@ type AccordionProps = Pick<
   preExpanded?: UUID[];
   allowMultipleExpanded: boolean;
   allowZeroExpanded: boolean;
+  className?: any;
```
azangru: Not a string?

msmanoj: I guess `className` is not required as it'll be in the `DivAttributes` anyways.

src/ensembl/src/shared/accordion/components/Accordion.test.tsx
msmanoj: Nope, I've removed them.

src/ensembl/src/shared/accordion/components/AccordionItemPermanentBlock.tsx
msmanoj: That's done.

src/ensembl/src/shared/accordion/components/Accordion.tsx
```diff
@@ -0,0 +1,62 @@
...
+      allowMultipleExpanded={allowMultipleExpanded}
+      allowZeroExpanded={allowZeroExpanded}
+      onChange={onChange}
+    >
+      <Consumer>{renderAccordion}</Consumer>
+    </Provider>
+  );
+};
+
+Accordion.diplayName = DisplayName.Accordion;
```
azangru: Given that display name of the component is inferred automatically from the name of the variable:

image

(https://reactjs.org/docs/react-component.html#displayname)

— why are we setting displayName explicitly? It does not differ from the name of components, as far as I can tell.

src/ensembl/src/shared/accordion/components/AccordionItemHeading.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+  };
+
+  useEffect(() => {
+    VALIDATE(ref);
+  });
+
+  const setRef = (reference: HTMLDivElement): void => {
+    ref = reference;
+  };
```
azangru: Idiomatically, this would be done using the `useRef` hook

src/ensembl/src/shared/accordion/components/AccordionItemPanel.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Consumer as ItemConsumer, ItemContext } from './ItemContext';
+import defaultStyles from '../css/Accordion.scss';
+import classNames from 'classnames';
+
+type Props = DivAttributes & {
+  extendDefaultStyles: boolean;
+};
+
+const AccordionItemPanel = (props: Props) => {
+  const renderChildren = ({ panelAttributes }: ItemContext): JSX.Element => {
```
azangru: Is this function leftover of a class method? Because there doesn't seem to be any reason for it to exist as a separate function.

src/ensembl/src/shared/accordion/components/AccordionItemButton.tsx
```diff
@@ -58,6 +57,6 @@ const AccordionItemButtonWrapper: React.SFC<WrapperProps> = (
   </ItemConsumer>
 );
 
-AccordionItemButtonWrapper.displayName = DisplayName.AccordionItemButton;
+AccordionItemButtonWrapper.displayName = 'AccordionItemButton';
```
azangru: Why not delete it at all? Shouldn't `AccordionItemButton` be inferred from the function name (AccordionItemButton)?

src/ensembl/src/shared/accordion/components/AccordionItemState.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Consumer as ItemConsumer, ItemContext } from './ItemContext';
+
+type Props = Pick<DivAttributes, Exclude<keyof DivAttributes, 'children'>> & {
+  children(
+    args: Partial<{ expanded: boolean; disabled: boolean }>
+  ): React.ReactNode;
+};
+
+const AccordionItemState = (props: Props) => {
+  const renderChildren = (itemContext: ItemContext): JSX.Element => {
```
azangru: This function also probably doesn't even need to exits. Also, props.children probably don't need to be wrapped in a fragment.

src/ensembl/src/shared/accordion/helpers/AccordionStore.ts
```diff
@@ -0,0 +1,130 @@
...
+    const isExpanded = this.isItemExpanded(uuid);
+    const isOnlyOneExpanded = this.expanded.length === 1;
+
+    return Boolean(isExpanded && !this.allowZeroExpanded && isOnlyOneExpanded);
+  };
+
+  public readonly isItemExpanded = (uuid: UUID): boolean => {
+    return (
+      this.expanded.findIndex((expandedUuid: UUID) => expandedUuid === uuid) !==
+      -1
```
azangru: `return this.expanded.includes(uuid)`?

src/ensembl/src/shared/accordion/components/AccordionItemPanel.tsx
msmanoj: I guess it exists as a separate function as the children has to be of type `(container: ItemContext) => ReactNode`

src/ensembl/src/shared/accordion/components/AccordionItemButton.tsx
msmanoj: That's done

src/ensembl/src/shared/accordion/helpers/AccordionStore.ts
msmanoj: Nice one. I've updated it

src/ensembl/src/shared/accordion/components/Accordion.tsx
msmanoj: Removed the setting of `displayName` from all the components.

src/ensembl/src/shared/accordion/components/AccordionItemHeading.tsx
msmanoj: That's done

src/ensembl/src/shared/accordion/components/AccordionItemPanel.tsx
msmanoj: Further to our discussion, I have removed the need for a separate function.

src/ensembl/src/shared/accordion/components/AccordionItemState.tsx
msmanoj: I've remove the function. The fragment is needed here as ItemContext expects only one child.

src/ensembl/src/shared/accordion/components/AccordionItemPanel.tsx
azangru: Have you pushed this change? I am still seeing this function here and in the AccordionItemState component.

msmanoj: Ahh! Yes, I forgot. Thanks for spotting it :)

PR #19, opened 2019-03-21
Add autocomplete search field

azangru:
- [x] Add SearchField as a component
- [x] Add AutosuggestSearchField as a component
- [x] Add tests for Input, SearchField, and AutosuggestSearchField

[Schematic to help visualise what approximately is going on here](https://i.imgur.com/SazZj4I.png)


src/ensembl/src/shared/search-field/SearchField.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const SearchField = (props: Props) => {
+  const { rightCorner } = props;
+  const className = classNames(styles.searchField, props.className);
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+    props.onSubmit && props.onSubmit(props.search);
```
msmanoj: Submission of empty string might be useful when the user wants to clear the input field to reset the results, but on the other hand we are dispatching an action which doesn't do anything. 
What is your opinion about preventing submission of empty string, but at the same time having an alternative way to reset the results?

azangru: This is a very good question, and the one that I do not have a clear answer to.  I did not make this choice consciously when I wrote this code, so it's a great opportunity to explore the options. 

On the one hand, the `SearchField` component, as well as the `AutocompleteSearchField` component, is entirely dependent upon its parent to store the string that goes into the input field. Upon every change of the input field, it calls the `onChange` handler that it receives from its parent. Which means that one way of clearing search results is just to delete the content of the input field, and then rely on the parent component to hide the search results when the search string is empty. So here's the alternative way of resetting the results.

On the other hand, the `onSubmit` handler offers a way of signalling to the parent that the user has made a clear decision (as opposed to the `onChanged` handler that fires at every intermediate step as well). So I am a bit hesitant to block this way of signalling when the field is empty.

Ultimately, it's up to the parent component to decide what to do when the search field is empty anyway. So I am leaning towards having the `onSubmit` fire even when the search field is empty, and then have the logic in the parent component for dealing with these events in appropriate manner.

src/ensembl/src/shared/autosuggest-search-field/AutosuggestSearchField.tsx
```diff
@@ -0,0 +1,230 @@
...
+      setCanShowSuggestions(true);
+    }
+    props.onFocus && props.onFocus();
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if ([keyCodes.UP, keyCodes.DOWN].includes(event.keyCode)) {
+      event.preventDefault();
+    }
+    if (event.keyCode === keyCodes.UP) {
```
msmanoj: Could this `if else` block be moved to be within the scope of the previous `if` block so that we can avoid checking for two additional conditions when the keys other than `up` and `down` are pressed?

azangru: It could, but then we'll have nested if-statements, which some developers dislike. I am happy to move it there if no-one minds.

Alternatively, of course, the first check could be the reverse of what it currently is:

```
if (![keyCodes.UP, keyCodes.DOWN].includes(event.keyCode)) {
      return;
}
event.preventDefault();

if (event.keyCode === keyCodes.UP) {
...
} else if (event.keyCode === keyCodes.DOWN) {
   setHighlightedItemIndex(getNextItemIndex(props, highlightedItemIndex));
}
```

Whichever you prefer.

@imran-ebi any preferences?

msmanoj: This approach would be even better. 
Also, we won't need `if (event.keyCode === keyCodes.DOWN)`

azangru: Updated.

imran-ebi: I too don't like nested if-statements much, but I am still fine if they are used.

PR #20, opened 2019-03-22
Fix tests

azangru:
**General:**
- [x] Update configuration files for running tests properly (tsconfig and webpack)

**Components**
Fix tests for
- [x] Content
- [x] Root
- [x] Launchbar
- [x] BrowserNavIcon
- [x] BrowserNavBar
- [x] Header _(**note**: deleted a couple of tests)_
- [x] HeaderButtons _(**note**: skip test for calling the toggleAccount function; unsure about current behaviour of the component)_
- [x] BrowserBar
- [x] App

**Delete tests that are useless**
- [x] Browser _(remove snapshot test; will try to come up with more useful tests)_
- [x] BrowserImage _(current tests don’t do anything useful; proper tests will be difficult, because this component extensively uses custom events)_
- [x] LaunchbarContainer
- [x] LaunchbarIcon _(they were copy-pasted tests for LaunchbarContainer anyway)_
- [x] browserActions and headerActions _(they were essentially testing the library we use to generate actions; and we shouldn't be testing third-party libraries)_
- [x] headerReducer and browserReducer _(reducers are very simple and hardly need testing)_
- [x] browserSelectors and headerSelectors _(test are very simple and testing things that Typescript will report anyway)_


src/ensembl/src/content/app/App.tsx
```diff
@@ -57,30 +67,33 @@ export const App: FunctionComponent<AppProps> = (props: AppProps) => {
...
-            from={`${url}/browser`}
-            to={{
-              pathname: `${url}/browser/GRCh38_demo/ENSG00000139618`,
-              search: '?region=13:32271473-32437359'
-            }}
-          />
-        </Switch>
-      </Suspense>
-    </Fragment>
+    <AppShell>
```
msmanoj: Is using `Shell` different from using `Wrapper`?

azangru: As a word? I don't think so. Just a different metaphor.

The phrase "app shell" is used a lot to describe a nearly empty box into which the app is going to be rendered.

imran-ebi: I like `AppShell` in this context. Will fit right into the initial rendered component if we use SSR.

src/ensembl/src/root/Root.test.tsx
```diff
@@ -1,29 +1,62 @@
...
+    const wrapper = mount(getRenderedRoot(defaultProps));
+    expect(wrapper.find('.privacyBanner').length).toBe(1);
+  });
+
+  test('does not show privacy banner if privacy cookie is set', () => {
+    cookiesMock.get = jest.fn(() => 'true');
+    const wrapper = mount(getRenderedRoot(defaultProps));
+
+    // useEffect will be run asynchronously; so simulate asynchrony here
+    setTimeout(() => expect(wrapper.find('.privacyBanner').length).toBe(0), 0);
```
msmanoj: Is this workaround similar to the following?
`// ugly hack: fall back to the end of event queue, giving priority to useEffect and useState
	      await new Promise((resolve) => setTimeout(resolve, 0));`

azangru: Yep :-)

msmanoj: Cool! I was wondering if it's better to keep it consistent so that it's easier to find and update these once enzyme starts to support it.

azangru: Updated for consistency.

_(Also, I have a nagging suspicion that this particular implementation was not even working properly. I suspect `jest` exited this test before the `expect` was called. But I didn't check to make sure that was the case.)_

src/ensembl/src/content/Content.tsx
```diff
@@ -15,29 +14,42 @@ type StateProps = {
...
 
-type ContentProps = RouteComponentProps & StateProps & OwnProps;
+type ContentProps = StateProps & OwnProps;
 
-export const getHeightClass = (launchbarExpanded: boolean): string => {
+const getHeightClass = (launchbarExpanded: boolean): string => {
   return launchbarExpanded ? styles.shorter : styles.taller;
 };
 
+const InnerContent = () => (
```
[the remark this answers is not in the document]
azangru: Renamed `InnerContent` to `ContentRoutes`

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -56,53 +57,45 @@ type OwnProps = {
...
-    }
+  const { navigator, reset } = browserInfoConfig;
+  const [showBrowserInfo, toggleShowBrowserInfo] = useState(
+    shouldShowBrowserInfo()
+  );
 
-    return classNames;
+  const setBrowserInfoVisibility = () => {
+    const shouldToggleVisibility = showBrowserInfo !== shouldShowBrowserInfo();
+    shouldToggleVisibility && toggleShowBrowserInfo(!showBrowserInfo);
```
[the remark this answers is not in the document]
azangru: Put the condition in an if-statement, though it was comfortable where it was and didn't really want to go.

PR #21, opened 2019-03-25
Refactor Redux (especially the Browser reducer/actions)

imran-ebi:
This pull request does a few things:

- Slices the browser reducer into many other reducers
- Separates reducers, actions et al into their own files for better maintenance
 - Add a common data thunk function in a separate file. This will make it easier to reuse this thunk for data fetching.
- Add google analytics to actions that aren't tracked yet


azangru:
I don't like the word `object`; its generality, vagueness, ambiguity (what is an "object"?), and general overload with various programming concepts just blow my mind. What are these thingies, and is there a way we could narrow down on a more meaningful name?

imran-ebi:
@azangru I agree it is vague, however that's the term which was used in the previous prototype for genes, transcripts, etc. And probably is still known in the same way. Maybe we will need to discuss this with Andrea and Steve to come up with a new term for it?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -23,26 +23,28 @@ import {
...
   getBrowserOpenState,
-  getDrawerOpened,
   getBrowserNavOpened,
   getChrLocation,
   getGenomeSelectorActive,
-  getBrowserActivated,
-  getExampleObjects
+  getBrowserActivated
 } from './browserSelectors';
+import { getChrLocationFromStr, getChrLocationStr } from './browserHelper';
```
msmanoj: Would it be better if we rename these functions to be `getChromosomeLocationFromStr' and `getChromsomeLocationStr`? 
Using `chr` creates a bit of confusion between chromosome and character.

imran-ebi: It was already long as it was so that is why I wanted to shorten it a bit. For character won't we be usually using `char` instead?

msmanoj: It's just that I've come across the term `chr` being used as a shorthand for character in the past.
If `Chr` would always mean chromosome in our case, it's cool.

ens-st3: The Location is not always going to be on Chromosome coordinates. depends on how good the assembly is. For example http://www.ensembl.org/Gopherus_agassizii/Location/View?r=PPEB01000001.1:490852-804510

getLocationFromStr ?

ens-st3:
> @azangru I agree it is vague and ambiguous and I didn't like using it either. However, that's the term which was used in the previous Ember.js prototype for genes, transcripts, etc. And probably is still known in the same way. Maybe we will need to discuss this with Andrea and Steve to come up with a new term for it?

We do still talk about 'Object Viewer' but this is not set in stone. Is the code where this is used going to be generic and apply to all the object types we (will) have (Gene, Transcript, Variation, Regulation, Location and prob others) ?

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: Won't location become ambiguous and too generic when we have other kinds of locations, such as URL locations, to work with?

imran-ebi:
>  Is the code where this is used going to be generic and apply to all the object types we (will) have (Gene, Transcript, Variation, Regulation, Location and prob others) ?

Yes that's the intention

src/ensembl/src/content/app/browser/Browser.tsx
ens-st3: getGenomeLocation...?

ens-st3:
> > Is the code where this is used going to be generic and apply to all the object types we (will) have (Gene, Transcript, Variation, Regulation, Location and prob others) ?
> 
> Yes that's the intention

EnsemblObject rather than Object ?

imran-ebi:
> EnsemblObject rather than Object ?

That sounds much better 😀

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: That would do!

src/ensembl/src/dataThunk.ts
```diff
@@ -0,0 +1,38 @@
...
+            })
+          );
+
+          return response.json();
+        },
+        (error) => dispatch(asyncActions.failure(error))
+      )
+      .then((json) => dispatch(asyncActions.success(json)));
+  };
+};
```
azangru: I am very conflicted about this file.

On the one hand, part of what's going on here — like concatinating the host name with the pathname, doing the fetch itself, and then figuring out how to provide the response (i.e. that the json method has to be called) — seems to belong to a dedicated `apiService`. Especially since the request to GoogleAnalytics is then made not through our redux middleware (which means that this could also be a part of the service).

It also seems that we are putting data about individual response time in Google Analytics, while the proper place for it would be a monitoring service (which could also send alerts if responses start taking too long time).

Personally, I wouldn't even extract the thunk into a separate file. Especially considering that, if we extract the `apiService`, the code of the thunk itself becomes very small, and we might need all the flexibility it can offer, as in this example:

```
const someAsyncAction = () => async (dispatch, getState) => {
  dispatch(asyncAction.request());
  const someStateValie = someSelector(getState());
  try {
    const response = await apiService.fetch(endpoint, options);
    const modifiedResponse = modify(response);
    dispatch(asyncActions.success(modifiedResponse));
  } catch (error) {
    dispatch(asyncActions.failure(error))
  }
}
```

imran-ebi:
I have renamed `Object` to `EnsObject` with ba4173d

src/ensembl/src/dataThunk.ts
imran-ebi: The main reason why I extracted this code into a separate thunk file was that it was tightly coupled with the analytics. However, if we use a monitoring service for APIs rather than analytics (which I also think is much better), then the code can be easily put into an API service instead.

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: I think this deserves to have a pull request of its own, since the change will affect a lot of files and also is not related to the Redux code.

PR #22, opened 2019-03-25
Add production server for both HTTP and HTTPS

imran-ebi:
This PR should get rid of the need to always run on HTTPS which will not work for users who haven't run the `certify` command beforehand.


imran-ebi:
Thanks for the approvals. I'll merge this into `dev` branch.

PR #27, opened 2019-03-29
Make site-down page work with current node setup.

ens-ds23:
The site-down page needs a few tweaks to work with the way nodejs is configured on the 2020 site. These are all changes to static assets. Has been tested on the live site.

Adding a base url and moving assets into a sub-directory makes it simpler to write express middleware to redirect all requests *except those to site-down assets* to the site-down page.

This is a temporary fix to get it working until we have time to review the nodejs setup.


azangru:
The jpegs are pretty huge. If you run them through [tinypng](https://tinypng.com/), you can get about 50% off in size:

image

If you run them through [Squoosh](https://squoosh.app), you can shave off about 80% of their size:

image

_(Both screenshots — from testing with 3.jpg)_

ens-ds23:
I'm not in a position to judge the quality degradation after this has been applied. I'm happy to update the PR with updated files if provided, though.

ens-ds23:
I've pushed the smaller files and made some layout changes which Andrea suggested.

PR #31, opened 2019-04-03
Add RoundButton and BadgedButton shared components

msmanoj:
- Added the RoundButton and BadgedButton shared components
- Added the storybook entries

### DIFF
--- a/src/ensembl/src/shared/badged-button/BadgedButton.scss
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.scss
@@ -2,19 +2,19 @@
 
 .badgedButton {
   position: relative;
-  display: inline-flex;
+  display: inline-block;
 }
 
 .badgeDefault {
   position: absolute;
   background-color: $ens-green;
   height: 26px;
-  width: 26px;
+  min-width: 26px;
   top: -13px;
   right: -13px;
   padding: 2px 0 0 0;
   color: $ens-white;
-  border-radius: 100%;
+  border-radius: 13px;
   text-align: center;
   cursor: pointer;
 }
--- a/src/ensembl/src/shared/badged-button/BadgedButton.scss
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.scss
@@ -5,7 +5,7 @@
   display: inline-flex;
 }
 
-.badge {
+.badgeDefault {
   position: absolute;
   background-color: $ens-green;
   height: 26px;
@@ -16,4 +16,5 @@
   color: $ens-white;
   border-radius: 100%;
   text-align: center;
+  cursor: pointer;
 }
--- a/src/ensembl/src/shared/badged-button/BadgedButton.scss
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.scss
@@ -1,0 +1,19 @@
+@import 'src/styles/common';
+
+.badgedButton {
+  position: relative;
+  display: inline-flex;
+}
+
+.badge {
+  position: absolute;
+  background-color: $ens-green;
+  height: 26px;
+  width: 26px;
+  top: -13px;
+  right: -13px;
+  padding: 2px 0 0 0;
+  color: $ens-white;
+  border-radius: 100%;
+  text-align: center;
+}
--- a/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
@@ -34,7 +34,7 @@ describe('BadgedButton', () => {
   });
 
   it('assigns the "badgeDefault" class to the badge by default', () => {
-    const renderedBadge = wrapper.find('div.badgeDefault');
+    const renderedBadge = wrapper.find('.badgeDefault');
     expect(renderedBadge).toHaveLength(1);
   });
 
@@ -46,7 +46,7 @@ describe('BadgedButton', () => {
       children: <Button onClick={onClick}>{faker.lorem.words()}</Button>,
       className: fakeClassName
     });
-    const renderedBadge = wrapper.find('div.badgeDefault');
+    const renderedBadge = wrapper.find('.badgeDefault');
 
     expect(renderedBadge.hasClass(fakeClassName)).toBe(true);
   });
@@ -57,7 +57,7 @@ describe('BadgedButton', () => {
       children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
     });
 
-    const renderedBadge = wrapper.find('div.badgeDefault');
+    const renderedBadge = wrapper.find('.badgeDefault');
 
     expect(renderedBadge.text()).toBe('abc');
   });
@@ -68,7 +68,7 @@ describe('BadgedButton', () => {
       children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
     });
 
-    const renderedBadge = wrapper.find('div.badgeDefault');
+    const renderedBadge = wrapper.find('.badgeDefault');
 
     expect(renderedBadge.text()).toBe('99+');
   });
--- a/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import faker from 'faker';
+import BadgedButton from './BadgedButton';
+import Button from '../button/Button';
+
+const onClick = jest.fn();
+
+const defaultProps = {
+  badge: faker.lorem.words()
+};
+
+const renderButton = (
+  ButtonComponent: React.FunctionComponent<any>,
+  props: any = defaultProps
+) => mount(<ButtonComponent {...props} />);
+
+describe('BadgedButton', () => {
+  let wrapper: any;
+
+  beforeEach(() => {
+    wrapper = renderButton(BadgedButton, {
+      ...defaultProps,
+      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
+    });
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('renders the passed in button', () => {
+    expect(wrapper.find('button').length).toEqual(1);
+  });
+
+  it('assigns the "badgeDefault" class to the badge by default', () => {
+    const renderedBadge = wrapper.find('div.badgeDefault');
+    expect(renderedBadge).toHaveLength(1);
+  });
+
+  it('extends the badge class', () => {
+    const fakeClassName = faker.lorem.word();
+
+    const wrapper = renderButton(BadgedButton, {
+      ...defaultProps,
+      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>,
+      className: fakeClassName
+    });
+    const renderedBadge = wrapper.find('div.badgeDefault');
+
+    expect(renderedBadge.hasClass(fakeClassName)).toBe(true);
+  });
+});
--- a/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
+++ b/src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
@@ -50,4 +50,26 @@ describe('BadgedButton', () => {
 
     expect(renderedBadge.hasClass(fakeClassName)).toBe(true);
   });
+
+  it('trims the longer strings to three characters', () => {
+    const wrapper = renderButton(BadgedButton, {
+      badgeContent: 'abcd',
+      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
+    });
+
+    const renderedBadge = wrapper.find('div.badgeDefault');
+
+    expect(renderedBadge.text()).toBe('abc');
+  });
+
+  it('formats number greater than 99 to "99+"', () => {
+    const wrapper = renderButton(BadgedButton, {
+      badgeContent: 100,
+      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
+    });
+
+    const renderedBadge = wrapper.find('div.badgeDefault');
+
+    expect(renderedBadge.text()).toBe('99+');
+  });
 });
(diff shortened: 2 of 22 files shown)


azangru:
Out of two options:

1) Having a Button component, a Badge component, and a component combining the two:

image

vs:

2) Having a Button component and a button-with-a-badge component:

image

why, do you think, the second option is preferable?

src/ensembl/src/shared/badged-button/BadgedButton.scss
```diff
@@ -0,0 +1,20 @@
+@import 'src/styles/common';
+
+.badgedButton {
+  position: relative;
+  display: inline-flex;
```
azangru: Why do we want flex here?

src/ensembl/src/shared/badged-button/BadgedButton.scss
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.badgeDefault {
+  position: absolute;
+  background-color: $ens-green;
+  height: 26px;
+  width: 26px;
```
azangru: Scary. How many characters will fit into this badge? Does Andrea offer any suggestions about what the badge should look like if there are several characters here? Is it possible that some kind of a rounded square (which can safely grow in width) will be a safer shape for the badge than a circle?

src/ensembl/src/shared/badged-button/BadgedButton.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import defaultStyles from './BadgedButton.scss';
+import classNames from 'classnames';
+
+type Props = {
+  children: React.ReactChild;
+  badge?: string | number | undefined;
```
azangru: If the component explicitly says in its name that it is a BadgedButton, then why should the badge field be optional (or its value be undefined)?

src/ensembl/src/shared/round-button/RoundButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+export enum RoundButtonStatus {
+  ACTIVE = 'active',
+  INACTIVE = 'inactive',
+  DISABLED = 'disabled',
+  DEFAULT = 'default'
+}
+
+type Props = {
+  onClick: () => void;
+  buttonStatus: RoundButtonStatus;
```
azangru: just `status`; since the component itself is called a button, it is obvious that it will be the status of this button

src/ensembl/src/shared/badged-button/BadgedButton.scss
msmanoj: At the moment it fits in three characters (eg. 999).
If I set `min-width: 26px` and `border-radius: 13px` it seems to expand safely. 
I'll check with Andrea if we need to allow more than 3 characters and make an update.

src/ensembl/stories/shared-components/badged-button/BadgedButton.stories.scss
```diff
@@ -0,0 +1,21 @@
...
+  padding: 40px;
+}
+
+.active {
+  border: none;
+  background-color: $ens-dark-grey;
+  color: $ens-white;
+}
+
+.imageButtonWrapper {
```
azangru: why **image**Button?

src/ensembl/src/shared/badged-button/BadgedButton.tsx
msmanoj: It is optional because a value for the badge might not always be available. Let's take unread messages count for example, it will have a value set only when there are some unread messages.

src/ensembl/stories/shared-components/badged-button/BadgedButton.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ImageButtonStatus
+} from 'src/shared/image-button/ImageButton';
+
+const onClick = action('button-click');
+
+storiesOf('Components|Shared Components/BadgedButton', module)
+  .add('badged Button', () => (
+    <div className={styles.wrapper}>
+      <BadgedButton badge={':)'}>
+        <SecondaryButton onClick={onClick}>Primary button</SecondaryButton>
```
azangru: Just to avoid confusion, perhaps the text inside the secondary button should not say "Primary button"? 😄

src/ensembl/stories/shared-components/badged-button/BadgedButton.stories.scss
msmanoj: I'm trying to display a story with all the available buttons to make use of BadgedButton.

`imageButtonWrapper` styling is just used in the story for imageButton.

src/ensembl/src/shared/round-button/RoundButton.tsx
msmanoj: That's done.

src/ensembl/stories/shared-components/badged-button/BadgedButton.stories.tsx
msmanoj: Haha 😄 
Yes, I've update it.

src/ensembl/src/shared/badged-button/BadgedButton.scss
msmanoj: Not sure why I tried to use flex instead of `inline-block`. I've updated it.

src/ensembl/src/shared/badged-button/BadgedButton.tsx
azangru: Fair point.

There are two other alternatives:

a) Choose in the parent component which button to show — either the one with a badge, or the one without, based on some conditions
b) Use the `showBadge: boolean` prop.

Alternative (a) is probably too cumbersome; but alternative (b) might actually be better than the current implementation. Especially given that the jsx:

`{props.badge && <div className={className}>{props.badge}</div>}`

will show a "0" (in a wrong place, too) if there are 0 unread messages, while you will probably not want to show any badge at all.

src/ensembl/stories/shared-components/badged-button/BadgedButton.stories.scss
azangru: Ah, I see.

msmanoj:
Doing it this way, the syntax will be:
<BadgedButton>
	<Button/>
<BadgedButton>

Would it be as simple as this if we do it the other way?

azangru:
```
const button = <Button>Something</Button>
const badge = <Badge>lol</Badge>

<BadgedButton
  button={button}
  badge={badge}
/>
```
This way looks simple as well.

msmanoj:
It does but the current structure does not require us to define two additional variables.

azangru:
On the other hand, it leaves us with only one type of badge — green circle `¯\_(ツ)_/¯`

src/ensembl/src/shared/badged-button/BadgedButton.tsx
msmanoj: > will show a "0" (in a wrong place, too) if there are 0 unread messages

I think `props.badge` would return false if the badge value is 0 and not render the badge.

Do we still require an additional parameter when the parent can actually control wether to show the badge or not, just by setting the right value?

msmanoj:
As discussed, the badge can be customised using the className prop.

src/ensembl/src/shared/badged-button/BadgedButton.tsx
azangru: > I think props.badge would return false if the badge value is 0 and not render the badge.

Try it 😄 
if `props.badge` is 0, it will return 0; and since 0 is falsy, it will not evaluate the second part of the expression and render 0. At least that's what it _should_ do.

azangru: On a related note, if `badge` prop is not a component, but something that will be placed inside a badge, then perhaps the `badgeContent` name will be more suitable?

msmanoj: Ahh! yes, that's right. I thought of it like a standard if condition. 
I could alternatively change it to:
`{props.badge? <div className={className}>{props.badge}</div> : null}`

azangru: What if you want to display the 0? Is this something we need to consider?

src/ensembl/src/shared/round-button/RoundButton.tsx
msmanoj: That's why we allow the type string. We can use '0' to display 0 and also ' ' to display an empty badge.

src/ensembl/src/shared/badged-button/BadgedButton.tsx
msmanoj: > What if you want to display the 0? Is this something we need to consider?

That's why we allow the type string. We can use '0' to display 0 and also ' ' to display an empty badge.

> Otherwise, just casting 0 to boolean (e.g. Boolean(props.badge), or the slightly less explicit !!props.badge) will work

Yes, this seems like a better solution.

msmanoj: I thought` badge` will be simpler but `badgeContent` seems to make it clear. I've updated it.

azangru: > That's why we allow the type string. We can use '0' to display 0 and also ' ' to display an empty badge.

Erm. I don't know if we want to be able to display a `0` in the badge (and if we don't, then we don't care); but just in case we do, what will the code look like?

```
<BadgedButton
  badgeContent={props.messageCount}  
>
 <Button>Something</Button>
</BadgedButton>
```

If `props.messageCount` is a number, then 0 will be a number (and will not show up). So if we want it to show up, will we to cast `props.messageCount` to string like this?

```
<BadgedButton
  badgeContent={`${props.messageCount}`}  
>
 <Button>Something</Button>
</BadgedButton>
```

msmanoj: I don't think there will be many cases where we'll be displaying a 0. If we need to do in some cases, I guess it's better to leave it to the parent component to pass it as a string instead. 
So yes, we will have to cast it to a string.

src/ensembl/src/shared/badged-button/BadgedButton.scss
msmanoj: As discussed with Steve and Andrea, we will not be having more than 3 characters displayed inside the badge.

src/ensembl/src/shared/badged-button/BadgedButton.tsx
```diff
@@ -11,12 +11,21 @@ type Props = {
 const BadgedButton = (props: Props) => {
   const className = classNames(defaultStyles.badgeDefault, props.className);
 
+  let badgeContent = props.badgeContent;
+
+  if (typeof badgeContent === 'number') {
+    if (badgeContent > 99) {
```
azangru: Can this condition go inside the previous one (`typeof badgeContent === 'number' && badgeContent > 99`) to avoid the nested if-statement?

src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
```diff
@@ -50,4 +50,26 @@ describe('BadgedButton', () => {
...
     expect(renderedBadge.hasClass(fakeClassName)).toBe(true);
   });
+
+  it('trims the longer strings to three characters', () => {
+    const wrapper = renderButton(BadgedButton, {
+      badgeContent: 'abcd',
+      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
+    });
+
+    const renderedBadge = wrapper.find('div.badgeDefault');
```
azangru: Here and elsewhere: is there any reason to use tag name in the CSS selector? It looks entirely redundant.

msmanoj: There will be multiple div elements inside this wrapper but I want to select only one in specific.
Is there a better way to do it?

src/ensembl/src/shared/badged-button/BadgedButton.tsx
msmanoj: That's done.

src/ensembl/src/shared/badged-button/BadgedButton.test.tsx
azangru: `wrapper.find('.badgeDefault')`?

msmanoj: Would it not be better to be a bit more specific by using `div.badgeDefault`?

azangru: Why?

It's good to be specific when there are different competing elements with the same level of CSS specificity and you need to select a particular one of those. But if we expect that there is only a single element with the class `badgeDefault `, then there is no point in increasing specificity of its CSS selector by adding its tag name.

msmanoj: We know that it's just going to be one element and also a div tag. I thought it will make the selection a little bit faster by specifying the tag it needs to look for. I'm aware that there will not be a big difference with just one test, but if we have 50+ tests using CSS selectors, I'm sure it will make the tests to run a little faster.

azangru: I would be curious to see whether there will be any detectable performance difference at all when using different kinds of CSS selectors during testing; and how many DOM reads there should be in order for the difference to be noticeable. It's the first time I've heard that concern.

In any case, we know that Jest runs its test suites (test files) in parallel.

What I've observed among CSS developers is that they prefer to use only classes by default. Extra selectors are added only when a class is not specific enough to do the job. Don't know if anyone has codified this preference in some kind of a style guide.

azangru: Ah! This has been reflected in some style guides after all!

Example:

image
(https://google.github.io/styleguide/htmlcssguide.html#Type_Selectors)

msmanoj: Makes more sense that `ul#example` filters `#example` first and then looks for `ul`. In our case, assuming that jest will select `.badgeDefault` at first and then checks if it belongs to a `div`, it would be better to remove the div from the selector to make it faster.

PR #32, opened 2019-04-04
Refactor Launchbar

azangru:
- [x] Get rid of the config file (launchbar is fairly simple and can be easily described by JSX alone)
- [x] Use inline svgs for icons, and style them accordingly
- [x] Use ReactRouter's `NavLink` to indicate whether the user has navigated to the route of a given app
- [x] Rename `LaunchbarIcon` component to `LaunchbarButton`
- [x] Delete tests for Launchbar (they were not particularly useful)


imran-ebi:
I notice the tooltips are gone. Can you add them back?

azangru:
Put the tooltips back. Thanks for noticing.

PR #35, opened 2019-04-05
Add Checkbox component

msmanoj:
- Added the Checkbox shared component.


src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+import defaultStyles from './Checkbox.scss';
+
+export enum CheckboxStatus {
+  CHECKED = 'checked',
+  DISABLED = 'disabled',
+  UNCHECKED = 'unchecked'
+}
+
+type Props = {
+  onClick?: () => void;
```
azangru: 1) Why would this function be optional? The component is useless without it.
2) Let's better call it `onChange` (to imitate the native input type checkbox more closely)

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  status: CheckboxStatus;
+  classNames?: { [key in CheckboxStatus]: string };
+  label?: string;
+  labelClassName?: string;
```
azangru: labelClassName is useless without the label. If label is optional, then perhaps label and labelClassName should be combined in a single type (so that if there is no label, then the component should not accept the  labelClassName prop)?

```
type WithoutLabelProps = {
  status: ...,
  onChange: ...,
}

type WithLabelProps = WithoutLabelProps & {
  label: ...
  labelClassName?: ...
}

type Props = WithoutLabelProps | WithLabelProps;
```

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const labelClassName = classNames(
+    defaultStyles.defaultLabel,
+    props.labelClassName
+  );
+
+  return (
+    <div onClick={handleOnClick} className={defaultStyles.checkboxHolder}>
+      <div className={className} />
+      {props.label && <div className={labelClassName}>{props.label}</div>}
+    </div>
```
azangru: Perhaps better write this component using browser's native `label` and `input` tags?

(since `input` isn't properly styleable, it will have to be hidden via display: none, and styles will apply to a div/span element close to it, see e.g. discussion [here](https://stackoverflow.com/questions/4148499/how-to-style-a-checkbox-using-css))

imran-ebi: @azangru In the same thread, there is this answer: https://stackoverflow.com/a/32876109/8899157. It states that we should not use `display: none`. I think this is because of accessibility reasons. Maybe using negative `z-index` as suggested in that answer is a better approach?

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export enum CheckboxStatus {
+  CHECKED = 'checked',
+  DISABLED = 'disabled',
+  UNCHECKED = 'unchecked'
+}
+
+type Props = {
+  onClick?: () => void;
+  status: CheckboxStatus;
```
azangru: While I understand the appeal of an enum, I think here it would be better to use two props: `checked: boolean` and `disabled?: boolean`, again to make the behaviour of the component closer in line with behaviour of the native input type checkbox.

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // Toggle the checked status
+    if (checkedStatus === CheckboxStatus.CHECKED) {
+      toggleCheckedStatus(CheckboxStatus.UNCHECKED);
+    } else {
+      toggleCheckedStatus(CheckboxStatus.CHECKED);
+    }
+
+    // Call the onClick function if we have one
+    if (props.onClick) {
+      props.onClick();
```
azangru: in line with my other comments, this will be `props.onChange(!props.checked)`.

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: > Why would this function be optional? The component is useless without it.
We don't have to worry about passing an onChange function if it is disabled?

> Let's better call it onChange (to imitate the native input type checkbox more closely)
Cool, will update it.

azangru: > We don't have to worry about passing an onChange function if it is disabled?

Usually the value of the `disabled` prop will be computed at render time and passed to the component along with other properties that it needs:

```
<Checkbox
  checked={props.something}
  onChange={props.someFuncton}
  disabled={someCondition}
/>
```
which, typically, means that `onChange` is going to be passed to the component anyway, so this shouldn't be a problem. Curious whether you've noticed other patterns.

_(The `Input` component currently has `onChange` as a mandatory prop)_

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: That's all done.

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: That's done.

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: That's done. 
TypeScript is not happy for some reason as it says:
```
Property 'label' does not exist on type 'Props'.
  Property 'label' does not exist on type 'WithoutLabelProps'.
```

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: > which, typically, means that onChange is going to be passed to the component anyway

Yes, that's right. It's always going to be passed regardless of the state. I'll set it to be mandatory.

src/ensembl/src/shared/checkbox/Checkbox.tsx
imran-ebi: @msmanoj That is because `WithLabelProps` may not exist. You will need add some checks before using `label` and `labelClassName`.

imran-ebi: This should be resolved by 48819b6

imran-ebi:
@msmanoj @azangru Seems like all the issues of this PR have been resolved now. So can we merge this to `dev`?

src/ensembl/src/shared/checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    it('calls the onChange function if the label is clicked', () => {
+      const label = faker.lorem.words();
+      const wrapper = mount(<Checkbox label={label} onChange={onChange} />);
+
+      wrapper.find('.defaultLabel').simulate('click');
+
+      expect(onChange).toHaveBeenCalled();
+    });
```
azangru: Ideally, three tests above would check what was passed to the `onChange` function (`toHaveBeenCalledWith(stuff)`)

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [checkedStatus, toggleCheckedStatus] = useState(props.checked);
+
+  const handleOnChange = () => {
+    // Do nothing if it is disabled
+    if (props.disabled) {
+      return;
+    }
+
+    // Toggle the checked status
+    toggleCheckedStatus(!checkedStatus);
```
azangru: If the `checked` prop is coming from the props, is there any point in keeping a parallel `checkedStatus` value in the state? Why not just rely on the `checked` value from the props?

src/ensembl/stories/shared-components/checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <div className={styles.wrapper}>
+        <Checkbox disabled={true} onChange={action('checkbox-toggled')} />
+      </div>
+    );
+  })
+  .add('with label', () => {
+    return (
+      <div className={styles.wrapper}>
+        <Checkbox
+          label={'I am a lable'}
```
azangru: lable -> label

src/ensembl/src/shared/checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }).not.toThrow();
+  });
+
+  it('applies the style "defaultCheckbox" by default', () => {
+    const wrapper = mount(<Checkbox onChange={onChange} />);
+    expect(wrapper.find('.defaultCheckbox')).toHaveLength(1);
+  });
+
+  describe('prop label', () => {
+    it('does not display any lable by default', () => {
```
azangru: lable -> label

src/ensembl/src/shared/checkbox/Checkbox.tsx
msmanoj: The `checked` parameter is optional so it may not always be passed from the parent component.

src/ensembl/src/shared/checkbox/Checkbox.test.tsx
msmanoj: That's done.

src/ensembl/stories/shared-components/checkbox/Checkbox.stories.tsx
msmanoj: That's done.

src/ensembl/src/shared/checkbox/Checkbox.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/checkbox/Checkbox.tsx
```diff
@@ -58,7 +53,7 @@ const Checkbox = (props: Props) => {
         type="checkbox"
         className={defaultStyles.hiddenInput}
         onChange={handleOnChange}
-        defaultChecked={checkedStatus}
+        defaultChecked={props.checked}
```
azangru: `defaultChecked={props.checked}` -> `value={props.checked}`

msmanoj: I did try that before but I was getting the following error:

> Type 'boolean' is not assignable to type 'string | number | string[] | undefined'.

msmanoj: `checked={props.checked}` works fine. I've updated it.

azangru: Right. Makes sense.

PR #36, opened 2019-04-08
update ADRs for state management and unit tests

imran-ebi:


azangru:
Could you rebase this PR on the current `dev`? Failing tests suggest that you've branched off quite a while ago. I know it shouldn't matter, but still.

azangru:
Perhaps ADRs should also include alternative solutions that we considered but eventually rejected; so that it's clearer why we chose a particular option. For instance, what is it that we like about Enzyme as opposed to react-testing-library? What is it that we like about redux as opposed to whatever else was considered?

imran-ebi:
> Could you rebase this PR on the current `dev`? Failing tests suggest that you've branched off quite a while ago. I know it shouldn't matter, but still.

That is done now.

>Perhaps ADRs should also include alternative solutions that we considered but eventually rejected; so that it's clearer why we chose a particular option. For instance, what is it that we like about Enzyme as opposed to react-testing-library? What is it that we like about redux as opposed to whatever else was considered?

I'll update the ADRs with them.

src/ensembl/doc/architecture/decisions/0004-unit-tests.md
```diff
@@ -16,7 +16,7 @@ Write unit tests using Jest and Enzyme.
 
 [Jest](https://jestjs.io/) is used as the main testing framework.
 
-[Enzyme](https://airbnb.io/enzyme/) is used as a utility of Jest to test React components.
+[Enzyme](https://airbnb.io/enzyme/) is used as a utility of Jest to test React components. Also, `react-testing-library` was briefly discussed as an alternative. It forces the developer to test the behaviour of React components and limits what can be tested. This means certain aspects of the component cannot be tested (only shallow rendering is possible). Due to this limitation, it was decided not to use this library for testing.
```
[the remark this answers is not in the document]
imran-ebi: I have updated the ADR with the suggested text

src/ensembl/doc/architecture/decisions/0003-state-management.md
```diff
@@ -14,8 +14,14 @@ The state of data and UI will grow in the application as the development goes on
 
 [Redux](https://redux.js.org/) will be used for the state management. It is a centralised state management container, that can handle the state for data and UI of the application.
 
+Using React.js state only, instead of any state management library, was considered. However, this would make it difficult to share state between sibling components and manage global state as well.
+
+Also, Mobx was discussed about. However, the magic nature of how things are done with it compared to Redux made the team decide not to go ahead with it.
```
azangru: My attempt:

There are various ways of state management in a React application. React itself provides two mechanisms:

- component's local state (useful for tracking transient localised state that most parts of the application are not concerned with)
- React context (useful for sharing state between unrelated components)

Both of these mechanisms, while useful, are generally insufficient for tracking state in a large application, which resulted in proliferation of numerous third-party libraries for state management. Of those, two most notable are:

- redux
- mobX

These two libraries represent two different philosophies of state management (and programming in general). Redux follows the functional programming approach, in which state itself is immutable, and state changes are a series of new copies of state produced in response to message events ("actions") dispatched within the store. In MobX, state is a mutable object that is changed via getters and setters (in latest implementations, proxies), which also notify subscriptions of the change.

While MobX's approach is generally regarded as easier to use, we feel that it also makes this library more "magical" and therefore more difficult to debug as compared to Redux. Also, Redux's reliance on events makes it a convenient target for plugging various kinds of middleware (i.e. analytics middleware, redux-saga or redux-observable) for better encapsulation of state-dependent logic.

imran-ebi: Thanks. This looks great. I'll add it to the ADR.

src/ensembl/doc/architecture/decisions/0004-unit-tests.md
```diff
@@ -12,18 +12,22 @@ When changes are made after the initial implementation of components, there is a
...
 ## Decision
 
-Write unit tests using Jest and Enzyme. Include snapshot testing too, for additional testing.
+Write unit tests using Jest and Enzyme.
 
-[Jest](https://jestjs.io/) is used as the main testing framework. Snapshot testing, which is a feature of Jest, is also used.
+[Jest](https://jestjs.io/) is used as the main testing framework and [`ts-jest`](https://kulshekhar.github.io/ts-jest/) is used as the TypeScrpit preprocessor to run them..
 
-[Enzyme](https://airbnb.io/enzyme/) is used as a utility of Jest to test React components.
+[Enzyme](https://airbnb.io/enzyme/) is used as a utility of Jest to test React components. A recent and promising alternative library for unit-testing React is react-testing-library. Its philosophy differs from Enzyme in that it offers the developer a very limited API and encourages testing of components only after they have been mounted in the DOM.
```
azangru: One last remark is that the first sentence in this paragraph says that Enzyme is a utility of Jest, which isn't technically correct. Enzyme is just a library for testing React components, and Jest is just a test runner coupled with an assertion library (based on Jasmine, if I am not mistaken), and they don't really care about one another.

imran-ebi: Ah, I altered what was mentioned in the Enzyme website, but added Jest into the mix and made it wrong. I'll change this to be something meaningful.

imran-ebi: Made the changes in 469d5c8

PR #37, opened 2019-04-10
Add Select component

azangru:
**Component's behaviour:**
- [x] Can accept an array of options or an array of groups of options
- [x] Selected option is sorted to the top of the list and is not selectable any longer
- [x] Select options can be navigated using arrow keys (as well as, obviously, by hovering/clicking with a mouse). Pressing `enter` will submit the current option. The current option is highlighted.
- [x] If the options list is too tall and does not fit the part of the screen to the bottom of the closed select element, the options list will be scrollable. Native scrollbars will be hidden, but the list will show up and down arrows indicating that there are more options than what is currently visible. Mousing over an arrow will start scrolling in the direction where the arrow is pointing. Leaving the mouse from the arrow will stop scrolling.
- [x] Make sure that, when navigating the options using the keyboard, the highlighted option is always visible (i.e. when an option is selected using the keyboard, it always gets scrolled into view if needed).

[Design reference](https://xd.adobe.com/view/46903a85-c5d9-43ad-6bb2-9b7e3a2b098e-30a1/screen/a43c11b6-baeb-4db9-851e-5a3068f18685/label-list-8)

**TODO (for subsequent pull requests):**
- Disable `onMouseEnter` for options when the user is navigating the options using keyboard (probably — debounce all onMouseEnters within, say, half a second after a key press). This should solve the current problem when an option under the cursor gets highlighted immediately after keyboard navigation causes the options list panel to scroll.
- Extend keyboard-based behaviour (open the options panel when pressing `down` or `space` key; confirm selection by pressing the `space` key) 
- Add appropriate behaviour for disabled options (making use of the `isDisabled` filed in the option data structure)?
- Find an option when the user starts typing in its text (there should be a brief interval within which every subsequent keypress is treated as being a part of the same string that begins a word, and after which a new keypress will be treated as starting a new string)

### DIFF
--- a/src/ensembl/src/shared/select/Select.tsx
+++ b/src/ensembl/src/shared/select/Select.tsx
@@ -48,11 +48,11 @@ type CommonProps = {
 };
 
 type OptionsSelectProps = CommonProps & OptionsSpecificProps;
-type OptionGroupssSelectProps = CommonProps & OptionGroupsSpecificProps;
+type OptionGroupsSelectProps = CommonProps & OptionGroupsSpecificProps;
 
-type SelectAdapterProps = OptionsSelectProps | OptionGroupssSelectProps;
+type SelectAdapterProps = OptionsSelectProps | OptionGroupsSelectProps;
 
-type SelectProps = OptionGroupssSelectProps & { placeholder: string };
+type SelectProps = OptionGroupsSelectProps & { placeholder: string };
 
 // TODO: think of a better name for this component?
 const ClosedSelect = (props: ClosedSelectProps) => {
@@ -158,8 +158,8 @@ Select.defaultProps = {
 // the purpose of the adapter is to unify props
 // to be consumed by the Select component
 const SelectAdapter = (props: SelectAdapterProps) => {
-  if ((props as OptionGroupssSelectProps).optionGroups) {
-    return <Select {...props as OptionGroupssSelectProps} />;
+  if ((props as OptionGroupsSelectProps).optionGroups) {
+    return <Select {...props as OptionGroupsSelectProps} />;
   } else {
     const { options, title, ...otherProps } = props as OptionsSelectProps;
     const optionGroups = [
--- a/src/ensembl/src/shared/select/Select.tsx
+++ b/src/ensembl/src/shared/select/Select.tsx
@@ -92,7 +92,7 @@ const Select = (props: SelectProps) => {
       {isOpen && (
         <SelectOptionsPanel
           optionGroups={props.optionGroups}
-          onSelect={(thing) => console.log('thing', thing)}
+          onSelect={handleSelect}
           onClose={closePanel}
         />
       )}
--- a/src/ensembl/src/shared/select/Select.tsx
+++ b/src/ensembl/src/shared/select/Select.tsx
@@ -1,7 +1,5 @@
 import React, { useState, useEffect } from 'react';
 
-import useDeepCompareEffect from 'src/shared/hooks/useDeepCompareEffect';
-
 import {
   findSelectedIndexForOptions,
   findSelectedIndexForOptionGroups,
@@ -30,7 +28,8 @@ export type GroupedOptionIndex = [
 ];
 
 type ClosedSelectProps = {
-  label: React.ReactNode;
+  selectedOption: Option | null;
+  placeholder: string;
   onClick: () => void;
 };
 
@@ -45,6 +44,7 @@ type OptionGroupsSpecificProps = {
 
 type CommonProps = {
   onSelect: (value: any) => void;
+  placeholder: string;
 };
 
 type OptionsSelectProps = CommonProps & OptionsSpecificProps;
@@ -63,7 +63,7 @@ const arrowHead = (
 const ClosedSelect = (props: ClosedSelectProps) => {
   return (
     <span className={styles.selectClosed} onClick={props.onClick}>
-      {props.label}
+      {props.selectedOption ? props.selectedOption.label : props.placeholder}
       {arrowHead}
     </span>
   );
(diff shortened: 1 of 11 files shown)


src/ensembl/src/shared/select/Select.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+type OptionsSelectProps = CommonProps & OptionsSpecificProps;
+type OptionGroupsSelectProps = CommonProps & OptionGroupsSpecificProps;
+
+type SelectAdapterProps = OptionsSelectProps | OptionGroupsSelectProps;
+
+type SelectProps = OptionGroupsSelectProps & { placeholder: string };
+
+// TODO: think of a better name for this component?
+const ClosedSelect = (props: ClosedSelectProps) => {
```
msmanoj: Something like SelectController or SelectHolder ?

src/ensembl/src/shared/select/Select.tsx
```diff
@@ -0,0 +1,175 @@
...
+  };
+
+  const handleFocus = () => {
+    focusRef.current = true;
+  };
+
+  const handleBlur = () => {
+    focusRef.current = false;
+    if (isOpen) {
+      setIsOpen(false);
```
msmanoj: Would it be better to make use of `closePanel()` here and on the lines below?

src/ensembl/src/shared/select/Select.tsx
```diff
@@ -0,0 +1,175 @@
...
+Select.defaultProps = {
+  placeholder: 'Select'
+};
+
+// the purpose of the adapter is to unify props
+// to be consumed by the Select component
+const SelectAdapter = (props: SelectAdapterProps) => {
+  if ((props as OptionGroupsSelectProps).optionGroups) {
+    return <Select {...props as OptionGroupsSelectProps} />;
+  } else {
```
msmanoj: The `else` part could just be put after the `if` without {}.

imran-ebi: Or use an early return without using `else`?

src/ensembl/src/shared/select/SelectOptionsPanel.tsx
```diff
@@ -0,0 +1,307 @@
...
+      return getPreviousItemIndex(state, action.payload);
+    case HighlightActionType.SET:
+      return action.payload;
+    default:
+      return state;
+  }
+};
+
+const SelectOptionsPanel = (props: Props) => {
+  const [shouldShowTopScrollButton, setShouldShowTopScrollButton] = useState(
```
msmanoj: How about just `showTopScrollButton`?

src/ensembl/src/shared/select/SelectOptionsPanel.tsx
```diff
@@ -0,0 +1,307 @@
...
+      dispatch({ type: HighlightActionType.NEXT, payload: props.optionGroups });
+    } else if (event.keyCode === keyCodes.ENTER) {
+      highlightedItemIndexRef.current &&
+        props.onSelect(highlightedItemIndexRef.current);
+    }
+
+    // scroll option into view after the state gets updated
+    setTimeout(() => {
+      if (!optionsListRef.current) {
+        // keypress has caused the options panel to close
```
msmanoj: Does this block gets executed only when `enter` key is pressed? If yes, I think we could use a `return` within the previous `if (event.keyCode === keyCodes.ENTER) {..` block and remove this check.

src/ensembl/src/shared/select/Select.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    if (event.keyCode === keyCodes.ENTER) {
+      setIsOpen(true);
+    } else if (event.keyCode === keyCodes.ESC) {
+      setIsOpen(false);
+    }
+  };
+
+  const handleSelect = (optionIndex: GroupedOptionIndex) => {
+    const [groupdIndex, itemIndex] = optionIndex;
```
imran-ebi: Is `groupdIndex` intentionally named like that or is it a typo?

azangru: That's a typo; should be groupIndex.

src/ensembl/src/shared/select/Select.tsx
azangru: I checked what they call this component in React-Select, and it is called `Control` there:

image

I am not sure I like that name, but it may be better than ClosedSelect.

src/ensembl/src/shared/select/Select.tsx
azangru: Yes to `closePanel`.
I think in this component I prefer the granularity of `openPanel`/`closePanel` to `togglePanel`

src/ensembl/src/shared/select/SelectOptionsPanel.tsx
azangru: This is a Typescript thing. `optionsListRef` is defined above as `HTMLDivElement | null`. If we move the `return` statement into the `Enter` branch, we will still need to convince Typescript that `optionsListRef.current` is not empty. We can do that, of course, by telling Typescript that we know better than it (`const optionsListElement = optionsList.current as  HTMLDivElement`), but at the moment Typescript, together with the if-statement, are doing this for us, and are safer than any manual override.

src/ensembl/src/shared/select/Select.tsx
imran-ebi: @azangru Is this going to be changed or stay as it is?

imran-ebi: @azangru Sorry, I checked the wrong file commit. The latest commit has this changed.

PR #38, opened 2019-04-12
Update packages and unignore package-lock.json

azangru:
- [x] Update packages (primarily because `react-redux` got updated to version 7, in which, they say, they resolved performace issues; but also updated other packages)
- [x] Unignore `package-lock.json`


.gitignore
```diff
@@ -12,7 +12,6 @@ built-docs/
 .DS_Store
 .vstags
 
-package-lock.json
```
imran-ebi: I thought this was ignored already. I wonder how it got back 🤔

imran-ebi:
@ens-ma7 Can you check if the package installation causes any problems in your Windows machine?

imran-ebi:
I just tested this on my machine and did not come across any issues

.gitignore
imran-ebi: As discussed with @azangru offline, we need to this file for continuous integration (CI).

PR #39, opened 2019-04-17
Replace TSLint with ESLint

imran-ebi:
ESLint has recently added support, hence TSLint is now in maintenance mode only. It is only a matter of time before the latter gets deprecated. This PR removes TSLint and adds ESLint for Ensembl 2020 frontend.

Links to the rules used for ESLint are:

- ESLint recommended: https://eslint.org/docs/rules/
- TypeScript rules: https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin
- React: https://www.npmjs.com/package/eslint-plugin-react
- React hooks: https://www.npmjs.com/package/eslint-plugin-react-hooks

To activate ESLint for TypeScript linting, you will need to add the following configuration in VSCode's `settings.json` (ref: https://github.com/typescript-eslint/typescript-eslint/issues/326):

```
"eslint.validate": [
    "javascript",
    "javascriptreact",
    "typescript",
    "typescriptreact"
]
```


src/ensembl/.eslintrc.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  extends: [
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react/recommended',
+    'prettier/@typescript-eslint', // Uses eslint-config-prettier to disable rules that conflict with prettier
+    'plugin:prettier/recommended' // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
```
azangru: Do we even want to display `prettier` errors? Shouldn't prettier just format the code without making us even think about it?

_(Personally, I am not a fan of `prettier`, because I often find its formatted code harder to read than the original, but if we must use `prettier`, then perhaps it should just do its work quietly?)_

src/ensembl/webpack/webpack.common.js
```diff
@@ -31,7 +31,7 @@ module.exports = (isDev, moduleRules, plugins) => ({
       // babel-loader is used for typescript as it has good tree shaking support compared to tsc
       {
         test: /.tsx?$/,
-        loader: 'babel-loader',
+        use: ['babel-loader', 'eslint-loader'],
```
azangru: Doesn't this mean that code with linting errors will fail to even compile and run? If so, let's not?

src/ensembl/.eslintrc.js
imran-ebi: We can get rid of prettier if we think it is too intrusive. However, I think it is good to know that the format is a bit off in my code, though, it does format the code for us when we commit.

src/ensembl/webpack/webpack.common.js
imran-ebi: Ah, that's a good question. I just checked it and turns out it compiles but does not display any changes made to the code with HMR. However, works on a reload (probably means it compiled?).  Anyway, I agree this is not ideal and I'll try to find a better alternative to this.

src/ensembl/.eslintrc.js
imran-ebi: Turns out this actually helps to silence all those formatting errors that ESLint complains about 😁 I changed the value of `prettier/prettier` (all prettier rules) from `1` to `0` to off it and voila, no more formatting errors.

src/ensembl/webpack/webpack.common.js
azangru: Alternative: run eslint as part of pre-commit hook (with husky).

imran-ebi: This is now fixed by using the `emitWarning` option. ESLint will print these errors that don't break the code as warnings in the browser and the console. However, it will clearly state if they are errors or warnings.

azangru:
Unless you also want to fix all the linting errors/warning in this PR, I think it is good for merging.

imran-ebi:
@azangru have you already changed the underscore variables to the camel case?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerTranscript.tsx
```diff
@@ -52,8 +52,8 @@ const DrawerTranscript: FunctionComponent<DrawerTranscriptProps> = (
             <Fragment>
               <p>
                 MANE Select transcripts match GRCh38 and are 100% identical
-                between Ensembl-GENCODE and RefSeq for 5' UTR, CDS, splicing and
-                3' UTR.
+                between Ensembl-GENCODE and RefSeq for 5&rsquo; UTR, CDS,
+                splicing and 3&rsquo; UTR.
```
azangru: Aren’t actual Unicode symbols better than their representation in html entities?

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -11,7 +11,7 @@ type Props = {
 const SpeciesSearchField: FunctionComponent<Props> = () => {
   return (
     <div className={styles.speciesSearchField}>
-      It's a stub. See the feature/species-selector-screen branch for
+      It&rsquo;s a stub. See the feature/species-selector-screen branch for
```
azangru: :disappointed:

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerTranscript.tsx
imran-ebi: From a typography perspective it is good to be using HTML entities as some of the keyboard characters aren't the real characters that we need to be using. In this case the apostrophe is not the same as the right-side quote and they also look slightly different.

azangru: I would always just key in the appropriate symbol. Shift-option-right_square_bracket gives you an `’`, which is the proper apostrophe/closing quotation mark.

But if we start considering typographically appropriate characters (slight shudder), then `5&rsquo; UTR` is not ever right. It's not 5-apostrophe (as in `it’s`, unicode U+0027) or 5-single_quote (unicode U+2019), but rather 5-prime (5′, unicode U+2032) 😞

imran-ebi: In that case we'll need to use the correct character then 😄 I think we will need to have a list of these commonly used special characters written somewhere for our reference.

imran-ebi: Or maybe refer this document: https://brajeshwar.github.io/entities/

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
imran-ebi: This has been disabled

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerTranscript.tsx
imran-ebi: @azangru This has been disabled now

PR #40, opened 2019-04-18
Improve test suite speed

imran-ebi:
This PR improves the test speed slightly by removing Jest snapshot testing and also adding an option to the Jest commands: `--maxWorkers=1`. The speed improvement is roughly around 10% for running all the test cases, whereas in watch mode it was around 300% when tested with `Account.test.tsx`.


src/ensembl/package.json
```diff
@@ -27,9 +27,9 @@
     "storybook": "start-storybook -p 9001 -c .storybook",
     "deploy-docs": "./deploy-docs.sh",
     "check-types": "tsc",
-    "jest": "jest",
-    "test": "jest --verbose",
-    "test:watch": "jest --watch",
+    "jest": "jest --maxWorkers=1",
```
azangru: Why do we have this npm script anyway? Doesn't the current implementation of `npm test` make it redundant?

imran-ebi: I used it to get the console output as Jest didn't print them in verbose mode. Now that Jest doesn't have that issue anymore, it is redundant indeed.

src/ensembl/package.json
```diff
@@ -94,8 +94,8 @@
     "@types/storybook__addon-actions": "3.4.2",
     "@types/storybook__react": "4.0.1",
     "@types/webpack-env": "1.13.9",
-    "babel-core": "7.0.0-bridge.0",
     "babel-loader": "8.0.5",
+    "babel-plugin-dynamic-import-node": "2.2.0",
```
azangru: :smile:

src/ensembl/babel.config.js
```diff
@@ -16,5 +16,21 @@ module.exports = {
...
+        [
+          '@babel/env',
+          {
+            useBuiltIns: 'usage',
+            corejs: 3,
+          }
+        ]
+      ],
+      plugins: [
+        'babel-plugin-dynamic-import-node'
```
azangru: Won't we also need class properties and object-rest-spread here?

src/ensembl/jest.config.js
```diff
@@ -16,9 +16,7 @@ module.exports = {
   },
   preset: 'ts-jest',
```
azangru: I am still a bit confused though. If this PR still uses ts-jest, then what was the point in updating babel.conf? If, however, we switch to babel for tests, then shouldn't we also make changes here?

imran-ebi: Yes, we should remove `ts-jest`.

imran-ebi:
@azangru @msmanoj I have now switched to using `babel-jest` instead of `ts-jest`. Tested on my machine after fixing NPM issues via `npm audit fix`. The time taken to run the tests is now ~10s.

src/ensembl/package.json
```diff
@@ -27,11 +27,9 @@
     "storybook": "start-storybook -p 9001 -c .storybook",
     "deploy-docs": "./deploy-docs.sh",
     "check-types": "tsc",
-    "jest": "jest",
-    "test": "jest --verbose",
-    "test:watch": "jest --watch",
-    "coverage": "jest --coverage",
-    "snapshot": "jest --updateSnapshot"
+    "test": "jest --verbose --maxWorkers=1",
+    "test:watch": "jest --watch --maxWorkers=1",
```
azangru: Do we still care about the maxWorkers option? We added it for ts-jest, but now that we switch to babel-jest, is it still relevant?

imran-ebi: Yes, when I tested without it the test suite runs slower.

PR #44, opened 2019-04-26
Add prepush hook for master branch

imransl:
This pull request will introduce a `prepush` hook. It will check if the current branch is `master` using `git-branch-is` package. And if it is, then `git rebase dev` will be run to make sure `master` is not being pushed without the latest commits from `dev`.

As this is a WIP, any suggestions and improvements are welcome 😄


azangru:
I think we also need a pre-commit hook that checks that the name of the branch and does not allow direct commits into `master`.

imran-ebi:
> I think we also need a pre-commit hook that checks that the name of the branch and does not allow direct commits into `master`.

The latest commit (a71e2e4) adds this.

src/ensembl/package.json
```diff
@@ -39,7 +39,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "if git-branch-is master; then exit 1; else lint-staged; fi",
```
azangru: We probably also want to echo a warning; something like ("no committing into master") 😄 
I wonder if having dedicated shell files to run in these two hooks will be better than typing the script inline.

imran-ebi: Yes, we'll probably need to have a file for it if we are going to include more lines of code 😄

imran-ebi:
IIRC @ens-ds23 was working on a `git` hooks in one of the recent PRs that got merged. I am not sure if it had the prepush hook that I was trying to implement with this PR. If it did, then I think we can close this PR.

azangru:
The original idea of this PR, as far as I remember it, was to disallow developers to commit directly to master. I don't know if we still want anything like this; but this is not what Dan was working on. His work related not to hooking into the git lyfecycle, but to update the build script for the Rust project.

imran-ebi:
I remember that @ens-ds23 asked us the `git` hooks that we need and I told him about this one. Maybe it was for some other purpose.

azangru:
That might have been during the first attempt to remove built wasms out of the repo, which was to use git-lfs. Which turned out to be surprisingly messy — git-lfs was competing with husky for the same hooks. That's where the question about which hooks we use might have come from.

imransl:
We have not come across any incident where someone in the team unintentionally pushed to `master` since creating this PR. So we have been careful and this is not necessary anymore. So closing this PR.

PR #46, opened 2019-05-06
Feature/species tabs

azangru:
This PR adds a header bar that shows a list of selected species and treats them as tabs (i.e. clicking on a species item will "activate" it, which will be accompanied by some change in the interface). The view that needs this bar the most at the moment is Genome Browser.

### Technical implementation and problems
The current implementation of required behaviour (a single line of tabs that adapt to the width of the screen) is done with CSS only (using flexbox). The problem with this approach is that instead of making all tabs the same width in cases where their collective width is larger than the width of the container, tabs are shrunk by the same fraction (which is how flexbox apparently works). This results in some tabs being wider and other tabs being shorter. While we (Andrea and me) recognise this defect, we agree that it is acceptable to have this header behaviour until we come up with a better one.

The current behaviour of the header can be observed in the [deployed storybook](https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-speciestabbar--multiple-species).

See also [design mockup](https://xd.adobe.com/view/b0b60cbe-c41f-45e3-4e8e-3fd73d0090d5-c687/screen/aa57f3ec-6c3b-4729-8090-e3d139792a30/species-selector-8-on/) to get a picture of designer's thinking.

### DIFF
--- a/src/ensembl/src/shared/species-tab/SpeciesTab.scss
+++ b/src/ensembl/src/shared/species-tab/SpeciesTab.scss
@@ -11,6 +11,7 @@ $speciesTabPadding: 6px 20px;
   cursor: pointer;
   overflow: hidden;
   line-height: 1;
+  user-select: none;
   text-overflow: ellipsis;
   flex-grow: 0;
   flex-shrink: 1;
--- a/src/ensembl/src/shared/species-tab/SpeciesTab.scss
+++ b/src/ensembl/src/shared/species-tab/SpeciesTab.scss
@@ -9,6 +9,10 @@ $speciesTabPadding: 6px 20px;
   border: 1px solid $ens-blue;
   border-radius: $speciesTabBorderRadius;
   cursor: pointer;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  flex-shrink: 1;
+  flex-basis: 100%;
 
   &:not(:last-child) {
     margin-right: 1em;
@@ -27,6 +31,7 @@ $speciesTabPadding: 6px 20px;
 
 .speciesTabActive {
   background-color: $ens-black;
+  border: none;
 
   .name,
   .assembly {
@@ -35,4 +40,6 @@ $speciesTabPadding: 6px 20px;
 }
 
 .speciesTabFullSize {
+  flex-shrink: 0;
+  flex-basis: auto;
 }
--- a/src/ensembl/src/shared/species-tab/SpeciesTab.scss
+++ b/src/ensembl/src/shared/species-tab/SpeciesTab.scss
@@ -1,13 +1,10 @@
 @import 'src/styles/common';
 
-$speciesTabBorderRadius: 20px;
-$speciesTabPadding: 6px 20px;
-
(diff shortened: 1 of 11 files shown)


src/ensembl/src/shared/species-tab/SpeciesTab.scss
```diff
@@ -0,0 +1,49 @@
+@import 'src/styles/common';
+
+$speciesTabBorderRadius: 20px;
+$speciesTabPadding: 6px 20px;
+
```
imran-ebi: Are these two variables necessary?

src/ensembl/src/shared/species-tab/SpeciesTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+    genome_id,
+    common_name,
+    scientific_name,
+    assembly_name
+  } = props.species;
+
+  const toggleHoverState = (newHoverState: boolean) => {
+    if (newHoverState !== isHovering) {
+      setIsHovering(newHoverState);
+    }
```
imran-ebi: What if `newHoverState` is equal to `isHovering`? Wouldn't you wanna toggle the hover state?

azangru:
One thing I forgot to mention in the description: according to the mockup, there should be a "change" link that will take the user back to Species Selector, where they can add, remove or enable/disable species:

image

It feels that this link should be a part of the SpeciesTabBar component, but since it is will be dependant on React Router, I don't know whether it's a good idea to try to show it in Storybook — and currently Storybook is the only place where this tab bar can be rendered. Perhaps we should add this link later, when we finally put this tab bar in Genome Browser's header?

src/ensembl/src/shared/species-tab/SpeciesTab.tsx
azangru: From lines 45-50:

```
    <div
      className={className}
      onMouseEnter={() => toggleHoverState(true)}
      onMouseLeave={() => toggleHoverState(false)}
      onClick={handleClick}
    >
```

I want to switch isHovering to `true` only on onMouseEnter. And I want to switch it to `false` only on onMouseLeave. If it is already `true` when onMouseEnter is fired or `false` when onMouseLeave is fired (I can't imagine how this can happen, but just in case it can), then I am ignoring the call to the function.

Most likely, I could have done just with:

```
      onMouseEnter={toggleHoverState}
      onMouseLeave={toggleHoverState}
```

and then in the `toggleHoverState` function always switch the state to its opposite, but it feels a little bit less robust than the current version.

imran-ebi: Ah sorry, I misunderstood the functionality. This makes sense.

imran-ebi:
@azangru I think adding it later should be alright

src/ensembl/src/shared/species-tab/SpeciesTab.scss
azangru: 😄 probably not. Removed.

PR #48, opened 2019-05-09
Update packages

azangru:
It's been a long several weeks since we last updated our dependencies.

After update, the tests pass and the app seems to work fine on dev machine.


src/ensembl/package.json
```diff
@@ -68,76 +68,76 @@
...
     "html-webpack-plugin": "3.2.0",
-    "husky": "1.3.1",
+    "husky": "2.2.0",
     "identity-obj-proxy": "3.0.0",
     "image-webpack-loader": "4.6.0",
-    "jest": "24.7.1",
+    "jest": "24.8.0",
     "jest-fetch-mock": "2.1.2",
-    "jest-snapshot": "24.7.1",
+    "jest-snapshot": "24.8.0",
```
imran-ebi: I thought we got rid of this package 🤔

azangru: Turns out, we didn't 😄 
Should we?

imran-ebi: Yes please

imran-ebi:
The update was smooth for me, however the tests run a bit slower now at ~13s. Everything else works fine.

src/ensembl/package.json
azangru: Removed.

PR #49, opened 2019-05-09
Add browser storage and preserve selected species

azangru:
This PR adds the storage service for keeping client-side data either in local storage or in session storage. It also uses this storage service for storing selected species between browser reloads.

Related JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-72

### DIFF
--- a/src/ensembl/src/content/app/species-selector/components/species-commit-button/SpeciesCommitButton.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/species-commit-button/SpeciesCommitButton.tsx
@@ -5,7 +5,7 @@ import {
   hasCurrentSpecies,
   canCommitSpecies
 } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
-import { commitSelectedSpecies } from 'src/content/app/species-selector/state/speciesSelectorActions';
+import { commitSelectedSpeciesAndSave } from 'src/content/app/species-selector/state/speciesSelectorActions';
 
 import { PrimaryButton } from 'src/shared/button/Button';
 
@@ -39,7 +39,7 @@ const mapStateToProps = (state: RootState) => ({
 });
 
 const mapDispatchToProps = {
-  onCommit: commitSelectedSpecies
+  onCommit: commitSelectedSpeciesAndSave
 };
 
 export default connect(
--- a/src/ensembl/src/services/storage-service.ts
+++ b/src/ensembl/src/services/storage-service.ts
@@ -9,12 +9,12 @@ export enum StorageType {
 }
 
 type PrimitiveValue = string | number | boolean | null | undefined;
-type ArrayValue = PrimitiveValue[];
+type ArrayValue = PrimitiveValue[] | ObjectValue[];
 type ObjectValue = {
   [key: string]: PrimitiveValue | ArrayValue | ObjectValue;
 };
 
-type ValueForSaving = PrimitiveValue | ArrayValue | ObjectValue;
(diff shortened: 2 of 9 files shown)


src/ensembl/src/services/storage-service.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export enum StorageType {
+  LOCAL_STORAGE = 'localstorage',
+  SESSION_STORAGE = 'sessionstorage'
+}
+
+type PrimitiveValue = string | number | boolean | null | undefined;
+type ArrayValue = PrimitiveValue[] | ObjectValue[];
+type ObjectValue = {
+  [key: string]: PrimitiveValue | ArrayValue | ObjectValue;
```
imran-ebi: Why is `ObjectValue` used here?

src/ensembl/src/content/app/species-selector/components/species-commit-button/SpeciesCommitButton.tsx
```diff
@@ -5,7 +5,7 @@ import {
   hasCurrentSpecies,
   canCommitSpecies
 } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
-import { commitSelectedSpecies } from 'src/content/app/species-selector/state/speciesSelectorActions';
+import { commitSelectedSpeciesAndSave } from 'src/content/app/species-selector/state/speciesSelectorActions';
```
imran-ebi: Are we going to be appending `AndSave` to names of actions that save data to web storage?

src/ensembl/src/services/storage-service.ts
azangru: Because the value of an object can be an object 😄 

```
{
  foo: 'bar',
  foos: ['foo', 'bar'],
  foobj: { bar: 'quoox' }
}
```

src/ensembl/src/content/app/species-selector/components/species-commit-button/SpeciesCommitButton.tsx
imran-ebi: We can make a convention out of this!

src/ensembl/src/services/storage-service.ts
imran-ebi: Ah makes sense

src/ensembl/src/content/app/species-selector/components/species-commit-button/SpeciesCommitButton.tsx
azangru: What if we have an action that combines lots of different stuff: makes one or several requests, transforms data, saves transformed data and fires off another action? I'm a bit scared of enforcing naming conventions on actions.

imran-ebi: That's too much work for an action but as you say we can't rule it out. How about using this convention only for actions that do a simple save without any data transformation? Or maybe we can look into this later.

azangru: > That's too much work for an action 

In the world of thunks — I don't think so. In the world of sagas (redux-saga), or epics (redux-observable), or other clever libraries (`redux-logic`, `redux-loop`) this additional logic would probably be delegated to those sagas or epics, etc.; but in the world of thunks, since thunks are actions, this complex work by necessity has to be done in an action.

PR #51, opened 2019-05-21
Fix state of tracks in browser app

imran-ebi:
This PR fixes the browser tracks not being saved when navigated away from the browser app. Also, the following are persisted via web storage:

- Track list tabs
- Track list groups

TODO:

- [ ] Write tests

The related JIRA ticket is: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-83

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -3,7 +3,8 @@ import React, {
   useCallback,
   useRef,
   useEffect,
-  Fragment
+  Fragment,
+  useState
 } from 'react';
 import { withRouter, RouteComponentProps } from 'react-router-dom';
 import { connect } from 'react-redux';
@@ -44,6 +45,8 @@ import { toggleDrawer } from './drawer/drawerActions';
 import styles from './Browser.scss';
 
 import 'static/browser/browser.js';
+import browserStorageService from './browser-storage-service';
+import { TrackStates } from './track-panel/trackPanelConfig';
 
 type StateProps = {
   browserActivated: boolean;
@@ -84,12 +87,19 @@ export const Browser: FunctionComponent<BrowserProps> = (
   props: BrowserProps
 ) => {
   const browserRef: React.RefObject<HTMLDivElement> = useRef(null);
+  const [trackStatesFromStorage, setTrackStatesFromStorage] = useState<
+    TrackStates
+  >({});
 
   const dispatchBrowserLocation = (chrLocation: ChrLocation) => {
     if (browserRef.current) {
       props.changeBrowserLocation(chrLocation, browserRef.current);
     }
   };
+
+  useEffect(() => {
+    setTrackStatesFromStorage(browserStorageService.getTrackStates());
+  }, []);
 
   useEffect(() => {
     const { stableId } = props.match.params;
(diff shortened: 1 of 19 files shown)


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -3,7 +3,8 @@ import React, {
   useCallback,
   useRef,
   useEffect,
-  Fragment
+  Fragment,
```
azangru: Not that it's important, but can we please remove the fragment from jsx below? It doesn't seem to be doing anything useful here.

azangru:
If I am reading the code correctly, it seems to assume that there is only one species that is ever going to be shown by the genome browser. However, In our case of several (multiple) species, should the genome browser have the same track settings for all of these species (as I suspect will be the effect of this code)?

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
```diff
@@ -103,14 +108,31 @@ const TrackPanelList: FunctionComponent<TrackPanelListProps> = (
     };
   };
 
-  const getTrackListItem = (track: TrackPanelItem | null) => {
+  const getDefaultTrackStatus = (categoryName: string, trackName: string) => {
+    const statesOfCategory = props.trackStates[categoryName];
+
+    let defaultTrackStatus = ImageButtonStatus.ACTIVE;
```
azangru: Very confused why a track status is defined by an image button status.

src/ensembl/src/content/app/browser/browser-storage-service.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      );
+      const result = browserStorageService.getTrackListToggleStates();
+
+      expect(result).toEqual({});
+
+      mockStorageService.get.mockRestore();
+    });
+  });
+
+  describe('.updateTrackListTogglestates()', () => {
```
azangru: capital s in States

imran-ebi:
>If I am reading the code correctly, it seems to assume that there is only one species that is ever going to be shown by the genome browser. However, In our case of several (multiple) species, should the genome browser have the same track settings for all of these species (as I suspect will be the effect of this code)?

This code will not be covering it. I will create another PR for it as it is part of a different task that needs more work. This PR, however, is to fix the state of tracks.

azangru:
But why do we want to even touch this if it will need immediate rework? It’s not like we are publishing this to the live site until the end of June? Or are we?

imran-ebi:
> But why do we want to even touch this if it will need immediate rework? It’s not like we are publishing this to the live site until the end of June? Or are we?

We are publishing this by end of June and Dan requires this to work on the genome browser to show tracks based on the saved track states.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
imran-ebi: That's what was used previously and I thought of keeping it for now. But I understand what you mean. This is semantically wrong and I probably need to create a new enum for the track states.

ens-st3:
Can you add a link to this to the relevant JIRA @imran-ebi ?

imran-ebi:
As discussed with @azangru offline, a new branch will be created based on this branch to include species related state as well. So will be closing this PR.

PR #53, opened 2019-05-21
Add popular species section

azangru:
Adds the "popular species" section to the species selector page.

_Note: this PR might seem huge (2500 lines), but that's because it contains a lot of svgs (about 40) as mock data._

[Design reference](https://xd.adobe.com/view/38482aac-ac76-4d51-57e7-35bd3d85a424-4e79/screen/70e13f69-a680-4186-999b-dd1c1dcb6bad/ss/)

### Expected behaviour

- On the species selector page, user sees the list of 42 popular species icons (3 rows, 14 items per row)
- Available popular species are blue; unavailailable popular species are grayed out
- Buttons for unavailable popular species are not clickable; buttons for available speices are
- If user clicks on an active button, the search field is populated with corresponding species, and the button itself goes into the "selected" state (visually, dark-gray)
- If user clicks on the selected button, the search field is cleared, and the button goes into the initial state (visually, blue)
- If user has committed a species, the popular species button goes into the "committed" state (visually, white with black image)
- If user clicks on the committed button, the species is removed from the list of the committed species, and the button goes into the initial state (visually, blue)

### Notable changes

- Added InlineSVG component for requesting svg's over the network and then embedding them as inline svg's into the DOM
- Removed PopularSpeciesButton story from the storybook (the PopularSpeciesButton component is not used anywhere other than on the species selector page, and it's now more functional there than in the storybook; so keeping a story for this component became kind of pointless). 

### What's not included

The popular species buttons in this pull request do not show a tooltip when hovered over. I'll add the tooltip in another pull request (it'll probably need some thinking, because it seems to be the same component as what will later be used for z-menus).

[Related JIRA ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-79)


src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -1,43 +1,100 @@
...
+  clearSelectedSpecies: () => void;
+  deleteCommittedSpecies: (genome_id: string) => void;
+};
+
+// FIXME: this should be moved to a file with general functions
+const ensureHost = (url: string) => {
+  return /^https?/.test(url)
+    ? url
+    : `${location.protocol}//${location.host}${url}`;
 };
```
azangru: This is what I'd like your opinion about. We currently have `api-service`, which used to send requests only to the api backend provided. This is unusable e.g. in development, when sample svgs are served by the running webpack dev server. What I am currently doing is I updated the api service so that it can take an optional `host` parameter to use it instead of the default host of the backend, and it also checks whether the provided url is absolute or relative (if it's absolute, then the api-service won't concatenate it with the api host url).

It works, but I am still not sure as to whether this is a good interface for the service.

src/ensembl/src/shared/inline-svg/InlineSvg.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const src = faker.random.image();
+    const wrapper = mount(<InlineSVG src={src} />);
+
+    // FIXME: currently, the line below produces a warning (because of React state update)
+    // it can be fixed by wrapping it into the asynchronous act function
+    // introduced in react v.16.9.0 alpha, like so:
+    // await act(async () => {
+    //   await new Promise((resolve) => setTimeout(resolve, 0));
+    // })
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
azangru: Regarding the Fixme above, we can upgrade to React v.16.9.0 alpha, if we are feeling adventurous (I am 😄 ), and the warning that is currently displayed in the console:

image

should go away. What do you say?

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
msmanoj: > What I am currently doing is I updated the api service so that it can take an optional host parameter to use it instead of the default host of the backend

Having the `host` as an optional parameter sounds like a a better solution to this and I couldn't think of any particular reason why the `api-service` should always point to a single `host`.

src/ensembl/src/shared/inline-svg/InlineSvg.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import styles from './InlineSvg.scss';
+
+type Props = {
+  src: string; // url of the svg
+};
+
+const InlineSVG = (props: Props) => {
+  const [svg, setSvg] = useState<string | null>(null);
+  const [isSet, setIsSet] = useState(false);
```
imran-ebi: `setIsSet` sounds a bit weird, but that's probably due to `isSet`. How about `isSvgSet` and `setIsSvgSet`?

src/ensembl/src/shared/inline-svg/InlineSvg.test.tsx
imran-ebi: I think that it would be *too adventurous* to upgrade to an alpha version.

azangru: According to their changelog, the **only** change that happened in 16.9.0 alpha compared to 16.8.whatever was updating their testing utility.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -30,15 +33,21 @@ export const fetchStrainsAsyncActions = createAsyncAction(
...
+)<undefined, { popularSpecies: PopularSpecies[] }, Error>();
+
 export const fetchAssembliesAsyncActions = createAsyncAction(
   'species_selector/assemblies_request',
   'species_selector/assemblies_success',
   'species_selector/assemblies_failure'
 )<undefined, { assemblies: Assembly[] }, Error>();
 
-export const setSelectedSearchResult = createStandardAction(
+export const setSelectedSpecies = createStandardAction(
```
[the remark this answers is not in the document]
azangru: I thought of updating as changing some fields on the same entity (A -> A′; e.g. there is a user; we update his name; now the user has an updated name, but is still the same user). Here we are changing the entity itself (A -> B; e.g. there was one user, now there is a different user). You might still think about it as updating, but I suspect this verb might be misleading in this case.

src/ensembl/src/shared/inline-svg/InlineSvg.test.tsx
imran-ebi: I still think we should not do this just to get rid of a test warning. And we probably need to have a guideline/consensus on what versions (alpha, beta, rc, etc.) of the libraries we can use in the project.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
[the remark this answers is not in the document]
imran-ebi: Let's leave this as it is for now.

src/ensembl/src/shared/inline-svg/InlineSvg.tsx
azangru: Renamed the variables.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -1,43 +1,100 @@
...
-  onClick: () => void;
-  onStrainSelect: () => void;
+  isCommitted: boolean;
+  handleSelectedSpecies: (species: PopularSpecies) => void;
+  clearSelectedSpecies: () => void;
+  deleteCommittedSpecies: (genome_id: string) => void;
+};
+
+// FIXME: this should be moved to a file with general functions
+const ensureHost = (url: string) => {
```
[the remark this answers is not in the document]
azangru: Added the `shared/helpers` folder. Moved this to `shared/helpers/urlHelper`.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
azangru: I guess I'll leave the updates to the api service as they are then?

src/ensembl/src/shared/inline-svg/InlineSvg.test.tsx
azangru: As agreed, will not upgrade React to 16.9.0.alpha in this PR. If we get bothered by this warning, we'll update React in a separate PR.

src/ensembl/src/shared/helpers/urlHelper.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// FIXME: this should be moved to a file with general functions
+export const ensureAbsoluteUrl = (url: string) => {
+  if (typeof window === 'undefined') {
+    // this helper only makes sense for client-side rendering
+    return url;
+  }
+  const location = windowService.getLocation();
+
+  return /^https?/.test(url)
```
msmanoj: Few questions on this one:
1. Do we need the `s?` here? 
2. Are we making the protocol to be mandatory?
3. Would it be better to check if the path starts with `/` instead?

azangru: Actually, I am being stupid.

The only reason I added this helper in this PR was to enable the `InlineSVG` component to request svg's from localhost (during development). Because in production I will be getting real data with absolute urls to svg's.

For some reason, I decided to convert relative urls to absolute urls here. But perhaps a better solution would have been to allow `apiService` to use relative urls as well.

What if instead of adding this helper I update apiService from [this](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/services/api-service.ts#L40):

```
const url = `${this.host}${endpoint}`;
```

to something like this:

```
const url = options.useUnmodifiedEndpoint ? endpoint : `${this.host}${endpoint}`;
```
?

We might still need a function that converts relative urls to absolute urls at some later point (e.g. when we need to add server-side rendering), but for now update to the apiService should be enough to do the job.

msmanoj: How about using `options.preserveEndpoint` to keep it short?

azangru: Updated.

PR #54, opened 2019-05-23
Add error boundaries to avoid white screen of death

azangru:
Design guidelines — see the [June 2019 prototype, section 1 (Error messages for browser support)](https://xd.adobe.com/view/38482aac-ac76-4d51-57e7-35bd3d85a424-4e79/).

### What‘s included
This PR adds the `ErrorBoundary` component that is intended to stop the errors that may occur during rendering of a component from propagating further up the React component tree (because if the error reaches the top React node the whole child DOM tree will be unmounted, which will result in the entirely empty screen ("white screen of death").

The ErrorBoundary component receives a fallback component as a prop, and will show this fallback component if rendering of its children produced an error. This PR adds two fallback components: one (called `NewTechError`) to use for the parts of the site where we are relying on new technology (such as web assembly) and expect that some browsers will fail to show these components. The other fallback component (called `GeneralErrorScreen`) is the ultimate fallback for when something entirely unpredictable happens.

### What‘s not included
When the error boundary catches an error, it should report this error to an error monitoring service. This PR adds the `error-service` file to define the interface for reporting errors, but since we have not yet chosen an error monitoring service, this file isn't doing anything particularly useful at the moment.


src/ensembl/src/shared/error-screen/ErrorScreen.tsx
```diff
@@ -0,0 +1,47 @@
...
+        Sorry, but it seems your browser can’t display the new site
+      </div>
+      <a href="https://www.ensembl.org" className={styles.errorLinkButton}>
+        Go to Ensembl classic
+      </a>
+    </div>
+  );
+};
+
+export const GeneralErrorScreen = () => (
```
imran-ebi: Won't it be better to have these two components in separate files instead in the same one?

azangru: No idea 😄 We don't seem to have a convention for exporting closely related things. We may use an `index.ts` file in the error-screen folder, which will import and export the `GeneralErrorScreen` component and the `NewTechError` component from sibling files sitting in the same folder, if you prefer that approach.

imran-ebi: That sounds good to me. Will make our import statements shorter 😄

azangru: Ok, moved components to separate files and expose them via index.ts

imran-ebi:
Unless @msmanoj wants some some changes to be made, I think this is good to be merged into `dev`.

PR #55, opened 2019-05-24
[ENSWBSITES-119] Add new stylings to launchbar icons

msmanoj:
Applied new stylings to the launchbar icons and used ImageButton component.

Related JIRA ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-119

### DIFF
--- a/src/ensembl/src/header/launchbar/Launchbar.scss
+++ b/src/ensembl/src/header/launchbar/Launchbar.scss
@@ -44,10 +44,11 @@
   margin: 0 12px;
 
   svg {
-    fill: $ens-blue;
     width: 100%;
     height: 100%;
     transform: scale(1.1);
+    fill: $ens-white;
+    background-color: $ens-blue;
   }
 
   img {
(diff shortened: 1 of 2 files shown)


azangru:
I am a little confused about the semantics of the words "active" and "highlighted". If we press on a button corresponding to a route, and the route gets activated, is the button then "active" or "highlighted"?

msmanoj:
It felt more like you are 'highlighting' the state one button to look different from others.
If you have a look at the storybook entry for [ImageButton](https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-imagebutton-trackpanel--bookmarkicon), it seems to be matching with way it is implemented here.

imran-ebi:
Or we can use `selected` instead?

azangru:
I feel that "highlighted" is more about visual presentation while "active" or "selected" are more about the state that a thing is in. I would go with either "active" (which is idiomatic for different react routers) or "selected" here.

msmanoj:
Cool, I've renamed it to 'selected'.

PR #56, opened 2019-05-24
Add index file for shared components

msmanoj:
Added an index file inside the shared components folder that makes the importing of components a little easier.

**For example:**
We can now use
`import Button from 'src/shared';`

instead of:
`import Button from 'src/shared/button/Button';`

### DIFF
--- a/src/ensembl/src/shared/index.tsx
+++ b/src/ensembl/src/shared/index.tsx
@@ -1,0 +1,65 @@
+import {
+  Accordion,
+  AccordionItem,
+  AccordionItemButton,
+  AccordionItemHeading,
+  AccordionItemPanel,
+  AccordionItemState,
+  AccordionItemPermanentBlock
+} from './accordion';
+import Checkbox from './checkbox/Checkbox';
+import Input from './input/Input';
+import BadgedButton from 'src/shared/badged-button/BadgedButton';
+import Button, {
+  PrimaryButton,
+  SecondaryButton
+} from 'src/shared/button/Button';
+import RoundButton, {
+  RoundButtonStatus
+} from 'src/shared/round-button/RoundButton';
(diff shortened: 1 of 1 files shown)


azangru:
As our shared components will inevitably grow, the index file will also grow to become one giant list of exports. I am not sure I like it; the arrangement we had before looked neater and more self-contained. I will not terribly mind if this is what we decide to have, but my first gut reaction is, no, this index is too large and the components are too unrelated.

imran-ebi:
I agree with @azangru that we might end up with a lot of imports and exports in the same file, though this would be more convenient for us when importing these share components.

msmanoj:
The main motive is to make it easier to import shared components.
 
Instead of having to write multiple import statements like:
```
import Button from 'src/shared/button/Button';
import Select from 'src/shared/select/Select';
Import Input from 'src/shared/select/Input';
```

We can now simply use:
`Import {Button, Input, Select} from 'src/shared';`

Although it makes the index file messier, I feel it will be useful for other modules.

> components are too unrelated

They are unrelated but I thought they could be collectively called as shared components.

azangru:
I remember we discussed with Imran that it may be a good idea to move the components even one level deeper (shared/components/button/Button), because we have code that wants to sit in the shared folder, but is not related to components (like shared types, for example).

imran-ebi:
That way we can have multiple shorter index files for shared.

azangru:
We could alias `shared-components` in webpack to `/src/shared/components` if you hate typing out the path that much.

msmanoj:
I'm happy to type the path but I felt it would be a bit neat to use one import statement to import all the shared components we want.

imran-ebi:
@azangru @msmanoj I think we will need to close this PR, now that it has been lying around for a while

msmanoj:
Closing this as we will be creating multiple shorter index files rather than having one large file.

PR #60, opened 2019-06-02
Smart tooltip

azangru:
### Context
We currently have a shared component called `Dropdown`, which has not yet been used in any of the production screens. It seems that `Dropdown` is not a particularly suitable name for this component; so it will be renamed here to `Tooltip` (or should it be `PointerBox` to make its purpose more generic?). Also, the `Dropdown` component that is currently in the `dev` branch can be positioned only in one way relative to the element it is pointing at. This makes it all too likely that in certain situations the dropdown will get out of the viewport boundaries.

### Description
A tooltip (or pointer box) is a container that points at the centre of an item, like so:

image

It will usually point at the top or bottom side of an item (but for completeness’ sake can also point at its right or left side). Its position relative to the item is configured with the `position` prop that consists of two words: one indicating the side of the item that the tooltip is pointing at and the other indicating the direction in which the tooltip’s width can grow, e.g.:

image

The tooltip accepts a property that tells it whether to render as is or to check its position relative to its container (which can be explicitly passed to the tooltip, or, if absent, is assumed to be the `window`). If the tooltip gets outside of the bounds of its container, it will try (using a very unsophisticated algorithm) to find a position that will place it within the bounds of its container, and if such a position is possible, it will adopt it.

### Working example
I've deployed Storybook with the Tooltip stories to [Github Pages](https://ensembl.github.io/ensembl-client/storybook/?path=/story/components-shared-components-tooltip--variants) (if the Actions panel gets too big, just close it; I don't know why it gets so high).

### Note
Tests are not included, because much of the behaviour is DOM-related. I hope Tooltip stories in Storybook can count as a sort of test.


src/ensembl/src/shared/tooltip/tooltip-helper.ts
```diff
@@ -0,0 +1,137 @@
...
+    bottom: anchorBottom,
+    height: anchorHeight
+  } = anchorBoundingRect;
+  const anchorCentreX = anchorLeft + anchorWidth / 2;
+  const anchorCentreY = anchorTop + anchorHeight / 2;
+
+  const { width, height } = boundingClientRect;
+
+  let predictedLeft, predictedRight, predictedTop, predictedBottom;
+  if (position === Position.TOP_LEFT) {
```
imran-ebi: Can't this be changed into a `switch` statement?

src/ensembl/src/shared/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+  const handleClickInside = (e: React.MouseEvent | React.TouchEvent) => {
+    e.stopPropagation();
+  };
+
+  const handleClickOutside = (e: Event) => {
+    if (!parentRef.current) return;
+
+    let target;
+    if (e.type === 'touchend' && (e as TouchEvent).touches) {
```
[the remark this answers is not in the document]
azangru: We seem to have it in package.json and not use it anywhere in the code.

How does it help us with Typescript though? We need to get the `target` out of the `event`, and Typescript doesn't know what kind of event this is, and will still require checks despite what that what-input thing might say.

src/ensembl/src/shared/tooltip/tooltip-helper.ts
azangru: It can, but is it worth it? There is no `default` case, and every case will have to end with the nasty `break`. What benefit will the switch bring?

imran-ebi: As discussed offline we'll leave this as it is.

src/ensembl/src/shared/tooltip/Tooltip.tsx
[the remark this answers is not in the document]
imran-ebi: Fair enough. We'll anyway need type checking to be done as we discussed, so let's not go with `what-input` for this.

PR #64, opened 2019-06-06
Feature/custom downloads

msmanoj:
This PR is to add the Custom Downloads feature that works with the mock data.

Related JIRA Epic: 
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-105

**Note**: The data for the result section alone is currently loaded form the Genesearch API.

Changes made:
- Enabled the CD link on the launchbar
- Created a root level state for CD
- Added the shared Radio component
- Fixed minor UI issue with the Select component
- Updated $ens-green to ' #47d147' from '#61dc14'
- Added 'GENESEARCH_API_ENDPOINT' to .env

Things that will be done in the subsequent PRs (or commits if time permits):

1. Radio button styling & Storybook entry
2. Using real API endpoints to get the data for attributes, filters & orthologues.
3. More tests
4. Storybook entry for CD components (CheckboxGrid, CheckboxwithSelects & CheckboxWithRadios)
5. Google analytics ( need to be improved )


src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { FunctionComponent } from 'react';
+import { connect } from 'react-redux';
+import PreFilterPanel from 'src/content/app/custom-download/containers/pre-filters-panel/PreFilterPanel';
+import { getShowPreFilterPanel } from './state/customDownloadSelectors';
+import { RootState } from 'src/store';
+import CustomDownloadHeader from './containers/header/CustomDownloadHeader';
+import CustomDownloadContent from './containers/content/CustomDownloadContent';
+
+type StateProps = {
+  showPreFiltersPanel: any;
```
azangru: `any` -> `boolean`

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+describe('<CheckboxGrid />', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  let wrapper: any;
+  const defaultProps = {
+    gridData: gridData,
+    checkboxOnChange: checkboxOnChange
+  };
```
azangru: Better with shorthand:
```
const defaultProps = {
    gridData,
    checkboxOnChange
  };
```

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  it('renders without error', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    expect(wrapper.find(CheckboxGrid).length).toEqual(1);
+  });
+
+  it('renders N number of checkbox based on the gridData', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+
+    expect(wrapper.find(Checkbox).length).toEqual(12);
```
azangru: Instead of using the magic number 12, can it be calculated from the gridData object?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    gridData: gridData,
+    checkboxOnChange: checkboxOnChange
+  };
+
+  it('renders without error', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    expect(wrapper.find(CheckboxGrid).length).toEqual(1);
+  });
+
+  it('renders N number of checkbox based on the gridData', () => {
```
azangru: "of checkbox" -> "of checkboxes"

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    wrapper
+      .find(Checkbox)
+      .first()
+      .find('.defaultCheckbox')
+      .simulate('click');
+    expect(checkboxOnChange).toHaveBeenCalled();
+  });
+
+  it('sorts the checkboxs alphebatically based on the label', () => {
```
azangru: checkboxs -> checkboxes

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+
+    const firstGridContainer = wrapper.find('.checkboxGridContainer').first();
+
+    const firstCheckbox = firstGridContainer.find(Checkbox).first();
+    expect(firstCheckbox.prop('label')).toEqual('End');
+
+    const lastCheckbox = firstGridContainer.find(Checkbox).last();
+    expect(lastCheckbox.prop('label')).toEqual('Symbol');
```
azangru: can `End` and `Symbol` be calculated dynamically?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const firstCheckbox = wrapper.find(Checkbox).first();
+    firstCheckbox.find('.defaultCheckbox').simulate('click');
+
+    expect(checkboxOnChange).toHaveBeenCalledWith(true, 'default', 'end');
+  });
+
+  it('does not display the `Default` title', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    const firstGridTitle = wrapper.find('.checkboxGridTitle').first();
+    expect(firstGridTitle.text()).not.toBe('Default');
```
azangru: Does this expectation mean that there is no text or that there is some text which is other than "Default"? Not clear for the reader.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const lastCheckbox = firstGridContainer.find(Checkbox).last();
+    expect(lastCheckbox.prop('label')).toEqual('Symbol');
+  });
+
+  it('calls the checkboxOnChange when a checkbox is unchecked', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    const firstCheckbox = wrapper.find(Checkbox).first();
+    firstCheckbox.find('.defaultCheckbox').simulate('click');
+
+    expect(checkboxOnChange).toHaveBeenCalledWith(true, 'default', 'end');
```
azangru: - why was the "calls the checkboxOnChange when a checkbox is checked" test checked with toHaveBeenCalled and this one is checked with toHaveBeenCalledWith?
- it's not really clear to the reader what 'default' and 'end' are.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+import CheckboxGrid from './CheckboxGrid';
+import Checkbox from 'src/shared/checkbox/Checkbox';
+
+const checkboxOnChange = jest.fn();
+const gridData = {
+  default: {
+    symbol: {
+      id: 'symbol',
+      label: 'Symbol',
+      checkedStatus: false
```
azangru: can we rename this field to `isChecked` instead of `checkedStatus`?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+  it('does not display the `Default` title', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    const firstGridTitle = wrapper.find('.checkboxGridTitle').first();
+    expect(firstGridTitle.text()).not.toBe('Default');
+  });
+
+  it('hides the unchecked checkboxes when hideUnchecked is true', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} hideUnchecked={true} />);
+
+    expect(wrapper.find(Checkbox).length).toBe(2);
```
azangru: Really unclear from the test what is going on. Can the magic number 2 be calculated from gridData?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+  it('hides the title when hideTitles is true', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} hideTitles={true} />);
+
+    expect(wrapper.find('.checkboxGridTitle').length).toBe(0);
+  });
+
+  it('draws N number of columns based on the `column` parameter', () => {
+    wrapper = mount(<CheckboxGrid {...defaultProps} />);
+    const firstGridContainer = wrapper.find('.checkboxGridContainer').first();
+    expect(firstGridContainer.children().length).toBe(3);
```
azangru: Where is the column parameter in this test?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
+import Checkbox from 'src/shared/checkbox/Checkbox';
+
+import styles from './CheckBoxGrid.scss';
+
+type Props = {
+  gridData: any;
```
azangru: Can this be more type-specific (at least to the level that it's an object or an object containing nested objects)?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
...
+  columns: number;
+  hideUnchecked?: boolean;
+  hideTitles?: boolean;
+  checkboxOnChange: (status: boolean, subSection: string, id: string) => void;
+};
+
+export const filterCheckedAttributes = (attributes: any) => {
```
azangru: Can attributes be more type-specific?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
...
+  let totalAttributes = 0;
+
+  if (!attributes || Object.keys(attributes).length === 0) {
+    return 0;
+  }
+  Object.keys(attributes).forEach((section) => {
+    totalAttributes += Object.keys(attributes[section]).length;
+  });
+
+  return totalAttributes;
```
azangru: - can attributes be typed more precisely?
- what does !attributes mean (is it `undefined` or `null`)?
- depending on what !attributes mean, we can either use default argument (`{}`) for when there are no attributes or do something like `attributes = attributes || {}`
- the body of the function looks like a good candidate for a `reduce`

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
...
+};
+
+const renderCheckBoxList = (
+  checkboxList: any,
```
azangru: add type to `checkboxList`

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
...
+  props: Props,
+  subSection: string
+) => {
+  if (!Object.keys(checkboxList).length) {
+    return null;
+  }
+
+  const checkboxListIDs = Object.keys(checkboxList).sort();
```
azangru: Not that it's important, but probably better not repeat the same operation (`Object.keys(checkboxList)`) twice. Get the keys once and save them as a variable?

azangru:
General concern: right now, when I check/uncheck a checkbox (e.g. BioGRID), I see a loader in each of the left-column elements. This shouldn't be happening.

image

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
```diff
@@ -0,0 +1,15 @@
+@import 'src/styles/common';
+
+.checkboxGridContainer {
+  width: 100%;
+  display: flex;
```
azangru: is this flex used for anything (e.g. is it better than `display: inline-block; width: 100%`)?

_Also, I am not quite sure I understand why `display: block` doesn't work here (tried in my browser)._

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -0,0 +1,146 @@
...
+    for (let j = 0; j < props.columns; j++) {
+      if (totalCheckbox <= 0) {
+        break;
+      }
+      totalCheckbox -= 1;
+      if (checkboxListIDs[i + j]) {
+        gridMatrix[j] += 1;
+      }
+    }
+  }
```
azangru: Is this loop the same as saying:

```
const rows = Math.ceil(checkboxListIDs / props.columns);
const gridMatrix = Array(props.columns).fill(rows);
```
(and, judging by how the `gridMatrix` is used below, you probably don't even need it):

```
import range from 'lodash/range';

...

<div className={styles.checkboxGridContainer}>
  { range(0, columns).map((column) => ...)}
</div>
```

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+const onChange = jest.fn();
+
+const options: radioOptions = [
+  {
+    value: 'one',
+    label: 'one'
+  },
+  {
+    value: 'two',
+    label: 'two'
```
azangru: `faker.lorem.word()` for values and labels?

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      .find(Checkbox)
+      .find('.defaultCheckbox')
+      .simulate('click');
+
+    wrapper
+      .find('input[type="radio"]')
+      .first()
+      .simulate('change');
+
+    expect(onChange).toHaveBeenLastCalledWith('one');
```
azangru: Do we expect several calls of onChange here? Why do we check the last call?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useCallback } from 'react';
+
+import defaultStyles from './Radio.scss';
+
+export type radioOption = {
```
azangru: Here and below: can we name types starting with capital letters?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,66 @@
...
+  value: string;
```
azangru: Can the value be a number? Can it be a boolean?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,66 @@
...
+      if (props.disabled) {
+        return;
+      }
+      props.onChange(event.target.value);
+    },
+    [props.selectedOption]
+  );
+
+  return (
+    <>
```
azangru: First, why fragment?
Second, why table? A radio button with a label should not be a table.

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleOnChange = useCallback(
+    (selectedOption: string) => {
+      props.onChange(selectedOption);
+    },
+    [props.selectedOption]
+  );
+
+  return (
+    <>
```
azangru: no need for a fragment when there's a single child

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.tsx
```diff
@@ -0,0 +1,66 @@
...
+                <Radio
+                  onChange={handleOnChange}
+                  radioOptions={props.radioOptions}
+                  selectedOption={props.selectedOption}
+                />
+              </td>
+            )}
+          </tr>
+        </tbody>
+      </table>
```
azangru: Should this be a table? Won't a couple of div's (perhaps with `display: table` if necessary) cut it?

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.tsx
```diff
@@ -0,0 +1,66 @@
...
+    },
+    [props.selectedOption]
+  );
+
+  const handleOnChange = useCallback(
+    (selectedOption: string) => {
+      props.onChange(selectedOption);
+    },
+    [props.selectedOption]
+  );
```
azangru: Not sure what's going on with the three functions above:

1) for the first function (useEffect) is it just the same as if we wrote, when we instantiated the state:
```
const [checkedStatus, setCheckedStatus] = useState(Boolean(props.selectedOption));
```
?
(also, perhaps change checkedStatus to isChecked? isChecked definitely reads like a true/false value, while checkedStatus reads as something more complicated)

2) for the third function, is it the same as just passing `props.onChange` to the Radio component as the onChange prop?

3) As for the second function, what's the practical purpose of the `useCallback` hook? Why is it not just a function that sets state and calls a prop function?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    value: 'four',
+    label: 'four',
+    isSelected: false
+  },
+  {
+    value: 'five',
+    label: 'five',
+    isSelected: false
+  }
+];
```
azangru: use faker to generate fake data?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    wrapper = mount(<CheckboxWithSelects {...defaultProps} />);
+
+    wrapper
+      .find(Checkbox)
+      .find('.defaultCheckbox')
+      .simulate('click');
+    expect(wrapper.find(Select).length).toBe(1);
+  });
+
+  it('automatically checks the checkbox if atleast one option is selected', () => {
```
azangru: atleast -> at least

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  it('displays the remove button next to the Select if an option is selected', () => {
+    wrapper = mount(
+      <CheckboxWithSelects {...defaultProps} selectedOptions={['one']} />
+    );
+
+    expect(wrapper.find('.removeIconHolder').length).toBe(1);
+  });
+
+  it('displays the Plus button when one option is selected', async () => {
```
azangru: why is this test async?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    wrapper
+      .find('.addIconHolder')
+      .find(ImageButton)
+      .simulate('click');
+    expect(
+      wrapper
+        .find(Select)
+        .last()
+        .prop('options')
+    ).toHaveLength(4);
```
azangru: The description of this test talks about displaying another Select element; while the test checks for the number of options in the select element. Either the description is incorrect, or the test is.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+import Checkbox from 'src/shared/checkbox/Checkbox';
+import Select, { Option } from 'src/shared/select/Select';
+import styles from './CheckboxWithSelects.scss';
+
+import ImageButton from 'src/shared/image-button/ImageButton';
+import { ReactComponent as AddIcon } from 'static/img/browser/zoom-in.svg';
+import { ReactComponent as RemoveIcon } from 'static/img/shared/clear.svg';
+
+type Props = {
+  selectOptions: Option[];
```
azangru: perhaps just call this prop `options` instead of `selectOptions`? A bit hard to tell the difference between `selectOptions` and `selectedOptions`.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+  label: string;
+  selectedOptions: string[];
+  onChange: (selectedOptions: []) => void;
```
azangru: the type of `selectedOptions` is an empty array?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+const CheckboxWithSelects = (props: Props) => {
+  const [checkedStatus, setCheckedStatus] = useState(false);
+  const [shouldShowExtraOption, setShowExtraOption] = useState(false);
+
+  useEffect(() => {
+    if (props.selectedOptions.length > 0) {
+      setCheckedStatus(true);
+    }
+  }, []);
```
azangru: instead of `useEffect`, make this check in `useState`?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const handleCheckboxOnChange = useCallback(
+    (checkedStatus: boolean) => {
+      setCheckedStatus(checkedStatus);
+      props.onChange([]);
+    },
+    [props.selectedOptions]
+  );
```
azangru: does it have to be wrapped in useCallback updated by a change in props.selectOptions?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const handleOnSelect = useCallback(
+    (value: string, selectIndex?: number) => {
+      const newSelectedOptions: any = [...props.selectedOptions];
```
azangru: can newSelectedOptions be typed more precisely?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+                      description={'Add'}
+                      image={AddIcon}
+                    />
+                  </div>
+                </td>
+              </tr>
+            )}
+        </tbody>
+      </table>
+    </>
```
azangru: - no need for fragment
- does it need to be a table?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -0,0 +1,207 @@
...
+      setShowExtraOption(false);
+      props.onChange(newSelectedOptions);
+    },
+    [props.selectedOptions]
+  );
+
+  const selectedOptionsClone: any = [...props.selectedOptions];
+
+  const newSelectOptions: any = [...props.selectOptions].filter(
+    (option: any) => {
```
azangru: too many any's in this file 😞

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { RootState } from 'src/store';
+import PreviewDownload from './preview-download/PreviewDownload';
+
+import { getSelectedAttributes, formatResults } from './result-holder/helpers';
+
+type Props = {
+  selectedTabButton: string;
+  showPreview: boolean;
+  previewResult: any;
+  attributes: any;
```
azangru: could we clarify these any's?

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const Content = (props: Props) => {
+  const selectedAttributes: any = getSelectedAttributes(props.attributes);
+  let formatedPreviewResult = [];
```
azangru: formated -> formatted

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
```diff
@@ -0,0 +1,72 @@
...
+import ResultHolder from './result-holder/ResultHolder';
+import { RootState } from 'src/store';
+import PreviewDownload from './preview-download/PreviewDownload';
+
+import { getSelectedAttributes, formatResults } from './result-holder/helpers';
+
+type Props = {
+  selectedTabButton: string;
+  showPreview: boolean;
+  previewResult: any;
```
azangru: is it a verb (preview result)? is it a noun (preview of the result)? perhaps just `preview` is sufficient?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -0,0 +1,236 @@
...
+} from './sections';
+
+import styles from './AttributesAccordion.scss';
+
+type Props = StateProps & DispatchProps;
+
+const Attributes = (props: Props) => {
+  useEffect(() => {
+    props.fetchAttributes();
+  }, []);
```
azangru: 1) for default exports, can we name components the same way the file is named?
2) Is fetching attributes the responsibility of the attributes accordion or of the content section of custom download?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/genes/Genes.tsx
```diff
@@ -0,0 +1,88 @@
...
+type DispatchProps = {
+  setGeneAttributes: (setGeneAttributes: {}) => void;
+};
+
+const mapDispatchToProps: DispatchProps = {
+  setGeneAttributes
+};
+
+type StateProps = {
+  geneAttributes: any;
```
azangru: 😞

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/genes/Genes.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+    const checkedAttributes = filterCheckedAttributes(props.geneAttributes);
+
+    if (Object.keys(checkedAttributes).length === 0) {
+      return null;
+    }
+
+    return (
+      <div className={styles.checkboxGridWrapper}>
+        <CheckBoxGrid
```
azangru: Checkbox is one word (and in other files you do not capitalize the Box)

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/helpers.ts
```diff
@@ -0,0 +1,40 @@
+export const getMatchedSpeciesList = async (searchTerm: string, props: any) => {
```
azangru: any – 😞

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/helpers.ts
```diff
@@ -0,0 +1,40 @@
...
+    ) {
+      filteredSpecies[species.name] = {
+        id: species.name,
+        label: species.display_name,
+        checkedStatus: false
+      };
+    }
+  });
+
+  const newOrthologueFilteredSpecies: any = {};
```
azangru: any - 😞

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/helpers.ts
azangru: why is this function async?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/location/Location.tsx
```diff
@@ -0,0 +1,88 @@
...
+type DispatchProps = {
+  setLocationAttributes: (setLocationAttributes: {}) => void;
+};
+
+const mapDispatchToProps: DispatchProps = {
+  setLocationAttributes
+};
+
+type StateProps = {
+  locationAttributes: any;
```
azangru: 😞

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -0,0 +1,235 @@
+import React, { useCallback } from 'react';
```
azangru: should the commented code in this file stay or can it be removed?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -0,0 +1,235 @@
...
+              </div>
+            );
+          }
+        )}
+    </div>
+  );
+};
+
+type DispatchProps = {
+  setOrthologueAttributes: (setOrthologueAttributes: {}) => void;
```
azangru: what's the type of setOrthologueAttributes?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const attributesOnChangeHandler = useCallback(
+    (status: boolean, subSection: string, attributeId: string) => {
+      const newOrthologueAttributes = { ...props.orthologueAttributes };
+
+      newOrthologueAttributes[subSection][attributeId].checkedStatus = status;
+
+      props.setOrthologueAttributes(newOrthologueAttributes);
+    },
+    [props.orthologueAttributes]
+  );
```
azangru: Here and elsewhere: what is the purpose of using the `useCallback` hook? According to the docs, it returns a memoized function, which would have been handy if we did some heavy computation inside this function, but we are not; it is used for the sole purpose of calling a callback passed through props. Why not just have a regular function then, instead of the one wrapped in `useCallback`?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -0,0 +1,235 @@
...
+          JSON.stringify(orthologueAttributes)
+        );
+      } else {
+        delete newOrthologueAttributes[sectionHeader];
+      }
+
+      props.setOrthologueAttributes(newOrthologueAttributes);
+    },
+    [props.orthologueSpecies, props.orthologueAttributes]
+  );
```
azangru: Very confused about what’s happening in this handler. Do I understand correctly that the body of this function is a temporary hack for using hardcoded mock data and that it will change in the future?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -0,0 +1,225 @@
...
+  (resolve) => {
+    return (expandedPanels: []) =>
+      resolve(
+        expandedPanels,
+        getCustomDownloadAnalyticsObject(
+          'Toggle variation attributes accordion panel'
+        )
+      );
+  }
+);
```
azangru: Regarding all the analytics events in this file: they are very granular and related to interactions with different elements of the interface. Are all of these events needed? Is anyone going to analyze them?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -0,0 +1,225 @@
...
+    dispatch(setAttributes.success(attributes));
+  } catch (error) {
+    dispatch(setAttributes.failure(error));
+  }
+};
+
+export const setGeneAttributes = createAction(
+  'custom-download/set-gene-attributes',
+  (resolve) => {
+    return (geneAttributes: {}) =>
```
azangru: types for attributes?

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: I thought it should always be a string?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Value

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: Added some comments

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: Sorry, this is supposed to be testing for the default number of columns. I've updated it.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: That's done.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: That's done.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: Removed the fragment.

I preferred table here as I thought it will be easy to control the alignment of the radios. For example, at the moment it draws them horizontally, we can update this later on to support vertical alignment and multiple columns.

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
msmanoj: It shouldn't be. I've removed it.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: That's updated

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: Removed the useCallback wrapper.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: 1. That's done.

2. I initially had it in a way that content fetches the attributes, but then I thought it would be better to move it to the place where it is mainly used.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/genes/Genes.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/genes/Genes.tsx
msmanoj: That's now renamed

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/location/Location.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
msmanoj: That's now updated

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
msmanoj: I have used `flex` so that the columns inside this container gets displayed horizontally.
Is there any particular reason to avoid using it?

src/ensembl/src/shared/radio/Radio.tsx
azangru: I am almost certain that when you create an input element in JSX, you can pass a number as its value, and it will work (i.e. it will be automatically stringified). Which means that `e.target.value` in the onChange handler will be a string; but if instead of `e.target.value` you pass the original RadioOption value to the `onChange` prop (by finding the appropriate radio option by its index), you can have the radio elements work not only with strings.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/helpers.ts
msmanoj: Missed to remove it. Initially I had it loading the data directly from the rest API.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: Removed the fragment.

I feel it is easier to control the vertical alignment of the Select, add Icon & remove icon, by using a table.

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: As discussed, range cannot be used here as the structure of the gridMatrix is slightly different.
We can get it replaced once we have a simpler logic.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: Updated the tests

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
msmanoj: > Do I understand correctly that the body of this function is a temporary hack for using hardcoded mock data and that it will change in the future?

Yes, that's right. This whole section will get re-implemented once we have the new design.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
msmanoj: Removed all the unnecessary useCallbacks.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: Cool, I've updated it to support number and boolean.

src/ensembl/src/content/app/custom-download/components/checkbox-with-radios/CheckboxWithRadios.tsx
msmanoj: I've rectified all the above three issues.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: At the moment I am tracking almost all the actions that are dispatched within custom downloads. This is because I thought we might need something to track for the june-alpha release and we do not have enough time to wait for the spec and then add the tracking afterwards.

src/ensembl/src/content/app/custom-download/containers/content/preview-download/PreviewDownload.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <table className={styles.previewDownloadTable}>
+        <tbody>
+          <tr className={styles.previewDownloadHeader}>
+            <td>Species</td>
+            <td>Attributes</td>
+            <td>Filters</td>
+          </tr>
+          <tr>
+            <td>
+              <table>
```
azangru: Can’t check what this visually corresponds to at the moment, but a `table` tag within a `table` tag looks suspicious. Unless it’s really, semantically, supposed to be a table within a table, can’t we use a more modern styling technique than table layout?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
```diff
@@ -0,0 +1,141 @@
...
+type Props = StateProps & DispatchProps;
+
+const ResultHolder = (props: Props) => {
+  const selectedAttributes: AttributesSection = getSelectedAttributes(
+    props.attributes
+  );
+
+  useEffect(() => {
+    if (!selectedAttributes.length && props.preview.results) {
+      props.clearPreviewResult([]);
```
azangru: `clearPreviewResult` is defined below as `setPreviewResult.success`. And in `customDownloadActions.ts` `setPreviewResult.success` is described as a function that takes an argument of  `{ preview: {} }`, if I am reading that correctly. How does it agree with this clearPreviewResult that takes an empty array?

azangru:
As a general consideration: there are quite a few files here that are called simply `helper`. In the future, could you please name them more specifically (`result-holder-helper` or whatever is more appropriate). The reason being that files with more specific names are 1) easier to find; 2) easier to identify in the editor when multiple tabs are open.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  if (props.preview && !props.preview.results && props.isLoadingResult) {
+    return (
+      <>
+        <div className={styles.loaderWrapper}>
+          <CircleLoader />
+        </div>
+        <div className={styles.wrapper}>
+          {Array(15)
+            .fill('')
```
azangru: Instead of `Array(15).fill()`, `[...Array(15)]`.

what is 15, by the way?

src/ensembl/src/content/app/custom-download/containers/content/tab-buttons/TabButtons.scss
```diff
@@ -0,0 +1,10 @@
+@import 'src/styles/common';
+
+.wrapper {
+  padding-top: 40px;
+  // text-align: center;
```
azangru: delete commented line

src/ensembl/src/content/app/custom-download/containers/content/tab-buttons/TabButtons.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <div>
+        <BadgedButton
+          badgeContent={getTotalSelectedAttributes(props.attributes)}
+        >
+          <RoundButton
+            onClick={() => {
+              props.toggleTabButton('attributes');
+            }}
+            status={dataButtonStatus}
+            // classNames={{inactive: styles.inactive}}
```
azangru: delete commented line?

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -0,0 +1,265 @@
...
+  //   isSelected: false
+  // }
+];
+
+const getFormattedTotal = (total: number) => {
+  if (!total) {
+    return 0;
+  }
+  return total.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
+};
```
azangru: What is this function doing? Is it formatting numbers in a way that adds commas as a thousands separator? If so, then it looks like a very generic function that deserves to be in one of the shared helpers (and deserves to be tested).

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -0,0 +1,265 @@
...
+    props.setShowPreview(false);
+  };
+
+  const resultCount = getFormattedTotal(props.preview.resultCount);
+
+  const handleDownloadTypeSelect = (option: string) => {
+    props.setDownloadType(option);
+  };
+
+  const options = [...downloadTypeoptions].map((option: Option) => {
```
azangru: What's the purpose of the array spread here?

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -0,0 +1,265 @@
...
+    if (optionClone.value === props.downloadType) {
+      optionClone.isSelected = true;
+    }
+    return optionClone;
+  });
+
+  let downloadButtonStatus = true;
+  if (props.downloadType !== '') {
+    downloadButtonStatus = false;
+  }
```
azangru: `const downloadButtonStatus = props.downloadType !== ''`?

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  let previewButtonDisabled = true;
+  if (props.preview.resultCount > 0) {
+    previewButtonDisabled = false;
+  }
```
azangru: similarly here, is it the same as `const previewButtonDisabled = props.preview.resultCount === 0`?

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const selectedAttributes: SelectedAttribute[] = getSelectedAttributes(
+    props.attributes
+  );
+  const selectedFilters: any = getSelectedFilters(props.filters);
+
+  const getFormattedResult = () => {
+    if (resultCount === 0) {
+      return <>No results found</>;
```
azangru: why fragment?

src/ensembl/src/content/app/custom-download/containers/header/helper.ts
```diff
@@ -0,0 +1,42 @@
...
+    endpointFields +
+    '&sort=id&array=true&accept=' +
+    downloadType;
+  try {
+    setTimeout(() => {
+      const response = {
+        file: endpoint
+      };
+      window.open(response.file);
+    }, 100);
```
azangru: - Why is this wrapped in try-catch?
- Why the setTimeout?
- What’s the purpose of the `response` object?

src/ensembl/src/content/app/custom-download/containers/pre-filters-panel/PreFilterPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+          </RoundButton>
+          <RoundButton
+            onClick={() => {
+              filterOnClick('Variation');
+            }}
+            status={
+              RoundButtonStatus.DISABLED
+              // props.selectedPreFilter === 'Variation'
+              //   ? RoundButtonStatus.ACTIVE
+              //   : RoundButtonStatus.INACTIVE
```
azangru: Is the above commented code relevant?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: I'm trying to display 15 empty result cards on the initial load just so the loader appears nicely.

src/ensembl/src/content/app/custom-download/containers/pre-filters-panel/PreFilterPanel.tsx
msmanoj: I've temporarily disabled it as we don't support these filters yet.

src/ensembl/src/content/app/custom-download/containers/content/preview-download/PreviewDownload.tsx
msmanoj: Replaced the inner table with Div tags.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: I've updated the above line to pass {} instead of [] and changed `preview: {}` to `preview: any` for the time being as I do not have a proper structure for it.

src/ensembl/src/content/app/custom-download/containers/content/tab-buttons/TabButtons.scss
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/tab-buttons/TabButtons.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
msmanoj: Moved to shared folder and added some tests.

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
msmanoj: Not sure why I left it there. I've removed it now.

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
msmanoj: Updated it to span.

src/ensembl/src/content/app/custom-download/containers/header/helper.ts
msmanoj: Forgot to clean these up. I was initially trying to use a fake endpoint for the downloads but then I changed it to use the genesearch API itself.

src/ensembl/src/content/app/custom-download/state/customDownloadSelectors.ts
```diff
@@ -0,0 +1,22 @@
...
+  state.customDownload.preFilter.showPreFiltersPanel;
+
+export const getSelectedTabButton = (state: RootState): string =>
+  state.customDownload.tabButton.selectedTabButton;
+
+export const getPreviewResult = (state: RootState): string =>
+  state.customDownload.resultHolder.preview;
+
+export const getIsLoadingResult = (state: RootState): boolean =>
+  state.customDownload.resultHolder.isLoadingResult;
```
azangru: Regarding naming of things. Redux is a data layer; while such things as tabButton or resultHolder (if I understand correctly how they are used in the code) are interface elements. It's ok to keep ui-related information in the data layer, but isn't this information more appropriately described with words such as `results` or `tab`, without tight coupling to the interface elements that are used for displaying this data? Also, the `tab` property (a string?), instead of sitting in the `tabButton` part of the state (which is too concrete), can sit in the `ui` part of the state.

src/ensembl/src/shared/helpers/numberFormater.test.ts
```diff
@@ -0,0 +1,31 @@
+import { getCommaSeparatedNumber } from './numberFormatter';
+
+describe('getCommaSeparatedNumber', () => {
+  it('returns 1234 for the input 1234', () => {
```
azangru: 1) incorrect description
2) too specific description (what's so special about sequence 1234?); it would be better if the description said something to the effect of "separates thousand group in a single-thousand number", "separates thousand group in a ten-thousand number", etc. The number can then be generated randomly with faker (`faker.random.number({min:1000, max:9999})`.

src/ensembl/src/shared/helpers/numberFormater.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  it('returns -1,234,567 for the input -1234567', () => {
+    expect(getCommaSeparatedNumber(-1234567)).toBe('-1,234,567');
+  });
+
+  it('returns 0 if the inout is an empty string', () => {
+    expect(getCommaSeparatedNumber('')).toBe('0');
+  });
+
+  it('returns 0 if the inout is undefined', () => {
```
azangru: this and previous: inout -> input. Also, should  this actually work for empty strings or undefineds? Shouldn't the function just receive type number?

src/ensembl/src/shared/radio/Radio.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+const onChange = jest.fn();
+
+const options: RadioOptions = [
+  {
+    value: 'one',
+    label: 'one'
+  },
+  {
+    value: 'two',
+    label: 'two'
```
azangru: use faker?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,62 @@
...
+export type RadioOption = {
+  value: string | number | boolean;
+  label: string;
+};
+
+export type RadioOptions = RadioOption[];
+
+type Props = {
+  onChange: (selectedOption: string | number | boolean) => void;
+  classNames?: any;
```
azangru: surely not `any`? Is it a string (then className instead of classNames) or an object of strings (and if it's an object, then how is it used)?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,62 @@
...
+  radioOptions: RadioOptions;
+  selectedOption: string;
```
azangru: if radioOptions can be string, number, or boolean, then why is the selectedOption a string?

src/ensembl/src/shared/radio/Radio.tsx
```diff
@@ -0,0 +1,62 @@
...
+                />
+                <label className={styles.radioLabel} htmlFor={`radio_${index}`}>
+                  {option.label}
+                </label>
+              </td>
+            );
+          })}
+        </tr>
+      </tbody>
+    </table>
```
azangru: please not a table?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
azangru: This is really strange (and would probably cause a scroll to appear on a small screen?). Perhaps we could just have a layer with a spinner and overlay it atop the results area while loading results (and the results area meanwhile will either be completely blank or show previous results while waiting for the update to happen)?

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: 1. corrected it

2. although it is too specific, I feel it is more easier to understand when we say "it converts 1000 to 1,000" than saying "separates thousand group in a single-thousand number".

Also, if I generate a random number with faker, what will be the value for the expected output?

azangru: >  if I generate a random number with faker, what will be the value for the expected output

If you generate a random number within a 1000-9999 range, you would expect the resultant string, when split at the comma, to have 1 character in the first string and 3 characters in the last string, wouldn't you? 😉 

(you could even concatenate the result of the split and compare it with the original number, and make sure they are the same 😄 )

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: I was thinking what if numbers are passed in as a string, but I can see that it can also accept inputs like 'hello', which is not good.
I'll change it to only accept numbers.

src/ensembl/src/shared/radio/Radio.test.tsx
msmanoj: That's done.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: Updated the type. It just overrides the default classnames.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: Well spotted. I've updated the type.

src/ensembl/src/shared/radio/Radio.tsx
msmanoj: Removed the table

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: The first method will work but are you sure we should do this? It feels like we might need to come up with another test to test if this test works fine 😄 

The second will give false positive if the formatted result is 12,34.

azangru: But the first and the second checks will be done together.

`1234` will be formatted as `1,234`, which will split into a single-digit group (`'1'`) and as three-digit group `'234'`, which, when joined together, will result in `'1234'`, which will be equal to stringified `1234`.

Isn't it the operational definition of what we are doing to these numbers?

src/ensembl/src/content/app/custom-download/state/customDownloadSelectors.ts
msmanoj: That's done.

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: The tests are now updated.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: Updated the loader as discussed.

src/ensembl/src/shared/helpers/numberFormater.test.ts
```diff
@@ -1,23 +1,103 @@
...
+    // Check if there are two elements in the array
+    expect(numberSplitByComma.length).toBe(3);
+
+    // Check if the length of the first element is 1
+    expect(numberSplitByComma[0].length).toBe(1);
+
+    // Check if the length of the second element is 3
+    expect(numberSplitByComma[1].length).toBe(3);
+
+    // Check if the length of the second element is 3
```
azangru: third

src/ensembl/src/shared/helpers/numberFormater.test.ts
```diff
@@ -1,23 +1,103 @@
...
-  it('returns 1,234,567 for the input 1234567', () => {
-    expect(getCommaSeparatedNumber(1234567)).toBe('1,234,567');
+  it('returns x,xxx,xxx for the input number xxxxxxx', () => {
+    const randomNumber = faker.random.number({ min: 1000000, max: 9999999 });
+
+    const formattedRandomNumber = getCommaSeparatedNumber(randomNumber);
+
+    const numberSplitByComma = formattedRandomNumber.split(',');
+
+    // Check if there are two elements in the array
```
azangru: three

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: Updated

src/ensembl/src/shared/helpers/numberFormater.test.ts
msmanoj: Updated

PR #65, opened 2019-06-06
Update packages and fix webpack config

azangru:
### Update all our packages to their latest version.

**Most important changes:**
- Update rxjs to 6.5.2, because the 6.5.1 version has a bug in their `fromFetch` observable. Basically, when you receive a response from `fromFetch` and call `.json()` on it, this aborts the request and generates an error. Discussed [here](https://github.com/ReactiveX/rxjs/issues/4739).
- Update typescript to 3.5.x. Microsoft was very [enthusiastic](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/) about this release, saying that, among other things, it improves speed of type checking. I actually experienced the speed boost when I ran the `npm run check-types` script.

The update produced one type error (fixed in the PR).

### Update webpack config
While checking that this PR didn't break anything, I ran `npm run build`, and realized that we have a problem in our dev branch. Webpack prod config targets only images that are in the static folder, and the images I added in the error-screen folder were not picked up by `file-loader` and were breaking the build. We have either an option to move them to the static folder, or to update webpack config. I prefer the second one.

## Checks
- ran tests,
- ran type checker
- ran development build
- ran production build

Everything seems to be working fine.


src/ensembl/webpack/webpack.config.prod.js
```diff
@@ -24,7 +24,7 @@ const moduleRules = [
   // image loader should compress the images
   // then file loader takes over to copy the images into the dist folder
   {
-    test: /static\/img\/.*\.(gif|png|jpe?g)$/i,
+    test: /.*\.(gif|png|jpe?g)$/i,
     use: [
       {
         loader: 'file-loader',
```
azangru: Right now, this preserves the original folder structure when copying files to `dist` folder (so that images in `src/shared/error-screen/images` folder end up in `dist/src/shared/error-screen/images`). Perhaps we might use the `outputPath` option of fileLoader to put all image files in something like `dist/static/images` (though I never tried this).

imran-ebi:
> While checking that this PR didn't break anything, I ran npm run build, and realized that we have a problem in our dev branch. Webpack prod config targets only images that are in the static folder, and the images I added in the error-screen folder were not picked up by file-loader and were breaking the build. We have either an option to move them to the static folder, or to update webpack config. I prefer the second one.

@azangru I agree. Let's make Webpack recognise all images inside `src` and `static` folders.

src/ensembl/webpack/webpack.config.prod.js
imran-ebi: We will need to do something like that for images that come from the `src` directory. However, I think we will need to maintain the path of the `static` folder images as they are divided into separate directories.

azangru: Updated the config file so that now all images go into the `dist/images` folder (will change it to `dist/static/images` in the `cicd-test` branch, but that's for another PR).

Sample output (3 files from `src/shared/error-screen`, and one from `static/images/...`):

image

PR #66, opened 2019-06-07
Add species selector banner to the homepage

msmanoj:
Added the banner to the homepage that will be displayed when no species are selected.

Related JIRA ticket: 
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-136

### DIFF
--- a/src/ensembl/src/content/home/Home.tsx
+++ b/src/ensembl/src/content/home/Home.tsx
@@ -6,12 +6,11 @@ import styles from './Home.scss';
 import { fetchExampleEnsObjectsData } from 'src/ens-object/ensObjectActions';
 import { getExampleEnsObjects } from 'src/ens-object/ensObjectSelectors';
 import { getCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
-import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
 import { RootState } from 'src/store';
 
 type StateProps = {
   exampleEnsObjects: {};
-  selectedSpecies: CommittedItem[];
+  totalSelectedSpecies: number;
 };
 
 type DispatchProps = {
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/home/Home.tsx
```diff
@@ -57,6 +59,19 @@ const Home: FunctionComponent<HomeProps> = (props: HomeProps) => {
...
   return (
     <div className={styles.home}>
+      {!props.totalSelectedSpecies && (
+        <>
+          <span className={styles.speciesSelectorBannerText}>
+            7 species now available
+          </span>
+          <Link
+            className={styles.speciesSelectorBannerLink}
+            to={'/app/species-selector'}
```
azangru: I have a general question about links. Or rather about paths. Should we have some central place where to store / build them? For example, a url-helper, which will have methods or functions that build the required path?

```
const speciesSelector = () => '/app/speicies-selector';
```
and then in a component file:

```
import * as urlFor from 'path-to-url-helper';

<Link to={urlFor.speciesSelector()} />
```

@imran-ebi , what's your opinion about this?

imran-ebi: @azangru It is a good idea. Alternatively, we can use constants too for this.

azangru: I don't like the idea of constants, because urls will often contain dynamic elements.

msmanoj: Are we creating the helper in this PR or leaving it to a separate PR once we have more links available?

azangru: Good question. We probably might as well do it as a separate PR; but the links are creeping in in various places, so I would rather we do it pretty soon.

imran-ebi: @msmanoj If you create this now, please do let me know. I too might be able to make use of this function for the bootstrap code.

msmanoj: @azangru @imran-ebi added the `urlHelper` file to the path `src/shared/helpers/urlHelper.ts`.

PR #67, opened 2019-06-07
Replace the track reset icon

msmanoj:
This PR is to reset the track reset icon used in the BrowserBar.

Related JIRA ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-107


src/ensembl/src/content/app/browser/browser-reset/BrowserReset.tsx
```diff
@@ -39,10 +41,16 @@ export const BrowserReset: FunctionComponent<BrowserResetProps> = (
...
   return (
-    <dd className={styles.resetButton} onClick={resetBrowser}>
-      <button>
-        <img src={getResetIcon()} alt={details.description} />
-      </button>
+    <dd className={styles.resetButton}>
+      <div className={styles.imageWrapper}>
+        <ImageButton
+          buttonStatus={getResetIconStatus()}
+          description={'reset browser image'}
```
azangru: Is this the "approved" text? As a user, I am not sure I would understand what the "browser image" means.

msmanoj: Good question. I didn't think of changing it as I was more focused on getting the image updated. I've now updated the text to 'Reset browser image' after confirming it with Andrea.

andrea-ebi:
'Reset browser location' is a better description of what will happen...

PR #68, opened 2019-06-07
Feature/browser size handling fix

ens-ds23:
A mixture of tidying fixes to make browser animations smoother. Among others:
1. compute as much as possible in floats, even pixel sizes;
2. don't change canvas width/height except on redraw, and always do it before redraw;
3. change CSS to 100% during animation to avoid excess browser work (Manoj's idea, thanks!);
4. check for resize more often;
5. only do expensive, visually subtle fixes (eg changes which only affect font rendering quality) when no resize stuff is going on (after ~500ms timeout, say).


ens-st3:
What do you want me to review @ens-ds23 ?

ens-ds23:
Hi Steve, it's mainly a formality/FYI at the moment for things like scoping, etc, so just taking a look at the English in commits, PR, etc is ok. I thought I should keep to the agreed procedure of review even though there maybe little to tell from the code itself, partly for me to be in the habit of doing stuff by PR. I'm not sure who to choose as reviewers yet so it can seem a bit random. I'd discussed this with you a bit  so I put you in.

PR #69, opened 2019-06-10
Updates to Species selector

azangru:
## Description
- [x] Use the real api endpoint to get search-related data
- [x] Update typings of search-related data (to account for the added `assembly_name` field and removed `subtype` field)
- [x] Show message when no matches were found
- [x] Add an interface element for clearing the search field
- [x] Show a link to Genome Browser when at least one species has been selected
- [x] Update the way that the data about selected assembly is passed in redux action
- [x] Rename CloseButton to ClearButton (better reflects its functionality)
- [x] Show tooltip when hovering over popular species buttons
- [x] Show tooltip when hovering over the hint button (right-hand corner of species search field)
- [x] Remove the remaining SpeciesSelector stories (they don't serve any real purpose anymore)
- [x] Fix tests
- [x] Remove unused mock data for species selector

## Views affected
Species Selector (/app/species-selector).

## Other important changes
This pull request starts using real backend data for Species Selector. Currently, this data is accessed from a docker container that runs locally and listens on a different port. Webpack dev config proxies requests to api endpoints (which will be prefixed by the `/api` namespace) to the running docker container. In the future, by manipulating the proxy config we will be able to use various backends for development.

## Related JIRA Issue(s)
- [Show message when no results are found](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-141)
- [Add a button for clearing search field](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-142)
- [Add the "View in Genome browser" link](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-143)


src/ensembl/src/services/api-service.ts
```diff
@@ -23,6 +23,9 @@ type FetchOptions = {
   preserveEndpoint?: boolean;
```
azangru: Generally, regarding the api-service, I think it will have to be updated in a way that it doesn't have a default `host` (and, by extension, does not have the `preserveEndpoint` option), because it looks likely that most requests will be sent to the current domain anyway; but this should probably be done in a separate PR.

src/ensembl/src/content/app/species-selector/components/assembly-selector/AssemblySelector.test.tsx
```diff
@@ -49,14 +50,29 @@ describe('<AssemblySelector />', () => {
...
     const selectedOption = options.find(
       ({ isSelected }: { isSelected: boolean }) => isSelected
     );
+    const selectedOptionIndex = options.findIndex(
+      ({ isSelected }: { isSelected: boolean }) => isSelected
+    );
 
     expect(options.length).toEqual(defaultProps.assemblies.length);
-    expect(selectedOption.value).toBe(defaultProps.genomeId);
+    expect(selectedOption.value).toBe(selectedOptionIndex);
```
msmanoj: The description says that it should render a select if multiple assemblies are available, which is done on the line:
`expect(selectElement.length).toBe(1);`
Would it be better to add some comments about what is being checked on this line?

azangru: Good point. Split the test into two to make it more obvious.

src/ensembl/src/services/api-service.ts
azangru: To have a rough idea, see the changes in the cicd-test draft PR: https://github.com/Ensembl/ensembl-client/pull/70/files#diff-ddff3a47f9c4b497baea0d3bb6d1a2bf

We will also leverage webpack dev server's proxy option for proxying requests to appropriate backend hosts in development, instead of encoding that into the api-service.

src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
```diff
@@ -54,6 +55,9 @@ const SelectedSpeciesList = (props: Props) => {
           onRemove={props.onSpeciesDelete}
         />
       ))}
+      <div className={styles.genomeBrowserLinkContainer}>
+        <Link to="/app/browser">View in Genome Browser</Link>
```
msmanoj: urlFor shared helper is now available on dev:

import * as urlFor from 'src/shared/helpers/urlHelper';
urlFor.browser()

imran-ebi: Thanks @msmanoj. I am already using it in the bootstrap branch.

azangru: Rebased onto the fresh dev and added the urlHelper here.

_(I think, in the near future we will need to update this helper method to be able to generate links to specific genome, region and focus object)_

PR #70, opened 2019-06-14
Cicd test

kamaldodiya:
## Description

- Update CI/CD scripts in the gitlab-ci config file. Frontend code is tested and built, then the build artifacts are copied in an nginx image (together with static files from the [ensembl-2020-static-assets](https://github.com/Ensembl/ensembl-2020-static-assests) repo, then the image is deployed in our docker repository, and then a script is executed to deploy this image to a Kubernetes cluster
- Do not specify default api host; expect that the api host is the one in the window address bar
- Update webpack config in such a way that all static assets are saved in the `dist/static` folder, while the `index.html` file is saved in the `dist` folder directly. This should make it easier to write nginx routing rules as opposed to having the index.html and all the rest of static files in the same folder.
- Update typescript config not to throw errors if there are unused local variables in a file. These errors do not prevent assets from building, but when they occur, then the exit code of the build process is other than 0, which fails the cicd script. Besides, shaming us for having unused variables in the source code should be the job of the linter.


src/ensembl/tsconfig.json
```diff
@@ -9,8 +9,8 @@
     "module": "esnext",
     "moduleResolution": "node",
     "noEmit": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
azangru: Changed this so that we do not **fail** the build if there are unused variables in code. Unused variables should be pointed out by the linter and dealt with at that stage.

src/ensembl/webpack/webpack.config.prod.js
```diff
@@ -97,13 +96,7 @@ const plugins = [
     skipWaiting: true
   }),
 
-  new RobotstxtPlugin(),
-
-  // analyse the file sizes of bundled files
-  new BundleAnalyzerPlugin({
-    analyzerMode: 'static',
-    reportFilename: path.join(__dirname, '../bundle-report.html')
-  })
```
azangru: No need to have the bundle analyzer plugin in our regular production script.

src/ensembl/.eslintrc.js
```diff
@@ -27,7 +27,8 @@ module.exports = {
     'react/prop-types': 0,
     'react/no-unescaped-entities': 0,
     'react-hooks/rules-of-hooks': 2,
-    'prettier/prettier': 0
+    'prettier/prettier': 0,
+    'no-unused-vars': ['warn', { args: 'after-used' }]
```
azangru: To prevent the linter from complaining when arguments before a used argument are unused.

Example:

```
function(foo, bar, baz) {
    return baz;
}
```
in the example above, linter will not complain of the variables foo and bar.

.gitlab-ci.yml
```diff
@@ -42,10 +45,13 @@ Nginx:
...
   - git clone --depth 1 https://github.com/Ensembl/ensembl-2020-static-assests.git
   - git clone --depth 1 https://gitlab.ebi.ac.uk/kamal/ensembl-client-nginx.git
-  - docker build -t ${CONTAINER_IMAGE} -f ensembl-client-nginx/Dockerfile --no-cache .
+  - docker build -t ${CONTAINER_STAGE_IMAGE} -f ensembl-client-nginx/Dockerfile --no-cache .
+  - docker build -t ${CONTAINER_PROD_IMAGE} -f ensembl-client-nginx/Dockerfile.prod --no-cache .
   - echo "$GITLAB_REGISTRY_TOKEN" | docker login -u "$GITLAB_REGISTRY_USER" --password-stdin https://"$GITLAB_REGISTRY_URL"
-  - docker push ${CONTAINER_IMAGE}
-  - docker rmi ${CONTAINER_IMAGE}
+  - docker push ${CONTAINER_STAGE_IMAGE}
+  - docker push ${CONTAINER_PROD_IMAGE}
```
azangru: Why would a push to staging cause a production image to be built at well?

kamaldodiya: The way nginx redirects requests to backend API is through local.conf (nginx) file which is embedded inside the nginx image during build time. At the moment we need seprate nginx (local.conf). Its not good solution but it will change once we have ingress controller in place in kubernetes cluster.

PR #71, opened 2019-06-16
Bootstrap setup and allowing multiple instances

imran-ebi:
## Description
- Create bootstrap setup: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-101
- Allow multiple instances: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-87
- Fix state of tracks: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-83. There are two parts of this that was not fixed. One of it is the track settings and the other is the zoom level. For the latter, the zoom level in the genome browser needs to be exposed to the site chrome.
- Example links on homepage: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-127
- Persist data about explored genetic region: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-140
- Show details of location for focussed region: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-130
- Track panel with no focus object should not container object tracks: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-117
- Supporting locations (regions) as focus object: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-115
- Make genome browser use selected species: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-70

TODO:

- [ ] Add and update tests
- [ ] Save state of track zoom level (as mentioned above)
- [ ] Save state of track settings (as mentioned above)
- [ ] Resolving conflicts. I came across this when creating this pull request. I am not really sure which parts are actually necessary in the conflicted files. So I'd appreciate if any of you can have a look at it and let me know what changes need to go in.


src/ensembl/src/shared/helpers/urlHelper.ts
```diff
@@ -0,0 +1,4 @@
+export const speciesSelector = () => '/app/species-selector';
+export const customDownload = () => '/app/custom-download';
+export const browser = (genomeId: string, focus: string, location: string) =>
+  `/app/browser/${genomeId}?focus=${focus}&location=${location}`;
```
azangru: This can’t be right. It should be possible to generate an "empty" genome browser url (`/app/browser`) that is going to show a screen with example objects for the first selected species.

azangru:
Closing because of git-related grievances. The contents of this pull request, after a cleaner merge, have been replicated in a [separate pull request](https://github.com/Ensembl/ensembl-client/pull/72).

PR #72, opened 2019-06-17
Bootstrap setup, multiple genomes and zmenus in genome browser

azangru:
## Note
This pull request re-creates https://github.com/Ensembl/ensembl-client/pull/71 after merging dev and resolving conflicts.

## Description
- Create bootstrap setup: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-101
- Allow multiple instances: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-87
- Fix state of tracks: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-83. There are two parts of this that was not fixed. One of it is the track settings and the other is the zoom level. For the latter, the zoom level in the genome browser needs to be exposed to the site chrome.
- Example links on homepage: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-127
- Persist data about explored genetic region: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-140
- Show details of location for focussed region: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-130
- Track panel with no focus object should not container object tracks: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-117
- Supporting locations (regions) as focus object: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-115
- Make genome browser use selected species: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-70

TODO:

- [ ] Add and update tests
- [x] Save state of track zoom level (as mentioned above)
- [x] Save state of track settings (as mentioned above)
- [x] Resolving conflicts. I came across this when creating this pull request. I am not really sure which parts are actually necessary in the conflicted files. So I'd appreciate if any of you can have a look at it and let me know what changes need to go in.


src/ensembl/src/shared/helpers/urlHelper.ts
```diff
@@ -1,3 +1,4 @@
-export const speciesSelector = () => '/app/speicies-selector';
+export const speciesSelector = () => '/app/species-selector';
 export const customDownload = () => '/app/custom-download';
-export const browser = () => '/app/browser';
+export const browser = (genomeId: string, focus: string, location: string) =>
+  `/app/browser/${genomeId}?focus=${focus}&location=${location}`;
```
azangru: This can’t be right. It should be possible to generate an "empty" genome browser url (/app/browser) that is going to show a screen with example objects for the first selected species. I suggest the browser url builder receive an optional params object with genomeId, focus and location properties.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -96,32 +119,64 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+
+      urlFor.browser(
+        props.activeGenomeId,
+        props.activeEnsObjectId[props.activeGenomeId],
+        locationStr
+      );
+    } else {
+      setTrackStatesFromStorage(browserStorageService.getTrackStates());
+    }
+  }, []);
```
azangru: I guess this is temporary solution until https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-122

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -96,32 +119,64 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+  }, [props.browserQueryParams.focus]);
 
   useEffect(() => {
-    const [, chrStart, chrEnd] = props.chrLocation;
+    const chrLocationForGenome = props.chrLocation[props.activeGenomeId] || [
+      '',
+      0,
+      0
+    ];
+    const [, chrStart, chrEnd] = chrLocationForGenome;
```
azangru: perhaps slightly clearer and cleaner will be:
```
const chrLocationForGenome = props.chrLocation[props.activeGenomeId] || [];
const [, chrStart = 0, chrEnd = 0] = chrLocationForGenome;
```
_(no need to put the first item in the default array if we don't care about it anyway)_

src/ensembl/src/global/globalHelper.ts
```diff
@@ -9,3 +9,17 @@ export const getBreakpoint = (width: number): BreakpointWidth => {
...
+  let paramMap: { [key: string]: string } = {};
+
+  params.forEach((param: string) => {
+    const [key, value] = param.split('='); // split param pair into key value
+
+    paramMap[key] = value; // add the param pair into the param map
+  });
+
+  return paramMap;
+};
```
azangru: Not sure if for this PR, but do we possibly want to use a special library for parsing and generating query strings? Like [query-string](https://www.npmjs.com/package/query-string)?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -3,50 +3,62 @@ import { Link } from 'react-router-dom';
...
+    } = props;
+
+    if (
+      genomeExampleEnsObjects.length === 0 ||
+      genomeExampleEnsObjects[0].genome_id !== activeGenomeId
+    ) {
+      fetchGenomeExampleEnsObjects(activeGenomeId);
     }
-  }, [props.exampleEnsObjects]);
+  }, []);
```
azangru: Not sure what's going on here:
- should the possible data fetching happen only once, when the component is mounted? Are there any props that, when changed, should prompt re-fetching?
- how can it happen that `genomeExampleEnsObjects[0].genome_id !== activeGenomeId`? Shouldn't example objects always correspond to the active genome id?

src/ensembl/src/content/app/AppBar.scss
```diff
@@ -1,57 +0,0 @@
-@import 'src/styles/common';
-
-.appBar {
-  font-size: 12px;
-  padding: 8px 20px;
-  min-height: 76px;
```
azangru: Note that with removal of this component we are losing the minimum height that Andrea wanted this section to have. Need to make sure that all the individial app bars have this minimal height.

src/ensembl/src/global/globalHelper.ts
imran-ebi: Yes, I think that'll be better.

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: Yes, it is only temporary.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
imran-ebi: The data should be refetched if the focused object changes. This can happen if a user clicks on an example object link in the bookmarks modal of the track panel.

As for the check, it is to refetch data when the genome ID changes. In hindsight, this probably should be in a `useEffect` that fires whenever the genome ID changes.

src/ensembl/src/global/globalHelper.ts
msmanoj: Will it just be a replacement for these few lines? I'm not sure if the rest of the functions like `stringify` will be useful for us.

azangru: Why do you think stringify is not going to be useful for us? We can use it to:
- build query strings when updating the url in the browser window (e.g. genome browser url)
- build the query part of urls for api endpoints

src/ensembl/src/shared/helpers/urlHelper.ts
imran-ebi: I see @msmanoj has fixed this. So I have marked this as resolved.

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: @azangru I have created the intermediate view but how it is related to the ticket ENSWBSITES-122?

azangru: Now that I'm re-reading this piece of code, I am not sure what I meant. I probably was confused. But the if-block of this code isn't doing anything at the moment? It just generates a url string and returns `undefined`?

PR #78, opened 2019-06-20
Display date of the build instead of the hard-coded month

azangru:
## Description
Save time of the build and show it in the `Month yyyy` format. Use browser’s Intl api for formatting the date. Add polyfill for Intl for (very) outdated browsers.

Also, add polyfill for IntersectionObserver that is already used in the dev branch (added there with the Tooltip component).

## Reference
[Discussion](https://genomes-ebi.slack.com/archives/CC2RKR525/p1561037429460300) on Slack
image

### DIFF
--- a/src/ensembl/config.ts
+++ b/src/ensembl/config.ts
@@ -8,5 +8,7 @@ export default {
   googleAnalyticsKey: process.env.GOOGLE_ANALYTICS_KEY,
 
   // Genesearch endpoint
-  genesearchAPIEndpoint: process.env.GENESEARCH_API_ENDPOINT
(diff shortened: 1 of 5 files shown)


msmanoj:
@azangru if we make a minor fix on July, would the build date get updated to July? Is it possible to keep it as June and continue to make fixes?

ens-st3:
Decided to drop the month and introduce versioning instead. For this release we should use 'pre-release version 2.0'

azangru:
Closing in favour of https://github.com/Ensembl/ensembl-client/pull/81

PR #80, opened 2019-06-23
Do not show disabled species on genome browser screen

azangru:
## Type
- Bug fix
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-154

## Description
- [x]  Exclude species that have been marked with "do not use" from the list of species shown on the genome browser screen
- [x] Persist the data regarding whether to use or not to use particular species

## Views affected
Genome browser screen (/app/browser).

### DIFF
--- a/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
@@ -4,7 +4,7 @@ import { Link } from 'react-router-dom';
 
 import { getCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
 import {
-  toggleSpeciesUse,
+  toggleSpeciesUseAndSave,
   deleteSpeciesAndSave
 } from 'src/content/app/species-selector/state/speciesSelectorActions';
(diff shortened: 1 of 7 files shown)


src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
```diff
@@ -48,6 +48,8 @@ export const SpeciesSelectorAppBar = (props: Props) => {
 };
 
 const SelectedSpeciesList = (props: Props) => {
+  const shouldLinkToGenomeBrowser =
+    props.selectedSpecies.filter(({ isEnabled }) => isEnabled).length > 0;
```
msmanoj: @azangru we may have to disable the launchbar button based on selected enabled species count. Do you think it can be done in this PR?

azangru: Yes.

msmanoj: Cool, are you okay to make that change?

azangru: Yep. Done.

PR #81, opened 2019-06-24
Keep release version number in config and show it in header

azangru:
## Type

Replaces https://github.com/Ensembl/ensembl-client/pull/78

## Description
Instead of showing the month of the release in the header:

image

we decided to show the version of the release:

image

The versions are numbered arbitrarily, but resemble semantic version numbers (first number is a very major release; second number is something more minor; third number is something entirely trivial).

## Alternative solutions
We could use package.json to keep the version number, which is what people do when publishing their packages as libraries. But since our website isn't a publishable library anyway, there doesn't seem to be any benefit to keeping the version number in package.json (while it will require extra lines of code for getting the version number from package.json in webpack, saving this number in environment variable and reading it back in config.js).

### DIFF
--- a/src/ensembl/config.ts
+++ b/src/ensembl/config.ts
@@ -1,4 +1,7 @@
 export default {
+  // version number
+  version: '0.2.0',
+
   // environment
   isDevelopment: process.env.NODE_ENV === 'development',
   isProduction: process.env.NODE_ENV === 'production',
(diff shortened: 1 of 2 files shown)


ens-st3:
Can you think of anyway to update these three numbers automatically ?

azangru:
I can't think of a clean way to do this automatically.

We could probably write a script called `deploy` or something, which will update the version number, push to dev, and also push to master. But then we will have to think of the way to make sure no-one pushes to master other than by running this script. And we will also need to devise options for this script so that it knows which of the three numbers in the version number to bump.

It all looks flaky, one way or the other.

PR #83, opened 2019-06-25
Add environment env variable

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-125

## Importance
Need to be merged with dev for the June-alpha release

## Description
- Added the key ENVIRONMENT to .env
- Created a shared helper file under src/shared/helpers/environment
- Enabling custom download launchbar link only when the ENVIRONMENT is 'development' or 'internal'

### Other effects

### DIFF
--- a/src/ensembl/.env.example
+++ b/src/ensembl/.env.example
@@ -1,6 +1,9 @@
 # Node environment
 NODE_ENV=development
 
+# Environment
+ENVIRONMENT="development"
+
 # Keys for external services
 GOOGLE_ANALYTICS_KEY="UA-58710484-17"
 
--- a/src/ensembl/config.ts
+++ b/src/ensembl/config.ts
@@ -1,7 +1,11 @@
 export default {
-  // environment
+  // Node environment
   isDevelopment: process.env.NODE_ENV === 'development',
   isProduction: process.env.NODE_ENV === 'production',
+
+  // Deployment environment
+  environment: process.env.ENVIRONMENT,
+
   apiHost: process.env.API_HOST,
 
   // keys for services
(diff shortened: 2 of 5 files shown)


azangru:
Several things I wanted to comment on:

**1. Conceptually**

We aren't planning to have many different environments, I hope. So I would rather make the environment helper have a number of methods/functions that answer the question which environment we are currently in:

```typescript

export const isDevelopment = () => {
  return config.isDevelopment;
}

export const isProductionDeployment = () => {
  return config.environment === Environment.PRODUCTION;
}

export const isInternalDeployment = () => {
  return config.environment === Environment.PRODUCTION;
}
```

then, in the launchBar:

```
import environment from ...
// (or could be shortened to env)

...

      enabled={environment.isDevelopment || environment.isInternalDeployment}
```

**2. Comments on the current code**
_(these will be entirely unnecessary if we change the code in accordance with the previous section)_

a) naming
`getEnvironmentAvailability` -> `isEnvironment`

b) wordiness
```
  if (environment.indexOf(config.environment as Environment) !== -1) {
    return true;
  }

  return false;
```

better:

```
return environment.includes(config.environment)
```

even better:

instead of overloading the `getEnvironmentAvailability` function, so that it accepts either a string or an array of strings, it's easier if it accepts only one type of arguments (in this case, array of strings).

msmanoj:
> getEnvironmentAvailability -> isEnvironment

- Done

> return environment.includes(config.environment)

- Done

> instead of overloading the getEnvironmentAvailability function, so that it accepts either a string or an array of strings, it's easier if it accepts only one type of arguments (in this case, array of strings).

- Updated it to only accept array of strings

azangru:
Didn't like section 1 then? Ok 😞 Although it looks easier to write

```
import {isDevelopment, isInternalDeployment} from ...

if(isDevelopment() || isInternalDeployment()){...} 
```

than: 
```
import {isEnvironment, Environment} from ...

if(isEnvironment([Environment.DEVELOPMENT, Environment.INTERNAL])){...} 
```

msmanoj:
I liked both but I didn't see any big difference so I just left it as it is.

PR #87, opened 2019-06-26
Fix focus object reset location

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Importance
Required for June-alpha release

## Description
Reseting the location using the browser reset icon takes the wrong location from the store.

Steps to reproduce:
1) Load genome browser for a species
2) Select the example gene
3) Open the bookmark panel
4) Change the focus to a region
5) Click on the Browser reset icon
6) The location will be reset to the gene's default location, having the region as the focus

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -190,10 +190,10 @@ export const Browser: FunctionComponent<BrowserProps> = (
       return;
     }
 
+    props.updateBrowserActiveEnsObjectIdAndSave(focus);
     dispatchBrowserLocation(parsedLocation);
     props.fetchEnsObject(focus, genomeId);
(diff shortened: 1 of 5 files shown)


azangru:
There are still some problems with it. Here is a scenario:

- go to species selector, clear localStorage, refresh the page
- select human
- navigate to genome browser
- choose BRCA2
- in track panel, click on the bookmark icon and then click on the region object
- drag genome browser content to a different position
- click on the reset button 

An error will appear in the console:

image

azangru:
The reason for this error being that `chrLocation` here somehow becomes undefined:

https://github.com/Ensembl/ensembl-client/blob/bootstrap-setup-merge/src/ensembl/src/content/app/browser/browserActions.ts#L131

msmanoj:
That should be fixed now.

PR #93, opened 2019-06-27
Fix no description for focus genes

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-185

## Importance
June-alpha hotfix

## Description
Fixed the issue with the focus description not being displayed in the drawer.

## Views affected
Browser Chrome - > Trackpanel -> Drawer

### DIFF
--- a/src/ensembl/src/content/app/browser/drawer/Drawer.tsx
+++ b/src/ensembl/src/content/app/browser/drawer/Drawer.tsx
@@ -18,7 +18,7 @@ import DrawerGC from './drawer-views/DrawerGC';
 
 import closeIcon from 'static/img/track-panel/close.svg';
 
-import { get, find } from 'lodash';
+import { find } from 'lodash';
 import styles from './Drawer.scss';
 import SnpIndels from './drawer-views/SnpIndels';
 
--- a/src/ensembl/src/content/app/browser/drawer/Drawer.tsx
+++ b/src/ensembl/src/content/app/browser/drawer/Drawer.tsx
@@ -18,7 +18,7 @@ import DrawerGC from './drawer-views/DrawerGC';
 
 import closeIcon from 'static/img/track-panel/close.svg';
 
-import { find } from 'lodash';
+import find from 'lodash/find';
 import styles from './Drawer.scss';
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/browser/drawer/Drawer.tsx
```diff
@@ -15,12 +18,16 @@ import DrawerGC from './drawer-views/DrawerGC';
 
 import closeIcon from 'static/img/track-panel/close.svg';
 
+import { find } from 'lodash';
```
azangru: By doing so, you import the whole lodash and then pick just one method off it (unless webpack has learned how to tree-shake it properly). And we are not using lodash-webpack-plugin that will remove unused lodash code. Don't do that. Import specific functions (`import find from 'lodash/find'`).

src/ensembl/src/content/app/browser/drawer/Drawer.tsx
```diff
@@ -32,12 +39,23 @@ type OwnProps = {};
...
+  };
+
   const getDrawerViewComponent = () => {
     switch (props.drawerView) {
-      case 'gene':
+      case 'gene-feat':
         return <DrawerGene ensObjectInfo={props.ensObjectInfo} />;
-      case 'transcript':
-        return <DrawerTranscript ensObjectInfo={props.ensObjectInfo} />;
+      case 'gene-feat-1':
```
azangru: wow, where did these names come from? Is whoever sending them going to change them?

src/ensembl/src/content/app/browser/drawer/Drawer.tsx
msmanoj: That's done.

src/ensembl/src/content/app/browser/drawer/Drawer.tsx
msmanoj: I don't think this will change until we support multiple transcripts. At the moment it all seems to be 'gene-feat-1'.

PR #94, opened 2019-06-28
Get gene description from the API response

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-185

## Importance
June-alpha hotfix

## Description
- Getting the gene description from the API response to display within the drawer.
- Fixed a minor issue with the browser that throws an error when we directly hit the URL app/browser/ without having any species selected.
- Reordered CICD pipelines as per Kamal's request

## Views affected
1) Browser -> TrackPanel -> Drawer
2) Browser

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -158,8 +158,8 @@ export const Browser: FunctionComponent<BrowserProps> = (
       ) {
         changeSelectedSpecies(activeGenomeId);
       } else {
-        if (props.committedSpecies[0]) {
-          changeSelectedSpecies(props.committedSpecies[0].genome_id);
+        if (committedSpecies[0]) {
+          changeSelectedSpecies(committedSpecies[0].genome_id);
         }
       }
     }
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -158,7 +158,9 @@ export const Browser: FunctionComponent<BrowserProps> = (
       ) {
         changeSelectedSpecies(activeGenomeId);
       } else {
-        changeSelectedSpecies(props.committedSpecies[0].genome_id);
+        if (props.committedSpecies[0]) {
+          changeSelectedSpecies(props.committedSpecies[0].genome_id);
+        }
       }
     }
     setTrackStatesFromStorage(browserStorageService.getTrackStates());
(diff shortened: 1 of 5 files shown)


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -158,7 +158,9 @@ export const Browser: FunctionComponent<BrowserProps> = (
       ) {
         changeSelectedSpecies(activeGenomeId);
       } else {
-        changeSelectedSpecies(props.committedSpecies[0].genome_id);
+        if (props.committedSpecies[0]) {
```
azangru: committedSpecies have already been destructured from props several lines above; might as well use them directly

msmanoj: That's done

src/ensembl/src/ens-object/ensObjectTypes.ts
```diff
@@ -24,6 +24,7 @@ export type EnsObjectTrack = {
   ensembl_object_id?: string;
   support_level?: string | null;
   track_id: string;
+  description: string;
```
azangru: Wondering about the types of EnsObjectTrack:
1) is it really so that most of the fields can be absent? Or is it simply that their value can be null?
2) have you agreed with the backend that the description field will always be present and non-null?

azangru: Ok, it looks like various fields indeed can be absent.
Still wondering whether the description field will always be present though.

msmanoj: At the moment the description field is present in all the objects we currently support. 
@sanjay-boddu could you confirm if a value will always be there for description, please?

msmanoj: @azangru we are not very sure whether the middleware will return null or a default value for missing descriptions at this point. Should we leave this for now and come back to it later?

azangru: ok. If we are that unsure, I think it will be safer to describe this field as `string | null`.

msmanoj: That's done.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerGene.tsx
```diff
@@ -31,7 +33,9 @@ const DrawerGene: FunctionComponent<DrawerGeneProps> = (
 
       <dd className="clearfix">
         <label htmlFor="">Description</label>
-        <div className={styles.details}>{ensObjectInfo.description}</div>
+        <div className={styles.details}>
+          {ensObjectInfo.description || '--'}
```
azangru: Do we actually use double hyphens (`--`) instead of em-dash (`—`)?

msmanoj: Yes, Andrea preferred it to be '--'.

azangru: 👌

PR #95, opened 2019-06-28
Track panel settings design tweaks

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-163
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-164
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-165
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-166

## Importance
Minor

## Description
Applying design tweaks from Andrea.

## Views affected
Genome browser track panel settings
As foundation black is changed to use ens-black, you could see small difference in black texts everywhere in the site.

- [ ] I have checked any requirements for Google Analytics  N/A
- [ ] I have created any required tests  N/A

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
+++ b/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
@@ -25,6 +25,7 @@ import trackHighlightBtn from 'static/img/browser/icon_tracks_highlight_grey.svg
 import trackMoveBtn from 'static/img/browser/icon_tracks_move_grey.svg';
 import { RootState } from 'src/store';
 import { CogList } from '../browserState';
+import Checkbox from 'src/shared/checkbox/Checkbox';
 
 type StateProps = {
   applyToAll: boolean;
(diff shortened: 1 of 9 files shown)


src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -98,28 +99,41 @@ const BrowserTrackConfig: FunctionComponent<BrowserTrackConfigProps> = (
     props.updateApplyToAll(!applyToAll);
   }, [applyToAll, updateApplyToAll]);
 
+  const checkboxStyles: any = {
```
azangru: not `any`, please. Any is the last resort (and even then perhaps `unknown` is better, I don't know). Here the type is `{[key: string]: string}`. But it doesn't look like you even need to declare any type here; typescript should be capable of inferring it by itself.

src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -98,28 +99,41 @@ const BrowserTrackConfig: FunctionComponent<BrowserTrackConfigProps> = (
...
             onChange={applyToAllToggle}
           />
-          <label htmlFor="">All tracks</label>
         </dd>
       </dl>
       <dl>
         <dd>
-          <label htmlFor="">Track name</label>
-          <button className={styles.trackConfigSlider} onClick={toggleName}>
+          <label htmlFor={'trackConfig-trackName'}>Track name</label>
```
azangru: The `htmlFor` attribute on a label (which translates into the `for` attribute of the [html `label` tag](https://www.w3schools.com/tags/tag_label.asp) makes sense only if we are binding the label to an element with the corresponding id (usually — to an input). Here we are not; so it's useless.

Frankly, I do not understand this makrup. Why did Imran use the definition list (dl, dd) tags, when it's not a definition list? Why are we using label tags, although there is no corresponding input or select? I would have changed it to just divs or spans. Not necessarily now; perhaps during a later refactor.

jyothishnt: Regarding label `for`: In our case I added it to turn on/off trackConfigSlider when clicking on label itself, just like any other input. 

dd/dl:  True that it is not a def list. Divs are better in this case. I dont think we need to wait for refactoring. Lets clean it as we move.

jyothishnt: Changes pushed.

src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -98,28 +99,40 @@ const BrowserTrackConfig: FunctionComponent<BrowserTrackConfigProps> = (
...
+          <button
+            id={'trackConfig-trackName'}
```
azangru: use simple string

PR #97, opened 2019-07-01
Fix broken trackpanel tabs

msmanoj:
## Type
- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-189

## Importance
June-alpha hotfix

## Description
Fixed the issue with changing trackpanel tabs that currently exists on staging.

## Views affected
Browser -> Trackpanel -> Tabs (Genomic, Variation & Expression)

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
+++ b/src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
@@ -66,7 +66,7 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
     });
 
     setClickHandlers(callbacks);
-  }, [props.drawerOpened, props.genomeSelectorActive]);
+  }, [props.drawerOpened, props.genomeSelectorActive, props.ensObjectInfo]);
 
   return (
     <dl className={`${styles.browserTabs} show-for-large`}>
--- a/src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
+++ b/src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent, useEffect, useState } from 'react';
+import React, { FunctionComponent } from 'react';
 
 import { TrackType } from '../track-panel/trackPanelConfig';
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
```diff
@@ -66,7 +66,7 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
     });
 
     setClickHandlers(callbacks);
-  }, [props.drawerOpened, props.genomeSelectorActive]);
+  }, [props.drawerOpened, props.genomeSelectorActive, props.ensObjectInfo]);
```
azangru: Why do we want clickHandlers to be inside the component state updated with useEffect? Why create clickHandlers directly inside the body of the component function?

msmanoj: I'm not sure why it is done this way but I think this function could just be put inside the body of the component function.

Also, I think it is better to do the refactoring on a separate PR as it could slow down the hot fix we are trying to make and also introduce new bugs.

azangru:
```
  const handleTabClick = (value: TrackType) => {
    if (props.genomeSelectorActive || !props.ensObjectInfo.genome_id) {
      return;
    }

    if (props.drawerOpened === true) {
      props.toggleDrawer(false);
    }

    props.selectBrowserTabAndSave(value);
  }
```

and in the `dd` element:

```
        <dd
          className={getBrowserTabClasses(value)}
          key={value}
          onClick={() => handleTabClick(value)}
        >
```

no need for an object of click handlers. Isn't going to slow anything down. And it's just a translation of what's written inside useEffect into a slightly more recognisable form.

msmanoj: That's done.

azangru: 🎉

PR #98, opened 2019-07-01
Refactor/tracks persistence

azangru:
## Type

## Description
Read track states that have been saved in localStorage into redux store during initialization of the store, and pass them down as props to the components that need them. Do not read the values stored in localStorage directly from components.

## Views affected
Genome browser page

## Remains to be done in subsequent PRs
- move track expand status ([see here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx#L119-L125)) to redux as well
- throttle/debounce the call to genome browser to actually show/hide the respective track (it is an expensive operation, and is currently performed on the main thread)
- update types in such a way that we do not use ImageButtonStatus in the code that doesn't need to know about ImageButton (such as in actions, etc.)
- `fetchGenomeInfo` function in `genomeActions` — the way it relies now on the first item in the response array:
```
        dispatch(
          fetchGenomeInfoAsyncActions.success({
            [species.genome_id]: response.genome_info[0]
          })
        );

```
is bad. Either we should iterate over the response array and make sure the item in this array match the requested genomeId, or the response should be an object instead of an array, keyed by genomeId.

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -339,10 +339,7 @@ export const Browser: FunctionComponent<BrowserProps> = (
                 />
               </div>
             </animated.div>
-            <TrackPanel
-              browserRef={browserRef}
-              trackStates={trackStatesFromStorage}
-            />
+            <TrackPanel browserRef={browserRef} />
           </div>
         </section>
       )}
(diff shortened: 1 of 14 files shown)


src/ensembl/webpack/webpack.config.dev.js
```diff
@@ -72,7 +72,7 @@ const devConfig = {
         secure: false
       },
       '/browser': {
-        target: 'http://localhost:4000',
+        target: 'https://staging-2020.ensembl.org',
```
azangru: this reverts the change introduced into dev [here](https://github.com/Ensembl/ensembl-client/commit/9dd304381acf06f9d3647d3824e679d30397e679)

PR #99, opened 2019-07-01
Remove useless scrollbars in the drawer

azangru:
## Type
- Bug fix

## Description
The content of the drawer has a useless (and duplicate!) scrollbar, which is invisible on a Mac, but obvious on Windows or Linux:

image

After the fix:

image

## Views affected
Genome browser page (open drawer)

## Notice
Layout of drawer views does not make much sense (`clearfix` class, `label` elements, definition list elements) and will have to be updated, but that’s the task for another PR.

### DIFF
--- a/src/ensembl/webpack/webpack.config.dev.js
+++ b/src/ensembl/webpack/webpack.config.dev.js
@@ -72,7 +72,7 @@ const devConfig = {
         secure: false
       },
       '/browser': {
-        target: 'http://localhost:4000',
+        target: 'https://staging-2020.ensembl.org',
         changeOrigin: true,
         secure: false
       }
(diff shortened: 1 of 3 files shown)


src/ensembl/webpack/webpack.config.dev.js
```diff
@@ -72,7 +72,7 @@ const devConfig = {
         secure: false
       },
       '/browser': {
-        target: 'http://localhost:4000',
+        target: 'https://staging-2020.ensembl.org',
```
azangru: this reverts the change introduced into dev [here](https://github.com/Ensembl/ensembl-client/commit/9dd304381acf06f9d3647d3824e679d30397e679)

PR #100, opened 2019-07-02
Fix browser bar missing focus description

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-192

## Importance
June-alpha hotfix

## Description
Fixed the missing focus description in the BrowserBar and created the type BrowserChrLocation.

## Views affected
Brower -> BrowserBar -> BrowserInfo

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -129,7 +129,8 @@ export const Browser: FunctionComponent<BrowserProps> = (
   };
 
   const changeSelectedSpecies = (genomeId: string) => {
-    const chrLocationForGenome = props.chrLocation[genomeId];
+    const chrLocationForGenome =
+      props.chrLocation[props.activeEnsObjectId[genomeId]];
 
     let newUrl: string;
     if (chrLocationForGenome) {
@@ -197,11 +198,9 @@ export const Browser: FunctionComponent<BrowserProps> = (
   }, [props.browserQueryParams.focus]);
 
   useEffect(() => {
-    const chrLocationForGenome = props.chrLocation[props.activeGenomeId] || [
-      '',
-      0,
-      0
-    ];
+    const chrLocationForGenome = props.chrLocation[
+      props.activeEnsObjectId[props.activeGenomeId]
+    ] || ['', 0, 0];
     const [, chrStart, chrEnd] = chrLocationForGenome;
 
     if (props.browserActivated && chrStart > 0 && chrEnd > 0) {
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -15,6 +15,7 @@ import { RootState } from 'src/store';
 import {
   BrowserOpenState,
   BrowserNavStates,
+  BrowserChrLocation,
   ChrLocation
 } from './browserState';
 import {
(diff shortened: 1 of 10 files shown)


src/ensembl/src/content/app/browser/browserState.ts
```diff
@@ -23,6 +23,8 @@ export type BrowserNavStates = [
 
 export type ChrLocation = [string, number, number];
 
+export type BrowserChrLocation = { [ensObjectId: string]: ChrLocation };
```
azangru: the word "browser" in `BrowserChrLocation` doesn't really help much, does it?
We used to have a single chromosome location; now we have multiple chromosome locations keyed by genomeId; does it mean that the type is `ChrLocations`? Which probably means that the `chrLocation` field of the state becomes `chrLocations`?

Also, I don't know whether it's better to have a state that looks like this:

```
{
  chrLocations: {
    id1: [1, 234, 345]
    id2: [1, 334, 445]
  },
  defaultChrLocations: {
    id1: [1, 234, 345]
    id2: [1, 334, 445]
  }
}
```

or:

```
{
 id1: {
     chrLocation: [1, 234, 345],
     defaultChrLocation: [1, 234, 345]
  },
 id2: {
     chrLocation: [1, 334, 445],
     defaultChrLocation: [1, 334, 445]
  },
}
```

It's all so confusing. Can you offer any suggestions, @ens-ds23 ?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -197,11 +199,9 @@ export const Browser: FunctionComponent<BrowserProps> = (
   }, [props.browserQueryParams.focus]);
 
   useEffect(() => {
-    const chrLocationForGenome = props.chrLocation[props.activeGenomeId] || [
-      '',
-      0,
-      0
-    ];
+    const chrLocationForGenome = props.chrLocation[
+      props.activeEnsObjectId[props.activeGenomeId]
```
azangru: How did we get to the point when a prop called something-something-id is in fact a hash of ids? Should we be passing this whole hash to the Browser component? Should we not extract the relevant id in the mapStateToProps function?

(Don't know whether this grumble belongs to this PR or to a separate refactoring effort, but it just reads wrong.)

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -197,11 +199,9 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+    ] || ['', 0, 0];
```
azangru: Is there a point in having a default chromosome location ['', 0, 0] from the genome browser perspective, @ens-ds23?

src/ensembl/src/content/app/browser/browserState.ts
msmanoj: `ChrLocations` and `ChrLocation`? Wouldn't it be hard to distinguish?

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: How about renaming it to `activeEnsObjects` that will have hash of ids?

Also, I would prefer to get this refactoring done in a separate PR.

src/ensembl/src/content/app/browser/browserState.ts
azangru: Don't know. But the name `BrowserChrLocation` doesn't indicate in any way how this is different from a `ChrLocation` (which is also used by genome browser, so, in a sense, it is also a `BrowserChrLocation`).

PR #107, opened 2019-07-04
Refactor browser state

azangru:
## Type
- Bug fix

## Related JIRA Ticket(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-232

### 1. Refactoring of part of redux state

**Before:**

image

**After:**

image

**Changes:**

- for browser/browserLocation field:
  - `chrLocations` now contain current locations of genome browser keyed by genome id
  - remove `defaultChrLocations` (because its information is already present  in ensObjects)
- for `ensObject` field:
  - renamed to `ensObjects`
  - is a simple key-value storage of `ensObjectId: ensObject`
  - remove other fields contained in it (`ensObjectInfo`, `ensObjectTracks`, `exampleEnsObjects`)
  - `exampleEnsObjects` are the same as any other ens objects; so they are stored at the root level
  - ensObjectTracks are stored in an ensObject instead of independently
- changes in types:
  - `activeGenomeId` is now nullable
  - `activeEnsObjectId` (accessed through `getBrowserActiveEnsObjectId` selector) is now nullable
  - `chrLocation` (accessed through `getChrLocation` selector) is now nullable
  - `defaultChrLocation` (accessed through `getDefaultChrLocation` selector) is now nullable

### 2. Rely on url as the main source of information about active genome id, focus object id and chromosome location

Trying to use the url as the ultimate messenger of what the genomeId, activeObjectId and chrLocation should be (see the `setDataFromUrl` function in Browser.tsx). This removes the situations such as [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/Browser.tsx#L230-L237) when the value of a parameter was taken either from the url or from redux state. 

### 3. Updates to Rust code (Dan)
- temporarily disable "settle" code which is visually less appealing but was introducing rounding errors. Hard to debug in present state. Will reintroduce when easier to diagnose.
- do not reset cache for a stick upon receiving a stick event if the stick is the same as is currently used
- separately handle intended and actual location for their different purposes (state&user-feedback).
- separate "stage" code (ie current position and zoom handling) into own module to reduce method publicity to enable reasoning and refactoring.
- Allow use of Lamport Clock to avoid stale messages being acted upon (not currently used in react in this PR, but backwards-compatible).
- allow users not using a debug-compatible server to use debug builds by correctly handling errors in debug POST requests.
- move action runner code into OO style to allow  Lamport Clock impl and because it's generally a good idea.
- only log "important" events on console in debug mode to avoid clutter.
- necessary code tidying to implement the above.

### 4. Fix bug related to previously viewed objects
Steps to reproduce the bug:
- select a species, navigate to genome browser, choose an object to view
- by dragging the genome map or zooming in/out, move to a different chromosome location
- in the right-hand sidebar (a.k.a. track panel), open bookmarks tab, and click on the same object you are currently viewing
- expected: genome browser should move its image to show default position of the focus object
- actual: nothing happens

## Views affected
Genome browser page


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -4,65 +4,51 @@ import { connect } from 'react-redux';
...
 import { getChrLocationFromStr, getChrLocationStr } from './browserHelper';
 import { getDrawerOpened } from './drawer/drawerSelectors';
 import { getEnabledCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
 import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
-import {
-  fetchEnsObject,
-  fetchEnsObjectTracks,
-  fetchExampleEnsObjects
-} from 'src/ens-object/ensObjectActions';
+import { fetchEnsObject } from 'src/ens-object/ensObjectActions';
```
msmanoj: fetchEnsObject is unused.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -71,37 +57,34 @@ import styles from './Browser.scss';
...
-  fetchEnsObjectTracks: (ensObjectId: string, genomeId: string) => void;
-  fetchExampleEnsObjects: (genomeId: string) => void;
-  fetchGenomeInfo: (genomeId: string) => void;
-  fetchGenomeTrackCategories: (genomeId: string) => void;
+  fetchGenomeData: (genomeId: string) => void;
   replace: Replace;
   toggleDrawer: (drawerOpened: boolean) => void;
-  updateBrowserActiveGenomeIdAndSave: (genomeId: string) => void;
-  updateBrowserActiveEnsObjectIdAndSave: (objectId: string) => void;
   updateBrowserNavStates: (browserNavStates: BrowserNavStates) => void;
```
msmanoj: updateBrowserNavStates is unused.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -122,37 +105,83 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+      remove the genomeId === lastGenomeId check in the if-statement below
+      and move dispatchBrowserLocation(genomeId, chrLocation) above the if-statement
+    */
+    if (
+      !genomeId ||
+      (genomeId === lastGenomeId &&
+        genomeId === props.activeGenomeId &&
+        focus === props.activeEnsObjectId &&
+        isEqual(chrLocation, props.chrLocation))
+    ) {
```
msmanoj: Will the second part of this IF condition be true only when we click the reset button?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -164,116 +193,78 @@ export const Browser: FunctionComponent<BrowserProps> = (
           changeSelectedSpecies(committedSpecies[0].genome_id);
```
msmanoj: Here and in the if block above, if we use `return`, we can prevent reading from the localStorage twice.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -164,116 +193,78 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+      props.match.params.genomeId === props.activeGenomeId &&
+      props.browserQueryParams.focus === props.activeEnsObjectId
+    ) {
+      updateLocationInUrl();
+    }
+  }, [props.chrLocation]);
+
+  // helps distinguish first render from subsequent renders
+  useEffect(() => {
+    isFirstRenderRef.current = false;
```
msmanoj: Looks like this is unused?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -223,15 +222,12 @@ export const BrowserNavigatorButton = (props: BrowserNavigatorButtonProps) => (
...
-  activeObjectId: getBrowserActiveEnsObjectId(state)[
-    getBrowserActiveGenomeId(state)
-  ],
   browserActivated: getBrowserActivated(state),
   browserNavOpened: getBrowserNavOpened(state),
   chrLocation: getChrLocation(state),
   defaultChrLocation: getDefaultChrLocation(state),
   drawerOpened: getDrawerOpened(state),
-  ensObjectInfo: getEnsObjectInfo(state),
+  ensObject: getBrowserActiveEnsObject(state),
```
msmanoj: Would it be better to call this as `activeEnsObject` like we have `activeGenomeId`?

src/ensembl/src/content/app/browser/Browser.tsx
azangru: Will remove it for now. We'll need to think how best to fetch it if it's missing (currently it's all being done via `fetchGenomeData`, but that won't work when we start navigating to arbitrary objects from zmenu), but that can wait until the moment we actually need it.

src/ensembl/src/content/app/browser/Browser.tsx
azangru: Good catch! It wasn't used before either.

src/ensembl/src/content/app/browser/Browser.tsx
azangru: Second part of the if-statement (`(genomeId === lastGenomeId && genomeId === props.activeGenomeId && focus === props.activeEnsObjectId && isEqual(chrLocation, props.chrLocation))`) will actually be true most of the times `props.match.params.genomeId` or `props.location.search` change 😃 

The sequence of events is as follows:
- user drags browser image
- genome browser sends an event to React wrapper telling it that chromosome location has changed
- this event is captured in redux store; and chrLocation gets updated
- react wrapper sees that the chrLocation in redux store has updated and updates the url
- this causes `props.location.search` string to change. But the values encoded in the url are now the same as we keep in redux store; so this if-statement will evaluate to true; so we will not do anything.

src/ensembl/src/content/app/browser/Browser.tsx
azangru: I am actually not sure what `setTrackStatesFromStorage` is doing here 😄 This useEffect now depends on props.match.params.genomeId and props.location.search; while setTrackStatesFromStorage should depend on `genomeId` and `activeEnsObjectId`? I'll move it in its own useEffect.

src/ensembl/src/content/app/browser/browser-genome-selector/BrowserGenomeSelector.tsx
```diff
@@ -74,6 +69,9 @@ const BrowserGenomeSelector: FunctionComponent<BrowserGenomeSelectorProps> = (
 
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
+    if (!activeGenomeId) {
```
msmanoj: Is it possible to submit this form without having an `activeGenomeId`?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
azangru: 🤷‍♂  I don't really have any preferences regarding this.

msmanoj: We could either rename `activeGenomeId` to just `genomeId` or `ensObject` to `activeEnsObject` to keep them consistent. 
I think using the word `active` makes it more clear?

azangru: Turns out this `ensObject` is used in multiple places and, what's worse, is passed in other components, which also call it ensObject. I'd much rather leave its name as it is for now. As well as the names of `activeGenomeId` and `activeEnsObjectId`.

src/ensembl/src/content/app/browser/browserReducer.ts
```diff
@@ -51,10 +52,23 @@ export function browserEntity(
...
 ): BrowserEntityState {
   switch (action.type) {
+    case getType(browserActions.setDataFromUrl): {
+      const { activeGenomeId, activeEnsObjectId } = action.payload;
+      const newState = {
+        ...state,
+        activeGenomeId
+      };
+      if (!activeEnsObjectId) {
+        delete newState.activeEnsObjectIds[activeGenomeId];
```
msmanoj: I used a debugger here but it looks like whenever this gets executed, there are no matching entries to delete.
Do you know when this will actually delete an entry?

msmanoj:
Can the example links be spaced out a bit more?

Screenshot 2019-07-12 at 09 55 10


Currently on prod:
Screenshot 2019-07-12 at 09 57 55

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: Cool, thanks 👍

azangru:
Thanks for reminding about the spacing! I kept saying to myself that this needs to be fixed and then forgetting that.

src/assets/browser/app/src/composit/combinedstickmanager.rs
```diff
@@ -23,7 +23,6 @@ impl CombinedStickManager {
 
 impl StickManager for CombinedStickManager {
     fn get_stick(&mut self, name: &str) -> Option<Stick> {
-        console!("csm: {:?} ({:?})",name,self.internal.get(name).is_some());
```
ens-ds23: removes useless warn from debugging mode.

src/assets/browser/app/src/composit/compositor.rs
```diff
@@ -96,10 +96,18 @@ impl Compositor {
     }
 
     pub fn update_report(&self, report: &Report) {
+        if let Some(stick) = self.train_manager.get_stick() {
+            report.set_status("i-stick",&stick.get_name());
+        }
```
ens-ds23: Separate state into intended and actual state.

src/assets/browser/app/src/composit/compositor.rs
```diff
@@ -96,10 +96,18 @@ impl Compositor {
...
         self.train_manager.update_report(report);
     }
 
     pub fn set_stick(&mut self, st: &Stick) {
+        if let Some(stick) = self.train_manager.get_stick() {
+            if stick == st {
+                return;
+            }
+        }
```
ens-ds23: Don't update stick if current stick: too much cache invalidation.

src/assets/browser/app/src/composit/mod.rs
```diff
@@ -6,12 +6,10 @@ mod componentset;
 mod landscape;
 mod leaf;
 pub mod source;
-mod stage;
 mod stick;
 mod stickmanager;
-mod zoom;
-mod position;
 mod wrapping;
+mod stage;
```
ens-ds23: (Whole file) separate "stage" position management into own module. Only lightly coupled with rest of package and easier to factor when methods are private to a small module.

src/assets/browser/app/src/composit/stage/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod stage;
```
ens-ds23: (Whole file) separate "stage" position management into own module. Only lightly coupled with rest of package and easier to factor when methods are private to a small module.

src/assets/browser/app/src/composit/stage/position.rs
```diff
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use composit::Zoom;
+use super::zoom::Zoom;
```
ens-ds23: (Whole file) separate "stage" position management into own module. Only lightly coupled with rest of package and easier to factor when methods are private to a small module.

src/assets/browser/app/src/composit/stage/position.rs
```diff
@@ -48,19 +48,21 @@ impl Position {
...
     
-    pub fn set_zoom(&mut self, z: f64) {
+    pub(super) fn set_zoom(&mut self, z: f64) {
         self.zoom.set_zoom(z);
+        self.check_own_limits();
     }
     
-    pub fn settle(&mut self) {
+    pub(super) fn settle(&mut self) {
+        return; /* temporary, degrades visual output but buggy */
```
ens-ds23: Temporarily disable settling (buggy) to be re-enabled in a later PR (probably). If not code should be deleted.

src/assets/browser/app/src/composit/stage/position.rs
```diff
@@ -92,8 +94,8 @@ impl Position {
         }
     }
 
-    pub fn get_limit_of_middle(&self, which: &Direction) -> f64 {
-        self.get_limit_of_edge(which) -  self.middle_to_edge(which,true)
+    fn get_limit_of_middle(&self, which: &Direction) -> f64 {
```
ens-ds23: Incorrectly accounted for bumpers when doing limit checks.

src/assets/browser/app/src/composit/stage/stage.rs
```diff
@@ -1,12 +1,13 @@
 use std::collections::HashMap;
```
ens-ds23: (Whole file) separate "stage" position management into own module. Only lightly coupled with rest of package and easier to factor when methods are private to a small module.

src/assets/browser/app/src/composit/stage/stage.rs
```diff
@@ -15,29 +16,45 @@ pub struct Stage {
     dims: Dot<f64,f64>,
     mouse_pos: CPixel,
     base: f64,
-    pos: Position,
+    apos: Position,
```
ens-ds23: Separate intended from actual position.

src/assets/browser/app/src/controller/global/app.rs
```diff
@@ -40,7 +40,8 @@ pub struct App {
     als: AllLandscapes,
     size: Option<Dot<f64,f64>>,
     last_resize_at: Option<f64>,
-    stage_resize: Option<Dot<f64,f64>>
+    stage_resize: Option<Dot<f64,f64>>,
```
ens-ds23: Lamport Clock for messages to avoid late replays.

src/assets/browser/app/src/controller/global/booting.rs
```diff
@@ -143,7 +143,7 @@ impl Booting {
             global.register_app_now(&self.key,ar.clone());
         }
         let app = ar.clone().state();
-        actions_run(&mut app.lock().unwrap(),&initial_actions());
+        app.lock().unwrap().run_actions(&initial_actions(),None);
```
ens-ds23: Switch to more OO model which also has better access to members needed.

src/assets/browser/app/src/controller/input/action.rs
```diff
@@ -42,7 +42,7 @@ fn exe_pos_range_event(app: &App, x_start: f64, x_end: f64, y: f64) {
     let middle = Dot((x_start+x_end)/2.,y);
     let (pos,zoom) = app.with_stage(|s| { 
         s.set_screen_in_bp(x_end-x_start);
-        s.set_pos_middle(&middle); 
+        s.set_pos_middle(&middle);
```
ens-ds23: (Whole file) simplify action code by removing useless alternative calls.

src/assets/browser/app/src/controller/input/action.rs
```diff
@@ -98,13 +99,14 @@ fn exe_component_add(a: &mut App, name: &str) {
 
 fn exe_set_stick(a: &mut App, name: &str) {
     if let Some(stick) = a.with_stick_manager(|sm| sm.get_stick(name)) {
-        a.with_compo(|co| co.set_stick(&stick));
         a.with_stage(|s| {
             s.set_limit(&LEFT,0.);
             s.set_limit(&RIGHT,stick.length() as f64);
             s.set_wrapping(&stick.get_wrapping());
+            s.set_pos_intent(false);
```
ens-ds23: (whole file) avoid stale replays with Lamport Clocks.

src/assets/browser/app/src/controller/input/direct.rs
```diff
@@ -1,6 +1,10 @@
+use std::convert::TryInto;
```
ens-ds23: (whole file) tidying.

src/assets/browser/app/src/controller/input/direct.rs
```diff
@@ -114,14 +119,23 @@ fn custom_make_events(j: &JSONValue) -> Vec<Action> {
...
+        None
+    }
+}
+
 pub fn run_direct_events(app: &mut App, j: &JSONValue) {
     let evs = custom_make_events(&j);
+    console!("receive/A {}",j.to_string());
     console!("receive/B {:?}",evs);
-    actions_run(app,&evs);
+    app.run_actions(&evs,extract_counter(&j));
```
ens-ds23: Switch to more OO model which also has better access to members needed.

src/assets/browser/app/src/controller/input/direct.rs
```diff
@@ -114,14 +119,23 @@ fn custom_make_events(j: &JSONValue) -> Vec<Action> {
             out.append(&mut custom_make_one_event_key(k,v));
         }
     }
-    console!("receive/A {}",j);
+    out.push(Action::Settled);
     out
 }
 
+fn extract_counter(j: &JSONValue) -> Option<f64> {
```
ens-ds23: Lamport Clocks to avoid stale messages.

src/assets/browser/app/src/controller/input/domevents.rs
```diff
@@ -28,7 +28,7 @@ impl EventListener<()> for DomEventListener {
             }
             _ => Vec::<Action>::new()
         };
-        actions_run(&mut self.cg.lock().unwrap(),&evs);
+        self.cg.lock().unwrap().run_actions(&evs,None);
```
ens-ds23: Switch to more OO model which also has better access to members needed.

src/assets/browser/app/src/controller/input/optical.rs
```diff
@@ -5,8 +5,9 @@ use types::CDFraction;
```
ens-ds23: (while file) Lamport clocks to avoid stale messages.

src/assets/browser/app/src/controller/input/physics.rs
```diff
@@ -10,6 +10,7 @@ pub struct MousePhysicsImpl {
     mouse_pos: Option<CDFraction>,    /* spring handle */
```
ens-ds23: (whole file) Lamport clock to avoid stale messages.

src/assets/browser/app/src/controller/input/user.rs
```diff
@@ -51,9 +51,9 @@ impl UserEventListener {
```
ens-ds23: Switch to more OO model which also has better access to members needed.

src/assets/browser/app/src/controller/output/report.rs
```diff
@@ -28,7 +28,8 @@ struct StatusOutput {
     last_value: Option<JSONValue>,
```
ens-ds23: (whole file) Lamport clock to avoid stale messages.

src/assets/browser/app/src/controller/output/report.rs
```diff
@@ -28,7 +28,8 @@ struct StatusOutput {
...
     jigsaw: StatusJigsaw,
     last_sent: Option<f64>,
-    send_every: Option<f64>    
+    send_every: Option<f64>,
```
ens-ds23: (whole file) separate actual and intended location.

src/assets/browser/app/src/controller/output/report.rs
```diff
@@ -186,11 +250,23 @@ impl ReportImpl {
...
+            self.try_update_counter();
+            if !self.delayed {
+                for (k,s) in &self.outputs {
+                    if s.is_always_send() {
+                        if let Some(value) = self.make_value(&s.jigsaw) {                
+                            out.insert(k.to_string(),value.clone());
+                        }
+                    }
+                }
+                if vital {
```
ens-ds23: Only console log important events even in debug compiles.

src/assets/browser/app/src/data/blackbox/httpblackboxdriver.rs
```diff
@@ -23,6 +23,27 @@ impl BlackBoxResponseConsumer {
             interval
         }
     }
+
```
ens-ds23: Allow debug builds to be used without debug backend)

src/assets/browser/app/src/debug/testcards/base.rs
```diff
@@ -37,5 +37,5 @@ fn debug_initial_actions(name: &str) -> Vec<Action> {
 
 pub fn select_testcard(a: &mut App, stick_name: &str) {
     let acts = debug_initial_actions(stick_name);
-    actions_run(a,&acts);
+    a.run_actions(&acts,None);
```
ens-ds23: Switch to more OO model which also has better access to members needed.

src/assets/browser/app/src/model/train/trainmanager.rs
```diff
@@ -55,7 +55,7 @@ impl TrainManager {
     
     pub fn update_report(&self, report: &Report) {
         if let Some(ref stick) = self.stick {
-            report.set_status("stick",&stick.get_name());
+            report.set_status("a-stick",&stick.get_name());
```
ens-ds23: separate intended and actual location.

src/assets/browser/app/src/model/train/trainmanager.rs
```diff
@@ -77,6 +77,8 @@ impl TrainManager {
         }
     }
     
+    pub fn get_stick(&self) -> &Option<Stick> { &self.stick }
```
ens-ds23: Don't change stick when current stick.

src/ensembl/src/content/app/browser/browserReducer.ts
azangru: Good question; and I can't remember my thinking at the moment. I'll remove this check then.

PR #108, opened 2019-07-12
Browser messaging service

azangru:
## Type

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-177
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-174

## Description
1. This PR switches communication mechanism between GenomeBrowser and Browser Chrome from using custom events dispatched on the wrapper of GenomeBrowser to window messages sent via the `postMessage` api. For that purpose, `browser-messaging-service` is added. **Notice** that this also changes browser bootstrap code (and that we are currently sending a selector for genome browser to find the appropriate DOM element to use as its container)

2. The semantics of `chrLocation` is split into the "intended" chromosome location (that keeps the old name, `chrLocation`) and the "actual" chromosome location (named `actualChrLocation`). "Intended chromosome location" means the final location after the user has interacted with genome browser (e.g. dragged browser image and released the mouse button). "Actual chromosome location" means the location that the browser is displaying at the moment. Actual chromosome location updates much more frequently than the intended location (e.g. it changes while the user drags the content of the browser or zooms in/out). Intended chromosome location changes only after the navigation has completed, and is used to update the url.

3. Add `messageCounter` that keeps track of the sequence of messages exchanged between GenomeBrowser and BrowserChrome. MessageCounter is usually updated when the intended location is updated. BrowserChrome uses the last value of messageCounter it received from GenomeBrowser when it sends its messages to GenomeBrowser. GenomeBrowser will discard messages from BrowserChrome if their messageCounter is lower than the latest counter in GenomeBrowser (which means that the message from BrowserChrome is based on outdated data and should not be acted upon).

4. Move the code for updating GenomeBrowser page url from Browser component to browserActions.

5. Update `.gitignore` rules (ignore all files that happen to get in the `.vscode` folder except for `settings.json`, which instructs VSCode to use typescript from `node_modules`)

## Views affected
Genome browser


src/ensembl/src/content/app/browser/BrowserCogList.tsx
```diff
@@ -36,40 +32,36 @@ type DispatchProps = {
...
-  const listenBpaneScroll = useCallback((event: Event) => {
-    const bpaneScrollEvent = event as BpaneScrollEvent;
-    if (
-      bpaneScrollEvent.detail.delta_y ||
-      bpaneScrollEvent.detail.delta_y === 0
-    ) {
-      props.updateCogList(bpaneScrollEvent.detail.delta_y);
+  const listenBpaneScroll = (payload: BpaneScrollPayload) => {
+    const { delta_y, track_y } = payload;
+    if (typeof delta_y === 'number') {
```
msmanoj: Do we need to check if it is a number here as it will always be a number if it is defined?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -20,9 +20,7 @@ type StateProps = {
 
 type DispatchProps = {};
 
-type OwnProps = {
-  browserElement: HTMLDivElement;
-};
+type OwnProps = {};
```
msmanoj: Do we need the above two definitions?

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -37,19 +45,17 @@ export const updateBrowserActivated = createStandardAction(
...
-      detail: {
-        'config-url': `${protocol}${host}/browser/config`,
-        key: 'main'
-      }
-    });
 
-    browserEl.dispatchEvent(activateEvent);
+    const payload = {
+      'config-url': `${protocol}${host}/browser/config`,
+      key: 'main', // TODO: remove this field after we confirmed that it is redundant
```
msmanoj: Is this something you wanted to confirm now or later?

src/ensembl/src/content/app/browser/BrowserCogList.tsx
azangru: It can be a number of undefined. The check above did 

```
if (
      bpaneScrollEvent.detail.delta_y ||
      bpaneScrollEvent.detail.delta_y === 0
    )
```
because if delta_y is 0 it will not pass the test of `if (delta_y)` (because 0 is falsy). But we want 0 to make it to the body of the condition block.

So our options are:
- either check `if (delta_y || delta_y ===0)`, which is confusing
- or check `if (typeof delta_y === 'number')`, which I did
- or check `if (delta_y !== undefined)`

If you prefer the third variant of check to the second one, then I am happy to change it.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -146,12 +152,35 @@ export const updateChrLocation = createStandardAction(
   'browser/update-chromosome-location'
 )<ChrLocations>();
 
+export const updateActualChrLocation = createStandardAction(
+  'browser/update-actual-chromosome-location'
+)<{ [genomeId: string]: ChrLocation }>();
```
msmanoj: Can we use `ChrLocations` here?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
azangru: I think this is @imran-ebi 's preferred way of writing prop types :-) I prefer to just lump everything in a single Props type. Am happy to get rid of these empty types (but then I would rename StateProps to Props).

src/ensembl/src/content/app/browser/browserActions.ts
azangru: Later. @ens-ds23 wanted to get rid of the key field, since we are now passing the `selector` field in the payload; but he says it's too deeply rooted in genome browser code at the moment and will require some time to refactor out.

src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx
```diff
@@ -29,74 +29,67 @@ const ZmenuController = (props: Props) => {
...
       ...zmenus,
-      [zmenuData.id]: newZmenu
+      [payload.id]: newZmenu
     });
   };
 
-  const handleZmenuDestroy = (event: ZmenuDestroyEvent) => {
-    setZmenus(pickBy(zmenus, (value, key) => key !== event.detail.id));
+  const handleZmenuDestroy = (payload: ZmenuDestroyPayload) => {
+    setZmenus(pickBy(zmenus, (value, key) => key !== payload.id));
```
msmanoj: Would it be better clone `zmenus`, delete zmenus[payload.id] and call setZmenus?
By this way we don't have to iterate over it.

azangru: Cloning implies iteration though.

Which cloning method did you have in mind? `Object.assign({}, zmenus)`?

Iteration is a normal way of doing things within the functional programming paradigm. It shouldn't be a concern unless the structure you iterate over gets very large. In this particular case the number of entries in the `zmenu` state is in single digits. Most likely, `1`.

src/ensembl/src/content/app/browser/BrowserCogList.tsx
msmanoj: Can `delta_y` have negative values? 
If yes, we could check `if (delta_y > -1)`

If not, I think the third variant will be better than the second one.

src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx
msmanoj: I was thinking of the spread operator but I didn't realise that it implies iteration.
Let's leave this as it is.

src/ensembl/src/content/app/browser/BrowserCogList.tsx
azangru: Replaced type check with a check for `undefined`.

PR #113, opened 2019-07-16
Add analytics to species selector

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-73

## Importance
July release

## Description
Added analytics to the species selector according to the following spec:
https://docs.google.com/document/d/1LxxphliJIw_2G2Q4cJfPNUEVjSczZwo6SqqutdbX_dI/edit#

## Views affected
https://analytics.google.com/analytics/web/#/realtime/rt-overview/a58710484w195756165p190896376/

### Other effects
The callback handler used in both PopularSpecies and SpeciesSearchAutocomplete uses the same action `setSelectedSpecies` and as a result, it is not possible to know where the action is getting dispatched from. To fix this, I have added an additional parameter called 'source' just to let the handler know where the action is getting dispatched from.

## Possible complications
I have removed the code that was creating Analytics objects for individual apps from the file:
` src/analyticsHelper.ts`. This means that the category is now string that needs to be passed while building the Analytics object.


azangru:
My biggest question related to changes in SpeciesSelector is:

- whether to send analytics events through redux or directly from components
- if (big if) we do need to send analytics events through redux, is it better to dispatch a separate analytics action rather than try to fit analytics logic into other actions?

My concern is that action creators are getting larger and in some cases receive parameters that they wouldn't otherwise care about (like `handleSelectedSpecies` receiving the `source` parameter).

msmanoj:
We initially decided to send analytics events through redux as we wanted to track almost all the events. 
If we make the action creators simpler by moving all the analytics logics into its own components, we are still making the components larger. 

I agree that `handleSelectedSpecies` is a special case here as it needs to know where the action is getting triggered from. One other solution for this would be is to create separate actions for each source, but I'm not sure if it's worth doing that either.

I feel action creator is the best place to keep this logic as it will help us avoid a lot of code duplication.

imran-ebi:
@azangru Agree that we don't need to track all of the actions.

> It doesn't mean that we have to remove analytics middleware from redux. But, equally, having analytics middleware in redux doesn't mean we have to send all analytics events via redux.

If we don't want all of these GA calls to not to go through Redux, will we need to differentiate between the ones that go through and don't go through Redux?

azangru:
> will we need to differentiate between the ones that go through and don't go through Redux

I don't think there has to be a clear-cut rule. If an analytics event needs data that only redux can provide it (something it wants to grab from the state; or something that some other middleware has dispatched), then an analytics event has to be dispatched through redux. However, if a component has all the data it needs to report to analytics, I am very doubtful that it's better to send it via redux rather than send the analytics event directly from the component.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -21,6 +21,8 @@ import {
   PopularSpecies
 } from 'src/content/app/species-selector/types/species-search';
 
+import GoogleAnalyticsTracking from 'src/services/analytics-service';
```
azangru: Can you call it analyticsService (throughout the PR)? The service acts as an interface between the code that sends analytics events and the actual analytics backend to which we are sending this event. If we change from GoogleAnalytics to any other service, we won't need to update files that use the `analytics-service` file.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -149,15 +135,14 @@ export const updateTrackStatesAndSave: ActionCreator<
 export const toggleBrowserNav = createAction(
   'browser/toggle-browser-navigation',
   (resolve) => {
-    return () => resolve(undefined, getBrowserAnalyticsObject('Navigation'));
+    return () => resolve(undefined);
```
imran-ebi: Since nothing is passed to `resolve` can this action be created using `createStandardAction`?

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -24,6 +24,9 @@ import {
   SearchMatch,
   SearchMatches
 } from 'src/content/app/species-selector/types/species-search';
+
+import GoogleAnalyticsTracking from 'src/services/analytics-service';
```
imran-ebi: This also should be changed: `GoogleAnalyticsTracking` -> `analyticsService`

src/ensembl/src/services/analytics-service.ts
```diff
@@ -1,15 +1,18 @@
...
 
-const { googleAnalyticsKey = '' } = config;
+const { googleAnalyticsKey = '', isTest } = config;
 
 class AnalyticsTracking {
   private reactGA: typeof ReactGA;
 
   public constructor() {
-    ReactGA.initialize(googleAnalyticsKey);
+    ReactGA.initialize(googleAnalyticsKey, {
```
azangru: Typescript is upset here:

image

(also, it's generally a good thing to run `npm run check-types` before submitting a PR 😉  )

src/ensembl/src/analyticsHelper.ts
```diff
@@ -1,63 +1,21 @@
...
-export const buildAnalyticsObject = (category: AnalyticsCategory) => (
-  data: AnalyticsOptions | string
-): AnalyticsType => {
-  data = typeof data === 'string' ? { label: data } : data;
-
+const buildAnalyticsObject = (data: AnalyticsOptions): AnalyticsType => {
   return {
     ga: {
-      category,
       ...data
```
azangru: Is this the same as writing `ga: data`?

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -149,15 +135,14 @@ export const updateTrackStatesAndSave: ActionCreator<
...
 );
 
 export const updateBrowserNavStates = createAction(
   'browser/update-browser-nav-states',
   (resolve) => {
-    return (browserNavStates: BrowserNavStates) =>
-      resolve(browserNavStates, getBrowserAnalyticsObject('Navigation'));
+    return (browserNavStates: BrowserNavStates) => resolve(browserNavStates);
   }
 );
```
azangru: Most of the actions here, which just take a value and resolve this value without doing anything else can be re-written with `createStandardAction`. I don't know whether we _should_ (for some actions, you may argue that you will soon re-write them again, addind analytics data), but we definitely could. For those actions that aren't expected to be linked to analytics reporting, it will be a good idea.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -158,42 +206,129 @@ export const handleSelectedSpecies: ActionCreator<
...
 
-export const toggleSpeciesUse = createStandardAction(
-  'species_selector/toggle_species_use'
-)<string>();
+export const toggleSpeciesUse = createAction(
+  'species_selector/toggle_species_use',
+  (resolve) => {
+    return (
+      speciesName: string,
+      currentStatus: string,
```
azangru: Can't this be retrieved from the committedSpecies object (committedSpecies.isEnabled ? 'use' : 'do_not_use' in the body of the function)? Why the extra parameter?

azangru: Same goes for the speciesName argument

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -270,18 +249,14 @@ export const updateApplyToAll = createAction(
   'browser/update-apply-to-all',
   (resolve) => {
     return (yn: boolean) => {
```
azangru: If `yn` means yes/no, as I suspect it does, can we change this argument's name to `yesNo`?

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -158,42 +206,129 @@ export const handleSelectedSpecies: ActionCreator<
...
   speciesSelectorStorageService.saveSelectedSpecies(committedSpecies);
 };
 
-export const deleteSpecies = createStandardAction(
-  'species_selector/delete_species'
-)<string>();
+export const deleteSpecies = createAction(
+  'species_selector/delete_species',
+  (resolve) => {
+    return (speciesName: string, committedSpecies: CommittedItem[]) => {
```
azangru: As above, speciesName is easily calculable from committedSpecies.

src/ensembl/src/global/globalActions.ts
```diff
@@ -1,12 +1,10 @@
...
 
 import { BreakpointWidth } from './globalConfig';
-import { getGlobalAnalyticsObject } from '../analyticsHelper';
 
 export const updateBreakpointWidth = createAction(
   'browser/update-breakpoint-width',
   (resolve) => {
-    return (breakpointWidth: BreakpointWidth) =>
-      resolve(breakpointWidth, getGlobalAnalyticsObject('Default Action'));
+    return (breakpointWidth: BreakpointWidth) => resolve(breakpointWidth);
```
azangru: I am sure we won't ever track this 😄  Can just as well use createStandardAction here.

src/ensembl/src/header/headerActions.ts
```diff
@@ -1,25 +1,48 @@
...
     return (currentApp: string) =>
-      resolve(currentApp, getHeaderAnalyticsObject('Navigation'));
+      resolve(
+        currentApp,
+        buildAnalyticsObject({
+          category: 'Header',
+          action: 'change-current-app',
+          label: currentApp
+        })
+      );
```
azangru: Do we still want to track the three actions in this file?

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: committedSpecies will not have the species anymore as we delete it beforehand.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: It can be but we have to loop through the committedSpecies again to find out which species was toggled.

src/ensembl/src/services/analytics-service.ts
msmanoj: Not sure why it's upset here?
```
export function initialize(trackingCode: string, options?: InitializeOptions): void;
export function initialize(trackers: Tracker[], options?: InitializeOptions): void;
```

src/ensembl/src/header/headerActions.ts
msmanoj: Removed these.

msmanoj: Yes, the URL changes are tracked.

src/ensembl/src/global/globalActions.ts
msmanoj: That's done

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: Changed this createStandardAction.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: Update it to use createStandardAction.

src/ensembl/src/analyticsHelper.ts
msmanoj: Yes, updated it.

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
msmanoj: That's done.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: That's done.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
msmanoj: That's done.

src/ensembl/src/services/analytics-service.ts
azangru: Interesting. Will `ReactGA.initialize(googleAnalyticsKey as string` help?

msmanoj: Tried it. It's still the same.

azangru: It seems that they overwrite the first export of the `initialize` type with the second export. If I change the order of exports in the def file (which we can't do, because it's third-party), it works.

Only we also need to update our typings/config.d.ts file:

```
declare module 'config' {
  const config: {
    isDevelopment: boolean,
    isProduction: boolean,
    isTest: boolean,

    [key: string]: string
  };
  export = config;
}
```

msmanoj: Updated the `typings/config.d.ts` file and this error is now gone 👍

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -1,14 +1,21 @@
-import { createStandardAction, createAsyncAction } from 'typesafe-actions';
```
azangru: I seem to have forgotten a console.log in the `updateSearch` action in this file. Could you please remove it?

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
azangru: I see. I missed the fact that committedSpecies is an array of species, not a single species 😄

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: That's done

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -158,42 +205,129 @@ export const handleSelectedSpecies: ActionCreator<
...
+    }
+  });
+  dispatch(toggleSpeciesUse(speciesName, currentStatus, committedSpecies));
   speciesSelectorStorageService.saveSelectedSpecies(committedSpecies);
 };
 
-export const deleteSpecies = createStandardAction(
-  'species_selector/delete_species'
-)<string>();
+export const deleteSpecies = createAction(
```
azangru: Don't `deleteSpecies` and `toggleSpecies` actions now do exactly the same thing? They receive updated list of committedSpecies and pass it to the reducer (while also sending analytics)? If that's how they had to change, shouldn't we rename this function (`updateCommittedSpecies` or something) and unite them in one?

Are there other actions that are now doing this same thing?

azangru: It can then have a signature of `updateCommittedSpecies(committedSpecies, analyticsPayload)`.

msmanoj: That's done

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -48,19 +48,24 @@ export const PopularSpeciesButton = (props: Props) => {
...
+    const speciesName = species.common_name || species.scientific_name;
+
     if (!is_available) {
       return;
     } else if (isSelected) {
       props.clearSelectedSpecies();
+      analyticsTracking.trackEvent({
+        category: 'popular_species',
+        action: 'unpreselect',
+        label: speciesName
```
azangru: Let's move this out of the body of the component function into a separate function (`sendAnalytics`?) that takes action (preselect/unpreselect) and speciesName and sends a track event?

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -312,7 +267,13 @@ export const deleteSpeciesAndSave: ActionCreator<
     return true;
   });
 
-  dispatch(deleteSpecies(speciesName, committedSpecies));
+  analyticsTracking.trackEvent({
+    category: categories.SELECTED_SPECIES,
+    label: speciesName,
+    action: 'unselected'
```
azangru: the doc says, unselect rather than unselected

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
msmanoj: How would creating a separate function be useful than directly calling `analyticsTracking.trackEvent`?

azangru: There are two calls to it in the body of the component function: one in the else if branch, another in the else branch. Moving this bit of code out of the body of the component function (I don't mean to a different file) will tighten up the code slightly.

msmanoj: I'm not sure if it's a good idea to create another wrapper for the analytics just to reduce 2 lines of code?

azangru: 🤷‍♂ 

ok, let's leave it as it is then.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: Updated it 👍

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -57,6 +60,15 @@ export const SpeciesSearchField = (props: Props) => {
 
   const onMatchSelected = (match: SearchMatch) => {
     props.onMatchSelected(match);
+
+    const speciesName = `${match.common_name || match.scientific_name} - ${
+      match.assembly_name
+    }`;
```
azangru: Why is the species name here set differently than in PopularSpeciesButton or in speciesSelectorActions (where it doesn't include assembly name)? Is it intentional?

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -81,14 +109,23 @@ export const setSelectedSpecies = createStandardAction(
...
 )<SearchMatch | PopularSpecies>();
 
+export const clearSearchResults = createAction(
+  'species_selector/clear_search_results',
+  (resolve) => {
+    return () =>
+      resolve(
+        undefined,
+        buildAnalyticsObject({
+          category: categories.POPULAR_SPECIES,
```
azangru: Erm. Why is it using the PopularSpecies category?

azangru: Also, it seems we don't need to track this action at all.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -158,42 +195,100 @@ export const handleSelectedSpecies: ActionCreator<
...
-  dispatch(deleteSpecies(genomeId));
-  const committedSpecies = getCommittedSpecies(getState());
+  let speciesName = '';
+  const committedSpecies = getCommittedSpecies(getState()).filter((item) => {
+    if (item.genome_id === genomeId) {
+      speciesName = item.common_name || item.scientific_name;
+      return false;
+    }
+    return true;
+  });
```
azangru: To avoid doing side effects outside of the filter function (which is not something one would expect), could we do:

```
const committedSpecies = getCommittedSpecies(getState())
const deletedSpecies = find(committedSpecies, ({genome_id}) => genome_id === genomeId);
const deletedSpeciesName = deletedSpecies.common_name || deletedSpecies.scientific_name;
const updatedCommittedSpecies = committedSpecies.filter(({genome_id}) => genome_id !== genomeId);

...do remaining stuff
```

msmanoj: That's done

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: Yes, it shouldn't be. Not sure why I left it there. Removed it now.

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
msmanoj: It was just species name in the document but it makes sense to include the assembly name as well. I've updated it.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -48,7 +48,9 @@ export const PopularSpeciesButton = (props: Props) => {
 
   const handleClick = () => {
     const { genome_id, is_available } = species;
-    const speciesName = species.common_name || species.scientific_name;
+    const speciesName = `${species.common_name || species.scientific_name} - ${
+      species.assembly_name
+    }`;
```
azangru: It looks like we are now doing the same calculation in 4 different places? Perhaps extract into a helper (`getSpeciesAnalyticsName`)?

PR #117, opened 2019-07-19
Feature/privacy banner

jyothishnt:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-162

## Importance
New versions of GDPR privacy policies should be accepted by users. Thus we store version as the cookie value and compare against it.

## Description

- Added gdpr cookie version name
- Moved hardcoded entries to config
- Fixed flickering of banner due to the boolean being set to true on the initial component load

## Views affected
All pages - as banner should be shown on all pages until they are accepted.

### Other effects
renamed "version" (app version) in config and updated associated files where it is being used.

- [ ] I have checked any requirements for Google Analytics **N/A**
- [ ] I have created any required tests **N/A**
- [ ] The PR has as its final commit a WASM/JS update commit if any Rust files were updated. **N/A**

## Possible complications
None

### DIFF
--- a/src/ensembl/config.ts
+++ b/src/ensembl/config.ts
@@ -1,9 +1,6 @@
 export default {
   // Version numbers
   app_version: '0.2.0',
-  privacy_policy_version: '2.0.0',
-  privacy_policy_cookie_name: 'ENSEMBL_PRIVACY_POLICY',
-  privacy_policy_cookie_expiry: 'Sun, 01 Jan 2040 00:00:00 GMT',
 
   // Node environment
   isDevelopment: process.env.NODE_ENV === 'development',
--- a/src/ensembl/config.ts
+++ b/src/ensembl/config.ts
@@ -1,6 +1,9 @@
 export default {
-  // version number
-  version: '0.2.0',
+  // Version numbers
+  app_version: '0.2.0',
+  privacy_policy_version: '2.0.0',
+  privacy_policy_cookie_name: 'ENSEMBL_PRIVACY_POLICY',
+  privacy_policy_cookie_expiry: 'Sun, 01 Jan 2040 00:00:00 GMT',
 
   // Node environment
   isDevelopment: process.env.NODE_ENV === 'development',
@@ -11,7 +14,7 @@ export default {
 
   apiHost: process.env.API_HOST,
 
-  // keys for services
+  // Keys for services
   googleAnalyticsKey: process.env.GOOGLE_ANALYTICS_KEY,
 
   // Genesearch endpoint
--- a/src/ensembl/src/shared/privacy-banner/privacy-banner-service.ts
+++ b/src/ensembl/src/shared/privacy-banner/privacy-banner-service.ts
@@ -1,13 +1,7 @@
+import * as privacyConfig from './privacyConfig';
 import storageService, {
   StorageServiceInterface
 } from 'src/services/storage-service';
-
-export const PrivacyConfig = {
-  name: 'ensembl_privacy_policy',
-  version: '2.0.0',
-  policyUrl: 'https://www.ebi.ac.uk/data-protection/ensembl/privacy-notice',
-  termsUrl: 'https://www.ebi.ac.uk/about/terms-of-use'
-};
 
 export class PrivacyBannerService {
   private storageService: StorageServiceInterface;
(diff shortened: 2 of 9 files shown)


jyothishnt:
Working on tests now. :-)

ens-st3:
Shouldn't be accepted until tests are added I guess

src/ensembl/config.ts
```diff
@@ -1,6 +1,9 @@
 export default {
-  // version number
-  version: '0.2.0',
+  // Version numbers
+  app_version: '0.2.0',
+  privacy_policy_version: '2.0.0',
+  privacy_policy_cookie_name: 'ENSEMBL_PRIVACY_POLICY',
+  privacy_policy_cookie_expiry: 'Sun, 01 Jan 2040 00:00:00 GMT',
```
azangru: These three extra fields seem to serve one single purpose of showing the privacy banner. I would put them in a constants file in the `shared/privacy-banner` folder, not in the general config.

azangru: Another thing, which I talked about with @imran-ebi, is that we may want to change the storage mechanism for whether the user has seen and agreed to the banner from cookies to localStorage. LocalStorage is much more spacious than cookies (which are up to 4 kb or something?), and stays in the browser, whereas cookies are sent back and forth between the server and the browser. This doesn't look like something that belongs in cookies.

jyothishnt: Do you have an example of such usage that I can have a look?

jyothishnt:
> Shouldn't be accepted until tests are added I guess

Already done :)

src/ensembl/config.ts
azangru: > privacy_policy_cookie_expiry: 'Sun, 01 Jan 2040 00:00:00 GMT',

This looks like "never" 😃 If we move privacy banner data to localStorage, we won't have to specify the expiry date.

azangru: > Do you have an example of such usage that I can have a look?

@jyo search the project for browser-storage-service or species-selector-storage-service.

jyothishnt:
Working on PrivacyBannerService using LocalStorage

jyothishnt:
All good for review. Thanks.

jyothishnt:
Working on style changes mentioned on the ticket

src/ensembl/src/shared/privacy-banner/privacy-banner-service.ts
```diff
@@ -0,0 +1,33 @@
+import storageService, {
+  StorageServiceInterface
+} from 'src/services/storage-service';
+
+export const PrivacyConfig = {
```
imran-ebi: I am wondering if we should have the privacy config in its own file rather than inside the privacy banner service 🤔

azangru: I would say, in its own file inside the privacy-banner folder.

jyothishnt: shall I go ahead and export the configs in another file? Is that what you meant?

imran-ebi: Yes, go ahead.

src/ensembl/src/shared/privacy-banner/privacy-banner-service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  public getPolicyVersion(): string {
+    return this.storageService.get(PrivacyConfig.name);
+  }
+
+  public setPolicyVersion() {
+    this.storageService.save(PrivacyConfig.name, PrivacyConfig.version);
+  }
+
+  // Returns true if versions doesn't match
```
azangru: This is an unnecessary comment :-)

azangru: But if you read the body of this function you see that it returns true if `this.getPolicyVersion()` is the same as `privacyConfig.version` and false otherwise. This comment doesn't add anything to the understanding of the code.

In the test file, we can incorporate the comments into the variable names, which is always better than comments.  For example:

```
      const returnDifferentVersion = () =>faker.lorem.word();
      jest
        .spyOn(privacyBannerService, 'getPolicyVersion')
        .mockImplementation(returnDifferentVersion);

      const result = privacyBannerService.shouldShowBanner();
```

```
      const privacyBannerService = new PrivacyBannerService(mockStorageService);
      const returnSameVersion = () => privacyConfig.version;
      jest
        .spyOn(privacyBannerService, 'getPolicyVersion')
        .mockImplementation(returnSameVersion);

      const result = privacyBannerService.shouldShowBanner();
```

PR #118, opened 2019-07-19
Fix drawer state

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4296

## Description
The drawer state currently is not unique to a species instance. So if the user open the drawer in a particular species and then clicks on another species, the drawer will be at the same state as the previous species. This is not ideal user behaviour. This PR makes the drawer state unique to a species instance, thus making the drawer open/close only if the user has already done so for that species.

## Views affected
Genome browser


imran-ebi:
I noticed when working on this bug that a lot of the browser components don't have tests. I think it warrants a PR of its own, so I didn't try to add any tests for now.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -55,15 +55,15 @@ import styles from './Browser.scss';
 import 'static/browser/browser.js';
 
 type StateProps = {
-  activeGenomeId: string | null;
```
azangru: But it can be null! If storage-service does not find active genome id in localStorage, it will return null:

https://github.com/Ensembl/ensembl-client/blob/9ac0015a723c8148fb3c5c5eaf70ab158d346442/src/ensembl/src/services/storage-service.ts#L55

The typing here is incorrect:
https://github.com/Ensembl/ensembl-client/blob/e963f12cd513415b7ceb7b1bdb1f9fbc4f428eae/src/ensembl/src/content/app/browser/browser-storage-service.ts#L29

That’s why I added the nulls during the last refactor.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -55,15 +55,15 @@ import styles from './Browser.scss';
...
+  activeGenomeId: string;
   activeEnsObjectId: string | null;
   allActiveEnsObjectIds: { [genomeId: string]: string };
   browserActivated: boolean;
   browserNavOpened: boolean;
   browserQueryParams: { [key: string]: string };
   chrLocation: ChrLocation | null;
   allChrLocations: { [genomeId: string]: ChrLocation };
-  drawerOpened: boolean;
+  drawerOpened: { [genomeId: string]: boolean };
```
azangru: Why are we getting the whole object of all drawer states for all genomeId? Aren’t we interested only in one, for the active genome?

I vote for having it as a boolean. Also, if possible, can we have it with an auxiliary verb (`isDrawerOpened`) so that it reads as a question?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -102,6 +102,7 @@ export const Browser: FunctionComponent<BrowserProps> = (
     TrackStates
   >({});
   const lastGenomeIdRef = useRef(props.activeGenomeId);
+  const drawerOpenedForGenome = props.drawerOpened[props.activeGenomeId];
```
azangru: This belongs in the selector, not in the component itself. The component shouldn't care about genomes other than active.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -1,16 +1,61 @@
...
+import { getBrowserActiveGenomeId } from '../browserSelectors';
 
-export const changeDrawerView = createAction(
-  'drawer/change-drawer-view',
+export const changeDrawerViewForGenome = createAction(
+  'drawer/update-drawer-view',
   (resolve) => {
-    return (drawerView: string) =>
+    return (drawerView: { [genomeId: string]: string }) =>
       resolve(drawerView, getDrawerAnalyticsObject('User Interaction'));
```
azangru: This is a useless analytics event.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -1,16 +1,61 @@
...
 );
 
-export const toggleDrawer = createAction('drawer/toggle-drawer', (resolve) => {
-  return (drawerOpened?: boolean) =>
-    resolve(drawerOpened, getDrawerAnalyticsObject('User Interaction'));
-});
+export const changeDrawerView: ActionCreator<
+  ThunkAction<void, any, null, Action<string>>
+> = (drawerViewForGenome: string) => (
+  dispatch: Dispatch,
```
azangru: the dispatch in the thunk is not of type `Dispatch`, but of type `ThunkDispatch`, or something like that. Functions of type `Dispatch` can not dispatch other thunks; while functions of type `ThunkDispatch` can. In fact, if you delete `Dispatch`, you will see that typescript can infer the correct type of `dispatch` by itself, from the `ThunkAction` type.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -1,16 +1,61 @@
...
+  };
+
+  dispatch(changeDrawerViewForGenome(drawerView));
+};
+
+export const toggleDrawerForGenome = createAction(
+  'drawer/toggle-drawer',
+  (resolve) => {
+    return (drawerOpened: { [genomeId: string]: boolean }) =>
+      resolve(drawerOpened, getDrawerAnalyticsObject('User Interaction'));
```
azangru: "User Interaction" is a useless analytics event

src/ensembl/src/content/app/browser/drawer/drawerReducer.ts
```diff
@@ -8,16 +8,22 @@ export default function drawer(
...
-        drawerView: action.payload
+        drawerView: { ...state.drawerView, ...action.payload }
       };
-    case getType(drawerActions.toggleDrawer):
+    case getType(drawerActions.toggleDrawerForGenome):
+      const genomeId = Object.keys(action.payload)[0];
+      let drawerOpenedValue = Object.values(action.payload)[0];
+      let drawerViewForGenome = drawerOpenedValue
+        ? state.drawerView[genomeId]
+        : '';
```
azangru: Very confused about what’s going on here:

1) If the payload structure of type `{ [genomeId: string]: boolean }` is not convenient enough, because you have to get the keys of this object (which is only one key) and all values of this object (which is only one value), then perhaps the payload of type `{ genomeId: string, isOpen: boolean }` will be better suited for the task?

2) don’t understand what’s going on with drawerViewForGenome

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
```diff
@@ -24,13 +25,15 @@ type TrackPanelBarProps = {
 const TrackPanelBar: FunctionComponent<TrackPanelBarProps> = (
   props: TrackPanelBarProps
 ) => {
+  const drawerOpenedForGenome = props.drawerOpened[props.activeGenomeId];
+
   const moveTrackPanel = useCallback(() => {
-    if (props.drawerOpened === true) {
+    if (drawerOpenedForGenome === true) {
```
azangru: can we, instead of writing `=== true`, write something like `if (isDrawerOpened)`? Firstly, it reads like a sentence; and secondly, typescript won’t allow the value that is declared to be of type boolean to be of any other potentially truthy type. So `if (isDrawerOpened)` will be exactly the same as `if (drawerOpenedForGenome === true)`, but less wordy.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
imran-ebi: Why is this useless? Don't we want to know what drawer views users see the most?

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
imran-ebi: I agree its name useless, but I think we'll still need to leave this to track how many times a user opens a drawer. So perhaps changing the event to a more meaningful name will resolve this?

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
azangru: If I read the correctly, the only information that will be sent here to GA is category drawer, label  "User Interaction", which doesn't tell us much.

This analytics event is removed in Manoj's PR https://github.com/Ensembl/ensembl-client/pull/113/files

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
imran-ebi: I am not really a fan of these truthy and falsy values as it is not used in most static typing languages. However, just to be consistent with the rest of the developers, I'll remove the explicit checks for `true` and `false`.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
imran-ebi: I have removed the analytics event for this. As discussed offline, we can add this later.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -253,9 +255,10 @@ export const Browser: FunctionComponent<BrowserProps> = (
   }, [props.chrLocation]);
 
   const closeTrack = () => {
-    if (props.drawerOpened === false) {
+    if (isDrawerOpened) {
```
azangru: But now it’s the opposite of what is deleted above?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -266,7 +269,7 @@ export const Browser: FunctionComponent<BrowserProps> = (
   }));
 
   const getBrowserWidth = (): string => {
-    if (props.drawerOpened) {
+    if (isDrawerOpened) {
       return 'calc(41px + 0vw)';
```
azangru: 0 is always 0; doesn't matter whether in pixels or in vws :smile:

src/ensembl/src/content/app/browser/browser-genome-selector/BrowserGenomeSelector.tsx
```diff
@@ -43,15 +43,15 @@ const BrowserGenomeSelector: FunctionComponent<BrowserGenomeSelectorProps> = (
   const getGenomeSelectorClasses = () => {
     let classNames = styles.browserGenomeSelector;
 
-    if (props.drawerOpened === true) {
+    if (isDrawerOpened) {
       classNames += ` ${styles.browserGenomeSelectorDisabled}`;
```
azangru: should we do this using the `classnames` package?

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
```diff
@@ -23,23 +22,21 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
...
-    const selectedBrowserTab =
-      props.selectedBrowserTab[activeGenomeId] || TrackType.GENOMIC;
+    const { isDrawerOpened, selectedBrowserTab, trackPanelModalOpened } = props;
     let classNames = styles.browserTab;
 
     if (
       props.ensObject.genome_id &&
       selectedBrowserTab === trackType &&
-      drawerOpened === false &&
+      isDrawerOpened === false &&
```
azangru: !isDrawerOpened ?

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
```diff
@@ -23,23 +22,21 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
...
       trackPanelModalOpened === false
```
azangru: !trackPanelModalOpened ?

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: Ah nice spot. Will fix it 😄

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: TypeScript does not infer the return type as string properly and that's why I explicitly added `string` as the return type there.

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
imran-ebi: I'll also change the variable to `isTrackPanelModalOpened`

src/ensembl/src/content/app/browser/drawer/drawerReducer.ts
imran-ebi: I have refactored this part of the code in 494761c. What I was trying to do here was to check if the drawer is closed and if so to reset the drawer view.

src/ensembl/src/content/app/browser/Browser.tsx
azangru: Sorry. What I meant here was that the `'calc(41px + 0vw)';` string doesn't make sense on several levels:
- `0vw` is the same as just 0
- 41px + 0 is 41 px
- therefore, calc is unnecessary

imran-ebi: Ah I get what you mean. I'll fix that.

imran-ebi: Unfortunately the fix for this breaks the `react-spring` animation. So I had to put the same value back. I found out from @msmanoj, who worked on this animation, that `react-spring` requires all widths to be in the same format.

azangru: Could you please add a comment (something like `calc and 0vw needed to preserve same format across different ReactSpring values`, so that we don't get confused the next time?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -58,24 +58,25 @@ type StateProps = {
   activeGenomeId: string | null;
   activeEnsObjectId: string | null;
   allActiveEnsObjectIds: { [genomeId: string]: string };
+  allChrLocations: { [genomeId: string]: ChrLocation };
```
azangru: As @msmanoj reminded me in browser-messaging-service PR, we now have `ChrLocations` type.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -211,11 +214,17 @@ export const Browser: FunctionComponent<BrowserProps> = (
     }
   }, [props.browserActivated]);
 
+  const closeDrawer = () => {
+    props.changeDrawerView('');
+    props.toggleDrawer(false);
+  };
```
azangru: Why does `closeDrawer` belong to `Browser` component? It is passed from Browser to BrowserBar and TrackPanel components. Given that both `BrowserBar` and `TrackPanel` are connected to redux, why can't they import `closeDrawer` action creator directly from browser actions?

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: @azangru I have added a comment in d15edff

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -75,7 +79,7 @@ const TrackPanelListItem: FunctionComponent<TrackPanelListItemProps> = (
       classNames += ` ${styles.main}`;
     }
 
-    if (drawerView === track.track_id) {
+    if (track.track_id === drawerView) {
       classNames += ` ${styles.currentDrawerView}`;
```
azangru: Let's use classNames?

imran-ebi: Made the changes in bab06f9

src/ensembl/src/content/app/browser/Browser.tsx
imran-ebi: Created a `closeDrawer` action in 8611954

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -51,3 +51,25 @@ export const toggleDrawer: ActionCreator<
...
+  dispatch(
+    toggleDrawerForGenome({
+      [activeGenomeId]: false
+    })
+  );
+
+  dispatch(
+    changeDrawerViewForGenome({
+      [activeGenomeId]: ''
+    })
```
azangru: We can batch these two together?

https://react-redux.js.org/api/batch

imran-ebi: This is done now

src/ensembl/src/content/app/browser/drawer/drawerReducer.ts
azangru: Now it's so much cleaner 👍

PR #120, opened 2019-07-22
Custom downloads July

msmanoj:
## Type
- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)

- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-196 - Get gene symbol from the API
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-197 - Add cDNA sequence attribute to the Sequence section
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-198 - Add a Biotype selector to the 'Transcript type' and Gene type' filters
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-199 - Create a Paste identifier component
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-200 - Re-think active filters displayed on the Preview Download screen
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-201 - Implement the option to reset the attribute/filter selections
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-204 - Split Gene stable ID version into Gene stable ID and version
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-262 - Orthologue selector bugs
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-202 - Use LocalStorage for Custom Downloads

## Importance
July release

## Description
- Covered all the tasks for first round of user testing mentioned in the ticket ENSWBSITES-194.
- Removed all analytics tracking from custom downloads

Created the following components:
- ContentBuilder
- PasteOrUpload
- CheckboxWithTextfields

A working copy if this branch is currently available on the following URL:
http://193.62.55.91:30180/app/custom-download

## Views affected
http://193.62.55.91:30180/app/custom-download

### Other effects
- [ ] The PR has as its final commit a WASM/JS update commit if any Rust files were


.gitlab-ci.yml
```diff
@@ -27,6 +28,7 @@ Test:
   except:
   - dev
   - master
+  - feature/custom-downloads-july
```
azangru: probably should delete before merging

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.test.tsx
```diff
@@ -39,11 +39,11 @@ describe('<CheckboxWithSelects />', () => {
     expect(wrapper.find(Checkbox).prop('checked')).toBe(false);
   });
 
-  it('does not display any Select when the checkbox is unchecked', () => {
+  it('displaya one Select when the checkbox is unchecked', () => {
```
azangru: displaya -> displays

azangru:
Styling problem (seen in Chromium on Linux):

- open Custom Download page
- switch to Filters
- open Genes and Transcripts accordion
- make sure the window is short enough to allow vertical scrolling
- scroll down
- notice how grey rectangles masking checkboxes travel down with the scroll

image

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
```diff
@@ -7,25 +7,32 @@ import ImageButton from 'src/shared/image-button/ImageButton';
...
+    }
   };
 
   const handleOnSelect = (value: string, selectIndex?: number) => {
     const newSelectedOptions: string[] = [...props.selectedOptions];
-    if (selectIndex) {
+    if (selectIndex !== undefined) {
       newSelectedOptions[selectIndex] = value;
     } else {
       newSelectedOptions.push(value);
```
azangru: line 48 (can’t leave comment directly): ` if (selectedOptionsClone.indexOf(option.value) === -1) ` ->  `if (selectedOptionsClone.includes(option.value))`

azangru: also, `removeSelection` looks like it could be written with a simple `filter` function

azangru:
Another general styling question — is this white space below the columns expected? Looks strange to me:

image

src/ensembl/src/content/app/custom-download/sample-data/attributes.ts
```diff
@@ -0,0 +1,208 @@
...
+      {
+        type: 'checkbox_grid',
+        label: '',
+        id: 'default',
+        options: [
+          {
+            id: 'name',
+            label: 'Gene name',
+            isChecked: true,
+            order: 1
```
azangru: What't the `order` field for? If it's to order the options one after another, then their relative positions inside the array should be enough for that.

msmanoj: If the options needs to be logically ordered, would it not be better to have this field rather than relying on the relative position in the array?

src/ensembl/src/content/app/custom-download/components/checkbox-with-selects/CheckboxWithSelects.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/sample-data/attributes.ts
azangru: That logical ordering would happen on the backend, presumably? Why wouldn't it sort the options appropriately and put them in the array in the correct order?

msmanoj: It would be great if it can but it shouldn't be a big problem to remove it later on?

azangru:
Thinking of components as `checkbox-with-something` (and there are many such components) feels really wrong. It seems that what the code and design are trying to say is that there are RadioList, SelectList, PasteOrUploadList (or whatever that last thing is; the word "upload" may be wrong as well); and all these components have a visibility switch, which can be modelled as:

```
<VisibilitySwitch> // <-- perhaps not a real name of the component
  <RadioLIst>
</VisibilitySwitch>
```

or

```
withVisibilitySwitch(RadioLIst)
```

or

```
const RadioList = () => {
  [isVisible] = useVisibilitySwitch(false);
}
```

I am not sure I am saying these components need to be re-written in this PR, but they feel wrong.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const newValue = faker.random.words();
+    wrapper
+      .find(PasteOrUpload)
+      .last()
+      .prop('onChange')(newValue);
+    expect(onChange).toBeCalledWith([newValue]);
+  });
+
+  it('updates the values when an inpus is removed', () => {
```
azangru: inpus -> input

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -0,0 +1,131 @@
...
+  values: string[];
+  label: string;
+  allowMultiple: boolean;
+  disabled?: boolean;
+  onChange: (values: string[]) => void;
+};
+
+const CheckboxWithTextfields = (props: CheckboxWithTextfieldsProps) => {
+  const [isChecked, setIsChecked] = useState(false);
+  const [values, setValues] = useState<(string | null)[]>(props.values);
```
azangru: Is there any reason to have these values in local state, when they are updated from props, and any change in the input calls onChange in the props?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const [shouldShowAddButton, setShowAddButton] = useState(
+    Boolean(values[values.length - 1])
+  );
```
azangru: Is this not directly calculable from props (as done in useEffect below)? Why store it in state?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const newValues: string[] = [...values] as string[];
+
+    newValues.splice(index, 1);
+    setShowAddButton(Boolean(newValues[newValues.length - 1]));
+
+    props.onChange(newValues);
+  };
+
+  const valuesWithoutFirst: string[] = [...values] as string[];
+  valuesWithoutFirst.shift();
```
azangru: `const valuesWithoutFirst = values.slice(1);`
but why do you need them?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -0,0 +1,131 @@
...
+                        value={value}
+                        onChange={(newValue) =>
+                          handleOnChange(newValue, key + 1)
+                        }
+                        onRemove={() => handleOnRemove(key + 1)}
+                        placeholder={'Paste data'}
+                      />
+                    </div>
+                  );
+                })}
```
azangru: What is the difference in treatment of first value as opposed to valuesWithoutFirst?

src/ensembl/src/content/app/custom-download/sample-data/attributes.ts
ens-st3: It will. But first we need a BE

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -0,0 +1,263 @@
...
+type Path = (string | number)[];
+let path: Path = [];
+
+type ContentBuilderProps = {
+  data: AttributeWithContent;
+  selectedData: JSONValue;
+  onChange: (type: string, path: Path, payload: PrimitiveOrArrayValue) => void;
+  contentState: JSONValue;
+  onContentStateChange: (
+    type: string,
```
azangru: The `type` argument doesn’t seem to be used by the callbacks in the parent components?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -14,26 +14,31 @@ import {
 import { getAttributesAccordionExpandedPanel } from './state/attributesAccordionSelector';
```
azangru: in attributesAccordionSelector there is an import statement with a relative path going 4 levels up (`import Attributes, { AttributeWithOptions } from '../../../../types/Attributes';`). Please use a path starting from `src` instead.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionState.ts
```diff
@@ -1,27 +1,37 @@
-import { CustomDownloadAttributes } from 'src/content/app/custom-download/types/Attributes';
+import Attributes, {
```
azangru: image

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionReducer.ts
```diff
@@ -90,20 +71,8 @@ function attributesAccordion(
...
-): CustomDownloadAttributes {
-  switch (action.type) {
-    case getType(attributesAccordionActions.setGeneAttributes):
-      return { ...state, gene: action.payload };
-    default:
-      return state;
-  }
-}
-
 function Orthologue(
```
azangru: We do not begin function names with capital letters. Except when they are components (or classes).

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionSelector.ts
```diff
@@ -1,47 +1,39 @@
...
 
-export const getAttributes = (state: RootState): {} =>
-  state.customDownload.attributesAccordion.attributes;
+import Attributes, { AttributeWithOptions } from '../../../../types/Attributes';
+import JSONValue from 'src/shared/types/JSON';
+import { CheckboxGridOption } from 'src/content/app/custom-download/components/checkbox-grid/CheckboxGrid';
 
-export const getGeneAttributes = (state: RootState): {} =>
-  state.customDownload.attributesAccordion.attributes.gene;
+export const getAttributes = (state: RootState): Attributes =>
```
azangru: Where is this selector used?

azangru:
General point: there are now parts of custom download redux state that:
- sit in the `containers` folder (as opposed to the `state` folder, which is also present)
- have the ui-related word `accordion` in their name:

image

If I understand correctly what these pieces of state are doing (I am not yet quite sure that I do), they would it would be much better if they:
- were called `attributesState` and `filtersState` (no `accordion` in the name)
- sat in sub-folders of the `custom-download/state` folder

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionState.ts
```diff
@@ -1,27 +1,37 @@
...
 
 export type AttributesAccordionState = Readonly<{
   expandedPanel: string;
-  expandedVariationPanels: string[];
-  attributes: CustomDownloadAttributes;
+  content: {
+    [key: string]:
+      | AttributeWithOptions
+      | { [key: string]: AttributeWithOptions };
+  };
```
azangru: what’s the difference between `content` and `contentState` (below)?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionReducer.ts
```diff
@@ -123,9 +92,9 @@ function Orthologue(
 }
 
 function OrthologueAttributes(
-  state: any = defaultAttributesAccordionState.attributes,
+  state: any = defaultAttributesAccordionState.content,
```
azangru: defaultAttributesAccordionState.content now has a type that is not `any`

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.scss
```diff
@@ -1,14 +1,21 @@
...
 
-.checkboxGridWrapper {
-  padding: 20px;
+.checkboxWithSelectWrapper {
+  padding-left: 20px;
+  padding-top: 10px;
 }
-.geneAccordion {
+
+.checkboxWitRadiosWrapper {
```
azangru: Wit -> With

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const buildCheckboxGrid = (entry: AttributeWithOptions, path: Path) => {
+    const currentPath = [...path, entry.id];
+
+    const selectedOptions = get(props.selectedData, currentPath, []);
+
+    const gridOptions = [...entry.options];
+
+    const gridClone: CheckboxGridOption[] = [];
+
+    Object.values(gridOptions as CheckboxGridOption[]).map((option) => {
```
azangru: Why are you running Object.values on an array? :scream_cat:

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -0,0 +1,263 @@
...
+      const optionClone = { ...option };
+      if (selectedOptions[option.id] !== undefined) {
+        optionClone.isChecked = selectedOptions[option.id];
+      } else if (optionClone.isChecked) {
+        onChangeHandler(entry.type, [...currentPath, optionClone.id], true);
+      }
+      gridClone.push(optionClone);
+    });
```
azangru: Please do not use `.map` as `.forEach`! Also, please do not do side effects inside a `.map`!

If you want to call `onChangeHandler` for all checked values while building the checkboxGrid (will it result in calling a redux action for every isChecked?), then run a separate `forEach`:

Also, there was no point in declaring the `gridClone` variable as an empty array. `.map` function returns a new array.

```
gridOptions.forEach((option) => {
  if (option.isChecked) {
     onChangeHandler(entry.type, [...currentPath, option.id], true);
   }

gridOptions = gridOptions.map(option => selectedOptions[option.id]  ? {
  ...option,
   isChecked: selectedOptions[option.id]
} : option;
```

src/ensembl/src/content/app/custom-download/components/paste-or-upload/PasteOrUpload.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  it('displays the text "Paste data or Upload file"', () => {
+    wrapper = mount(<PasteOrUpload {...defaultProps} />);
+
+    expect(wrapper.text()).toBe('Paste data or Upload file');
+  });
+
+  it('does not display the text "Paste data or Upload file" when a value is passed', () => {
+    wrapper = mount(<PasteOrUpload {...defaultProps} value={'foo'} />);
+
+    expect(wrapper.text()).not.toBe('Paste data or Upload file');
```
azangru: do we expect any `wrapper.text()` at all?

src/ensembl/src/content/app/custom-download/components/paste-or-upload/PasteOrUpload.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const [shouldShowInput, showInput] = useState(props.value !== null);
+  const [value, setValue] = useState<string | null>(null);
+
+  useEffect(() => {
+    setValue(props.value);
+  }, [props.value]);
+
+  const [shouldShowFileUpload, showFileUpload] = useState(false);
+  const onChangeHandler = (value: string) => {
+    props.onChange(value);
```
azangru: So if every change event of the input calls onChange on the props, and props, when changed, update component’s state, then what’s the point of having component’s state?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -14,26 +14,31 @@ import {
...
+import JSONValue from 'src/shared/types/JSON';
+
 type Props = StateProps & DispatchProps;
 
 const AttributesAccordion = (props: Props) => {
   useEffect(() => {
     props.fetchAttributes();
+    props.updateSelectedAttributes(
+      customDownloadStorageService.getSelectedAttributes()
+    );
```
azangru: Should reading from and writing to storage perhaps be moved to the redux layer?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -55,11 +60,36 @@ const AttributesAccordion = (props: Props) => {
...
+        <span className={styles.resetIcon} onClick={onReset}>
+          <ImageButton
+            buttonStatus={ImageButtonStatus.ACTIVE}
+            description={'Reset attributes'}
+            image={ResetIcon}
+          />
+        </span>
       </div>
       <Accordion
         preExpanded={Array(1).fill(props.expandedPanel)}
```
azangru: `Array(1).fill(props.expandedPanel)` -> `[props.expandedPanel]` ?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -55,11 +60,36 @@ const AttributesAccordion = (props: Props) => {
...
 
+  const onReset = () => {
+    props.resetSelectedAttributes();
+    customDownloadStorageService.saveSelectedAttributes({});
+  };
+
+  const buildSection = (
+    section: string,
+    hideTitles?: boolean,
+    hideUnchecked?: boolean
```
azangru: can we, instead of two boolean flags passed (sometimes) as individual parameters `buildSection('genes')`, `buildSection('genes', true, true)` use one argument for `buildSection` which is an object of shape:

```
{
  section: string, // or perhaps enum
  hideTitles?: boolean,
  hideUnchecked?: boolean
}
```

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/AttributesAccordionSection.tsx
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  return (
+    <ContentBuilder
+      data={allAttributes[props.section] as AttributeWithContent}
+      onChange={onChangeHandler}
+      contentState={props.contentState}
+      onContentStateChange={onContentStateChangeHandler}
+      selectedData={props.selectedAttributes}
+      contentProps={{ checkbox_grid: { hideUnchecked: props.hideUnchecked } }}
```
azangru: The name `contentProps` is confusing (so is `contentState`, for that matter, but when there is also `contentProps`, doubly so). Is it some kind of `presentationOptions` (not really related to the "content", or options that the user selects for passing to the backend)?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/AttributesAccordionSection.tsx
```diff
@@ -0,0 +1,90 @@
...
+    />
+  );
+};
+
+type DispatchProps = {
+  updateSelectedAttributes: (updateSelectedAttributes: JSONValue) => void;
+  updateContentState: (updateContentState: JSONValue) => void;
```
azangru: naming of arguments:

`updateSelectedAttributes: JSONValue` -> `attributes: JSONValue` ?
`updateContentState: JSONValue` -> `content: JSONValue` ?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -42,38 +47,56 @@ type Props = ownProps & StateProps & DispatchProps;
 const Orthologue = (props: Props) => {
   const attributesOnChangeHandler = (
     status: boolean,
-    subSection: string,
+    species: string,
     attributeId: string
   ) => {
-    const newOrthologueAttributes = { ...props.orthologueAttributes };
+    const newOrthologueAttributes: any = { ...props.orthologueAttributes };
```
azangru: why `any`?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
```diff
@@ -42,38 +47,56 @@ type Props = ownProps & StateProps & DispatchProps;
...
+      newOrthologueFilteredSpecies[modifiedSpeciesIndex].label;
 
-    newOrthologueFilteredSpecies[subSection][attributeId].isChecked = status;
+    newOrthologueFilteredSpecies[modifiedSpeciesIndex].isChecked = status;
 
     props.setOrthologueSpecies(newOrthologueFilteredSpecies);
 
-    const newOrthologueAttributes = { ...props.orthologueAttributes };
-    const sectionHeader = props.orthologueSpecies.default[attributeId].label;
+    const newOrthologueAttributes: any = { ...props.orthologueAttributes };
```
azangru: why `any`?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: This is just to implement the option to display a new input field when the add button is clicked, by pushing `null` to the values.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: The add button gets hidden when clicked ,displaying another input whilst the props remains the same. Does that make sense?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: That's not needed anymore. The logic was a bit different initially and I missed to remove this when I changed it.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.test.tsx
msmanoj: Done

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: Related to the comment above. This logic has been updated.

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
msmanoj: It is not used now but I think it might be useful in the future to identify what type of data is getting stored.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionState.ts
msmanoj: Removed it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionReducer.ts
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionSelector.ts
msmanoj: Updated the following file to use it:
`src/content/app/custom-download/containers/content/attributes-accordion/sections/AttributesAccordionSection.tsx`

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionReducer.ts
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.scss
msmanoj: Updated

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
msmanoj: Forget to remove it. It was an object before and then I changed it to be an array.

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
msmanoj: >will it result in calling a redux action for every isChecked? 
It looks like it is supposed to dispatch many actions but it wasn't. Not sure why.
I've anyways updated this logic to call the onChange once all the options are processed.

src/ensembl/src/content/app/custom-download/components/paste-or-upload/PasteOrUpload.test.tsx
msmanoj: Updated the test to check if the text is ''.

src/ensembl/src/content/app/custom-download/components/paste-or-upload/PasteOrUpload.tsx
msmanoj: It's not necessary. I've removed it.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: No idea what I was trying to do here 😄  I've updated it.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/AttributesAccordionSection.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/sections/orthologues/Orthologues.tsx
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: Done

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -1,12 +1,23 @@
-import { createAction, createAsyncAction } from 'typesafe-actions';
+import {
+  createAction,
+  createStandardAction,
+  createAsyncAction
+} from 'typesafe-actions';
 import { ThunkAction } from 'redux-thunk';
 import { ActionCreator, Action } from 'redux';
+import { orthologueSpecies as sampleOrthologueSpecies } from '../../../../sample-data/orthologue';
```
azangru: Instead of going four levels up the file tree, better use a path starting from `src`

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -15,7 +26,7 @@ export const setAttributes = createAsyncAction(
 )<undefined, {}, Error>();
 
 export const fetchAttributes: ActionCreator<
-  ThunkAction<void, any, null, Action<string>>
+  ThunkAction<void, Attributes, null, Action<string>>
```
azangru: Second type of ThunkAction is root state (that is returned by the getState function). It can’t be `Attributes`

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -26,142 +37,95 @@ export const fetchAttributes: ActionCreator<
...
   'custom-download/set-orthologue-species-request',
   'custom-download/set-orthologue-species-success',
   'custom-download/set-orthologue-species-failure'
-)<{ searchTerm: string }, {}, Error>();
+)<{ searchTerm: string }, CheckboxGridOption[], Error>();
 
 export const fetchOrthologueSpecies: ActionCreator<
-  ThunkAction<void, any, null, Action<string>>
-> = (searchTerm: string, orthologueSpecies: any) => async (dispatch) => {
+  ThunkAction<void, CheckboxGridOption[], null, Action<string>>
```
azangru: As above, wrong second type in ThunkAction

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -26,142 +37,95 @@ export const fetchAttributes: ActionCreator<
...
+      }
+    );
 
-    allSpecies.forEach((species: any) => {
+    allSpecies.forEach((species: Species) => {
       if (
+        searchTerm &&
         species.display_name.toLowerCase().indexOf(searchTerm.toLowerCase()) !==
-        -1
+          -1
```
azangru: let’s change indexOf !== -1 to either `includes` (with polyfill added in the polyfill.io link in the template file) or regex?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
```diff
@@ -26,142 +37,95 @@ export const fetchAttributes: ActionCreator<
...
+  dispatch
+) => {
   dispatch(setOrthologueSpecies.request({ searchTerm: searchTerm }));
   try {
     // This will be fetched from the API when we have one
-    let allSpecies = sampleOrthologueSpecies.species;
+    let allSpecies = sampleOrthologueSpecies.species as Species[];
 
-    let filteredSpecies: any = {};
+    let filteredSpecies: CheckboxGridOption[] = [...orthologueSpecies].filter(
```
azangru: unnecessary spread of orthologueSpecies

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -13,13 +13,30 @@ import {
...
+} from 'src/shared/image-button/ImageButton';
+import { ReactComponent as ResetIcon } from 'static/img/shared/trash.svg';
 
 type Props = StateProps & DispatchProps;
 
-const Filters = (props: Props) => {
+const FiltersAccordion = (props: Props) => {
+  useEffect(() => {
+    props.updateSelectedFilters(
+      customDownloadStorageService.getSelectedFilters()
```
azangru: Should this not better be done via redux layer?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/genes/Genes.tsx
```diff
@@ -1,245 +1,57 @@
...
-  const geneSourceFilterOnChange = (selectedOptions: string[]) => {
-    props.setGeneSourceFilters(selectedOptions);
-  };
-
-  const gencodeAnnotationFilterOnChange = (
-    selectedOption: string | number | boolean
-  ) => {
-    props.setGencodeAnnotationFilters(selectedOption);
-  };
+  useEffect(() => {}, [props.selectedFilters]);
```
azangru: Huh?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/proteins/Proteins.tsx
```diff
@@ -0,0 +1,63 @@
...
+} from '../../state/filterAccordionActions';
+
+import JSONValue from 'src/shared/types/JSON';
+import allFilters from 'src/content/app/custom-download/sample-data/filters';
+import { AttributeWithContent } from 'src/content/app/custom-download/types/Attributes';
+
+type Props = StateProps & DispatchProps;
+
+const Proteins = (props: Props) => {
+  useEffect(() => {}, [props.selectedFilters]);
```
azangru: ?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/proteins/Proteins.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const content = allFilters['proteins'] as AttributeWithContent;
+  return (
+    <div>
```
azangru: What’s the purpose of the wrapper div? ContentBuilder has its own top-level div.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/state/filterAccordionActions.ts
```diff
@@ -7,37 +9,20 @@ export const setFiltersAccordionExpandedPanel = createAction(
...
 
-export const setGeneSourceFilters = createAction(
-  'custom-download/set-gene-sourcefilters',
-  (resolve) => {
-    return (geneSource: {}) => resolve(geneSource);
-  }
-);
+export const updateContentState = createStandardAction(
+  'custom-download/update-filter-content-state'
+)<{}>();
```
azangru: Is `{}` the correct type?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/state/filterAccordionActions.ts
```diff
@@ -7,37 +9,20 @@ export const setFiltersAccordionExpandedPanel = createAction(
...
-);
-
-export const setTranscriptTypeFilters = createAction(
-  'custom-download/set-transcript-type-filters',
-  (resolve) => {
-    return (transcriptTypeFilters: {}) => resolve(transcriptTypeFilters);
-  }
-);
+export const resetSelectedFilters = createStandardAction(
+  'custom-download/reset..-selected-filters'
```
azangru: why two dots?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/state/filterAccordionSelector.ts
```diff
@@ -1,26 +1,13 @@
...
-export const getGeneSourceFilters = (state: RootState): {} =>
-  state.customDownload.filtersAccordion.filters.source;
-
-export const getGencodeAnnotationFilters = (state: RootState): string =>
-  state.customDownload.filtersAccordion.filters.gencode_basic_annotation;
+  state.customDownload.filters.filters;
 
-export const getGeneTypeFilters = (state: RootState): string =>
-  state.customDownload.filtersAccordion.filters.gene_type;
+export const getSelectedFilters = (state: RootState): {} =>
```
azangru: Is `{}` the correct type?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
```diff
@@ -6,10 +6,10 @@ import {
   getIsLoadingResult
 } from '../../../state/customDownloadSelectors';
 
-import { getAttributes } from '../attributes-accordion/state/attributesAccordionSelector';
-import { getFilters } from '../filter-accordion/state/filterAccordionSelector';
+import { getSelectedAttributes } from '../attributes-accordion/state/attributesAccordionSelector';
+import { getSelectedFilters } from '../filter-accordion/state/filterAccordionSelector';
 
-import AttributesSection from 'src/content/app/custom-download/types/Attributes';
+import { keys } from 'lodash';
```
azangru: This imports the whole lodash library!

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
```diff
@@ -20,32 +20,40 @@ import {
...
-
   useEffect(() => {
-    if (!selectedAttributes.length && props.preview.results) {
+    const flatSelectedAttributes: { [key: string]: boolean } = flattenObject(
+      props.selectedAttributes
+    );
+
+    const totalSelectedAttributes = keys(flatSelectedAttributes).length;
+    if (!totalSelectedAttributes && props.preview.results) {
       props.clearPreviewResult({});
```
azangru: Why does a function called `clearPreviewResult` accept an argument?

If it were declared in mapDispatchToProps as  `clearPreviewResult: () => setPreviewResult.success({})`, wouldn't it work?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -1,81 +1,96 @@
 import config from 'config';
 
-export const getEndpointUrl = (attributes: any) => {
-  let endpoint = config.genesearchAPIEndpoint + '/genes/query?query=';
-
-  let endpointFields = '';
-  attributes.forEach((attribute: any) => {
-    endpointFields += attribute[2] + ',';
+import { keys, mapKeys, set, get, trim } from 'lodash';
```
azangru: No. This imports the whole lodash library.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -1,81 +1,96 @@
...
+import JSONValue from 'src/shared/types/JSON';
+
+export const getProcessedAttributes = (flatSelectedAttributes: JSONValue) => {
+  const filteredAttributes = keys(flatSelectedAttributes).filter(
```
azangru: keys -> Object.keys?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -1,81 +1,96 @@
...
+    (key) => flatSelectedAttributes[key]
+  );
+  return filteredAttributes.map((value: string) => {
+    return value
+      .split('.default.')
+      .join('.')
+      .split('genes.')
+      .join('');
```
azangru: is this the same as `value = value.replace('.default.', '.'); value = value.replace('genes.', '')`?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -1,81 +1,96 @@
...
+  );
+
+  const selectedFilters = mapKeys(
+    flatSelectedFilters,
+    (value: boolean, key: string) => {
+      return key
+        .split('.default.')
+        .join('.')
+        .split('.genes.')
+        .join('.');
```
azangru: chain of `replaces`, as above?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -1,81 +1,96 @@
...
-  const selectedFilters: any = {};
+  // FIXME: Temporarily apply the filters locally
+  const gene_ids = get(processedFilters, 'genes.limit_to_genes');
+  const gene_biotypes = get(processedFilters, 'genes.biotype');
+  const gene_source = get(processedFilters, 'genes.gene_source');
+
+  if (gene_ids) {
+    endpointFilters.id = gene_ids
+      .join(',')
+      .split(',')
```
azangru: what does `.join(',').split(',')` do?

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
```diff
@@ -91,14 +106,14 @@ const flattenObject = (
       );
 };
 
-const formatResponseToArray = (responseData: any) => {
+const formatResponseToArray = (responseData: any): any => {
```
azangru: Why are there so many anys here? Is this intentional?

msmanoj: We do not have a proper structure of the result yet and this whole thing will be removed once we have the backend.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: It's doing this:
['1,2,3', '4,5,6'] => [1,2,3,4,5,6]

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: `.replace` only replaces the first match. This string could have multiple '.default.'.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: Updated it

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: It's not needed. Just to confirm, do we use loadash only when there are some compatibility issues?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/state/filterAccordionActions.ts
msmanoj: It was by mistake. Removed it.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/state/filterAccordionActions.ts
msmanoj: Updated it.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/proteins/Proteins.tsx
msmanoj: The ContentBuilder was initially returning undefined when the data was no defined. This was causing some error so I added a wrapper to get around it. The ContentBuilder was updated later on to return `null` and I missed to remove these wrappers.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/genes/Genes.tsx
msmanoj: Removed it.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/sections/proteins/Proteins.tsx
msmanoj: Removed it.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: Removed it.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: Updated it.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: Updated it.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/state/attributesAccordionActions.ts
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
azangru: > Just to confirm, do we use loadash only when there are some compatibility issues?
Primarily when it offers us some convenience. The `find` vs `Array.prototype.find` question is still something I don't have a strong opinion on. Perhaps we should use the native `Array.prototype.find` and add a polyfill for browsers that do not support it (that will be only IE11).

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: In this case, should we say that we can only use certain functions?

imran-ebi: Just wanted to add my two cents here. We (@msmanoj and I) had a chat about this in the morning. The native `find` is a lot faster than the lodash `find`: https://measurethat.net/Benchmarks/ShowResult/63691.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
imran-ebi: @msmanoj You can replace all with `.replace` if you use regex with the global flag. For example: `str.replace(/\.default/g, '.').replace(/genes\./g, '')`.

azangru:
I wonder whether there is a bug in checkbox-with-select:

If you select several options:

Screenshot 2019-07-26 at 17 41 51

And open any of the select dropdowns except the first

Screenshot 2019-07-26 at 17 41 59

you will see that available options exclude the ones that are selected.

But in the topmost select dropdown, all options remain available:

Screenshot 2019-07-26 at 17 42 05



**P.S.**: When I talked to Andrea, she agreed that checkbox-with-select doesn't look right and will probably need to go, because it is trying to do the same job as checkbox-grid, only does it worse.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
azangru: @imran-ebi @msmanoj I am ok if we agree to use Array.prototype.find instead of lodash's `find`. Lodash's `find` is allowed if ever we decide to use functions from `lodash/fp` (which is like ramda, only lodash 😄 )

I think, the general guideline is, prefer native functions over lodash's functions if native function do exactly the same job and there is no apparent benefit to using lodash's equivalents.

src/ensembl/src/content/app/custom-download/types/Attributes.ts
```diff
@@ -1,21 +1,36 @@
+import { RadioOptions } from 'src/shared/radio/Radio';
+import { Option } from 'src/shared/select/Select';
+import { CheckboxGridOption } from 'src/content/app/custom-download/components/checkbox-grid/CheckboxGrid';
+import JSONValue from 'src/shared/types/JSON';
+
 export type Attribute = {
-  isChecked: boolean;
-  id: string;
+  type: string;
```
azangru: Shouldn't this be an enum? Isn't there a very limited range of attribute types?

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -0,0 +1,272 @@
...
+    const gridOptions = [...entry.options] as CheckboxGridOption[];
+    let shouldUpdateSelectedData = false;
+
+    gridOptions.forEach((option) => {
+      if (option.isChecked && selectedOptions[option.id] === undefined) {
+        set(newSelectedData, [...currentPath, option.id], true);
+        shouldUpdateSelectedData = true;
+      }
+    });
+    if (shouldUpdateSelectedData) props.onChange(newSelectedData);
```
azangru: I think our style is to always use curly braces, even if it's just a one-liner.

src/ensembl/src/content/app/custom-download/sample-data/attributes.ts
azangru: @msmanoj if I understand your point correctly, you are saying that since we don't have a proper backend, we can use any method available to signify order of options.

To which I would say that since we don't have a proper backend, it's up to us to decide how we encode the information about the order of options; and since an array is already, by definition, an ordered list, we should rely on the order of options in an array and not introduce the order field, which is unnecessary and might confuse backend developers at a later point (if they use our mock data as a starting point to model api response).

azangru:
I think I have resigned to the `Attribute` type having the `content` field.

As for the `contentState` prop, if it describes the content of the accordion inside another accordion, can we rename it to `nestedContent`?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -11,24 +11,26 @@ import {
...
 
-import { getAttributesAccordionExpandedPanel } from './state/attributesAccordionSelector';
+import { getAttributesAccordionExpandedPanel } from '../../../state/attributes/attributesSelector';
 import {
   setAttributesAccordionExpandedPanel,
-  fetchAttributes
-} from './state/attributesAccordionActions';
+  fetchAttributes,
+  resetSelectedAttributes,
+  updateSelectedAttributes
```
azangru: Is it unused in this component?

src/ensembl/src/content/app/custom-download/state/attributes/attributesReducer.ts
```diff
@@ -0,0 +1,105 @@
...
+  AttributesAccordionState,
+  defaultAttributesAccordionState,
+  OrthologueState,
+  AttributeContentState
+} from './attributesState';
+
+function attributesAccordion(
+  state: AttributesAccordionState = defaultAttributesAccordionState,
+  action: ActionType<RootAction>
+): AttributesAccordionState {
```
azangru: `AttributesAccordionState` is the name that remains from the time when these files lived in containers/accordion. Should be renamed to `AttributesAccordionState`.

src/ensembl/src/content/app/custom-download/state/attributes/attributesState.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type AttributeContentState = {
+  [key: string]: AttributeWithOptions | { [key: string]: AttributeWithOptions };
+};
+
+export type AttributesAccordionState = Readonly<{
+  expandedPanel: string;
+  content: AttributeContentState;
+  selectedAttributes: JSONValue;
+  contentState: JSONValue;
```
azangru: Just when I thought I was starting to understand what is going on... 😄 What is the `contentState` field in redux state and how is it different from the `content` field?

Also, why do we keep `selectedAttributes` as a separate redux field instead of updating the `isChecked` field in the options kept in the `content` field?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: Removed it

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
msmanoj: Cool 👍 updated it.

src/ensembl/src/content/app/custom-download/state/attributes/attributesReducer.ts
msmanoj: Renamed it to `AttributesState`.

src/ensembl/src/content/app/custom-download/state/attributes/attributesState.ts
msmanoj: Renamed `contentState` to `uiState` on ContentBuilder and to `ui` on redux.

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/types/Attributes.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/result-holder/ResultHolder.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/result-holder/resultHolderHelper.ts
msmanoj: Updated it with replace.

msmanoj:
I guess all the requested changes are done in this PR?

azangru:
One last question: the settings that we store for custom download — do we want to keep them in `localStorage`, as we currently do (meaning that they will persist for a long time), or only in `sessionStorage` (meaning that they will persist only until the user closes the browser)? Intuitively, I would think that sessionStorage is more appropriate here. Any thoughts from Andrea about it?

src/ensembl/src/content/app/custom-download/state/filters/filterState.ts
```diff
@@ -0,0 +1,17 @@
+import { Attributes } from 'src/content/app/custom-download/types/Attributes';
+import JSONValue from 'src/shared/types/JSON';
+import customDownloadStorageService from 'src/content/app/custom-download/services/custom-download-storage-service';
+
+export type FilterAccordionState = Readonly<{
```
azangru: Same as with AttributeAccordionState — rename to FilterState

src/ensembl/src/content/app/custom-download/state/filters/filterState.ts
```diff
@@ -0,0 +1,17 @@
...
+  expandedPanel: string;
+  filters: Attributes;
+  selectedFilters: JSONValue;
+  ui: JSONValue;
+}>;
+
+export const defaultFilterAccordionState: FilterAccordionState = {
```
azangru: defaultFilterAccordionState -> defaultFilter(s)AccordionState

src/ensembl/src/content/app/custom-download/state/filters/filterReducer.ts
```diff
@@ -0,0 +1,28 @@
...
+    case getType(filterActions.updateUi):
+      return { ...state, ui: action.payload };
+    case getType(filterActions.resetSelectedFilters):
+      return { ...state, selectedFilters: {} };
+    default:
+      return state;
+  }
+}
+
+export default filterAccordion;
```
azangru: name

src/ensembl/src/content/app/custom-download/state/attributes/attributesReducer.ts
```diff
@@ -0,0 +1,103 @@
...
+): AttributeUi {
+  switch (action.type) {
+    case getType(attributesActions.setOrthologueAttributes):
+      return { ...state, orthologues: action.payload };
+    default:
+      return state;
+  }
+}
+
+export default attributesAccordion;
```
azangru: name

.gitlab-ci.yml
azangru: In fact, we should change this to:
```
- /^feature\/.*$/
```
so that branches that start with `feature` do not start the Test job (they will start `Test_N_Build:feature` instead)

azangru:
@msmanoj I thought you also wanted to remove the `contentProps` prop from ContentBuilder?

src/ensembl/src/content/app/custom-download/state/filters/filterState.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/state/filters/filterState.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/state/attributes/attributesReducer.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/state/filters/filterReducer.ts
msmanoj: That's done

msmanoj:
> I thought you also wanted to remove the contentProps prop from ContentBuilder?

Thanks for reminding. I've removed it now.

msmanoj:
@azangru made the changes to use session storage instead of local storage.

PR #123, opened 2019-07-25
Feature/focus object change

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-272

### On the Browser Chrome side of things
- upon click on the focus object link in zmenu (which, at the moment, is bound to be a gene), send the id of the focus object to genome browser so that it can move to this focus object
- request information about the focus object when it changes
- update api service to be able to handle error responses (throws an error if the response is an error)

## Views affected
Genome browser


.gitlab-ci.yml
```diff
@@ -188,7 +188,7 @@ Nginx:feature:
   - cd ensembl-client-nginx
   - git checkout june-release
   - cd ..
-  - sed -i "s/<DEPLOYMENT_ENV>/internal/g" ensembl-client-nginx/config/conf.d/local.conf
+  - sed -i "s/<DEPLOYMENT_ENV>/$feature_tag/g" ensembl-client-nginx/config/conf.d/local.conf
```
azangru: **NOTE TO SELF:** Remember to change back before merge

src/ensembl/webpack/webpack.config.dev.js
```diff
@@ -72,7 +72,7 @@ const devConfig = {
         secure: false
       },
       '/browser': {
-        target: 'https://staging-2020.ensembl.org',
+        target: 'http://193.62.55.91:30319', // FIXME: change back
```
azangru: **NOTE TO SELF:** Remember to change back before merge

src/ensembl/src/services/api-service.ts
```diff
@@ -67,12 +67,28 @@ class ApiService {
...
+      if (!response.ok) {
+        throw await this.handleError(response, fetchOptions);
+      }
       return await this.handleResponse(response, fetchOptions);
     } catch (error) {
       throw error;
     }
   }
 
+  private async handleError(response: Response, options: FetchOptions) {
```
imran-ebi: Since `options` isn't used anywhere in this function can it be removed?

src/ensembl/src/services/api-service.ts
```diff
@@ -67,12 +67,28 @@ class ApiService {
...
+    let errorJson;
+    try {
+      errorJson = await response.json();
+    } catch {
```
imran-ebi: As `catch` is empty how about removing it and moving the `return` statement into a `finally` clause?

src/ensembl/src/services/api-service.ts
azangru: removed

src/ensembl/src/services/api-service.ts
azangru: ok

PR #124, opened 2019-07-25
Fix toggling of species use status

azangru:
## Type
- Bug fix

## Description
Fixes regression introduced in https://github.com/Ensembl/ensembl-client/pull/113, which breaks toggling between using and not using selected species.

## Views affected
Species Selector

### DIFF
--- a/src/ensembl/src/content/app/species-selector/state/speciesSelectorSelectors.ts
+++ b/src/ensembl/src/content/app/species-selector/state/speciesSelectorSelectors.ts
@@ -1,4 +1,5 @@
 import get from 'lodash/get';
+import find from 'lodash/find';
 
 import { RootState } from 'src/store';
 import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/species-selector/state/speciesSelectorSelectors.ts
```diff
@@ -62,6 +63,16 @@ export const getCommittedSpecies = (state: RootState): CommittedItem[] => {
   return state.speciesSelector.committedItems;
 };
 
+export const getCommittedSpeciesById = (
+  state: RootState,
+  genomeId: string
+): CommittedItem | null => {
+  const allCommittedSpecies = getCommittedSpecies(state);
+  return (
+    find(allCommittedSpecies, ({ genome_id }) => genomeId === genome_id) || null
```
imran-ebi: Won't this be more readable if `genomeId === genome_id` is instead as `genomeId === species.genome_id`?

azangru: Updated the syntax of this line.

PR #125, opened 2019-07-26
Fix missing species on the header bar

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-74

## Importance
July release

## Description
Navigation to the genome browser by direct url now commits the species if it is not already in the committed species list and displays it on the header bar.

## Views affected
Browser -> Species List on the header bar

## Other effects
GA for browser will be done separately.

### DIFF
--- a/src/ensembl/src/genome/genomeActions.ts
+++ b/src/ensembl/src/genome/genomeActions.ts
@@ -1,17 +1,24 @@
 import { createAsyncAction } from 'typesafe-actions';
 import { ThunkAction } from 'redux-thunk';
 import { Action, ActionCreator } from 'redux';
-
+import find from 'lodash/find';
 import apiService from 'src/services/api-service';
 import { RootState } from 'src/store';
-import { GenomeInfoData, GenomeTrackCategories } from './genomeTypes';
+import {
+  GenomeInfoData,
+  GenomeInfo,
+  GenomeTrackCategories
+} from './genomeTypes';
+import omit from 'lodash/omit';
 
 import { fetchExampleEnsObjects } from 'src/ens-object/ensObjectActions';
-
+import { getEnabledCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
+import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
 import {
   getGenomeInfoById,
   getGenomeTrackCategories
 } from 'src/genome/genomeSelectors';
+import { updateCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorActions';
 
 export const fetchGenomeInfoAsyncActions = createAsyncAction(
   'genome/fetch_genome_info_request',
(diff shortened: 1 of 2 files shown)


src/ensembl/src/genome/genomeActions.ts
```diff
@@ -1,17 +1,24 @@
...
+import find from 'lodash/find';
 import apiService from 'src/services/api-service';
 import { RootState } from 'src/store';
-import { GenomeInfoData, GenomeTrackCategories } from './genomeTypes';
+import {
+  GenomeInfoData,
+  GenomeInfo,
+  GenomeTrackCategories
+} from './genomeTypes';
+import omit from 'lodash/omit';
```
azangru: according to lodash's roadmap for v5 (which will hopefully happen some time), they are going to remove `omit` and just use pickBy: https://github.com/lodash/lodash/wiki/Roadmap#v500-2019

Can we avoid  lodash's `omit` as well? (I personally love it, but so that we don't have to replace it later?)

src/ensembl/src/genome/genomeActions.ts
```diff
@@ -21,12 +28,30 @@ export const fetchGenomeInfoAsyncActions = createAsyncAction(
...
-> = (genomeId: string) => async (dispatch) => {
+> = (genomeId: string) => async (dispatch, getState: () => RootState) => {
   await Promise.all([
     dispatch(fetchGenomeInfo(genomeId)),
     dispatch(fetchGenomeTrackCategories(genomeId))
   ]);
 
+  const state = getState();
+  const committedSpecies: CommittedItem[] = getEnabledCommittedSpecies(state);
+  const genomeInfo: GenomeInfo | null = getGenomeInfoById(state, genomeId);
```
azangru: Don't we automatically get types for committedSpecies and genomeInfo from selectors?

src/ensembl/src/genome/genomeActions.ts
```diff
@@ -21,12 +28,30 @@ export const fetchGenomeInfoAsyncActions = createAsyncAction(
...
+  if (
+    genomeInfo &&
+    !find(
+      committedSpecies,
+      (species: CommittedItem) => species.genome_id === genomeId
```
azangru: There is a new selector `getCommittedSpeciesById` added to `dev` now. It will do the finding.

src/ensembl/src/genome/genomeActions.ts
```diff
@@ -21,12 +28,30 @@ export const fetchGenomeInfoAsyncActions = createAsyncAction(
...
+    )
+  ) {
+    const newCommittedSpecies = [
+      ...committedSpecies,
+      { ...omit(genomeInfo, ['example_objects']), isEnabled: true }
+    ];
+
+    dispatch(updateCommittedSpecies(newCommittedSpecies));
```
azangru: Should we also save the species?

src/ensembl/src/genome/genomeActions.ts
```diff
@@ -21,12 +28,30 @@ export const fetchGenomeInfoAsyncActions = createAsyncAction(
...
+    genomeInfo &&
+    !find(
+      committedSpecies,
+      (species: CommittedItem) => species.genome_id === genomeId
+    )
+  ) {
+    const newCommittedSpecies = [
+      ...committedSpecies,
+      { ...omit(genomeInfo, ['example_objects']), isEnabled: true }
+    ];
```
azangru: The added code in green does not really belong to a function whose name is `fetchGenomeData`. In the spirit of keeping functions small, can we move this code to its own thunk function (`addSpeciesFromGenomeId`?)?

src/ensembl/src/genome/genomeActions.ts
msmanoj: Updated it. Thanks 👍

src/ensembl/src/genome/genomeActions.ts
msmanoj: It's jus that there was no indication to say that it's not necessary. 
Removed it now.

src/ensembl/src/genome/genomeActions.ts
msmanoj: That's done

src/ensembl/src/genome/genomeActions.ts
msmanoj: Good point. That's done.

src/ensembl/src/genome/genomeActions.ts
msmanoj: That's done

PR #126, opened 2019-07-27
Mock react-ga library during tests

azangru:
## Type

## Description
Currently, tests try to call the real React-GA library and produce warnings:

image

By mocking the `react-ga` library we make it so that its `initialize` function becomes stubbed out and never get truly executed.

## For discussion
I think it would be also possible not to mock react-ga in individual test files, but to create a `__mocks__` directory at the same level as `node_modules`, and create a `react-ga` file with a mock version of `react-ga` (see [jest docs](https://jestjs.io/docs/en/manual-mocks#mocking-node-modules)). The advantage of this approach is that we won’t have to mock `react-ga` in individual test files. The disadvantage is that a root-level `__mocks__` folder looks unnecessary and ugly. Not sure which option to prefer.


imran-ebi:
I think the `__mocks__` folder would be a better solution to this than adding individual mock statements for `react-ga`.

msmanoj:
Although it looks ugly, I like the idea of having the `__mocks__` directory as it has a very good purpose. Since `ga` and similar functionalities will be used throughout the app, it would be better to go with this approach so that the tests can remain simple.

azangru:
I couldn’t make the `__mocks__` directory work for me. However, an alternative solution that I found was to put `jest.mock` command in a setup file, which looks, I think, even nicer than the solution with the `__mocks__` folder.

_(I also moved `setup-enzyme.ts` file in the `tests` folder)_

imran-ebi:
Are there any more changes that need to go into this PR?

azangru:
@imran-ebi yes, I just added the last commit to remove test awareness from analytics-service itself.

PR #127, opened 2019-07-29
Fix for browser tabs get hidden when track panel closed on large screens

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)

https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-258

## Description

Fixes for:
- Browser tabs not shown when track panel closed on large screens
- Browser tabs not shown on smaller screens when track panel is open

The XD for this behaviour is: https://xd.adobe.com/view/8b709d74-78a9-47ee-78eb-54f65cd02054-fd41/?fullscreen

## Views affected

Genome browser

## Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
@@ -126,13 +126,10 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
     props.toggleBrowserNav();
   };
 
-  const shouldShowBrowserTabs = () => {
-    return (
-      props.activeGenomeId &&
-      (props.isTrackPanelOpened ||
-        props.breakpointWidth === BreakpointWidth.LARGE)
-    );
-  };
+  const shouldShowBrowserTabs =
+    props.activeGenomeId &&
+    (props.isTrackPanelOpened ||
+      props.breakpointWidth === BreakpointWidth.LARGE);
 
   const className = classNames(styles.browserInfo, {
     [styles.browserInfoExpanded]: !props.isTrackPanelOpened,
@@ -175,7 +172,7 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
           )}
         </dl>
       </div>
-      {shouldShowBrowserTabs() && (
+      {shouldShowBrowserTabs && (
         <BrowserTabs
           closeDrawer={props.closeDrawer}
           ensObject={props.ensObject}
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
@@ -127,15 +127,11 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
   };
 
   const shouldShowBrowserTabs = () => {
-    if (
+    return (
       props.activeGenomeId &&
       (props.isTrackPanelOpened ||
         props.breakpointWidth === BreakpointWidth.LARGE)
-    ) {
-      return true;
-    } else {
-      return false;
-    }
+    );
   };
 
   const className = classNames(styles.browserInfo, {
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
@@ -23,8 +23,11 @@ import {
   getIsTrackPanelModalOpened,
   getIsTrackPanelOpened
 } from '../track-panel/trackPanelSelectors';
-import { selectBrowserTabAndSave } from '../track-panel/trackPanelActions';
-import { closeDrawer, toggleDrawer } from '../drawer/drawerActions';
+import {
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -119,6 +126,18 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
...
+    if (
+      props.activeGenomeId &&
+      (props.isTrackPanelOpened ||
+        props.breakpointWidth === BreakpointWidth.LARGE)
+    ) {
+      return true;
+    } else {
+      return false;
+    }
+  };
```
azangru: can we please immediately return the result of the equality check instead of checking whether the condition evaluates to true and then returning true?

E.g.

```
return props.activeGenomeId &&
      (props.isTrackPanelOpened ||
        props.breakpointWidth === BreakpointWidth.LARGE)
```

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
```diff
@@ -22,40 +26,43 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
...
+      !props.isDrawerOpened &&
+      !props.isTrackPanelModalOpened
     ) {
-      classNames += ` ${styles.browserTabActive} ${styles.browserTabArrow}`;
-    } else if (!props.ensObject.genome_id) {
-      classNames = styles.browserTabDisabled;
+      return true;
+    } else {
+      return false;
     }
```
azangru: Same as comment above: the condition checks whether something is true and then returns true. Why not just return the result of the boolean check that is the condition itself?

It's been the subject of various memes, e.g.:

image

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
imran-ebi: There are too many conditions to evaluate here and that is why I thought it would be neater to put them in `if-else` clauses. Anyway, I'll return them with a ternary operator.

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
imran-ebi: It is the same here too. There are quite a lot of evaluations that might make it difficult to find where the true and false statements are in the ternary operator 😄

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
imran-ebi: @azangru As we discussed offline I misunderstood that you suggested to change this to a ternary operator. I'll return this sans any if statements and ternary operators.

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
imran-ebi: As I mentioned above, I'll return without any if statements and ternary operators.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -119,6 +126,14 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
     props.toggleBrowserNav();
   };
 
+  const shouldShowBrowserTabs = () => {
+    return (
+      props.activeGenomeId &&
+      (props.isTrackPanelOpened ||
+        props.breakpointWidth === BreakpointWidth.LARGE)
+    );
+  };
```
azangru: Now that I noticed how `shouldShowBrowserTabs` is used, I think it should just be changed to a boolean constant instead of the function:

```
const shouldShowBrowserTabs = (
      props.activeGenomeId &&
      (props.isTrackPanelOpened ||
        props.breakpointWidth === BreakpointWidth.LARGE)
 );
```

Same goes for shouldShowBrowserInfo, I think.

imran-ebi: Done for both

src/ensembl/src/content/app/browser/browser-tabs/BrowserTabs.tsx
```diff
@@ -22,40 +26,37 @@ const BrowserTabs: FunctionComponent<BrowserTabsProps> = (
       return;
     }
 
+    if (!props.isTrackPanelOpened) {
+      props.toggleTrackPanel(true);
```
msmanoj: If the tabs are already hidden when then TrackPanel is closed, how is this onClick getting called?

imran-ebi: Tabs aren't always hidden on a large screen when the track panel is closed. So this listener can still be fired even when the track panel is closed.

msmanoj: Cool 👍

PR #128, opened 2019-07-29
Update consumption of data from api/object and api/genome_search endpoints

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-285

## Importance
Should ideally be merged by the end of July

## Description
Update the client to consume updated api endpoints
- `/api/ensembl_object` namespace has changed to `/api/object`
- `/api/object/info` parses region location and generates data for region focus object
- `ensembl_object_id` field in the response has been renamed to `object_id`
- `/api/genome_search` endpoint now accepts the `exclude` parameter (to exclude already selected genome id from search results) and the `limit` parameter (to limit the number of search matches in the response).

## Warning
Deploy simultaneously with updated backend (branch https://github.com/Ensembl/ensembl-2020-genome-search/commits/feature/july-2019)


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -307,16 +307,17 @@ const ExampleObjectLinks = (props: BrowserProps) => {
   if (!activeGenomeId) {
     return null;
   }
+  console.log('ExampleObjectLinks', props);
```
imran-ebi: Can this be removed?

azangru: Removed the console.log.

Also (because I forgot to do that before) I removed the `Spliced mRNA length` field from `BrowserBar` and updated the tests for BrowserBar.

src/ensembl/webpack/webpack.config.dev.js
```diff
@@ -67,7 +67,7 @@ const devConfig = {
     // rules to proxy requests to the backend server in development
     proxy: {
       '/api': {
-        target: 'https://staging-2020.ensembl.org',
+        target: 'http://193.62.55.91:30414',
```
azangru: **NOTE TO SELF:** remember to change back

src/ensembl/src/shared/helpers/regionFormatter.ts
```diff
@@ -0,0 +1,10 @@
+import { getCommaSeparatedNumber } from 'src/shared/helpers/numberFormatter';
+
+import { EnsObjectLocation } from 'src/ens-object/ensObjectTypes';
+
+export const getFormattedLocation = (location: EnsObjectLocation) => {
+  const start = getCommaSeparatedNumber(location.start);
+  const end = getCommaSeparatedNumber(location.end);
+
+  return `${location.chromosome}:${start}-${end}`;
+};
```
imran-ebi: Can this function go into `browserHelper.ts`?

azangru: Is this browser-specific though?

Won't we have an EntityViewer which might need to display coordinates of an Entity?

imran-ebi: True. Let's keep this separate then.

PR #129, opened 2019-07-30
Separate home and genome browser coordinates with commas

imran-ebi:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-214

## Description
The coordinates in the UI don't use comma separators for big numbers. This makes it difficult to read, so commas have been added to them. However, the input field numbers are not separated by commas.

## Views affected
Home and Genome Browser

## Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
@@ -27,6 +27,7 @@ import { selectBrowserTabAndSave } from '../track-panel/trackPanelActions';
 import { closeDrawer, toggleDrawer } from '../drawer/drawerActions';
 import { RootState } from 'src/store';
 import { EnsObject } from 'src/ens-object/ensObjectTypes';
+import { getCommaSeparatedNumber } from 'src/shared/helpers/numberFormatter';
 
 import BrowserReset from '../browser-reset/BrowserReset';
 import BrowserGenomeSelector from '../browser-genome-selector/BrowserGenomeSelector';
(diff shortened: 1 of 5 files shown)


src/ensembl/src/shared/helpers/numberFormatter.ts
```diff
@@ -2,6 +2,6 @@
     Formats the input number to comma separated representation
     eg: 10000 -> 10,000
 */
-export const getCommaSeparatedNumber = (input: number): string => {
+export const getCommaSeparatedNumber = (input: number | string): string => {
```
azangru: I am not sure extending the type of the input to `number | string` is a good thing. With input type as number we are guaranteed that whatever is passed to this function is actually a number. With allowing a string to be passed, we no longer have this guarantee.

imran-ebi: I have now made changes to `getCommaSeparatedNumber` to accept only `number` as before. There weren't any other changes required for the coordinates as all of them are of type `number`.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -207,7 +208,9 @@ export const BrowserInfo = ({ ensObject }: BrowserInfoProps) => {
           <dd className={styles.ensObjectLabel}>
             <label>Region: </label>
             <span className={styles.value}>
-              {`${ensObject.location.chromosome}:${ensObject.location.start}:${ensObject.location.end}`}
+              {`${ensObject.location.chromosome}:${getCommaSeparatedNumber(
+                ensObject.location.start
+              )}:${getCommaSeparatedNumber(ensObject.location.end)}`}
```
msmanoj: Would it be better to have a shared function that takes the `location` and returns the formatted string?

imran-ebi: As we discussed offline the `ensObject` shape is going to change, so I am thinking of waiting till that is finalised in order create this shared function.

PR #132, opened 2019-07-31
Use ImageButton in HeaderButtons (in masthead)

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/projects/ENSWBSITES/issues/ENSWBSITES-287

## Description
This PR adds the tooltip to the image buttons in the masthead (a.k.a. Header).

image

While making this change, it became apparent that the tooltip positioning algorithm had the following problems:

1. If the algorithm could not find position for the tooltip in which it will not extend beyond container boundaries, it would fall back to the initial position. However, it is possible that one of the available positions for the tooltip would actually have been better than the initial position (i.e. in that position, the tooltip will leave its container boundaries only by a tiny margin). The updated algorithm calculates the area of the tooltip that will be out of container's bounds in a given position, and returns the position in which this area is the smallest.

2. The tooltip was relying on IntersectionObserver to figure out whether it remains within bounds of its container. It became obvious that IntersectionObserver was completely unnecessary for this task. Moreover, removing IntersectionObserver makes it possible to put the invisible tooltip in some arbitrary place on the screen (e.g. `position: fixed, top: 0, left: 0`) while its optimal position is being calculated; and thus be sure that the tooltip will not extend beyond the window boundaries and will not cause scrollbars to appear at this point.

### DIFF
--- a/src/ensembl/src/shared/tooltip/Tooltip.tsx
+++ b/src/ensembl/src/shared/tooltip/Tooltip.tsx
@@ -18,7 +18,7 @@ import styles from './Tooltip.scss';
 type Props = {
   position: Position;
   container?: HTMLElement | null;
-  autoAdjust: boolean; // try to adapt position so as not to extend beyond screen bounds
+  autoAdjust: boolean; // try to adjust tooltip position so as not to extend beyond screen bounds
   delay: number;
   children: ReactNode;
   onClose: () => void;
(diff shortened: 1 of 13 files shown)


src/ensembl/src/shared/tooltip/Tooltip.tsx
```diff
@@ -81,50 +83,55 @@ const Tooltip = (props: Props) => {
...
       return;
     }
-    parentRef.current = parentElement;
 
-    setInlineStyles(getInlineStyles({ ...props, parentElement }));
+    const tooltipElement = tooltipElementRef.current;
+    const parentElement = tooltipElement && tooltipElement.parentElement;
 
-    if (!props.autoAdjust) {
+    if (!(tooltipElement && parentElement)) {
```
msmanoj: Do we need `tooltipElement` here?

azangru: We are passing `tooltipElement`, along with `parentElement` to the `adjustPosition` function below. `adjustPosition` relies on the `tooltipElement` being of type `HTMLDivElement`, but here, since `tooltipElement` is `tooltipElementRef.current`, it can be `HTMLDivElement` or `null`. So I would think yes, we want to assure Typescript that both `tooltipElement` and `parentElement` are not null before we proceed further.

msmanoj: Cool, I can see that Typescript is not happy if we remove it. I was thinking that if `tooltipElement` is null, `parentElement` will also be null in which case the `if` condition will return true even if we just check for `!parentElement`.

src/ensembl/src/shared/tooltip/tooltip-helper.ts
```diff
@@ -15,62 +15,67 @@ const leftSide = [Position.LEFT_TOP, Position.LEFT_BOTTOM];
...
   } else if (rightSide.includes(position)) {
-    return [...rightSide, ...leftSide];
+    return [...preferredFirst(rightSide, position), ...leftSide];
   } else {
     return [];
   }
 };
 
-const willStayWithinBounds = (params: FindOptimalPositionParams) => {
+const preferredFirst = (positions: Position[], preferredPosition: Position) => {
```
msmanoj: Is it possible to think of a better name for this function?

azangru: Renamed the function to `bringToFront` and moved it in its own file in src/shared/utils

PR #133, opened 2019-07-31
Do not lower-case gene location info in BrowserBar

azangru:
## Type
- Bug fix

## Description
**Bug**: letters in chromosome name in browser bar get lower-cased (although the case is actually significant):

**Example:** region of a wheat gene; expecting chromosome to be 3D, but instead it shows up as 3d

image

**Cause**: the `nonLabelValue` CSS class has the rule `text-transform: lowercase`.

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -99,6 +99,7 @@ type BrowserProps = RouteComponentProps<MatchParams> &
 export const Browser: FunctionComponent<BrowserProps> = (
   props: BrowserProps
 ) => {
+  console.log('Hello world!');
   const browserRef: React.RefObject<HTMLDivElement> = useRef(null);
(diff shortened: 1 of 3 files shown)


imran-ebi:
The `nonLabelValue` class is used a bit below with `browserInfoGreyed` to grey out all info in the browser bar when the drawer is open. So we might need to use a different class for this to grey out the gene location.

azangru:
@imran-ebi Removed the nonLabelValue class entirely, and cast text to lower-cas in the javascript file.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -99,6 +99,7 @@ type BrowserProps = RouteComponentProps<MatchParams> &
 export const Browser: FunctionComponent<BrowserProps> = (
   props: BrowserProps
 ) => {
+  console.log('Hello world!');
```
imran-ebi: Can you remove this one?

PR #135, opened 2019-08-01
Update workbox config to prevent caching of html files

azangru:
## Type
- Bug fix

## JIRA ticket
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-296

## Description
- Update Workbox config so that it does not cache html files. (_See detailed description of the mechanism of the bug in the linked JIRA ticket._)
- Exclude compressed files and sourcemap files from the workbox-generated manifest
- Do not pass default options to koa-static in local production server config (both `br` option for serving brotli-compressed files if the browser can consume them and `gzip` option to serve gzipped files if the browser can consume them are `true` by default in koa-static).

### DIFF
--- a/src/ensembl/webpack/webpack.config.prod.js
+++ b/src/ensembl/webpack/webpack.config.prod.js
@@ -103,7 +103,12 @@ const plugins = [
   new WorkboxPlugin.GenerateSW({
     swDest: '../service-worker.js', // save service worker in the root folder (/dist) instead of /dist/static
     clientsClaim: true,
-    skipWaiting: true
+    skipWaiting: true,
+    exclude: [/index.html/],
+    runtimeCaching: [{
+      urlPattern: ({ event }) => event.request.mode === 'navigate',
+      handler: 'NetworkOnly'
+    }]
   }),
 
   new RobotstxtPlugin()
--- a/src/ensembl/webpack/webpack.config.prod.js
+++ b/src/ensembl/webpack/webpack.config.prod.js
@@ -107,7 +107,7 @@ const plugins = [
     exclude: [/index.html$/, /\.gz$/, /\.br$/, /\.js\.map$/],
     runtimeCaching: [{
       urlPattern: ({ event }) => event.request.mode === 'navigate',
-      handler: 'NetworkOnly'
+      handler: 'NetworkFirst'
     }]
   }),
 
--- a/src/ensembl/webpack/webpack.config.prod.js
+++ b/src/ensembl/webpack/webpack.config.prod.js
@@ -104,7 +104,7 @@ const plugins = [
     swDest: '../service-worker.js', // save service worker in the root folder (/dist) instead of /dist/static
     clientsClaim: true,
     skipWaiting: true,
-    exclude: [/index.html/],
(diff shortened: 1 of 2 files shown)


src/ensembl/webpack/webpack.config.prod.js
```diff
@@ -103,7 +103,12 @@ const plugins = [
   new WorkboxPlugin.GenerateSW({
     swDest: '../service-worker.js', // save service worker in the root folder (/dist) instead of /dist/static
     clientsClaim: true,
-    skipWaiting: true
+    skipWaiting: true,
+    exclude: [/index.html/],
+    runtimeCaching: [{
+      urlPattern: ({ event }) => event.request.mode === 'navigate',
+      handler: 'NetworkOnly'
```
msmanoj: I think what we need here is `NetworkFirst` as `NetworkOnly` will cause the site to be completely down during the build time.

azangru: Two things here:

1) Our site will **not** be down during build time, because while the build is happening our site is still served by the container with the previous build.
2) I can replace NetworkOnly with NetworkFirst in hope that some of our pages may be operational offline (although I doubt any of them will be useful offline).

**TLDR:** NetworkFirst probably wouldn't be worse than NetworkOnly and might in some cases be slightly better.

msmanoj: > Our site will not be down during build time, because while the build is happening our site is still served by the container with the previous build.

Cool, that makes sense. Are we sure that `index.html` will always be available on the network during the deployment?

Since we exclude `index.html` from getting cached, if at some point it is not available (for a fraction of time), the handler is going to fallback and look for it in the cache where it won't exist.

azangru: > Are we sure that index.html will always be available on the network during the deployment?

That's up to Kamal to say; but if there are more than one containers with nginx (which there probably should be), then there shouldn't be a situation when index.html (or any static files) become unavailable during deployment.

msmanoj: Hi @kamaldodiya, any thoughts on this?

src/ensembl/webpack/webpack.config.prod.js
```diff
@@ -103,7 +103,12 @@ const plugins = [
   new WorkboxPlugin.GenerateSW({
     swDest: '../service-worker.js', // save service worker in the root folder (/dist) instead of /dist/static
     clientsClaim: true,
-    skipWaiting: true
+    skipWaiting: true,
+    exclude: [/index.html$/, /\.gz$/, /\.br$/, /\.js\.map$/],
+    runtimeCaching: [{
```
azangru: The only purpose of runtimeCaching is that we can open/reload our application when offline, and it will show something (very little, and likely broken, but still something rather than nothing). I am not quite sure whether it's useful.

If we remove runtimeCaching, our site will not show anything when offline, but all our static assets produced by webpack will still be requested and cached by service worker. Which is the most value out of service worker that we are currently getting anyway.

imran-ebi: If there is no harm in having this, why not? Or is there any disadvantages using this?

PR #136, opened 2019-08-02
Make track panel state of species indepedent

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-276

## Description
The genome browser for a species should display the track panel just as how the user left it before going to a different view. The track panel open/close state and the track panel modal state is made independent with this PR.

## Views affected
Genome Browser

## Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-storage-service.ts
+++ b/src/ensembl/src/content/app/browser/browser-storage-service.ts
@@ -75,7 +75,7 @@ export class BrowserStorageService {
     );
   }
 
-  public getSelectedBrowserTab() {
+  public getSelectedBrowserTab(): { [genomeId: string]: TrackType } {
     return this.storageService.get(StorageKeys.SELECTED_BROWSER_TAB) || {};
   }
 
--- a/src/ensembl/src/content/app/browser/browser-storage-service.ts
+++ b/src/ensembl/src/content/app/browser/browser-storage-service.ts
@@ -3,7 +3,7 @@ import storageService, {
 } from 'src/services/storage-service';
 import {
   TrackStates,
-  TrackType,
+  TrackSet,
   TrackToggleStates
 } from './track-panel/trackPanelConfig';
 import { ChrLocations } from './browserState';
@@ -75,12 +75,12 @@ export class BrowserStorageService {
     );
   }
 
-  public getSelectedBrowserTab(): { [genomeId: string]: TrackType } {
+  public getSelectedBrowserTab(): { [genomeId: string]: TrackSet } {
     return this.storageService.get(StorageKeys.SELECTED_BROWSER_TAB) || {};
   }
 
   public updateSelectedBrowserTab(selectedBrowserTabForGenome: {
-    [genomeId: string]: TrackType;
+    [genomeId: string]: TrackSet;
   }) {
     this.storageService.update(
       StorageKeys.SELECTED_BROWSER_TAB,
--- a/src/ensembl/src/content/app/browser/browser-storage-service.ts
+++ b/src/ensembl/src/content/app/browser/browser-storage-service.ts
@@ -15,7 +15,7 @@ export enum StorageKeys {
   DEFAULT_CHR_LOCATION = 'browser.defaultChrLocation',
   TRACK_STATES = 'browser.trackStates',
   TRACK_LIST_TOGGLE_STATES = 'browser.trackListToggleStates',
-  SELECTED_BROWSER_TAB = 'browser.selectedBrowserTab'
+  SELECTED_TRACK_PANEL_TAB = 'browser.selectedTrackPanelTab'
 }
 
 export class BrowserStorageService {
@@ -75,16 +75,16 @@ export class BrowserStorageService {
     );
   }
 
-  public getSelectedBrowserTab(): { [genomeId: string]: TrackSet } {
-    return this.storageService.get(StorageKeys.SELECTED_BROWSER_TAB) || {};
+  public getSelectedTrackPanelTab(): { [genomeId: string]: TrackSet } {
+    return this.storageService.get(StorageKeys.SELECTED_TRACK_PANEL_TAB) || {};
   }
 
-  public updateSelectedBrowserTab(selectedBrowserTabForGenome: {
+  public updateSelectedTrackPanelTab(selectedTrackPanelTabForGenome: {
     [genomeId: string]: TrackSet;
   }) {
     this.storageService.update(
-      StorageKeys.SELECTED_BROWSER_TAB,
-      selectedBrowserTabForGenome
+      StorageKeys.SELECTED_TRACK_PANEL_TAB,
+      selectedTrackPanelTabForGenome
     );
   }
 }
--- a/src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
+++ b/src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
@@ -1,6 +1,22 @@
 import React, { FunctionComponent, useCallback } from 'react';
+import { connect } from 'react-redux';
 
 import { trackPanelBarConfig, TrackPanelBarItem } from './trackPanelBarConfig';
+import { getBrowserActiveGenomeId } from '../../browserSelectors';
+import { getIsDrawerOpened } from '../../drawer/drawerSelectors';
+import {
+  getIsTrackPanelModalOpened,
+  getIsTrackPanelOpened,
+  getTrackPanelModalView
+} from '../trackPanelSelectors';
+import { RootState } from 'src/store';
+import { getLaunchbarExpanded } from 'src/header/headerSelectors';
+import {
+  toggleTrackPanel,
+  closeTrackPanelModal,
+  openTrackPanelModal
+} from '../trackPanelActions';
+import { closeDrawer } from '../../drawer/drawerActions';
 
 import TrackPanelBarIcon from './TrackPanelBarIcon';
 
@@ -9,18 +25,25 @@ import chevronRightIcon from 'static/img/shared/chevron-right.svg';
 
 import styles from './TrackPanelBar.scss';
 
-type TrackPanelBarProps = {
-  activeGenomeId: string;
-  closeDrawer: () => void;
-  closeTrackPanelModal: () => void;
+type StateProps = {
+  activeGenomeId: string | null;
   isDrawerOpened: boolean;
   isTrackPanelModalOpened: boolean;
   isTrackPanelOpened: boolean;
   launchbarExpanded: boolean;
-  openTrackPanelModal: (trackPanelModalView: string) => void;
-  toggleTrackPanel: (isTrackPanelOpened?: boolean) => void;
   trackPanelModalView: string;
 };
+
+type DispatchProps = {
+  closeDrawer: () => void;
+  closeTrackPanelModal: () => void;
+  openTrackPanelModal: (trackPanelModalView: string) => void;
+  toggleTrackPanel: (isTrackPanelOpened?: boolean) => void;
+};
+
+type OwnProps = {};
+
+type TrackPanelBarProps = StateProps & DispatchProps & OwnProps;
 
 const TrackPanelBar: FunctionComponent<TrackPanelBarProps> = (
   props: TrackPanelBarProps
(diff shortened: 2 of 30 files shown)


azangru:
I wonder which model of state is better — the current one:

image

or the one where root-level keys are genome ids:

image

For me, the second one is more natural (also, it's easier to generate a template with default values for every new species)

imran-ebi:
@azangru The model you suggested is neater and makes more sense. I'll change the track panel state to this model.

src/ensembl/src/content/app/browser/track-panel/trackPanelState.ts
```diff
@@ -1,18 +1,31 @@
...
-  selectedBrowserTab,
-  trackPanelModalView: ''
-};
+export type TrackPanelState = Readonly<{
+  [genomeId: string]: TrackPanelStateForGenome;
+}>;
+
+const selectedBrowserTabFromStorage = browserStorageService.getSelectedBrowserTab();
+
+// create the default state using the stored selected browser tab object
```
azangru: this comment is redundant — it follows from variable names that the defaultTrackPanelState is created and it somehow uses selectedBrowserTabFromStorage for its creation.

src/ensembl/src/content/app/browser/track-panel/trackPanelSelectors.ts
```diff
@@ -1,21 +1,30 @@
 import { RootState } from 'src/store';
 import { getBrowserActiveGenomeId } from '../browserSelectors';
 import { TrackType } from './trackPanelConfig';
+import { TrackPanelStateForGenome } from './trackPanelState';
+
+const getTrackPanelStateProp = (
```
azangru: getTrackPanelProperty?

azangru: Also, do I understand correctly that this function is only used in this selector?

Perhaps instead of casting return value of this to various types it will be better to just have `getActiveTrackPanel` (which returns `TrackPanelStateForGenome` or null) and then, in other selectors, pick relevant properties out of that object?

src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
```diff
@@ -7,39 +7,41 @@ export default function trackPanel(
   state: TrackPanelState = defaultTrackPanelState,
   action: ActionType<typeof trackPanelActions>
 ): TrackPanelState {
-  switch (action.type) {
-    case getType(trackPanelActions.toggleTrackPanel):
-      const isTrackPanelOpened =
-        action.payload === undefined
-          ? !state.isTrackPanelOpened
-          : action.payload;
+  // check if action payload is not undefined to assign activeGenomeId
```
azangru: this comment is redundant

src/ensembl/src/content/app/browser/browser-storage-service.ts
```diff
@@ -75,7 +75,7 @@ export class BrowserStorageService {
     );
   }
 
-  public getSelectedBrowserTab() {
+  public getSelectedBrowserTab(): { [genomeId: string]: TrackType } {
```
msmanoj: Could we rename`TrackType` to something like `TrackPanelTab`?

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -1,20 +1,47 @@
...
+    return;
+  }
+
+  dispatch(
+    toggleTrackPanelForGenome({
+      activeGenomeId,
+      isTrackPanelOpened
+    })
+  );
+};
```
azangru: Just a thought. If you had a `getActiveTrackPanel` selector (my comment for selectors file), you could do something like this:

```
  const state = getState();
  const activeTrackPanel = getActiveTrackPanel(state);

  // after check for nulls

  const payload = {
    [activeGenomeId]: {
      ...activeTrackPanel,
      isTrackPanelOpened
    }
   }

  dispatch(updateTrackPanels(payload));
```

I am not sure whether this approach will be better, but It seems that it will reduce the amount of code in actions and reducer.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.tsx
```diff
@@ -50,7 +57,7 @@ const TrackPanelModal: FunctionComponent<TrackPanelModalProps> = (
   return (
     <section className={getTrackPanelModalClasses()}>
       <button
-        onClick={props.closeTrackPanelModal}
+        onClick={() => dispatch(closeTrackPanelModal())}
```
azangru: Yuck! HOC is better.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.tsx
```diff
@@ -9,27 +8,35 @@ import PersonalData from './modal-views/PersonalData';
...
-  launchbarExpanded: boolean;
-  trackPanelModalView: string;
-};
+const TrackPanelModal: FunctionComponent = () => {
+  const { launchbarExpanded, trackPanelModalView } = useSelector(
+    (state: RootState) => ({
+      launchbarExpanded: getLaunchbarExpanded(state),
+      trackPanelModalView: getTrackPanelModalView(state)
+    })
+  );
```
azangru: 🤢

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.tsx
```diff
@@ -9,27 +8,35 @@ import PersonalData from './modal-views/PersonalData';
...
+import closeIcon from 'static/img/track-panel/close.svg';
+
 import styles from './TrackPanelModal.scss';
 
-type TrackPanelModalProps = {
-  closeTrackPanelModal: () => void;
-  launchbarExpanded: boolean;
-  trackPanelModalView: string;
-};
+const TrackPanelModal: FunctionComponent = () => {
```
azangru: No need to specify FunctionComponent type.

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
```diff
@@ -9,29 +25,41 @@ import chevronRightIcon from 'static/img/shared/chevron-right.svg';
...
+
+type DispatchProps = {
+  closeDrawer: () => void;
+  closeTrackPanelModal: () => void;
   openTrackPanelModal: (trackPanelModalView: string) => void;
   toggleTrackPanel: (isTrackPanelOpened?: boolean) => void;
-  trackPanelModalView: string;
 };
 
+type OwnProps = {};
```
msmanoj: Why do we need this?

imran-ebi: It's an unofficial convention that we have been using to define `OwnProps` in connected components even if it doesn't have any properties. I wonder if this needs to change and also if we need to have an agreed convention for this. Any thoughts @azangru?

src/ensembl/src/content/app/browser/track-panel/trackPanelSelectors.ts
imran-ebi: >Also, do I understand correctly that this function is only used in this selector?

Yes

>Perhaps instead of casting return value of this to various types it will be better to just have getActiveTrackPanel (which returns TrackPanelStateForGenome or null) and then, in other selectors, pick relevant properties out of that object?

That will be neater

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
imran-ebi: But won't this not tell us what part of the track panel state has changed in Redux (or analytics)?

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
azangru: Personally, I just throw everything into a single `Props` type; so if there is an unofficial convention, I have been violating it for quite a while now 😄 

I am mildly in favour of just a single Props type, because I can't see any real value behind splitting the Props type into the props that come from mapStateToProps, props that come from mapDispatchFromProps, and props that come from parent component.

imran-ebi: > Personally, I just throw everything into a single Props type; so if there is an unofficial convention, I have been violating it for quite a while now 😄
You probably have 😆 

The main reason why I chose to separate the prop types is so that it is easier to decipher where the individual props are coming from by just looking at the types.

On the other hand, I admit that letting TypeScript infer the type for `mapStateToProps` and `mapDispatchToProps` is better than writing it by hand.

azangru: >  I admit that letting TypeScript infer the type for mapStateToProps and mapDispatchToProps is better than writing it by hand

Erm. You'll still have to write them by hand. Unless you are using the `useSelector` hook (😐), when it will probably infer the type of the thing returned by the selector.

imran-ebi: > On the other hand, I admit that letting TypeScript infer the type for mapStateToProps and mapDispatchToProps is better than writing it by hand.
By this, what I meant was writing these types separately for the constants that we use at the bottom of the file. For example:

`const mapDispatchToProps: DispatchProps = { updateTrackState };`

azangru: Oh yes, we won't need to declare the type of the `mapDispatchToProps` object; `react-redux`'s types should do that for us.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.tsx
imran-ebi: I changed it to a connected component

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
imran-ebi: I have made changes to be somewhat similar to this.

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
imran-ebi: I have got rid of these different types for track panel related components.

src/ensembl/src/content/app/browser/browser-storage-service.ts
imran-ebi: I have renamed `TrackType` to be `TrackSet` and `BrowserTabs` to `TrackPanelTabs`. Also renamed the state property `selectedBrowserTab` to `selectedTrackPanelTab`.

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
```diff
@@ -20,51 +14,49 @@ type TrackPanelBarIconProps = {
...
-          buttonStatus={getViewIconStatus()}
-          description={props.iconConfig.description}
-          onClick={toggleModalView}
-          image={props.iconConfig.icon}
-        />
-      </dt>
-    );
-  }
-);
+const TrackPanelBarIcon = memo((props: TrackPanelBarIconProps) => {
```
azangru: `memo` is a premature optimization.

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
```diff
@@ -20,51 +14,49 @@ type TrackPanelBarIconProps = {
...
+  const [toggleState, setToggleState] = useState(false);
+
+  useEffect(() => {
+    if (!props.isTrackPanelModalOpened) {
+      setToggleState(false);
+    }
+  }, [props.isTrackPanelModalOpened]);
+
+  const toggleModalView = useCallback(() => {
```
azangru: `useCallback` is also a premature optimisation

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBarIcon.tsx
imran-ebi: Ah the old code has come back to haunt us 😄 I'll remove these.

PR #145, opened 2019-08-15
Recently viewed bookmarks

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-237

## Importance
August Sprint 3

## Description
Added the 'Previously viewed' section to the track panel bookmarks section.

## Views affected
TrackPanel -> Bookmarks

### Other effects


azangru:
General consideration: you use the general term "bookmark", but from what we discussed, it seems that we will have at least two types of bookmarks for genome browser:
1) bookmarks that are created automatically when the user changes focus object
2) bookmarks that are created manually and are a snapshot of current location + states of tracks

What this PR is about is bookmarks of type 1. They are going to be different from bookmarks of type 2; so we should name them accordingly. Perhaps `recentlyViewed` or `previouslyViewed` instead of `bookmarks`? I am talking of type `Bookmark` in trackPanelState, `bookmarks` field in redux state, `bookmarks` prop in TrackPanelBookmarks component, and so on. We should be able to tell one type of bookmarks from the other.

src/ensembl/src/content/app/browser/track-panel/track-panel-storage-service.test.ts
```diff
@@ -0,0 +1,113 @@
...
+          'gene-pc-fwd': ImageButtonStatus.INACTIVE
+        },
+        main: {
+          'gene-feat': ImageButtonStatus.INACTIVE,
+          'gene-feat-1': ImageButtonStatus.INACTIVE
+        }
+      }
+    }
+  ]
+};
```
azangru: If all that these test do is check that a particular `mockStorageService` method gets called with a particular argument, then what does it matter whether the argument has a shape real bookmarks? Anything would do, just type it as `any`.

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -82,6 +114,49 @@ export const changeTrackPanelModalViewForGenome: ActionCreator<
...
+> = () => (dispatch, getState: () => RootState) => {
+  const state = getState();
+  const activeGenomeId = getBrowserActiveGenomeId(state);
+  const activeEnsObject = getBrowserActiveEnsObject(state);
+  if (!activeGenomeId || !activeEnsObject) {
+    return;
+  }
+  const trackStates = getBrowserTrackStates(state)[activeGenomeId];
+
+  const activeGenomeBookmarks = [...getActiveGenomeBookmarks(getState())];
```
azangru: state is already extracted using getState above

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -16,24 +19,31 @@ import { getFormattedLocation } from 'src/shared/helpers/regionFormatter';
...
+  const getExampleObjLabel = (exampleObject: EnsObject | Bookmark) => {
     if (exampleObject.object_type === 'gene') {
       return exampleObject.label;
     } else {
       return getFormattedLocation(exampleObject.location);
     }
   };
 
-  const getPreviouslyViewed = () => {
+  const getExampleLinks = () => {
```
azangru: This looks like a React component (`<ExampleLinks ensObjects={props.exampleEnsObjects} />`) rather than a function inside another component

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -44,10 +54,43 @@ export const TrackPanelBookmarks = (props: TrackPanelBookmarksProps) => {
...
+          <span className={styles.previouslyViewedType}>
+            {' '}
+            {upperFirst(exampleObject.object_type)}
+          </span>
+        </dd>
+      );
+    });
+  };
+
+  const getBookmarkedLinks = () => {
```
azangru: This also looks like a React component (`<PreviouslyViewed items={props.previouslyViewed} />`) rather than a function inside another component.

src/ensembl/src/content/app/browser/track-panel/trackPanelState.ts
```diff
@@ -1,11 +1,32 @@
...
+  genome_id: string;
+  object_id: string;
+  object_type: string;
+  location: {
+    chromosome: string;
+    start: number;
+    end: number;
+  };
+  label: string;
+  trackStates: GenomeTrackStates;
```
azangru: If this is describing data for a previously viewed item, does it need anything at all other than just one single data point: ensembl object id?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-storage-service.test.ts
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/trackPanelState.ts
msmanoj: @azangru as per the Gherkin syntax mentioned in the ticket https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-237, we will have to store the track states along with the previously viewed links. This has already been approved by Andrea.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('renders without any error', () => {
+    expect(wrapper).toHaveLength(1);
+  });
+
+  it('renders N number of previously viewed links based on the length of previously viewed objects', () => {
+    const previouslyViewedLinksWrapper = wrapper.find(PreviouslyViewedLinks);
+
+    expect(previouslyViewedLinksWrapper.find('div.link').length).toBe(2);
```
azangru: - div is unnecessary? (if there is any concern about class collision, we could use more unique names for link classes (e.g. 'mock-link')
- 2 is a magic number here; better either to use `props. previouslyViewedObjects.length`, or a variable (numberOfPreviouslyViewedObjects) that will be used both when creating the props and when checking the result

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      .first()
+      .simulate('click');
+
+    expect(closeTrackPanelModalMock).toBeCalled();
+  });
+
+  it('renders N number of Links based on the length of example links', () => {
+    const exampleLinksWrapper = wrapper.find(ExampleLinks);
+
+    expect(exampleLinksWrapper.find('div.link').length).toBe(2);
```
azangru: same as with test for previously viewed links

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('calls the closeTrackPanelModalMock when the link is clicked', () => {
```
azangru: this test is called identically to the one before previous. Having identically named tests is not good. In fact, there is an eslint rule in jest plugin to point this out.

msmanoj: Renamed the test

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
msmanoj: Updated it

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
msmanoj: Removed the div and using a random length.

azangru:
General question: should the area with links to previously viewed features be scrollable (I can now get it beyond the window border):

image

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.scss
```diff
@@ -39,18 +39,34 @@
...
+    border-bottom: 1px solid $ens-grey;
   }
 
   dd {
     margin: 15px 0 0 15px;
+    position: relative;
   }
 }
+
+.previouslyViewedType {
```
azangru: What does "previouslyViewedType" mean, and how is it different from "previouslyViewed" above?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -4,61 +4,150 @@ import { connect } from 'react-redux';
...
+};
+
+type DispatchProps = {
   fetchExampleEnsObjects: (objectId: string) => void;
+  updateTrackStates: (trackStates: TrackStates) => void;
+  closeTrackPanelModal: () => void;
+};
+export type TrackPanelBookmarksProps = StateProps & DispatchProps;
+
+const getExampleObjLabel = (exampleObject: EnsObject | Bookmark) => {
```
azangru: 1) name of the function no longer describes what the function does
2) both gene and region objects now have the `label` field
3) is this function even necessary anymore?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -4,61 +4,150 @@ import { connect } from 'react-redux';
...
+    <div>
+      {[...props.previouslyViewedObjects]
+        .reverse()
+        .map((previouslyViewedObject, index) => {
+          const locationStr = `${previouslyViewedObject.location.chromosome}:${previouslyViewedObject.location.start}-${previouslyViewedObject.location.end}`;
+          const path = urlFor.browser({
+            genomeId: props.activeGenomeId,
+            focus: previouslyViewedObject.object_id,
+            location: locationStr
+          });
```
azangru: What we definitely don't want is the location parameter here. We want Genome Browser to take us to the default position of the focus object; and where this default position is, is something that only Genome Browser now knows about. So what we need to pass to Genome Browser is genome id and focus object id. It can be done either by creating React Router links (as here) or by sending direct message to Genome Browser. Either method will work here; I am not sure whether one of them is preferable over the other.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -4,61 +4,150 @@ import { connect } from 'react-redux';
...
+
+          const onClickHandler = () => {
```
azangru: please move click handler outside of the map function

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -3,9 +3,42 @@ import { ThunkAction } from 'redux-thunk';
...
+  ensObject: EnsObject,
+  trackStates: GenomeTrackStates
+): Bookmark => {
+  let bookmarkLabel = ensObject.object_type === 'gene' ? ensObject.label : '';
+
+  if (!bookmarkLabel && ensObject.stable_id) {
+    bookmarkLabel = ensObject.label;
+  } else if (!bookmarkLabel) {
+    bookmarkLabel = getFormattedLocation(ensObject.location);
+  }
```
azangru: Currently all ens objects, including regions, that we get from the backend have a `label` field. There is no need for all this logic.

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -88,10 +121,64 @@ export const changeTrackPanelModalViewForGenome: ActionCreator<
...
+      buildBookmark(activeEnsObject, trackStates)
+    );
+  } else if (existingIndex !== -1) {
+    // If it is already present, bump it to the end
+    activeGenomePreviouslyViewedObjects.push({
+      ...activeGenomePreviouslyViewedObjects[existingIndex],
+      trackStates
+    });
+    activeGenomePreviouslyViewedObjects.splice(existingIndex, 1);
+  }
```
azangru: 1) `else if` -> `else`
2) in order to avoid actually creating a new previously viewed object, would this approach work better?

```
const [previouslyViewedObject] = activeGenomePreviouslyViewedObjects.splice(existingIndex, 1);
activeGenomePreviouslyViewedObjects.push(previouslyViewedObject);
```
?

As a side note, the `activeGenome` part of `activeGenomePreviouslyViewedObjects` name looks completely unnecessary. Could we drop it?

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -101,17 +188,63 @@ export const changeHighlightedTrackId: ActionCreator<
...
+//   dispatch(
+//     updateTrackPanelForGenome({
+//       activeGenomeId,
+//       data: {
+//         ...getActiveTrackPanel(getState()),
+//         bookmarks: activeGenomeBookmarks
+//       }
+//     })
+//   );
+// };
```
azangru: Perhaps remove all this commented code? It probably is not correct anyway (we probably were not supposed to cap bookmarks at 20)

msmanoj:
> General question: should the area with links to previously viewed features be scrollable (I can now get it beyond the window border)?

I would guess so. I've now added overflow auto to TrackPanelModel to fix this. I didn't realise it wasn't there.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.scss
msmanoj: "previouslyViewedType" is the greyed out type display next to the link and "previouslyViewed" is the actual text used for the link.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: Not sure of label was available for reigns when I wrote this.
Removed this function now.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: > We want Genome Browser to take us to the default position of the focus object;

I initially thought that is the case but we are taking the user to the location they previously viewed. I will look for a way to do this without adding it to the link. Please let me know if you think of an alternative way to do this.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
msmanoj: Removed it. Thanks

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
msmanoj: Done these changes.

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
msmanoj: That's true. Remove it.

msmanoj: That's true. Removed it.

azangru:
I talked to Andrea regarding two issues:
1) What recently viewed links and in which order should appear in the recently viewed panel
2) Where should the user be taken when they click on a recently viewed link (see our [conversation](https://github.com/Ensembl/ensembl-client/pull/145#discussion_r321804529) above; we also asked Emily's opinion on what she would expect to happen when she clicks on a link).

**The conclusions:**
- the way recently viewed links is appearing is slightly wrong (Andrea will update the ticket with a schematic of which links and in which order should be shown)
- when user clicks on the link to a recently viewed object, he should be taken to the **default** position of this object, not to the position where he left (rationale: if user clicks on a link that says gene such and such, they expect to see this gene; if the link took them to the view where they left, it may not even contain this gene anymore, because user may have dragged the browser image in such a way that the focus gene is not visible any longer)

azangru:
I navigated to Genome Browser (with human as selected species) for the first time, and I already see BRCA2 in the list of previously viewed objects. This contradicts the first Gherkin scenario in the description of the ticket.

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -99,10 +106,62 @@ export const changeTrackPanelModalViewForGenome: ActionCreator<
...
+      previouslyViewedObject.object_id === activeEnsObject.object_id
+  );
+  if (existingIndex === -1) {
+    // IF it is not present, add it to the end
+    previouslyViewedObjects.push({
+      genome_id: activeEnsObject.genome_id,
+      object_id: activeEnsObject.object_id,
+      object_type: activeEnsObject.object_type,
+      label: activeEnsObject.label,
+      trackStates: { ...trackStates }
```
azangru: why the spread (...trackStates)?

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -99,10 +106,62 @@ export const changeTrackPanelModalViewForGenome: ActionCreator<
...
+    });
```
azangru: So, we are creating an infinite history of previously viewed objects... Is this _really_ what we want to do? I know Andrea says that we don’t know how many we want until we try, but realistically, how many previously viewed objects should we allow our application to store? A hundred? A thousand? Ten thousand? Where is the threshold after which they have definitely become no more than useless noise?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,57 @@
...
+  columns: 4;
+  position: relative;
+  width: calc(100% - 110px);
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  overflow-x: auto;
+
+  a {
+    font-weight: $normal;
```
azangru: why? overriding what?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
```diff
@@ -0,0 +1,86 @@
...
+                genomeId: previouslyViewedObject.genome_id,
+                focus: previouslyViewedObject.object_id
+              });
+
+              const onClickHandler = () => {
+                props.updateTrackStates({
+                  [previouslyViewedObject.genome_id]: {
+                    ...previouslyViewedObject.trackStates
+                  }
+                });
```
azangru: What is the reason for calling the `updateTrackStates` action here?

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -79,10 +82,14 @@ export const setDataFromUrlAndSave: ActionCreator<
...
     browserStorageService.updateChrLocation({ [activeGenomeId]: chrLocation });
-  activeEnsObjectId &&
+
+  if (activeEnsObjectId) {
     browserStorageService.updateActiveEnsObjectIds({
       [activeGenomeId]: activeEnsObjectId
     });
+
+    dispatch(updatePreviouslyViewedObjectsAndSave());
+  }
```
azangru: Having `updatePreviouslyViewedObjectsAndSave` both in `setDataFromUrlAndSave` and in `changeObject` makes it being called twice. It should either be called in `changeBrowserLocation` and `changeObject` or only in `setDataFromUrlAndSave`. Not sure which is better.

src/ensembl/src/content/app/browser/track-panel/trackPanelState.ts
```diff
@@ -1,13 +1,29 @@
...
+};
+
+export type Bookmarks = {
+  [genomeId: string]: Bookmark[];
+};
+
+// const bookmarksFromStorage: Bookmarks = trackPanelStorageService.getBookmarks();
+// const previouslyViewedObjectsFromStorage: Bookmarks = trackPanelStorageService.getPreviouslyViewedObjects();
```
azangru: leftover comments

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,57 @@
...
+    margin-bottom: 15px;
+    position: relative;
+  }
+
+  dd {
+    margin: 5px 0 0 15px;
+    position: relative;
+    width: 200px;
+    padding-right: 10px;
+    text-overflow: ellipsis;
```
azangru: Both `dd` and `a` have `text-overflow: ellipsis` — are both necessary? Also, `dd` has a max-width of 200px, but `a` (a child of `dd`) has a max-width of 215px. Is this correct?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,57 @@
...
+    font-size: 13px;
+  }
+
+  dt {
+    color: $dark-grey;
+    font-size: 12px;
+    border-bottom: 1px solid $grey;
+    margin-bottom: 15px;
+    position: relative;
+  }
```
azangru: There doesn't seem to be a `dt` element inside of the `.linkWrapper` class? Is this block of styles copy-pasted from `TrackPanelModal.scss`?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  dd {
+    margin: 5px 0 0 15px;
+    position: relative;
+    width: 200px;
+    padding-right: 10px;
+    text-overflow: ellipsis;
+    overflow: hidden;
+  }
+}
```
azangru: As a side note, why do we use definition list tags for the contents of this drawer? Why aren't they just spans?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import { RootState } from 'src/store';
+import { Bookmark } from 'src/content/app/browser/track-panel/trackPanelState';
+import { TrackStates } from 'src/content/app/browser/track-panel/trackPanelConfig';
+import { updateTrackStates } from 'src/content/app/browser/browserActions';
+import { closeTrackPanelModal } from 'src/content/app/browser/track-panel/trackPanelActions';
+import { getActiveGenomePreviouslyViewedObjects } from 'src/content/app/browser/track-panel/trackPanelSelectors';
+import { closeDrawer } from 'src/content/app/browser/drawer/drawerActions';
+import upperFirst from 'lodash/upperFirst';
+import styles from './DrawerBookmarks.scss';
```
azangru: _(style issue)_ Can we keep some order in the order of imports? Imports from node modules in the top; imports from styles in the very bottom; imports from helpers, actions, selectors, components, etc. clustered in their own groups?

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
```diff
@@ -99,10 +106,62 @@ export const changeTrackPanelModalViewForGenome: ActionCreator<
   );
 };
 
+export const updatePreviouslyViewedObjectsAndSave: ActionCreator<
+  ThunkAction<void, any, null, Action<string>>
+> = () => (dispatch, getState: () => RootState) => {
```
azangru: When is this action called — **after** we have switched to the new focus object in redux store or **before** we have? From the code, it seems that the action is called after redux store has been updated to the new focus object, but shouldn't it be called before, in order to capture the track states of the previous object?

I have also noticed that you call `updateTrackStates` in individual components; but if this action were called right before the focus object has changed, it could have incorporated snapshotting of track states within the code of this action, and there would be no need to call `updateTrackStates` from individual components? At least, I hope there wouldn't.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
msmanoj: I've changed it to span. I think I need to be more cautious and ask these to myself before I reuse stuff.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
msmanoj: I would have copy-pasted from `TrackPanelModal.scss`. I've removed it now.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
msmanoj: Removed max width from `a`.

src/ensembl/src/content/app/browser/track-panel/trackPanelState.ts
msmanoj: Removed it

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
msmanoj: Removed it.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: Removed it from `setDataFromUrlAndSave ` and added it to `changeBrowserLocation`. Keeping it only on `setDataFromUrlAndSave` doesn't seem to work when I change the focus from the zmenu.

src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts
azangru: If we want to have a completely new trackStates object here, you can use something like `cloneDeep` from lodash

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -30,6 +30,24 @@ export const changeDrawerView: ActionCreator<
...
+    return;
+  }
+
+  dispatch(
+    changeDrawerViewForGenome({
+      [activeGenomeId]: drawerView
+    })
+  );
+
+  dispatch(toggleDrawer(true));
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,57 @@
...
+
+.drawerTitle {
+  margin: 20px 0 0 30px;
+}
+
+.contentWrapper {
+  height: 80%;
+}
+
+.previouslyViewedObjectType {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.scss
```diff
@@ -38,19 +39,50 @@
   }
 }
 
-.previouslyViewed {
-  margin-top: 60px;
+.previouslyViewedObject {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/TrackPanelModal.scss
```diff
@@ -38,19 +39,50 @@
...
   }
 
   dd {
-    margin: 15px 0 0 15px;
+    margin: 5px 0 0 15px;
+    position: relative;
   }
 }
+
+.previouslyViewedObjectType {
```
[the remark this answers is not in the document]
msmanoj: That's done

azangru:
image

Is the title correct? It says "All previously viewed", but shows only 2 links (that exceed the limit of 20 links shown in the track panel). Should it be "other previously viewed", or something? A question to ask Andrea.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+type DispatchProps = {
+  updateTrackStatesAndSave: (trackStates: BrowserTrackStates) => void;
+  closeTrackPanelModal: () => void;
+  closeDrawer: () => void;
+};
+export type DrawerBookmarksProps = StateProps & DispatchProps;
+
+const DrawerBookmarks: FunctionComponent<DrawerBookmarksProps> = (
```
azangru: no need to use the FunctionComponent type

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
```diff
@@ -74,19 +69,30 @@ const TrackPanelList = (props: TrackPanelListProps) => {
...
-    const trackStatus = get(
-      props.trackStates,
-      `${activeGenomeId}.${categoryName}.${track_id}`,
-      defaultTrackStatus
-    );
+    let trackStatus = getDefaultTrackStatus();
+
+    if (activeEnsObject) {
+      // FIXME: Temporary hack until we have a set of proper track names
+      if (track_id.indexOf('track:gene') === 0) {
```
azangru: `startsWith`?

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -139,12 +142,32 @@ const TrackPanelListItem = (props: Props) => {
...
-    props.updateTrackStates({
-      genomeId: activeGenomeId as string,
-      categoryName,
-      trackId: track.track_id,
-      status: newStatus
-    });
+    if (!activeGenomeId || !activeEnsObjectId) {
+      return;
+    }
+    if (track.track_id.indexOf('track:gene') === 0) {
```
azangru: `startsWith`?

Does this also need a FIXME comment regarding temporary hack?

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -126,27 +137,36 @@ export const updateDefaultPositionFlag = createStandardAction(
...
+  const activeEnsObjectId = getBrowserActiveEnsObjectId(state);
+  const activeGenomeTrackStates = getBrowserActiveGenomeTrackStates(state);
+
+  const mergedTrackStates = merge(
+    {},
+    get(activeGenomeTrackStates, `objectTracks.${activeEnsObjectId}`),
+    get(activeGenomeTrackStates, 'commonTracks')
+  ) as TrackStates;
+
+  Object.values(mergedTrackStates).forEach((trackStates) => {
```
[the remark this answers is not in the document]
msmanoj: I tried using this but typescript is not happy with `Object.keys`. 
Also, the result of .map(Object.keys) will be [[trackIDs],[trackIDs]] so we can't directly use .forEach.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.linksWrapper {
+  margin: 20px 0 0 40px;
+  height: 100%;
+  columns: 4;
+  position: relative;
+  width: calc(100% - 110px);
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
```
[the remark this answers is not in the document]
msmanoj: I tried this but I get something like this:
Screenshot 2019-10-14 at 15 52 02

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
msmanoj: Removed it

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -195,6 +211,26 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+      get(trackStatesFromStorage, `${props.activeGenomeId}.commonTracks`)
+    ) as TrackStates;
+
+    Object.values(mergedTrackStates).forEach((trackStates) => {
+      Object.keys(trackStates).forEach((trackId) => {
+        const trackStatus: string =
+          trackStates[trackId] === ImageButtonStatus.ACTIVE ? 'on' : 'off';
+        browserMessagingService.send('bpane', { [trackStatus]: trackId });
+      });
+    });
```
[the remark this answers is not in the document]
msmanoj: I've replaced merge with object spread.

> Should all contents of this useEffect perhaps be extracted into one redux action?

I have moved the contents from useEffect to a separate action. However, the action `resetBrowserTrackStates` is more likely to be changed in the near future so I think it is better to think about defining a common function later on.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
msmanoj: Not sure why I called it here but I have removed it now as it's no longer required.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+type StateProps = {
+  previouslyViewedObjects: PreviouslyViewedObject[];
+};
+
+type DispatchProps = {
+  closeTrackPanelModal: () => void;
+  closeDrawer: () => void;
+};
+export type DrawerBookmarksProps = StateProps & DispatchProps;
```
imran-ebi: This can be just one props type

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const props = {
+    activeGenomeId: faker.random.word(),
+    exampleEnsObjects: times(numberOfExampleObjects, () => createEnsObject()),
+    previouslyViewedObjects: times(numberOfPreviouslyViewedObjects, () =>
+      createPreviouslyViewedLink()
+    ),
+    fetchExampleEnsObjects: fetchExampleEnsObjectsMock,
+    updateTrackStatesAndSave: updateTrackStatesAndSaveMock,
+    closeTrackPanelModal: closeTrackPanelModalMock,
+    changeDrawerViewAndOpen: changeDrawerViewAndOpenMock
```
imran-ebi: These can use the shortened format as both the property names and value match.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  it('calls closeTrackPanelModal when a previously viewed object link is clicked', () => {
+    const previouslyViewedLinksWrapper = wrapper.find(PreviouslyViewedLinks);
+
+    previouslyViewedLinksWrapper
+      .find('.link')
+      .first()
+      .simulate('click');
+
+    expect(closeTrackPanelModalMock).toBeCalled();
```
imran-ebi: Can we check the number of times the function is called with `toHaveBeenCalledTimes(1)` rather than whether it is just called?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  });
+
+  it('calls closeTrackPanelModal when an example object link is clicked', () => {
+    const exampleLinksWrapper = wrapper.find(ExampleLinks);
+    exampleLinksWrapper
+      .find('.link')
+      .first()
+      .simulate('click');
+
+    expect(closeTrackPanelModalMock).toBeCalled();
```
imran-ebi: Same as mentioned above. We can use `toHaveBeenCalledTimes` instead.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -1,64 +1,146 @@
...
+  previouslyViewedObjects: PreviouslyViewedObject[];
+};
+
+type DispatchProps = {
   fetchExampleEnsObjects: (objectId: string) => void;
+  updateTrackStatesAndSave: (trackStates: BrowserTrackStates) => void;
+  closeTrackPanelModal: () => void;
+  changeDrawerViewAndOpen: (drawerView: string) => void;
 };
+export type TrackPanelBookmarksProps = StateProps & DispatchProps;
```
imran-ebi: One props type again?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
msmanoj: Why do we need to check the number of times it is being called? Isn't that used where we need to check multiple calls?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imran-ebi: I misunderstood that this was used in a hook. Let's leave like this for now.

msmanoj:
> Is the title correct? It says "All previously viewed", but shows only 2 links (that exceed the limit of 20 links shown in the track panel). Should it be "other previously viewed", or something? A question to ask Andrea.

Changed the title to 'Previously viewed' after confirming with Andrea.

@imran-ebi @azangru could you please check if there is anything else pending with this PR just so we can keep it ready to be merged once the track-reset branch is merged?

msmanoj:
@imran-ebi @azangru just a gentle reminder about this PR. Could you please check if there is anything else pending?

imran-ebi:
@msmanoj Had a quick run through of the latest changes and I am fine with them. @azangru might like to check on some of the suggestions?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -127,20 +132,29 @@ export const Browser: FunctionComponent<BrowserProps> = (
...
+    props.setDataFromUrlAndSave(payload);
   };
 
   const dispatchBrowserLocation = (
     genomeId: string,
+    focus: string | null,
     chrLocation: ChrLocation
   ) => {
-    props.changeBrowserLocation(genomeId, chrLocation);
+    props.changeBrowserLocation(genomeId, focus, chrLocation);
```
azangru: Can we change the function signature for changeBrowserLocation and dispatchBrowserLocation, so that instead of a list of arguments they accept an object with an optional field?

calling a function like this:

`dispatchBrowserLocation({genomeId: 'blah', chrLocation: 'something'})` is much more readable than `dispatchBrowserLocation(genomeId, null, chrLocation)` because it avoids asking the question "what on earth does the `null`" do?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -1,64 +1,141 @@
...
-      });
+        return (
+          <dd key={exampleObject.object_id}>
+            <Link to={path} onClick={props.closeTrackPanelModal}>
+              {exampleObject.label}
+            </Link>
+            <span className={styles.previouslyViewedType}>
+              {upperFirst(exampleObject.object_type)}
+            </span>
+          </dd>
```
azangru: can we slowly get rid of dt's and dd's when do not have definition-related semantics?

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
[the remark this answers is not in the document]
msmanoj: With flex I get this:
Screenshot 2019-10-31 at 15 54 48


But with column fill, I get this:
Screenshot 2019-10-31 at 15 55 11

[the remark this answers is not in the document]
azangru: Oh, I see what you mean. Are you saying that if we just leave the columns rule, then they will never grow horizontally to more than 4?

I can't get rid of the feeling that we are using columns and flex incorrectly — it will confuse people who will read this CSS. It confused me 😃  I expected there to be 4 columns, but the top image shows five 😃 

What if we remove the columns altogether then, and only use flex, which, I think was what you initially started with.

Will these styles work: https://jsfiddle.net/4e0k51at/ ?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: That's done

azangru:
> @imran-ebi @azangru just a gentle reminder about this PR. Could you please check if there is anything else pending?

Point of order 😃  Please post such reminders on Slack — in long-standing pull requests with long conversation threads and multiple subsequent commits it is very easy to miss these additional notifications.

src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerBookmarks.scss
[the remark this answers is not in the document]
msmanoj: I've removed the columns and set the item width to 25% 👍

PR #146, opened 2019-08-16
Close Track Configuration box by clicking anywhere - 206

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-206

## Importance
Aug Sprint 2

## Description
Added a custom react hook to handle clicks made outside an element

### Other effects


src/ensembl/src/content/app/browser/BrowserCog.tsx
```diff
@@ -22,6 +24,11 @@ const BrowserCog: FunctionComponent<BrowserCogProps> = (
 ) => {
   const { cogActivated, updateSelectedCog, index } = props;
 
+  const ref = useRef(null);
+  useOutsideClick(ref, () => {
+    updateSelectedCog('');
+  });
```
imran-ebi: Since this is a one statement function how about writing it as `() => updateSelectedCog('')`

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -0,0 +1,21 @@
+import { useEffect } from 'react';
+
+export default function useOutsideClick<T extends HTMLElement>(
+  ref: React.RefObject<T>,
+  callback: () => void
+): void {
```
imran-ebi: `void` as return type is not necessary here

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -0,0 +1,21 @@
...
+      callback();
+    }
+  };
+
+  useEffect(() => {
+    // Bind the event listener
+    document.addEventListener('click', handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener('click', handleClickOutside);
```
imran-ebi: This function too (like above) can be written as a one liner.

src/ensembl/src/content/app/browser/BrowserCog.tsx
msmanoj: That's done

src/ensembl/src/shared/hooks/useOutsideClick.tsx
msmanoj: Removed it

src/ensembl/src/shared/hooks/useOutsideClick.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/BrowserCog.tsx
```diff
@@ -65,7 +70,9 @@ const BrowserCog: FunctionComponent<BrowserCogProps> = (
         return (
           item && (
             <animated.div key={key} style={style}>
-              <BrowserTrackConfig />
+              <div ref={ref}>
+                <BrowserTrackConfig />
+              </div>
```
azangru: If the only purpose of this div is to provide an anchor point for a ref, then perhaps, instead of adding a div, we could use forwardRef on BrowserTrackConfig?

msmanoj: I tried using `forwardRef` but I'm not able to make TypeScrips happy for some reason.
Although, this `div` here doesn't have much purpose, I feel if makes the code simpler than using `forwardRefs`. Do we really need `forwardRef` here?

azangru: > Do we really need forwardRef here?

Maybe not :smiley: 
Out of curiosity, why do we add this hook and this ref in the BrowserCog component rather than in the BrowserTrackConfig component?

msmanoj: By this way I'm just avoiding passing down `updateSelectedCog` to BrowserTrackConfig.

azangru: That’s right. But doesn’t the ability to close when someone clicks outside a component more naturally belong to this component rather than to its parent?
In other words, isn’t it reasonable for  <BrowserTrackConfig /> component to receive the onClose (or handleClose, etc.) prop that it uses when it decides that it needs to close?

msmanoj: That's a fair point. I've update it.

PR #147, opened 2019-08-20
Remove Zurb Foundation

azangru:
## Type
- code refactoring

## Related JIRA Issue(s)
ENSWBSITES-261

## Description
Ensembl is a highly customized web app that doesn’t seem to benefit much from using the Zurb Foundation CSS framework; so we decided to remove it from the list of our dependencies.

This PR keeps the few Foundation CSS rules that we use (such as normalization styles or global styles) and removes the rest.

## Views affected
Pretty much all the views

## Topics for Discussion
1. How best to encode styles at different breakpoints?
- should we use mixins for breakpoints, or is notation such as `@media #{$breakpoint-medium} {...}` good enough?
- should breakpoints define `max-width` or `min-width`?


ens-ap5:
It all looks OK to me. The issue of breakpoints is one I'll be looking into once this code change is merged - see ENSWBSITES-205

PR #149, opened 2019-08-27
Remove refs from Browser component

azangru:
## Type
- refactoring

## Related JIRA Issue(s)
ENSWBSITES-174

## Description
The task description says "Check why useRef is used in the Browser.tsx and remove it if it’s not required".

`useRef` was used for 2 purposes in the Browser component:

1) to hold reference to the html element into which genome browser renders the canvas element, and
2) to get reference to the last value of active genome id coming from the props

Regarding the first usage, it is still relevant (some React components, such as zmenu, may need to be able to access the wrapper around canvas directly to calculate the dimensions of the canvas), but the ref itself can be moved from `Browser` to `BrowserImage`.
Second usage seems to be redundant at the moment.

**Note:** `BrowserImage` contained code that communicated with genome browser using the old mechanism (sending custom events via a DOM element) to tell genome browser stuff about cogs. I've changed the communication method to the new one (via browser-messaging-service) and moved this code to `BrowserCogList` component.

## Views affected
Genome browser screen.

### DIFF
--- a/src/ensembl/src/content/app/browser/BrowserCogList.tsx
+++ b/src/ensembl/src/content/app/browser/BrowserCogList.tsx
@@ -14,6 +14,8 @@ import {
   getBrowserActivated,
   getBrowserCogList,
   getBrowserCogTrackList,
+  getTrackConfigNames,
+  getTrackConfigLabel,
   getBrowserSelectedCog
 } from './browserSelectors';
 
@@ -23,6 +25,8 @@ type BrowserCogListProps = {
   browserActivated: boolean;
   browserCogList: number;
   browserCogTrackList: CogList;
+  trackConfigNames: any; // FIXME
+  trackConfigLabel: any; // FIXME
   selectedCog: any;
   updateCogList: (cogList: number) => void;
   updateCogTrackList: (track_y: CogList) => void;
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/browser/BrowserCogList.tsx
```diff
@@ -55,6 +59,37 @@ const BrowserCogList = (props: BrowserCogListProps) => {
...
   }, []);
 
+  useEffect(() => {
+    if (props.browserCogTrackList) {
+      const ons: string[] = [];
+      const offs: string[] = [];
+
+      /* what the frontend and backend call labels and names is flipped */
+      Object.keys(props.browserCogTrackList).forEach((name) => {
+        /* undefined means not seen means on for names */
```
imran-ebi: This comment is a bit cryptic. Can this be rephrased?

src/ensembl/src/content/app/browser/BrowserCogList.tsx
```diff
@@ -55,6 +59,37 @@ const BrowserCogList = (props: BrowserCogListProps) => {
...
+        if (props.trackConfigNames[name]) {
+          ons.push(`${name}:label`);
+        } else {
+          offs.push(`${name}:label`);
+        }
+        /* undefined means not seen means off for labels */
```
imran-ebi: Same as above

src/ensembl/src/content/app/browser/BrowserCogList.tsx
azangru: I completely agree, but that's not my comment; it came along with the code copied from BrowserImage. I will ask @ens-ds23 what he meant 😄

azangru: @imran-ebi I have talked to Dan and updated the comments. I hope they make a bit more sense now.

imran-ebi: Yes that's much better.

imran-ebi:
Is there anything else that needs to be done in this PR?

PR #153, opened 2019-08-30
Remove 'ens-' prefix from colour variable names

ens-ap5:
## Type

- [ ] Bug fix
- [ ] New feature

(Not sure - really just a code cleanup.)

## Related JIRA Issue(s)

https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-205 (not directly relevant, but I want to make sure the CSS is clean before I start making major edits)

## Importance

This is only urgent in the sense that any new or updated components with colour variables in the CSS will be incompatible with this branch (see Complications, below).

## Description

I have removed the 'ens-' prefix from all the colour variable names. It made sense to use this prefix when we still used Foundation, to avoid name clashes, but now the longer names just make the CSS harder to read. There was also a potential confusion between $white (#fefefe) and $ens-white (#fff), which this PR fixes by renaming them $off-white and $white respectively.

## Views affected

All views - colours should be (and are) completely unchanged, as the code change is behind-the-scenes.

### Other effects

No automated tests needed, as this only affects cosmetic CSS. Browser and Storybook have been visually checked for issues and none found so far.

## Possible complications

If any new components are merged in from branches that don't have this update, they will appear as black. However this can be avoided (after this PR is merged) by merging dev into your branch and updating any CSS colours before submitting your PR.

### DIFF
--- a/src/ensembl/src/styles/_settings.scss
+++ b/src/ensembl/src/styles/_settings.scss
@@ -4,8 +4,9 @@
 $global-font-size: 14px;
 $global-lineheight: 1.5;
 $black: #17354e;
-$white: #fefefe;
-$body-background: $white;
+$white: #fff;
+$off-white: #fefefe;
+$body-background: $off-white;
 $body-font-color: $black;
 $body-font-family: Lato, 'Helvetica Neue', Helvetica, Roboto, Arial, sans-serif;
 $body-antialiased: true;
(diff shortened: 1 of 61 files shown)


src/ensembl/src/styles/_settings.scss
```diff
@@ -17,43 +18,40 @@ $global-button-cursor: auto;
...
+  dark-blue: $dark-blue;
+  red: $red;
+  dark-grey: $dark-grey;
+  medium-dark-grey: $medium-dark-grey;
+  grey: $grey;
+  medium-light-grey: $medium-light-grey;
+  light-grey: $light-grey;
+  green: $green;
+  white: $white;
+  shadow-color: $shadow-color;
```
azangru: Should off-white also go to exported values and to storybook then?

azangru: As discussed offline, let's merge this first and then see if we need to add extra colours to storybook (and whether we do indeed need two white colours)

ens-ap5: OK. Could you take a look at the conflict? It looks to me like a CSS classname has been changed, so the new classname is using the old colour name whereas the old classname has the new colour name! (which is what I was afraid might happen)

azangru: Yes, that's exactly it. The class `currentDrawerView` was renamed to `trackHighlighted`, and at the same time the background colour variable name changed from `$ens-light-blue` to `$light-blue`.

I pushed a commit resolving this conflict.

azangru:
The failed check is because our CI/CD system has troubles with underscores in branch names. This shouldn't affect the merge.

PR #154, opened 2019-09-02
Send 'activity outside' action when user clicks anywhere outside z-menu

azangru:
## Type

## Description
Zmenu needs to close when user clicks anywhere outside of it.
Currently, zmenu closes only when user click on some other point of genome browser image. This is because genome browser controls registers clicks and decides how to respond (e.g., it commands zmenu to close down). But since genome browser does not control area outside of its container, and since user can click anywhere on the screen, we need to delegate responsibility for listening of outside clicks to the zmenu itself, and when it registers an outside click it should notify genome browser of it, so that it can then send a command to close this zmenu.

## Views affected
Genome browser.

### DIFF
--- a/src/ensembl/src/content/app/browser/zmenu/zmenu-types.ts
+++ b/src/ensembl/src/content/app/browser/zmenu/zmenu-types.ts
@@ -25,6 +25,7 @@ export enum Markup {
 export enum ZmenuAction {
   CREATE = 'create_zmenu',
   DESTROY = 'destroy_zmenu',
+  ACTIVITY_OUTSIDE = 'zmenu_activity_outside',
   REPOSITION = 'update_zmenu_position',
   ENTER = 'zmenu-enter',
   LEAVE = 'zmenu-leave'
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/browser/zmenu/zmenu-types.ts
```diff
@@ -25,6 +25,7 @@ export enum Markup {
 export enum ZmenuAction {
   CREATE = 'create_zmenu',
   DESTROY = 'destroy_zmenu',
+  ACTIVITY_OUTSIDE = 'zmenu-activity-outside', // TODO: sometime later, unify underscores vs hyphens (together with Genome Browser)
```
msmanoj: Can this be `CLICK_OUTSIDE` to make it more specific?

imran-ebi: @msmanoj I think `CLICK_OUTSIDE` would be too specific. And in case we start supporting touch screens, it will need to be changed back to something more generic for outside activity.

msmanoj: Fair point. But even if we support touch screen, I think onClicks will still be called for touch events. My only concern is that `ACTIVITY_OUTSIDE` sounds like any activity that happens outside. For example, even moving the pointer could be an activity we perform outside zmenu but we don't close the zmenu for that.

azangru: The actual message string sent to genome browser contains the word "activity". I would prefer the constant to reflect that.

_(But we hope to have some kind of auditing of the messages passed between genome browser and browser chrome in the future)_

PR #157, opened 2019-09-05
Feature/focus button

azangru:
## Type
- Bug fix
- Refactoring

## Related JIRA Issue(s)
- ENSWBSITES-319
- ENSWBSITES-321

## Description
Change the internal wiring of the Browser reset button:
- Genome Browser (being the source of truth of focus object position) sends message to Browser Chrome informing it of whether or not the focus object is in default position
- Browser Chrome disables BrowserReset button if focus object is in default position and enables it otherwise. Click on BrowserReset button sends a message to GenomeBrowser asking it to put current focus object in default position.

## Views affected
GenomeBrowser


src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -57,6 +62,7 @@ type StateProps = {
   genomeSelectorActive: boolean;
   ensObject: EnsObject | null;
   selectedTrackPanelTab: TrackSet;
+  isFocusObjectInDefaultPosition: boolean;
```
msmanoj: Since we are importing a selector with the same name, would it be possible to change this to something like `focusObjectIsInDefaultPosition` ? It's not great but it will be less ambiguous.

azangru: I know what you mean, but I gave them the same name quite deliberately. One is a function, the other is a boolean; there is no context in which this will cause confusion. Typescript will warn us if we try to use any one of them improperly.

msmanoj: My thought is that we look at our code in places like Github and browser console and not just the IDEs. I agree that Typescript will warn us during development but should we depend on that do differentiate a function and a boolean?

Also, in the use case above, we could end up accidentally using `! isFocusObjectInDefaultPosition` instead of `!props. isFocusObjectInDefaultPosition` without getting any warning.

azangru: I updated the name of the selector, but in general this means that we feel uneasy about having variable names inside functions that are the same as variable names in the outer scope:

```
const x = 1;

function doSomething() {
  const x = 'foo'
  const y = 'bar'
  return x + y;
}
```

There is even a special eslint rule for that (https://eslint.org/docs/rules/no-shadow), but I've never felt any real need for it. Naming is hard as it is, to be prevented from reusing suitable names.

msmanoj: I agree to that but I think we should try and avoid if it can be for better readability.

imran-ebi: I might be a bit late into commenting on this. I agree with @msmanoj's suggestion. Maybe moving forward we could prefix selectors that return these boolean state properties with `get`?

azangru: @imran-ebi We are using `get` as a prefix in selectors, but it makes for cumbersome names. For example, we currently have `getIsTrackPanelOpened` as a method in trackPanelSelectors. This reads very unnatural, and I was trying to avoid that.

By arguing against having props named the same as functions outside the container, you are essentially arguing against having object fields named the same as a function:

```
const isOk = () => true;

const obj = {
  isOk: isOk()
}
```

Notice also that the argument that appeals to possible confusion falls apart in case of our mapDispatchToProps objects:

```
import {connect} from 'react-redux';
import doSomeAction from 'some-actions';

const Component = (props) => (
 <span onClick={props.doSomeAction}>Click me</span>
)

const mapDispatchToProps = {
  doSomeAction
}

export default connect(null, mapDispatchToProps)(Component);
```

@msmanoj 's point above is that we have no guarantee developer will write `onClick={props.doSomeAction}` instead of `onClick={doSomeAction}` (the latter won't work); but we happily write this kind of code and do not make many errors.

imran-ebi: Using `doSomeAction` for both the props and selector methods should be fine as you have described. However, I think we shouldn't be using the same name for the selectors that return a boolean if all it does is just return a property. It is more of a getter function. And a `isTrackPanelOpened` function returning a `isTrackPanelOpened` property doesn't sound right.

azangru: Returning a property is an implementation detail! We are asking a function to tell us, whether trackPanel is open; the fact that it then looks in the state and picks a particular property from it is irrelevant. It could have done some calculations along the way; it could have looked at two properties and given us some decision based on them. My point is, when we are interrogating a function, we are not interested in the way it gets the answer.

Compare some getter property on a class:

```

class Dog {
  ...
  isFriendly() {
   // do some thinking and return a boolean answer
  }
}
```

we would use this instance as `dog.isFriendly()`, even if the `.isFriendly()` method only returned the `isFriendly` property. We would not call the method `.getIsFriendly()` if it returned the `getFriendly` property. Ideally, we wouldn't care about how the `.isFriendly()` method comes to its decision whether the `dog` is friendly or not.

PR #162, opened 2019-09-11
Enhancements and new functionality for navigation bar

imran-ebi:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-209
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-210
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-211
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-215
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-175

## Description
This PR covers five JIRA tickets (mentioned above) of the change focus region epic (https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-275).

The genome selector, which was in the browser bar, is now replaced by a region field and region editor in the browser navigation bar. These new inputs will let users navigate to different regions and/or slice within the same feature.

Also, test cases were written for browser components that did not have any.

## Views affected
Genome browser

### Other effects


src/ensembl/webpack/webpack.config.dev.js
```diff
@@ -67,7 +67,9 @@ const devConfig = {
     // rules to proxy requests to the backend server in development
     proxy: {
       '/api': {
-        target: 'https://staging-2020.ensembl.org',
+        // IMPORTANT! NOTE: change this to the original url once the PR is approved
+        target: 'http://193.62.55.91:30667',
+        // target: 'https://staging-2020.ensembl.org',
```
imran-ebi: NOTE TO SELF: Change this to the original `target` URL once the PR has been approved

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -1,5 +1,95 @@
...
+    });
+
+    test('renders the genome browser and track panel when if there is a selected focus object', () => {
+      const wrapper = mountBrowserComponent({
+        browserQueryParams: { focus: faker.lorem.words() }
+      });
+      expect(wrapper.find(BrowserImage)).toHaveLength(1);
+      expect(wrapper.find(TrackPanel)).toHaveLength(1);
+    });
+  });
```
msmanoj: `when if` -> `only if`? 

Also, would it be better to test if the length is 0 when there is no focus?

imran-ebi: I didn't test that since we assume it is going to be the opposite when there is no selected focus object. However, I am curious if we should also test the opposite effects in our tests.

msmanoj: If the `BrowserImage` should get rendered only if the focus is present, there is a chance that it could always get rendered without failing any tests.

imran-ebi: That sounds right. I'll add these assertions to the same test case and change the wording of the description.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.scss
```diff
@@ -88,3 +88,65 @@
...
+    cursor: pointer;
+  }
+}
+
+.chrLocationView {
+  display: inline-block;
+}
+
+.chrCode {
+  appearance: none;
```
msmanoj: Do we need this rule?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.scss
```diff
@@ -88,3 +88,65 @@
...
+  border-radius: 0;
+  color: $white;
+  cursor: pointer;
+  display: inline-block;
+  font-weight: $bold;
+  margin: 0 5px;
+  padding: 1px 4px;
+  text-align: center;
+  text-align-last: center;
+  width: auto;
```
msmanoj: Same with the following rules:
```
border: none;
border-radius: 0;
text-align-last: center;
width: auto;
```

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.scss
imran-ebi: No we don't anymore. I'll remove it.

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -1,5 +1,95 @@
...
+    </MemoryRouter>
+  );
+
+  const mountBrowserComponent = (props?: Partial<BrowserProps>) =>
+    mount(<Browser {...defaultProps} {...props} />, { wrappingComponent });
+
+  describe('rendering', () => {
+    test('does not render when no activeGenomeId', () => {
+      const wrapper = mountBrowserComponent({ activeGenomeId: null });
+      expect(wrapper.find(AppBar)).toHaveLength(0);
```
azangru: If the test says that the component does not render at all, then why is the test checking specifically for the AppBar?

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -1,5 +1,95 @@
...
+import React from 'react';
+import { MemoryRouter } from 'react-router';
+import { Provider } from 'react-redux';
+import { mount } from 'enzyme';
+import faker from 'faker';
+
+import { Browser, BrowserProps, ExampleObjectLinks } from './Browser';
+import BrowserImage from './browser-image/BrowserImage';
+import TrackPanel from './track-panel/TrackPanel';
+import AppBar from 'src/shared/components/app-bar/AppBar';
```
azangru: Perhaps mock all these components during the import (so that we do not render the real ones).

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -1,5 +1,95 @@
...
+
+  describe('rendering', () => {
+    test('does not render when no activeGenomeId', () => {
+      const wrapper = mountBrowserComponent({ activeGenomeId: null });
+      expect(wrapper.find(AppBar)).toHaveLength(0);
+    });
+
+    test('renders links to example objects if there is no selected focus object', () => {
+      const wrapper = mountBrowserComponent();
+      expect(wrapper.find(ExampleObjectLinks)).toHaveLength(1);
```
azangru: how do we know ExampleObjectLinks won't be rendered if there is selected focus object?

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -1,5 +1,95 @@
...
+      expect(wrapper.find(TrackPanel)).toHaveLength(1);
+    });
+  });
+
+  describe('behaviour', () => {
+    test('fetches genome data when selected genome changes', () => {
+      const wrapper = mountBrowserComponent({ activeGenomeId: null });
+      wrapper.setProps({ activeGenomeId: faker.lorem.words() });
+      expect(wrapper.props().fetchGenomeData).toHaveBeenCalledTimes(1);
+    });
```
azangru: two things here:
1) test is not checking that fetchGenomeData has not been called before activeGenomeId was changed
2) the initial case in this test is with `activeGenomeId: null`. But will fetchGenomeData be called if we start with `activeGenomeId: some_string`?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -136,23 +102,30 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
     props.toggleBrowserNav();
   };
 
+  const [chrCode, chrStart, chrEnd] = props.actualChrLocation || [, , ,];
```
msmanoj: props.actualChrLocation || [] ?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -164,21 +137,20 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
...
-              navigator={navigator}
-              toggleNavigator={toggleNavigator}
-              icon={getBrowserNavIcon()}
-            />
+          {props.actualChrLocation && (
+            <dd className={browserRegionClassName}>
+              <label className="show-for-large">Chromosome</label>
+              <div className={styles.chrLocationView} onClick={toggleNavigator}>
+                <div className={styles.chrCode}>{chrCode}</div>
+                {displayChrRegion ? (
```
msmanoj: displayChrRegion  && () ?

src/ensembl/src/content/app/browser/browser-cog/BrowserCog.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const wrappingComponent = (props: any) => (
+    <Provider store={store}>{props.children}</Provider>
+  );
+
+  describe('rendering', () => {
+    test('renders browser track config', () => {
+      const wrapper = mount(<BrowserCog {...defaultProps} />, {
+        wrappingComponent
+      });
+      expect(wrapper.find(BrowserTrackConfig).length).toBeGreaterThan(0);
```
msmanoj: Is it going to be a specific number or it can be any number above 0?

src/ensembl/src/content/app/browser/Browser.test.tsx
imran-ebi: True, checking `AppBar` is not right. I will instead check if the wrapper HTML is `null`.

src/ensembl/src/content/app/browser/browser-cog/BrowserCog.tsx
```diff
@@ -9,10 +9,10 @@ import analyticsTracking from 'src/services/analytics-service';
...
 import cogOnIcon from 'static/img/shared/cog-on.svg';
 import cogOffIcon from 'static/img/shared/cog.svg';
-import BrowserTrackConfig from './browser-track-config/BrowserTrackConfig';
+import BrowserTrackConfig from '../browser-track-config/BrowserTrackConfig';
 import { useTransition, animated } from 'react-spring';
 
-type BrowserCogProps = {
+export type BrowserCogProps = {
   cogActivated: boolean;
   index: string;
```
msmanoj: Would it be possible to rename this to `trackName` or `trackId` ?

src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+describe('<BrowserCogList />', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  const defaultProps = {
+    browserActivated: true,
+    browserCogList: 0,
+    browserCogTrackList: { 'track:gc': 491 },
```
msmanoj: Does this value has to be 491?

src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    trackConfigNames: {},
+    trackConfigLabel: {},
+    selectedCog: 'track:gene-other-fwd',
```
msmanoj: Can this value be autogenerated using faker?

src/ensembl/src/content/app/browser/browser-cog/BrowserCog.test.tsx
imran-ebi: It can be any number above zero, and dependent on which tracks the user has enabled.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.scss
```diff
@@ -88,3 +88,65 @@
...
+  }
+
+  label {
+    margin-right: 5px;
+  }
+
+  input {
+    padding: 1px 6px;
+    width: 200px;
+  }
```
azangru: I don't see either `input` or `form` element in the .browserInfoRegion element in the corresponding tsx

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -136,23 +102,30 @@ export const BrowserBar: FunctionComponent<BrowserBarProps> = (
     props.toggleBrowserNav();
   };
 
+  const [chrCode, chrStart, chrEnd] = props.actualChrLocation || [, , ,];
+  const displayChrRegion = !(chrStart === 0 && chrEnd === 0);
```
azangru: `const displayChrRegion  = chrStart !== 0 && chrEnd !== 0` ? ;-)

src/ensembl/src/content/app/browser/browser-cog/BrowserCog.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const defaultProps: BrowserCogProps = {
+    cogActivated: true,
+    index: faker.lorem.words(),
+    updateSelectedCog: jest.fn()
+  };
+
+  const store = configureStore();
+  const wrappingComponent = (props: any) => (
+    <Provider store={store}>{props.children}</Provider>
```
azangru: What's the role of redux provider here?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    test('has an overlay on top when region field is active', () => {
+      const wrapper = mountBrowserImageComponent({ regionFieldActive: true });
+      expect(wrapper.find('.browserOverlay').length).toBe(1);
+    });
+
+    test('has an overlay on top when region editor is active', () => {
+      const wrapper = mountBrowserImageComponent({ regionEditorActive: true });
+      expect(wrapper.find('.browserOverlay').length).toBe(1);
+    });
```
azangru: duplicate test?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
```diff
@@ -2,20 +2,28 @@ import React from 'react';
...
+          chrLocation={chrLocation}
+          genomeKaryotypes={[]}
           isTrackPanelOpened={true}
+          regionEditorActive={true}
+          regionFieldActive={false}
+          toggleRegionEditorActive={toggleRegionEditorActive}
+          toggleRegionFieldActive={toggleRegionFieldActive}
         />
       ).hasClass(styles.browserNavBarExpanded)
     ).toBe(false);
```
msmanoj: Would it be better to move the shallow(...) part to a separate wrapper variable?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setActualChrLocation: jest.fn(),
+    updateMessageCounter: jest.fn(),
+    updateDefaultPositionFlag: jest.fn(),
+    changeHighlightedTrackId: jest.fn()
+  };
+
+  const store = configureStore();
+
+  const wrappingComponent = (props: any) => (
+    <Provider store={store}>{props.children}</Provider>
```
azangru: If Provider was used so that connected nested components could properly render, maybe mock these components while importing them?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
imran-ebi: They are for two different scenarios. The first is when the region field is selected by the user and the second is when the region editor is selected.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -1,50 +1,99 @@
...
+  regionFieldActive: boolean;
+  toggleRegionEditorActive: (regionEditorActive: boolean) => void;
+  toggleRegionFieldActive: (regionFieldActive: boolean) => void;
 };
 
-type BrowserNavBarProps = StateProps;
+export const BrowserNavBar = (props: BrowserNavBarProps) => {
+  // the region editor and field style should be reset so that it won't be opaque when nav bar is opened again
+  useEffect(
+    () => () => {
```
msmanoj: Is this correct?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -1,50 +1,99 @@
...
+      props.toggleRegionEditorActive(false);
+      props.toggleRegionFieldActive(false);
+    },
+    []
+  );
+
+  const shouldBeEnabled = (index: number) => {
+    const { browserNavStates, regionEditorActive, regionFieldActive } = props;
+    const maxState = browserNavStates[index];
```
msmanoj: What does `maxState` mean here?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
azangru: Oh, I see.
Then the descriptions of the tests should be different.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -1,50 +1,99 @@
...
             key={item.name}
             browserNavItem={item}
-            maxState={props.browserNavStates[index]}
+            enabled={shouldBeEnabled(index)}
           />
         ))}
-      </dl>
-    </div>
+      </dd>
+      <dd>{props.chrLocation ? <BrowserRegionField /> : null}</dd>
```
msmanoj: just `props.chrLocation && <BrowserRegionField />`?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.scss
```diff
@@ -12,4 +14,49 @@
   &.browserNavBarExpanded {
     width: 100vw;
   }
+
+  dd {
+    margin-left: 10px;
+  }
```
azangru: Is it a good time to gradually be removing `dd`s and `dl`s, at least from new / edited files? They are probably fine to be used for definitions (dl — definition list, dt — definition term, dd — definition description); but why do we keep using them for some other purposes?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.scss
```diff
@@ -12,4 +14,49 @@
...
+  input {
+    padding: 3px 9px;
+  }
+}
+
+.errorText {
+  color: $red;
+}
+
+.inputText {
```
azangru: this class is used only in the BrowserRegionField component. Why doesn't it sit in the scss file there?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -1,50 +1,99 @@
...
+  toggleRegionEditorActive,
+  toggleRegionFieldActive
+} from '../browserActions';
 import { getIsTrackPanelOpened } from '../track-panel/trackPanelSelectors';
-import { BrowserNavStates } from '../browserState';
-
-import BrowserNavIcon from './BrowserNavIcon';
+import { BrowserNavStates, ChrLocation } from '../browserState';
+import { getGenomeKaryotypes } from 'src/genome/genomeSelectors';
+import { GenomeKaryotype } from 'src/genome/genomeTypes';
```
azangru: The term "karyotype" describes the complete set of chromosomes of an organism. Therefore, a genome has one karyotype (not "karyotypes", as in the name of the `getGenomeKaryotypes ` selector); and the type that is called `GenomeKaryotype` in the line above is in fact a `GenomeKaryotypeItem`.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+          .last()
+          .props().value
+      ).toBe(locationEndInput);
+    });
+
+    test('validates region input on submit', () => {
+      wrapper
+        .find(Input)
+        .first()
+        .simulate('change', { target: { value: getCommaSeparatedNumber(0) } });
```
msmanoj: Can this be just '0'?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
imran-ebi: I agree. I'll change the descriptions to be more meaningful.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+import { ChrLocation } from '../browserState';
+import genomeKaryotypes from 'tests/data/browser/karyotypes';
+import { getCommaSeparatedNumber } from 'src/shared/helpers/numberFormatter';
+import { LoadingState } from 'src/shared/types/loading-state';
+import { createValidationInfo } from 'tests/fixtures/browser';
+
+const defaultProps = {
+  activeGenomeId: faker.lorem.words(),
+  chrLocation: ['13', 2315086, 32400266] as ChrLocation,
```
azangru: use faker for string - number - number ?

src/ensembl/src/content/app/browser/browser-image/BrowserImage.test.tsx
imran-ebi: In hindsight, I'll include both the scenarios in the same test case with separate assertions.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+    test('validates region input on submit', () => {
+      wrapper
+        .find(Input)
+        .first()
+        .simulate('change', { target: { value: getCommaSeparatedNumber(0) } });
+
+      wrapper.find('form').simulate('submit');
+
+      expect(wrapper.props().validateRegion).toHaveBeenCalledTimes(1);
```
azangru: toHaveBeenCalledWith is even more interesting.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
imran-ebi: It is a value that is returned from the genome browser when a maximum is hit on either of the navigation directions i.e. top, bottom, left and right. I just couldn't think of a better variable name for this when I wrote this.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+        .last()
+        .simulate('change', { target: { value: locationEndInput } });
+
+      expect(
+        wrapper
+          .find(Input)
+          .last()
+          .props().value
+      ).toBe(locationEndInput);
+    });
```
azangru: I am not sure this test is interesting.
Testing that when a start input and an end input are filled and a chromosome selected, and the apply button is pressed, then something happens — now that would be interesting.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+            error_code: null,
+            error_message: 'Start should be between 1 and 248956422',
+            is_valid: false,
+            value: 0
+          }
+        })
+      });
+
+      wrapper.update();
+      expect(wrapper.find('.startInputGroup').find(Tooltip)).toHaveLength(1);
```
azangru: and the tooltip content is...

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.tsx
```diff
@@ -1,50 +1,99 @@
...
+  // the region editor and field style should be reset so that it won't be opaque when nav bar is opened again
+  useEffect(
+    () => () => {
+      props.toggleRegionEditorActive(false);
+      props.toggleRegionFieldActive(false);
+    },
+    []
+  );
+
+  const shouldBeEnabled = (index: number) => {
```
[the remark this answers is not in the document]
imran-ebi: That sounds much better. I'll change it to that.

azangru:
General problem: if we put cursor anywhere inside the number representing the location and delete one of the characters to the right or to the left, the cursor will be thrown to the end of the number:

image
image

As a user, I would find it annoying.

One special case of this is that if we try to delete a comma, the number won't change, and the cursor will be thrown to the end of the number:
image
image

I would suggest not re-calculating the comma-separated representation of the number while the user changes the input, and do the recalculation only after they submit the input. I think Andrea agreed to that when we discussed the behaviour of the input.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -0,0 +1,340 @@
...
+  >(null);
+
+  const getKaryotypeOptions = () =>
+    props.genomeKaryotypes.map(({ name }) => ({
+      value: name,
+      label: name,
+      isSelected: regionInput === name
+    }));
+
+  const updateRegionInput = (value: string) => {
```
azangru: the word "region" is used in two senses within this file, which is confusing. In some places it means the whole combination of chromosome(or any other "stick")-start-end; but here it means the name of the chromosome (or any other "stick"). I suggest we change the name here — either to Chromosome, or (if we don't like to call chromosome what may turn out to be a plasmid or a scaffold, etc.) to Stick, or something. Or maybe change the Region part in RegionEditor to Location.

imran-ebi:
@azangru Andrea wanted the region field to behave so, but not the region editor's start and end. Therefore, the masking for these fields. But we can check with her again about this.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imran-ebi: Will make this change

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -0,0 +1,340 @@
...
+    setRegionInput(value);
+    updateLocationStartInput('1');
+
+    const karyotypeLength = props.genomeKaryotypes.filter(
+      ({ name }) => name === value
+    )[0].length;
```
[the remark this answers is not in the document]
imran-ebi: This is for a select input field, so all the fields should match the genome karyotype names returned by the endpoint.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: I'll change it to `chrInput` for now.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+    props.replace(urlFor.browser(params));
+  };
+
+  const updateBrowser = () => {
+    if (!isFormSubmitted) {
+      return;
+    }
+
+    const [region, ,] = props.chrLocation;
```
azangru: no need to write out the 2nd and 3rd items of the array if they are ignored.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+    test('has an overlay on top when region field is active', () => {
+      wrapper.setProps({ regionFieldActive: true });
+      expect(wrapper.find('.browserOverlay').length).toBe(1);
+    });
+  });
+
+  describe('behaviour', () => {
+    const locationStartInput = getCommaSeparatedNumber(10);
+    const locationEndInput = getCommaSeparatedNumber(10500);
```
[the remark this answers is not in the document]
imran-ebi: Yes, will do so.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+          setRegionFieldInput('');
+          setErrorMessages(null);
+          setIsFormSubmitted(false);
+          props.toggleRegionFieldActive(false);
+
+          updateBrowser();
+        }
+      }
+    }
+  }, [props.regionValidationInfo]);
```
azangru: This is very confusing.

Wouldn't it be easier to change the browser location or the focus object in the same action that sends request to validate the region — after the response comes back and is successful? Then there wouldn't be any need for the `isFormSubmitted` flag (which seems to be complicating things quite a bit).

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.scss
imran-ebi: Should we be instead using `ul` then?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { mount } from 'enzyme';
+import faker from 'faker';
+
+import { BrowserRegionField } from './BrowserRegionField';
+import Input from 'src/shared/components/input/Input';
+import Tooltip from 'src/shared/components/tooltip/Tooltip';
+
+import { ChrLocation } from '../browserState';
+import { LoadingState } from 'src/shared/types/loading-state';
+import genomeKaryotypes from 'tests/data/browser/karyotypes';
```
azangru: Perhaps use a function to generate random data that has the shape of a karyotype item?
The shape of these items will likely change; and what is a couple of lines change in the constructor function will otherwise be a lot of copy-pasting in the data file.
It's not like we need the exact values of the karyotype items here, is it?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { createValidationInfo } from 'tests/fixtures/browser';
+
+const defaultProps = {
+  activeGenomeId: faker.lorem.words(),
+  chrLocation: ['13', 2315086, 32400266] as ChrLocation,
```
azangru: faker

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+import styles from './BrowserRegionField.scss';
+import browserStyles from '../Browser.scss';
+import browserNavBarStyles from '../browser-nav/BrowserNavBar.scss';
+
+type BrowserRegionFieldProps = {
+  activeGenomeId: string | null;
+  chrLocation: ChrLocation;
+  isDrawerOpened: boolean;
+  regionEditorActive: boolean;
+  regionFieldActive: boolean;
```
azangru: Let's go over the props here.
Why does BrowserRegionField need to know stuff about other components (that the Drawer is opened or that the regionEditor is active)? Whether the Drawer is open or not, should be of no concern for the BrowserRegionField. What it should care about is whether it's active or not, and whether it is disabled (covered with overlay )or not. Which, I think, suggests that the props should be `isActive` and `isDisabled`.

Same goes for BrowserRegionEditor — its props also are informing it of the state of other components, which should be none of its business.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const activateForm = () => {
+    if (!props.regionFieldActive && !props.isDrawerOpened) {
+      props.toggleRegionFieldActive(true);
+    }
+  };
+
+  const changeRegionFieldInput = (value: string) => setRegionFieldInput(value);
+
+  const getRegionInputWithRegion = (input: string) => {
+    const [region, ,] = props.chrLocation;
```
azangru: `const [region]` should be enough

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
azangru: Same as for region editor — update browser from within react action after successful response?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+        <Input
+          type="text"
+          placeholder="0:1-1,000,000"
+          value={regionFieldInput}
+          onChange={changeRegionFieldInput}
+          onFocus={activateForm}
+          className={inputClassNames}
+        />
+        <span className={buttonsClassNames}>
+          <button type="submit" className="submitButton">
```
azangru: There isn't any style associated with this button. Same in RegionEditor.

If the only purpose of this class is to find this element in the test, then perhaps `button[type="submit"]` will be enough to identify it?

src/ensembl/src/content/app/browser/browser-reset/BrowserReset.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    test('renders nothing when focus feature does not exist', () => {
+      const wrapper = mount(
+        <BrowserReset {...defaultProps} focusObject={null} />
+      );
+      expect(wrapper.html()).toBe(null);
+    });
+  });
+
+  describe('behaviour', () => {
+    test('changes focus object when clicked on reset button', () => {
```
azangru: just "when clicked" :-)

src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  describe('rendering', () => {
+    test('renders the all tracks checkbox', () => {
+      expect(wrapper.find(Checkbox)).toHaveLength(1);
+    });
+
+    test('renders all the buttons', () => {
+      expect(wrapper.find('button')).toHaveLength(6);
+    });
+  });
```
azangru: these two tests aren't really useful. They are (partially?) covered in the behaviour section. Let's remove them.

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  getRegionEditorErrorMessages
+} from './browserHelper';
+import { createValidationInfo } from 'tests/fixtures/browser';
+import { RegionValidationResponse } from './browserState';
+import { RegionErrors } from './browserConfig';
+
+describe('browserHelper', () => {
+  describe('getChrLocationFromStr', () => {
+    test('outputs ChrLocation when input received as string', () => {
+      expect(getChrLocationFromStr('10:1-10000')).toStrictEqual([
```
azangru: use faker

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        '10',
+        1,
+        10000
+      ]);
+    });
+  });
+
+  describe('getChrLocationStr', () => {
+    test('outputs string when input received as ChrLocation', () => {
+      expect(getChrLocationStr(['10', 1, 10000])).toBe('10:1-10000');
```
azangru: use faker

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        RegionErrors.PARSE_ERROR
+      );
+    });
+
+    test('returns error if region is invalid', () => {
+      const newMockValidationInfo = Object.assign(
+        {},
+        mockValidationInfo,
+        invalidRegionInfo
+      );
```
azangru: the Object.assign syntax in this file can be replaced with a simpler notation of:
```
{
 ...something,
 ...somethingElse
}
```

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      expect(getRegionFieldErrorMessages(newMockValidationInfo)).toBe(
+        'Could not find region aaa'
```
azangru: invalidRegionInfo.errorMessage

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    test('returns error if start is invalid', () => {
+      const newMockValidationInfo = Object.assign(
+        {},
+        mockValidationInfo,
+        invalidStartInfo
+      );
+
+      expect(getRegionFieldErrorMessages(newMockValidationInfo)).toBe(
+        'Start should be between 1 and 248956422'
```
azangru: invalidStartInfo.errorMessage

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    test('returns error if end is invalid', () => {
+      const newMockValidationInfo = Object.assign(
+        {},
+        mockValidationInfo,
+        invalidEndInfo
+      );
+
+      expect(getRegionFieldErrorMessages(newMockValidationInfo)).toBe(
+        'End should be between 1 and 248956422'
```
azangru: as above

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    test('returns location start error message when start is invalid', () => {
+      const newMockValidationInfo = Object.assign(
+        {},
+        mockValidationInfo,
+        invalidStartInfo
+      );
+
+      expect(getRegionEditorErrorMessages(newMockValidationInfo)).toStrictEqual(
+        {
+          locationStartError: 'Start should be between 1 and 248956422',
```
azangru: as above

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      const newMockValidationInfo = Object.assign(
+        {},
+        mockValidationInfo,
+        invalidEndInfo
+      );
+
+      expect(getRegionEditorErrorMessages(newMockValidationInfo)).toStrictEqual(
+        {
+          locationStartError: null,
+          locationEndError: 'End should be between 1 and 248956422'
```
azangru: as above

src/ensembl/src/content/app/browser/browserHelper.ts
```diff
@@ -14,3 +20,47 @@ export function getChrLocationStr(
...
+      } else if (validationInfo.end && !validationInfo.end.is_valid) {
+        return validationInfo.end.error_message;
+      }
+    }
+
+    return null;
+  } catch {
+    return RegionErrors.REQUEST_ERROR;
+  }
+};
```
azangru: I am confused. Why the try-catch? What can possibly go wrong inside the try block that will produce an error that has to be caught?

src/ensembl/src/content/app/browser/browserSelectors.ts
```diff
@@ -50,19 +50,19 @@ export const getBrowserNavStates = (state: RootState) =>
...
   const locations = state.browser.browserLocation.actualChrLocations;
   const activeGenomeId = getBrowserActiveGenomeId(state);
   return activeGenomeId ? locations[activeGenomeId] : null;
 };
 
-export const getDefaultChrLocation = (state: RootState) => {
+export const getDefaultChrLocation = (state: RootState): ChrLocation | null => {
   const activeEnsObjectId = getBrowserActiveEnsObjectId(state);
   const activeEnsObject =
     activeEnsObjectId && getEnsObjectById(state, activeEnsObjectId);
```
azangru: Was typescript not smart enough to infer the return values for the three functions above itself?

src/ensembl/src/content/app/browser/browserSelectors.ts
```diff
@@ -71,15 +71,23 @@ export const getDefaultChrLocation = (state: RootState) => {
...
-  return state.browser.browserEntity.messageCounter;
-};
+export const getRegionEditorActive = (state: RootState) =>
+  state.browser.browserLocation.regionEditorActive;
+
+export const getRegionFieldActive = (state: RootState) =>
+  state.browser.browserLocation.regionFieldActive;
+
+export const getRegionValidationInfo = (state: RootState) =>
+  state.browser.browserLocation.regionValidationInfo;
```
azangru: I know we had this conversation, but here's an example, where I would prefer the selector to be named `isRegionEditorActive`, and the redux state field to be named the same way as well. `getRegionEditorActive` is very stilted (what am I getting?) :-(

Don't insist on this change if you are opposed to it; just indicating my preference.

src/ensembl/src/content/app/browser/browserState.ts
```diff
@@ -25,6 +26,38 @@ export type CogList = {
   [key: string]: number;
 };
 
+export type RegionValidationError = {
+  error_code: string | null;
+  error_message: string | null;
+  is_valid: boolean;
+};
```
azangru: why does a type that is called `error` have an is_valid field that is boolean (which means that it can be true)?

isn't it rather a ValidationResult or something?

src/ensembl/src/content/app/browser/browserState.ts
```diff
@@ -25,6 +26,38 @@ export type CogList = {
...
+  error: string;
+  genome_id: string;
+  region: string;
+  region_code: string;
+}>;
+
+export type RegionValidationResponse = Partial<{
+  end: RegionValidationValueError;
+  genome_id: RegionValidationValueError;
+  is_parseable: boolean;
```
azangru: shouldn't forget that this field will go away

src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const wrappingComponent = (props: any) => (
+    <Provider store={store}>{props.children}</Provider>
+  );
+
+  const mountBrowserImageComponent = (props?: Partial<TrackPanelProps>) =>
+    mount(<TrackPanel {...defaultProps} {...props} />, { wrappingComponent });
+
+  describe('rendering', () => {
+    test('track panel when active genome is present', () => {
```
azangru: does what?

src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      const wrapper = mountBrowserImageComponent({
+        activeGenomeId: faker.lorem.words()
+      });
+      expect(wrapper.html()).not.toBe(null);
+    });
+
+    test('track panel bar and track panel list when browser is activated and active feature is selected', () => {
```
azangru: again, what is being tested according to the description?

src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    isTrackPanelOpened: true,
+    selectedTrackPanelTab: TrackSet.GENOMIC,
+    genomeTrackCategories: createGenomeCategories(),
+    trackStates: createTrackStates()
+  };
+
+  const store = configureStore();
+
+  const wrappingComponent = (props: any) => (
+    <Provider store={store}>{props.children}</Provider>
```
azangru: Or mock the components that TrackPanel uses so that we don't need to wrap it in redux provider?

src/ensembl/src/content/app/browser/track-panel/TrackPanel.tsx
```diff
@@ -76,7 +76,7 @@ const TrackPanel = (props: TrackPanelProps) => {
           <TrackPanelBar />
           <TrackPanelList />
           {props.isTrackPanelModalOpened ? <TrackPanelModal /> : null}
-          {isDrawerOpened && <Drawer />}
+          {isDrawerOpened ? <Drawer /> : null}
```
azangru: Interesting :-) Your style and @msmanoj's style are diametrically opposite 😄

src/ensembl/src/content/app/browser/zmenu/Zmenu.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  describe('rendering', () => {
+    test('renders zmenu content', () => {
+      expect(wrapper.find(ZmenuContent)).toHaveLength(1);
+    });
+
+    test('renders zmenu tip', () => {
+      expect(wrapper.find(Tip)).toHaveLength(1);
+    });
```
azangru: Rendering of the tip is definitely not an interesting test 😄

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    test('renders zmenu content line', () => {
+      const wrapper = mountZmenuContent();
+      expect(wrapper.find('.zmenuContentLine')).toHaveLength(1);
+    });
+
+    test('renders zmenu content block', () => {
+      const wrapper = mountZmenuContent();
+      expect(wrapper.find('.zmenuContentBlock')).toHaveLength(2);
+    });
```
azangru: `toHaveLength` aren't very informative tests. Do we know what's actually rendered in those elements?

src/ensembl/src/content/app/browser/browserHelper.test.ts
imran-ebi: I used values so that there is no need define a lot of variables for region, start, end, chrLocation and chrLocationStr. Do we really need these to be random though?

azangru: random variables are a bit more convincing (they show that things aren't some magic numbers or magic strings), and they might, surprisingly, be easier to read.

`expect(getChrLocationFromString(`${chr}:${start}-${end}`)).toStrictEqual([chr, start, end])` is arguably more meaningful than `toStrictEqual([10, 1, 1000])`.

src/ensembl/src/content/app/browser/browserHelper.ts
imran-ebi: This was there when we didn't have a standard format for the  response and that caused some unexpected errors. Hence, the try-catch block. We probably don't need this anymore.

src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
imran-ebi: Thanks! That works like a charm.

src/ensembl/src/content/app/browser/zmenu/Zmenu.test.tsx
imran-ebi: It's one of those tests that I really wasn't sure whether to add or not? 😄 Visually it should be present and something that can be missed out, however I get what you are implying; it is more about UI testing than testing the code.

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { Provider } from 'react-redux';
```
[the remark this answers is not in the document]
imran-ebi: I'll mock this component so how we import it won't be an issue :smile:

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
imran-ebi: They aren't worth testing either so I'll remove these test cases

src/ensembl/src/content/app/browser/browserHelper.test.ts
imran-ebi: I'll add a function that creates both the tupple and string values into the browser fixtures for this

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: If I am not wrong, `jest` can't parse attributes properly. I tried all those approaches before settling with classes as they all ended up failing my test cases.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: The BrowserRegionField has to know whether the BrowserRegionEditor is opened as it should have an overlay on top of it to make it not clickable, when the latter is active. So if we are to store the status instead for BrowserRegionField, we will need to toggle the status of it in the Redux action that sets the active status for BrowserRegionEditor. Will this be a better approach? Same applies to the drawer status.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,226 @@
...
+import browserNavBarStyles from '../browser-nav/BrowserNavBar.scss';
+
+type BrowserRegionFieldProps = {
+  activeGenomeId: string | null;
+  chrLocation: ChrLocation;
+  isDrawerOpened: boolean;
+  regionEditorActive: boolean;
+  regionFieldActive: boolean;
+  regionValidationInfo: RegionValidationResponse | null;
+  regionValidationLoadingStatus: LoadingState;
```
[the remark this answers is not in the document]
imran-ebi: Isn't `isLoading` too ambiguous inside a component? Also I am making using of the `LoadingState` enum, hence why it isn't a `boolean`.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: Tried this again with quoting the attribute values with quotes e.g. `button[type="submit"]. That did work!

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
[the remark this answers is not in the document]
azangru: This test is **for** ZmenuComponent, so you can't mock it here 😄 
However, you can export an unconnected version of it in a named export.

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
azangru: Wow-wow, these tests are actually useful; they are showing that data passed to ZmenuContent ends up in correct places.
Only instead of checking the number of correct places, the test should check that these places contain correct data.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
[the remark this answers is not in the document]
azangru: But isn't the component in only one of two states — loading or not loading? If it receives the `isLoading` prop, it knows that its data is loading, and should behave appropriately. How is this ambiguous?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.scss
azangru: Is it a list though?

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
imran-ebi: Yes, however removed these test cases since it is useless testing whether they exist or not. I will write test cases to test the individual child components instead.

src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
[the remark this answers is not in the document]
imran-ebi: As mentioned below, I have removed (yet to be committed) the test cases that require `Provider` so there is no need for this anymore.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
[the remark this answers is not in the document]
imran-ebi: This isn't for the component data, rather for the validation data that it gets on submission. So `isLoading` will not be accurate.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.scss
imran-ebi: Yes, it is a list of icons.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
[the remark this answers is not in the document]
imran-ebi: I have changed `regionValidationInfo` to `validationInfo` and `regionValidationLoadingStatus` to `isValidationInfoLoading`. Is this better?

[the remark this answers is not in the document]
azangru: 👌

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: I have now changed the code after removing `isFormSubmitted`

imran-ebi: I have changed the code. Is it better now?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -139,16 +137,16 @@ export const BrowserBar = (props: BrowserBarProps) => {
         <dl className={styles.browserInfoRight}>
           {props.actualChrLocation && (
             <dd className={browserRegionClassName}>
-              <label className="show-for-large">Chromosome</label>
+              <div className={`${styles.chrLabel} show-for-large`}>
```
azangru: Does the `show-for-large` class serve any purpose? It’s not in scss.

src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
```diff
@@ -14,10 +15,10 @@ describe('<BrowserCogList />', () => {
   const defaultProps = {
     browserActivated: true,
     browserCogList: 0,
-    browserCogTrackList: { 'track:gc': 491 },
+    browserCogTrackList: { 'track:gc': Math.random() },
```
azangru: faker.random.number ? It will be an integer then, rather than a float between 0 and 1/

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -2,37 +2,42 @@ import React from 'react';
...
+  createValidationInfo,
+  createChrLocationValues
+} from 'tests/fixtures/browser';
 
 describe('<BrowserRegionEditor', () => {
   afterEach(() => {
     jest.resetAllMocks();
   });
 
+  const initialChrLocation = createChrLocationValues().tuppleValue;
```
azangru: 1) tuple, not tupple
2) perhaps `asTuple` and `asString` rather than `tupleValue` and `stringValue`?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -163,30 +155,36 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
+    resetForm();
+  };
+
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
-    const region = `${regionInput}:${locationStartInput}-${locationEndInput}`;
-
-    setIsFormSubmitted(true);
+    const region = `${chrInput}:${locationStartInput}-${locationEndInput}`;
```
azangru: Don’t we use browserHelper for this (`getChrLocationStr`)?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -214,22 +210,18 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
-          updateErrorMessages(null, null);
-          props.toggleRegionEditorActive(false);
-
+          resetForm();
           updateBrowser();
         }
       }
     }
-  }, [props.regionValidationInfo]);
+  }, [props.validationInfo]);
```
azangru: Doesn’t it essentially mean "every time the component receives the new props"? ValidationInfo is an object, and props comparison for objects is by reference (whether it’s the same object rather than whether it has the same values), and I suspect every time the component updates due to props change, `props.validationInfo` will be different. Does this effect behave differently with `props.validationInfo` than without any dependencies array? If yes, then something is wrong with the dependency.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -316,15 +309,18 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
-  regionEditorActive: getRegionEditorActive(state),
-  regionFieldActive: getRegionFieldActive(state),
-  regionValidationInfo: getRegionValidationInfo(state),
-  regionValidationLoadingStatus: getRegionValidationLoadingStatus(state)
-});
+const mapStateToProps = (state: RootState) => {
+  return {
+    activeGenomeId: getBrowserActiveGenomeId(state),
+    chrLocation: getChrLocation(state) as ChrLocation,
+    genomeKaryotypes: getGenomeKaryotypes(state) as GenomeKaryotype[],
```
azangru: The selector says that it can return null. Why do we convince it here that it can’t?
_(Same for `getChrLocation`)_

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
azangru: Still don’t like it that the logic for updating the location or focus of the browser sits inside the component and is not a part of the action.

Isn’t it much more natural to think of the thunk associated with this action in terms of "go check whether the input is valid, and if it is, then change the browser location/focus object"?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -0,0 +1,336 @@
...
+        <div className={styles.inputGroup}>
+          <label className="show-for-large">Chr</label>
+          <Select
+            onSelect={updateRegionInput}
+            options={getKaryotypeOptions()}
+          ></Select>
+        </div>
+        <div
+          className={styles.inputGroup}
+          role="startInputGroup"
```
azangru: Now, here we actually have a class (styles.inputGroup) that will let us target this div in the test. Why is it targeted by its role then?

(Same below)

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -202,15 +187,17 @@ export const BrowserRegionField = (props: BrowserRegionFieldProps) => {
...
-  regionFieldActive: getRegionFieldActive(state),
-  regionValidationInfo: getRegionValidationInfo(state),
-  regionValidationLoadingStatus: getRegionValidationLoadingStatus(state),
-  chrLocation: getChrLocation(state) as ChrLocation,
-  isDrawerOpened: getIsDrawerOpened(state)
-});
+const mapStateToProps = (state: RootState) => {
+  return {
+    activeGenomeId: getBrowserActiveGenomeId(state),
+    chrLocation: getChrLocation(state) as ChrLocation,
```
azangru: Same question as for BrowserRegionEditor

src/ensembl/tests/fixtures/browser.ts
```diff
@@ -1,18 +1,18 @@
...
-  'track:gene-pc-rev': 302,
-  'track:variant': 428
+  'track:contig': Math.random(),
+  'track:gc': Math.random(),
+  'track:gene-feat': Math.random(),
+  'track:gene-other-fwd': Math.random(),
+  'track:gene-other-rev': Math.random(),
+  'track:gene-pc-fwd': Math.random(),
+  'track:gene-pc-rev': Math.random(),
+  'track:variant': Math.random()
```
azangru: Math.random is a float between 0 and 1. Perhaps prefer faker.random.number?

src/ensembl/tests/fixtures/browser.ts
```diff
@@ -50,36 +50,51 @@ export const createZmenuContent = () => [
...
+      is_valid: true,
+      region_code: 'chromosome',
+      region_name: '13'
+    },
+    region_id: faker.lorem.words(),
+    start: {
+      error_code: null,
+      error_message: null,
+      is_valid: true,
+      value: randomValue + Math.random()
```
azangru: That's a less than one plus less than one.

src/ensembl/tests/fixtures/browser.ts
```diff
@@ -50,36 +50,51 @@ export const createZmenuContent = () => [
...
+};
+
+export const createChrLocationValues = () => {
+  const region = faker.lorem.words();
+  const start = faker.random.number();
+  const end = faker.random.number();
+
+  return {
+    tuppleValue: [region, start, end] as ChrLocation,
+    stringValue: `${region}:${start}-${end}`
```
azangru: Browser helper?

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
imran-ebi: It used to when we had Foundation. I'll replace it with the new class.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imran-ebi: I was tempted to use `asString`, however that would be slightly similar to `as string` in TypeScript so wanted to stay clear of it. And as for the function name, won't `createChrLocation` just mean creating one value that is a tuple rather than having two separate versions of it?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: Using it would complicate the code a bit as it expects a tuple: `[string, number, number]`. However, all three of these values are strings, so I didn't find the need to use a function that returns a string.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: The intention is to run this whenever `props.validationInfo` gets updated as that would mean the response has been returned. The validation happens only whenever a user clicks on the `submit` button, hence not many requests to validate. So I thought premature optimisation for this is not necessary.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: It's because it is good to have some kind of identifier for the actual set of elements that we are testing?

src/ensembl/tests/fixtures/browser.ts
imran-ebi: Changing this to `faker.random.number` as you suggested above should fix this

src/ensembl/tests/fixtures/browser.ts
imran-ebi: I didn't use it as it is not clear which version to use. Should it be `getChrLocationFromStr` that returns a tuple or `getChrLocationStr` that returns a `string`?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: I admit it wasn't a good idea to use type assertions here rather than inside the component. Both `BrowserRegionEditor` and `BrowserRegionField` are rendered only if `chrLocation` is not `null`, so I thought it was safe to assert `props.chrLocation`. This would not require to check if `props.chrLocation` is `null` or not in these components.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
azangru: This is not a matter of premature optimisation.
This is (if I understand correctly how it will work; I may be wrong) a misleading piece of code.
It suggests, like you say, "run this whenever props.validationInfo gets updated". But what it likely is actually doing (again, unless React is somehow smarter than I take it to be) is "run this whenever props get updated". Because every time the props are updated, the validationInfo object will be different.
If that's what it's doing (and if it's ok), then removing the list of dependencies will help avoid the misunderstanding.

Alternatively, there are [two custom hooks](https://github.com/Ensembl/ensembl-client/commit/bf65e4f0cc3fdcf796bbb3827be236c602a351de) that I wrote (well, copied mostly) a while back, but never had a chance to actually use. One of them, `useDeepCompareEffect`, does deep comparison of props. If you want to compare the validationInfo prop, we should use that.

I am still not sure this useEffect is doing the proper thing though. It:
- updates the state to show error messages (why do we keep them in the component's local state? why don't we show them directly as they come from props?
- updates the browser position (why don't we do it from inside the action that performs validation)
- clears the form (why don't we pass a callback to the action that does the validation, so that it clears the form on success?)

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
azangru: Why doesn't the `. inputGroup` class serve as an identifier?

imran-ebi: It is not an identifier to a particular set of elements i.e. in this case `startInputGroup`.

azangru: Oh, I see!
Ok, let it stay. I am a bit concerned we may be abusing the `role` attribute (which, eventually, has to do with accessibility), and if we end up really needing to identify otherwise indistinguishable elements in the DOM during tests, we will need to use `data-test-id`s (and a [babel plugin](https://www.npmjs.com/package/babel-plugin-jsx-remove-data-test-id) for their removal from production code).

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: As we discussed offline a couple of times about this, the `useEffect` is not used anymore and the `validateRegion` function has been moved into the browser helper.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imran-ebi: Removed these. The tests for validation and submission are already available.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: I changed this again to `stick` and the `regionInput` to be `stickInput` so it is a lot clearer.

src/ensembl/src/content/app/browser/Browser.test.tsx
imran-ebi: I'll add another assertion to test when the `activeGenomeId` is changed again

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: The code change has been done

src/ensembl/tests/fixtures/browser.ts
azangru: I would think `getChrLocationStr`.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -0,0 +1,274 @@
...
+  genomeKaryotype: GenomeKaryotypeItem[] | null;
+  isActive: boolean;
+  isDisabled: boolean;
+  changeBrowserLocation: (genomeId: string, chrLocation: ChrLocation) => void;
+  changeFocusObject: (objectId: string) => void;
+  toggleRegionEditorActive: (regionEditorActive: boolean) => void;
+};
+
+export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
+  const genomeKaryotypes = props.genomeKaryotype as GenomeKaryotypeItem[];
```
azangru: karyotypes — 🙅‍♂

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    if (activeGenomeId && regionFieldInput && chrLocation) {
+      setErrorMessage(null);
+
+      validateRegion({
+        regionInput: getRegionInputWithRegion(regionFieldInput),
+        genomeId: props.activeGenomeId,
+        onSuccess: onValidationSuccess,
+        onError: onValidationError
+      });
```
azangru: just curious why you decided to use this interface for `validateRegion` instead of having is an async function?

src/ensembl/src/content/app/browser/browserHelper.ts
```diff
@@ -14,3 +23,77 @@ export function getChrLocationStr(
...
+    },
+    regionId
+  };
+};
+
+export const validateRegion = async (params: {
+  regionInput: string;
+  genomeId: string | null;
+  onSuccess: (regionId: string) => void;
+  onError: (errorMessageList: any) => void;
```
azangru: any -> JSONValue (from types/JSON)?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
azangru: Yes, it’s gotten much simpler and easier to reason about :smiley:

azangru:
Not sure whether this needs to be addressed in this ticket (probably not), but the fact that the interface breaks on screens as wide as 1400px is worrying:

image

E.g. it's broken on my 13-inch laptop 😞 

image

ens-st3:
Thanks for spotting. It doesn't need fixing in this ticket, but obv does need fixing. I'll create another

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
imran-ebi: As discussed offline, I'll leave the code as it is for now.

src/ensembl/src/content/app/browser/browserHelper.ts
imran-ebi: I added a specific type for this with a new commit

imran-ebi:
@azangru @msmanoj I made two more commits after Kamal added the parse error message to the validation endpoint. Can you please have a look at the PR again (or the last few commits).

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -62,18 +61,18 @@ describe('browserHelper', () => {
         value: faker.random.number()
       }
     };
-    const mockValidationResult = createRegionValidationResult(); // all error messages are undefined by default
+    const mockValidationMessages = createRegionValidationMessages(); // all error messages are undefined by default
 
     test('returns all error messages as undefined', () => {
```
azangru: `undefined` no longer describes the code correctly?

(here and elsewhere in the file)

imran-ebi: Fixed it

src/ensembl/src/content/app/browser/browserHelper.ts
```diff
@@ -24,6 +24,19 @@ export function getChrLocationStr(
   return `${chrCode}:${startBp}-${endBp}`;
 }
 
+export type ErrorMessages = {
+  genomeIdError: string | undefined;
+  regionParamError: string | undefined;
+  regionError: string | undefined;
+  startError: string | undefined;
+  endError: string | undefined;
```
msmanoj: Here and below, would it be simpler to use `endError?: string` instead of specifying `undefined`?

imran-ebi: The types of these error properties have now changed to `string | null`. So all of them will exist in the returned `RegionValidationMessages` object.

src/ensembl/src/content/app/browser/browserHelper.test.ts
```diff
@@ -61,15 +61,15 @@ describe('browserHelper', () => {
...
       expect(getRegionValidationMessages(null)).toStrictEqual(
         mockValidationMessages
       );
     });
 
-    test('returns all error messages as undefined if all inputs are valid', () => {
+    test('returns all error messages as null if all inputs are valid', () => {
       expect(
         getRegionValidationMessages(mockValidationInfo).errorMessages
       ).toStrictEqual(mockValidationMessages.errorMessages);
```
msmanoj: Should this be `.toStrictEqual(null)` as we are not checking for null as mentioned in the description?

imran-ebi: They should all be `null`, as all of them are returned irrespective of the validation succeeding or failing.

PR #165, opened 2019-09-13
CSS breakpoints

ens-ap5:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-205

## Importance
We need to make our CSS breakpoints as flexible and easy to understand as possible. With the removal of Foundation I have implemented a new, more comprehensive set of breakpoints as described in the article mentioned in the ticket.

## Description
As per recommendations in the article:
* Extended the breakpoint CSS and used human-understandable names such as "for-phone-only" and "for-desktop-up"
* Changed one of the breakpoints from Foundation's 1440px to the more generic 1200px
* Switched to using mixins that wrap the media queries, which makes for cleaner CSS

## Views affected
The Header, BrowserBar and BrowserNavBar all use breakpoints, and have been updated to use the new breakpoint mixins.

### Other effects
None.

## Possible complications
If someone merges a feature branch that uses the old breakpoints, they will not see the expected behaviour. This seems unlikely, however, as we have not been prioritising responsive design yet.


src/ensembl/src/styles/_breakpoints.scss
```diff
@@ -1,16 +1,41 @@
...
+  @media (max-width: 599px) { @content; }
+}
+
+/* big phones, tablets in portrait mode */
+@mixin for-tablet-up {
+  @media (min-width: 600px) { @content; }
+}
+@mixin for-tablet-only {
+  @media (min-width: 600px) and (max-width: 899px) { @content; }
+}
```
azangru: One general consideration: it might be a good idea to share breakpoints with javascript (the same way we share colours). For that, we might want to have the breakpoints defined as variables that are then also exported from the SCSS file.

imran-ebi:
Are we waiting for any more changes or can it be merged after resolving the conflicts?

azangru:
@imran-ebi if you are ok with the names, then it only needs conflict resolution.

src/ensembl/src/styles/_breakpoints.scss
imran-ebi: Should we do this change here or in another PR?

imran-ebi:
I resolved the conflict and ran tests locally. It was successful. However, it fails when run on GitLab. Has this something to do with the branch name again?

PR #166, opened 2019-09-15
Update Home page

azangru:
## Type
- bug fix

## Related JIRA Issue(s)
ENSWBSITES-224
ENSWBSITES-186

## Description
This PR does two things:

1) Shows a list of previously viewed genomes on the Home page. Clicking on the link to a genome will take the user to the last location where they were in Genome Browser when viewing this species.
_(Also, some Home page styles were updated, as directed by Andrea. And removed the request for genome info on the home page, which was both failing and unnecessary)_

2) Fixes the AppBar for Genome Browser (when there is not enough space for all species tabs, they do not wrap to the next line, but instead shrink to fit the available space). Also, fixes the SpeciesTabBar component in the Storybook (which broke, because we added the Link component from react-router to SpeciesTabBar).

## Views affected
- Home page
- Genome browser
- Storybook (Shared components / SpeciesTabBar)

### DIFF
--- a/src/ensembl/src/content/home/homePageSelectors.ts
+++ b/src/ensembl/src/content/home/homePageSelectors.ts
@@ -1,0 +1,46 @@
+import * as urlFor from 'src/shared/helpers/urlHelper';
+import { getChrLocationStr } from 'src/content/app/browser/browserHelper';
+
+import {
+  getBrowserActiveEnsObjectIds,
+  getAllChrLocations
+} from 'src/content/app/browser/browserSelectors';
+import { getEnsObjectById } from 'src/ens-object/ensObjectSelectors';
+import { getCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
+
+import { RootState } from 'src/store';
+import { EnsObject } from 'src/ens-object/ensObjectTypes';
+
+export const getPreviouslyViewedGenomeBrowserObjects = (state: RootState) => {
+  const ensObjectIdsMap = getBrowserActiveEnsObjectIds(state);
+  const chrLocations = getAllChrLocations(state);
+  const committedSpecies = getCommittedSpecies(state);
+
+  const genomeIds = Object.keys(ensObjectIdsMap);
+
+  const ensObjectMap: { [genomeId: string]: EnsObject } = Object.keys(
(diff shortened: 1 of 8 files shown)


src/ensembl/src/content/home/homePageSelectors.ts
```diff
@@ -0,0 +1,91 @@
...
+        [genomeId]: {
+          speciesName,
+          assemblyName,
+          ensObjectId
+        }
+      };
+    }
+  }, {});
+
+  const areLoading: boolean = Object.keys(ensObjectIdsMap).reduce(
```
msmanoj: Is using `reduce` a good option here? If the first object is still loading, we don't really have to process the rest.

azangru: What alternative approaches do you have in mind? We don't have a `reduceWhile` (as Ramda has); and I would much rather use a reduce than a for-loop with external flags indicating whether or not to continue iteration. Execution of this iteration is going to be very fast (there are only a few committed species a user can reasonably have, and for each species it's just a lookup of data in a hash table).

imran-ebi: We can instead use `every`?

msmanoj: I agree that these blocks run very fast on our machines but there will be some slower systems that could work slightly better by avoiding the few extra iterations. 
I just did a site speed test and it says our website takes 3.6 seconds to load on 4G, which is already considered to be slow by google:
https://www.thinkwithgoogle.com/intl/en-gb/feature/testmysite/

I am worried that this value will only go higher as we keep adding features without considering these minor things.

azangru: > I just did a site speed test and it says our website takes 3.6 seconds to load on 4G

Loading time (on any G) depends on the amount of bytes that have to be downloaded. It does not depend on whether you run a longer or a shorter cycle :-)

In order to check the impact of iterating over a list of items, you would need to profile a specific piece of code. I am sure there would be no significant difference even if we had to loop over a thousand items; but in this case the number of items to iterate over will at most be a couple of dozens.

Following @imran-ebi 's suggestion, I replaced `reduce` with `some`.

PR #171, opened 2019-10-03
PasteOrUpload component refactoring and storybook entry

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-357

## Importance
2019 - Sprint 5 - Data Review

## Description
- Moved PasteOrUpload to shared -> components

## Views affected
- CustomDownlods -> Filters -> Limit By Gene ID
- Storybook -> Shared Components -> PasteOrUpload

### Other effects


src/ensembl/src/shared/components/paste-or-upload/PasteOrUpload.tsx
```diff
@@ -67,12 +73,12 @@ const PasteOrUpload = (props: PasteOrUploadProps) => {
       {shouldShowInput && (
         <div className={styles.fields}>
           <div className={styles.inputWrapper}>
-            <Input
+            <textarea
```
azangru: Let's start by creating our own Textarea component (similar to the Input component).

msmanoj: Would it not be better to do that in a separate PR?

azangru: 🤷‍♂ 
we can do a separate PR with just the Textarea, which will be considered before this PR. Or include the Textarea component in this PR.

imran-ebi: I think a separate PR would be better since `Textarea` will be a shared component

azangru:
Is this PasteOrUpload component reappearing in any application other than CustomDownload? Do we know what its behaviour will be?

For example, I am a bit surprised that this component "does not display the textarea by default" (why?). I also suspect that reading file contents and immediately filling the textarea with it may be too specific a behaviour (as may be incorrect to be called `upload`).

src/ensembl/src/shared/components/paste-or-upload/PasteOrUpload.tsx
msmanoj: I am thinking it will be better to create it later on and replace it here so that I can focus more on Custom Downloads tasks.

ens-st3: Agreed

azangru: This PR extracts PasteOrUpload into a shared component. Which means that you think that this component is generic enough to be used somewhere other than in custom download. Which means that you are un-focusing from the purely custom download task.

If you think this is a component that will be usable in other applications, please make it usable for other applications. If not, let it stay in custom download application and extract it to shared component later.

msmanoj: I think it is usable for other applications. Please let me know if there is anything that would prevent it from being used elsewhere.

msmanoj:
It is not reappearing anywhere at the moment. However, it will appear on other places in the future as we have more apps to be added like tools, for example. The current behaviour seems to work as expected, provided the limited backed we have. 

> "does not display the textarea by default" 
If we display the textarea by default, it will be same as having a Textarea component and the option to upload a file will not be visible to the user.

To my understanding, upload would mean that there is an option to select a file form the local disk. I wouldn't be worried too much about what happens to the file after selecting. I am happy to reword it if we can come up with a meaningful name.

src/ensembl/src/shared/components/paste-or-upload/PasteOrUpload.tsx
azangru: > Please let me know if there is anything that would prevent it from being used elsewhere.

Since we don't have any other applications that are currently relying on this component, it's hard to predict what they will need from it. But if we are considering this component to be a standalone shared component, then can we treat it a bit more carefully?

- let's create the Textarea component on which it depends (it will take 10-20 minutes or so)
- let's consider its name —  it is really PasteOrUpload or rather PasteOrReadFromFile?
- let's consider whether it always needs to hide the textarea initially?

msmanoj: > let's create the Textarea component on which it depends (it will take 10-20 minutes or so)

Create the component, look for improvements, write tests, add to storybook, create PR, get it reviewed, make changes and so on... No way that's all happening in 20 mins 😄 

> let's consider its name — it is really PasteOrUpload or rather PasteOrReadFromFile?

`ReadFromFile` makes sense as a name but I haven't come across many places where is says 'Read from file' and you are asked to select a file. I named it `PasteOrUpload` because the initial text says "Paste or Upload file".

msmanoj:
That's true. We need an Upload component as well.  I would prefer to create JIRA tickets for these two components and leave PasteOrUpload in the shared folder just so it gets more attention to be improvised in future. We already have many components that are only used in Custom Downloads but placed in the shared folder.

azangru:
> We already have many components that are only used in Custom Downloads but placed in the shared folder.

We have the Accordion shared component that is used only in Custom Download, but is sufficiently generic. We have the Checkbox shared component that is used only in Custom Download, but is sufficiently generic. ENSWBSITES-357 could have been accomplished by creating the Textarea shared component and using it in Custom Download's PasteOrUpload. If instead you propose to take PasteOrUpload out of Custom Download, I have no objections to that, but I do not understand why not do it properly.

I don't think there is much more I can say about this PR, and I understand that this conversation is itself distracting and time-consuming. If you prefer to treat PasteOrUpload as a very temporary component that generates subsequent tasks for its refactoring, then merge it; if you'd rather do it properly from the beginning, then don't.

msmanoj:
Okay, I will put this task on hold and get back to it once I complete the other CD tasks. I might be able to find some time to work on Textarea and Upload components later on. Worst case, I will put it back inside CD folder.

src/ensembl/src/content/app/custom-download/containers/header/customDownloadHeaderHelper.ts
```diff
@@ -75,16 +75,26 @@ export const getEndpointUrl = (
...
 
   // FIXME: Temporarily apply the filters locally
-  const gene_ids = get(processedFilters, 'genes.limit_to_genes', [])
-    .join(',')
-    .split(',')
-    .map(trim)
-    .filter(Boolean);
+  const gene_ids: string[] = [];
+
+  const limitToGenes = get(processedFilters, 'genes.limit_to_genes', []);
```
imran-ebi: What does `limitToGenes` mean here? Is it a limit to the number of genes?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -1,115 +1,148 @@
 import React, { useState, useEffect } from 'react';
```
azangru: Very confused by the name of this component now. It says "Checkbox with textfields", but now also contains logic about "files"

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -1,115 +1,148 @@
...
-import styles from './CheckboxWithTextfields.scss';
 
 import ImageButton from 'src/shared/components/image-button/ImageButton';
-import { ReactComponent as AddIcon } from 'static/img/browser/zoom-in.svg';
-import PasteOrUpload from '../paste-or-upload/PasteOrUpload';
+import { ReactComponent as RemoveIcon } from 'static/img/shared/clear.svg';
+
+import Textarea from 'src/shared/components/textarea/Textarea';
+import Upload, { ReadFile } from 'src/shared/components/upload/Upload';
+import { nextUuid } from 'src/shared/helpers/uuid';
```
azangru: I missed this previously, but we do we call this helper `uuid`? uuid is a universally unique identifier; it has several possible formats, but just being a regularly incremented number is not one of them.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -1,115 +1,148 @@
...
-    setShowAddButton(Boolean(props.values[props.values.length - 1]));
-  }, [props.values]);
+    let checkedStatus = false;
 
-  const handleCheckboxOnChange = (isChecked: boolean) => {
-    setIsChecked(isChecked);
-    if (!isChecked && values.length > 0) {
-      props.onChange([]);
+    if (props.textValue && props.textValue.length) {
+      showTextarea(true);
```
azangru: technically, this should only be done if textarea isn't already shown. Don't know whether updating true to true will cause a re-render though; perhaps React is smart and will prevent a re-render.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -1,115 +1,148 @@
...
-
-  const handleOnChange = (value: string, index: number) => {
-    const newValues: string[] = [...values] as string[];
-    newValues[index] = value;
+    if (props.files && props.files.length) {
+      checkedStatus = true;
+    }
 
-    setShowAddButton(Boolean(newValues[newValues.length - 1]));
+    setIsChecked(checkedStatus);
```
azangru: technically, this should only be done if checkedStatus is different from isChecked

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
```diff
@@ -1,115 +1,148 @@
...
-
-    newValues.splice(index, 1);
-    setShowAddButton(Boolean(newValues[newValues.length - 1]));
+    const newFiles = [...files];
+    newFiles.splice(index, 1);
+    props.onFilesChange(newFiles);
+  };
 
-    props.onChange(newValues);
+  const handleOnUpload = (uploadedFiles: ReadFile[]) => {
```
azangru: `handleOnUpload` --> either `handleUpload`, or `onUpload`

src/ensembl/src/shared/components/upload/Upload.scss
```diff
@@ -1,19 +1,19 @@
...
   padding: 10px;
   cursor: pointer;
-  transition: border 0.4s;
+  transition: background-color 0.4s;
 }
 
 .defaultUploadActive {
-  border: 1px dashed $blue;
-  transition: border 0.4s;
+  // border: 1px dashed $blue;
```
azangru: ✂️  ?

src/ensembl/src/content/app/custom-download/containers/header/customDownloadHeaderHelper.ts
msmanoj: This is just a temporary solution so I have just used the field name displayed in the accordion.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: I did notice it but I thought it would be better to refactor on a different PR than keep on extending this one.

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: Do you have any suggestions for what to call it as?

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: The textarea is hidden by default and is only shown when there is a value passed on or when the 'Paste data' is clicked.

src/ensembl/src/shared/components/upload/Upload.scss
msmanoj: Removed it

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-with-textfields/CheckboxWithTextfields.tsx
azangru: just an `id`? `generateSimpleId` instead of `nextUuid`?

msmanoj: How about `generateUniqueId` ? I feel like the main thing is missing if we say  `generateSimpleId`

azangru: Ids should be unique by definition 😄

msmanoj: That's true but would it not be a little bit more helpful if we use `Unique` instead of `Simple` rather than having to go by definition?

azangru: The word "unique" is redundant, and misleading. By calling a function `generateUniqueId`, we imply that there may also be non-unique id's that some other function may generate.

I am not particularly attached to the word "simple". `generateId`? `generateNumericId`? `generateId({ strategy: 'increment-number' })`? The `uuid` [package](https://www.npmjs.com/package/uuid) on npm refers to `uuid/v1`, `uuid/v3`, etc. (although each of these versions have technical definitions); so we could take inspiration from them and do something like `generateId({ version: 1 })`; or `getIdGenerator({ version: 1 })`, which returns a function that generates ids by incrementing numbers (if we expect any other id generation strategies to warrant introducing versions).

msmanoj: I guess we can go for `generateId` now and if we need to introduce versioning, it can be done in a separate PR.

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -176,6 +177,7 @@ const Upload = (props: UploadProps) => {
 };
 
 Upload.defaultProps = {
+  id: 'upload_' + generateId(),
```
azangru: This is much better, but why have the id at all? What happens if we remove it?

msmanoj: Are you saying we can set the id's only if it is passed in?

azangru: I am not sure we even need to set an id on the input field at all. Do we?

msmanoj: I've removed it 🤞

PR #173, opened 2019-10-08
Update dependencies

imran-ebi:
## Type
- Infrastructure

## Description
Update dependencies, especially react router which now has hooks. We can use them instead of wrapped components in our code. This will avoid the need to pass props when writing tests for components that use router API.

Also, react has got a new update. There was a breaking change for `use-resize-observer`, so I did not update it.

As for the dev dependencies, there are several packages that got updated. The main ones are the typings for react router (needed for using hooks), and stylelint. The latter is a breaking change, however nothing broke when I ran the checks mentioned in the section below. Eslint and its plugins have also been updated and they contain several fixes.

## Checks
- `npm test` - OK
- `npm run check-types` - OK
- dev build - OK
- prod build - OK
- storybook - OK


imran-ebi:
@azangru I had to push another commit after updating the dev dependencies. I'll update the description to reflect this.

azangru:
👌

PR #174, opened 2019-10-09
Custom downloads layout changes

msmanoj:
## Type

- [x] Bug fix
- [x] New feature

## Related JIRA Issue(s)
- Incorporate species selector into custom download:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-155

- Replace results preview with attributes accordion and remove tab buttons
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-356

- Introduce result cards in the preview downloads screen
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-358

- Display selected filters in the permanent block after folding the accordion
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-359

## Description
- Update the UI to match the new design provided by Andrea
- Incorporated the species selector
- Fixed on minor bug with the Accordion shared component

## Views affected
Custom Downloads


src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -1,22 +1,45 @@
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, useEffect } from 'react';
```
azangru: no need for FunctionComponent

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -1,22 +1,45 @@
...
+  getShowPreFilterPanel,
+  getCustomDownloadActiveGenomeId
+} from './state/customDownloadSelectors';
 import { RootState } from 'src/store';
 import CustomDownloadHeader from './containers/header/CustomDownloadHeader';
 import CustomDownloadContent from './containers/content/CustomDownloadContent';
+import CustomDownloadAppBar from './containers/app-bar/CustomDownloadAppBar';
+import { setActiveGenomeId } from './state/customDownloadActions';
+import { getCommittedSpecies } from '../species-selector/state/speciesSelectorSelectors';
+import { CommittedItem } from '../species-selector/types/species-search';
```
azangru: Perhaps absolute paths starting with src instead?

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -1,22 +1,45 @@
...
+type CustomDownloadProps = StateProps & DispatchProps;
 
 const CustomDownload: FunctionComponent<CustomDownloadProps> = (
   props: CustomDownloadProps
 ) => {
+  useEffect(() => {
+    if (!props.activeGenomeId && props.committedItems.length) {
+      props.setActiveGenomeId(props.committedItems[0].genome_id);
+    }
+  }, [props.activeGenomeId]);
```
azangru: Should this effect run only once, on component mount (if so, use empty array in the list of dependencies).

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -27,8 +50,26 @@ const CustomDownload: FunctionComponent<CustomDownloadProps> = (
...
 
-const mapStateToProps = (state: RootState): StateProps => ({
-  showPreFiltersPanel: getShowPreFilterPanel(state)
-});
+const mapStateToProps = (state: RootState): StateProps => {
+  const activeGenomeId = getCustomDownloadActiveGenomeId(state);
+
+  let showPreFiltersPanel = false;
+  if (activeGenomeId) {
+    showPreFiltersPanel = getShowPreFilterPanel(state);
```
azangru: Maybe better call it shouldShowPreFilterPanel (otherwise it reads as if I am getting a ShowPreFilterPanel). Also, won’t it be better to check for active genome id in the selector itself?

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -27,8 +50,26 @@ const CustomDownload: FunctionComponent<CustomDownloadProps> = (
...
+  }
+
+  return {
+    showPreFiltersPanel: showPreFiltersPanel,
```
azangru: shorthand?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
```diff
@@ -4,14 +4,70 @@
...
+  height: 45px;
+}
+
+.checkboxGridHeader + .checkboxGridContainer {
+  padding-top: 5px;
+}
+
+.label {
+  color: $grey;
+  float: left;
```
azangru: Float? Oh dear, do we really want floats?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
```diff
@@ -8,16 +12,33 @@ export type CheckboxGridOption = {
...
-  hideLabel?: boolean;
   label: string;
+  hideLabel?: boolean;
+  isCollapsable?: boolean;
+  isCollapsed?: boolean;
   onChange: (status: boolean, id: string) => void;
+  onCollapse?: (collapsed: boolean) => void;
 };
 
+type NonCollapsableProps = Modify<
```
azangru: I think the preferred spelling is collapsible with an i (see e.g. https://github.com/react-bootstrap/react-bootstrap/issues/425).

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { default as classNamesMerger } from 'classnames';
```
azangru: Why this style? It was simply import classNames from 'classnames' before.

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import closeIcon from 'static/img/shared/close.svg';
+
+import styles from './Panel.scss';
+
+type Props = {
+  title: string;
+  children: JSX.Element;
```
azangru: ReactNode, surely?

src/ensembl/src/content/app/custom-download/containers/app-bar/CustomDownloadAppBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import AppBar, {
+  HelpAndDocumentation
+} from 'src/shared/components/app-bar/AppBar';
+import { FocusableSelectedSpecies } from 'src/shared/components/selected-species';
+import SpeciesTabsWrapper from 'src/shared/components/species-tabs-wrapper/SpeciesTabsWrapper';
+
+import { RootState } from 'src/store';
+import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
+import { getCustomDownloadActiveGenomeId } from '../../state/customDownloadSelectors';
```
azangru: `../..` -> path starting with `src`

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -11,26 +11,58 @@ import {
   AccordionItemPermanentBlock
 } from 'src/shared/components/accordion';
 
-import { getAttributesAccordionExpandedPanel } from '../../../state/attributes/attributesSelector';
+import {
+  getAttributesAccordionExpandedPanels,
+  getSelectedAttributes
+} from '../../../state/attributes/attributesSelector';
```
azangru: `../../..` -> `src`

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
```diff
@@ -1,40 +1,60 @@
...
+  launchBarExpanded: boolean;
 };
 
+type DispatchProps = {
+  setShowExampleData: (showExampleData: boolean) => void;
+};
+
+type Props = StateProps & DispatchProps;
+
 const Content = (props: Props) => {
```
azangru: Name of the default exported component should be the same as the name of the file.

src/ensembl/src/content/app/custom-download/components/overlay/Overlay.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
azangru: Several points:
- This component is in custom download, but its name is very generic. Perhaps change it to CustomDownloadContentOverlay? Unless it's a component that wants to be shared
- It looks like what the Overlay component is doing is achievable with the `filter: opacity(0.3)` CSS rule applied to  the `CustomDownloadContent` component. On the plus side, that makes the overlay component unnecessary here. On the minus side, filter isn't supported on IE11. But we don't much care for IE11 either.

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: I was using the auto import feature to import these.
I am starting to think why we shouldn't implement an index file for types somewhere in the root level as it would make importing a lot easier. In the example above, it's not very easy to keep in mind where `CommittedItem` is defined. We also do not have a guarantee that the name `CommittedItem` will never be re-used on other apps.
Wouldn't it be better if we can use: ` import {CommittedItem} from 'src/types'` ? Any thoughts?

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
azangru: This component sits in CustomDownload, but has a very generic name. Should it be CustomDownloadPopup? Or do we expect this popup to reappear in other apps and suspect that it belongs in the shared folder?

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
```diff
@@ -0,0 +1,47 @@
...
+    : styles.panelHeaderDefault;
+  const panelBodyClassNames = classNames
+    ? classNamesMerger(styles.panelBodyDefault, classNames.bodyClassName)
+    : styles.panelBodyDefault;
+
+  return (
+    <div className={panelClassNames}>
+      <div className={headerClassNames}>{title}</div>
+      <div className={panelBodyClassNames}>
+        <div {...rest}></div>
```
azangru: What remains in the `rest` here? Just children? Would it not be clearer to explicitly write out the children prop?

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.scss
```diff
@@ -0,0 +1,33 @@
...
+    width: 100%;
+    position: relative;
+    clear: both;
+    .lineHeader {
+      text-align: right;
+      color: $dark-grey;
+      width: 35%;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      float: left;
```
azangru: image

It looks like a table. Perhaps instead of `float:left`, and `clear: both`, either the `table` markup or the `display: grid` will be more appropriate?

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
msmanoj: I've changed it to work with `display`. Just wondering why float is not good to use?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -11,26 +11,58 @@ import {
...
+};
 
 type Props = StateProps & DispatchProps;
 
+const getTotalSelectedAttributes = (
+  attributes: JSONValue,
+  totalSelectedAttributes = 0
+) => {
+  Object.keys(attributes).forEach((key) => {
+    if (typeof attributes[key] === 'boolean' && attributes[key] === true) {
```
azangru: was `typeof attributes[key] === 'boolean'` required by typescript? It looks excessive, considering the next check is `attributes[key] === true`

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
azangru: > We also do not have a guarantee that the name CommittedItem will never be re-used on other apps.

Regarding this point, it would be much easier if we try to avoid it, because otherwise it will get harder to find the correct thing. Particularly when something breaks in the interface, and you need to find out which component the broken piece corresponds to. Also, error messages will contain the names of components/files. So, would really, really like us to not reuse the same name for different things in different parts of our application.

As for the index-like `src/types` file, I am still not sold on it, to be honest. Perhaps I am just not using relevant editor's features, and so never found myself wanting this.

src/ensembl/src/content/app/custom-download/components/overlay/Overlay.tsx
msmanoj: > It looks like what the Overlay component is doing is achievable with the filter: opacity(0.3)

It also prevents interaction with the elements behind the overlay.

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
azangru: It's not that float is not good to use; it's just that it was historically used (abused) to achieve all sorts of tasks when the browser didn't have any better alternatives. Nowadays, with the much more powerful CSS rules, we can reserve it for what it was originally intended — text-wrap effects.

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -11,26 +11,58 @@ import {
...
+      totalSelectedAttributes++;
+    } else if (typeof attributes[key] === 'object') {
```
azangru: Just curious here: can `attributes[key]` be an array? If yes, it will be captured by this if statement. Will it be ok?

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: No longer applies

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: Removed the excessive check

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/app-bar/CustomDownloadAppBar.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.tsx
msmanoj: Renamed it

src/ensembl/src/content/app/custom-download/containers/content/customDownloadContentHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+  objectOrArray: JSONValue,
+  prefix = '',
+  formatter = (k: string) => k
+) => {
+  if (!objectOrArray) {
+    return {};
+  }
+  const nestedFormatter = (k: string) => '.' + k;
+
+  const nestElement = (prev: any, value: any, key: any): JSONValue =>
```
azangru: any, any, any => JSONValue, JSONValue, string?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
msmanoj: No, it cannot be an array at the moment.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.scss
```diff
@@ -2,15 +2,49 @@
...
+    font-weight: 700;
+  }
+}
+.previewCounter {
+  margin-left: 20px;
+  padding-top: 12px;
+  display: inline-block;
+}
+
+.boldResultCounter {
```
azangru: is this rule used anywhere?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.scss
```diff
@@ -34,3 +68,7 @@
   top: 10px;
```
azangru: What does the `tempPadding` rule above mean?  Why is it `temp`?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
 } from 'src/shared/components/accordion';
 
-import styles from './FiltersAccordion.scss';
-
-import { getFiltersAccordionExpandedPanel } from '../../../state/filters/filtersSelector';
+import { getPreviewResult } from '../../../state/customDownloadSelectors';
+import {
+  getFiltersAccordionExpandedPanels,
+  getSelectedFilters
+} from '../../../state/filters/filtersSelector';
```
azangru: ../../.. -> src

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+  preview: JSONValue;
+};
+
+type DispatchProps = {
+  setFiltersAccordionExpandedPanel: (expandedPanels: string[]) => void;
+  resetSelectedFilters: () => void;
+  updateSelectedFilters: (filters: JSONValue) => void;
+};
+
+const getTotalSelectedFilters = (
```
azangru: getSelectedFiltersCount?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+type Filter = {
+  [key: string]: boolean;
+};
+type SelectedFilters = {
+  [key: string]: boolean | string | string[] | Filter;
+};
+
+type StateProps = {
+  expandedPanels: string[];
+  selectedFilters: {};
```
azangru: Why not JSONValue?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+  filters: SelectedFilters,
+  totalSelectedFilters = 0
+) => {
+  if (!filters) {
+    return 0;
+  }
+  Object.keys(filters).forEach((key: string) => {
+    if (key === 'preExpanded') {
+      // Skip preExpanded keys
+    } else if (typeof filters[key] === 'boolean' && filters[key] === true) {
```
azangru: unnecessary first part of condition

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+      totalSelectedFilters++;
+    } else if (typeof filters[key] === 'string' && filters[key] !== '') {
```
azangru: unnecessary typeof

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+      totalSelectedFilters++;
+    } else if (
+      Array.isArray(filters[key]) &&
+      (filters[key] as string[]).length > 0
+    ) {
+      totalSelectedFilters++;
+    } else if (typeof filters[key] === 'object') {
+      totalSelectedFilters = getTotalSelectedFilters(
```
azangru: is this part correct? `= getTotalSelectedFilters`, not `+= getTotalSelectedFilters`?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
```diff
@@ -7,32 +7,87 @@ import {
...
+    }
+  });
+
+  return totalSelectedFilters;
+};
 
 type Props = StateProps & DispatchProps;
 
 const FiltersAccordion = (props: Props) => {
+  useEffect(() => {}, []);
```
azangru: what does this do?

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { useEffect } from 'react';
+import { connect } from 'react-redux';
+import { RootState } from 'src/store';
+import {
+  getPreviewResult,
+  getIsLoadingResult
+} from '../../../state/customDownloadSelectors';
```
azangru: ../../../ -> src

src/ensembl/src/content/app/custom-download/containers/content/customDownloadContentHelper.ts
msmanoj: TypeScript is not happy with that. This is just a temporary solution until the backend is available.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.scss
msmanoj: No, it's not. Removed it.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+import {
+  flattenObject,
+  getProcessedAttributes
+} from 'src/content/app/custom-download/containers/content/customDownloadContentHelper';
+
+const formatResponseToArray = (responseData: any): any => {
+  const preResult: any = [];
+
+  const responseArray = flattenObject(responseData);
+  Object.keys(responseArray)
```
azangru: `responseArray` is an object. Why is it called an array then?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.scss
msmanoj: It's to add padding to the temporary text that is displayed. Renamed it to `temporaryContentPadding`.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: I remember changing it but it got missed out somehow. It's done now.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: Removed it

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      if (!preResult[topID]) {
+        preResult[topID] = {
+          ...preResult[`${keySplit[0]}00`]
+        };
+      }
+      preResult[topID][id] = responseArray[key];
+    });
+
+  return preResult;
```
azangru: I am seeing a preResult that is and array with over 7000 elements, most of them empty:

image

image

Is this intended?

Also, why `pre`Result, why not result?

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: Yes, it is correct.

src/ensembl/src/content/app/custom-download/containers/content/filter-accordion/FiltersAccordion.tsx
msmanoj: Left it by mistake, sorry. Removed it now.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.tsx
msmanoj: That's done.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+};
```
azangru: Without any comments or tests, it's hard to understand what's supposed to be happening in the function above. Is this throw-away code that will have to be rewritten when the api updates?

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
msmanoj: I've renamed it to `result`.

Yes, it is intended. I've just implemented this as a temporary solution as all these will be done in the backend.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
msmanoj: This is just to cope up with a bug in the backend. We can scrap these once it is fixed. I think Marc recently mentioned that it's more likely to be getting fixed soon.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
msmanoj: Renamed it.

src/ensembl/src/content/app/custom-download/containers/content/preview-download/PreviewDownload.tsx
```diff
@@ -1,59 +1,55 @@
 import React from 'react';
+import get from 'lodash/get';
+
 import { connect } from 'react-redux';
 import { RootState } from 'src/store';
-import { Link } from 'react-router-dom';
+
+import { ReactComponent as closeIcon } from 'static/img/shared/close.svg';
 import { getSelectedAttributes } from '../../../state/attributes/attributesSelector';
 import { getSelectedFilters } from '../../../state/filters/filtersSelector';
```
azangru: ../../.. -> src

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
azangru: Wouldn't using object instead of an array (and returning `Object.values` of that object) be a more natural choice? Is there any specific array-like behaviour that you needed from the result while building it?

src/ensembl/src/content/app/custom-download/containers/content/result-loader/ResultLoader.tsx
```diff
@@ -57,64 +65,7 @@ const ResultHolder = (props: Props) => {
...
-      </div>
-    </>
-  );
-};
-
-type DispatchProps = {
-  fetchPreviewResult: (fetchPreviewResult: string) => void;
-  clearPreviewResult: () => void;
-  setIsLoadingResult: (setIsLoadingResult: boolean) => void;
+  return null;
```
azangru: Huh? Is this component no longer needed? Should it be deleted?

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+import styles from './PreviewCard.scss';
+
+type Props = {
+  selectedAttributes: JSONValue;
+  selectedFilters: JSONValue;
+  preview: JSONValue;
+  isLoadingResult: boolean;
+};
+
+const ResultLoader = (props: Props) => {
```
azangru: Name

src/ensembl/src/content/app/custom-download/state/attributes/attributesActions.ts
```diff
@@ -25,65 +27,243 @@ export const setAttributes = createAsyncAction(
...
+    })
+  );
+};
 
 export const setOrthologueSpecies = createAsyncAction(
   'custom-download/set-orthologue-species-request',
   'custom-download/set-orthologue-species-success',
   'custom-download/set-orthologue-species-failure'
-)<{ searchTerm: string }, CheckboxGridOption[], Error>();
+)<{ searchTerm: string }, undefined, Error>();
```
azangru: not `void`?
(although I am not sure what the difference is)

src/ensembl/src/content/app/custom-download/state/filters/filtersActions.ts
```diff
@@ -1,29 +1,103 @@
...
+import { ActionCreator, Action } from 'redux';
 
-import customDownloadStorageService from 'src/content/app/custom-download/services/custom-download-storage-service';
 import JSONValue from 'src/shared/types/JSON';
+import { updateActiveConfigurationForGenome } from '../customDownloadActions';
+import {
+  getCustomDownloadActiveGenomeConfiguration,
+  getCustomDownloadActiveGenomeId
+} from '../customDownloadSelectors';
+import set from 'lodash/set';
```
azangru: to the top of the file, with other imports from node_modules

src/ensembl/src/content/app/custom-download/containers/content/preview-download/PreviewDownload.tsx
msmanoj: That's done

src/ensembl/src/content/app/custom-download/state/attributes/attributesActions.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/state/attributes/attributesActions.ts
```diff
@@ -1,21 +1,23 @@
...
 
 import { AttributeWithOptions } from 'src/content/app/custom-download/types/Attributes';
 import JSONValue from 'src/shared/types/JSON';
 import { CheckboxGridOption } from 'src/content/app/custom-download/components/checkbox-grid/CheckboxGrid';
 import Species from 'src/content/app/custom-download/types/Species';
+import {
+  getCustomDownloadActiveGenomeId,
+  getCustomDownloadActiveGenomeConfiguration
+} from '../customDownloadSelectors';
+import set from 'lodash/set';
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.tsx
msmanoj: Renamed it

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
msmanoj: It might reappear as Andrea mentioned but I'm not sure what else it requires to be put in the shared folder. Also, I don't think it can be called as a popup because of the way it is used in the Preview Download screen.

src/ensembl/src/content/app/custom-download/containers/content/result-loader/ResultLoader.tsx
msmanoj: This is still responsible to load the results when the attributes or filters are changed. Not very sure if it needs to be inside an action on its own.

src/ensembl/src/content/app/custom-download/state/filters/filtersActions.ts
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/preview-card/PreviewCard.scss
msmanoj: Updated it to use table.

src/ensembl/src/content/app/custom-download/containers/content/result-loader/ResultLoader.tsx
azangru: Interesting. But using React’s component just for its `useEffect` hook without rendering anything is probably an abuse of React’s component model. I vaguely remember seeing react-router guys do something like that quite a while ago; but this probably can be expressed more idiomatically, without having a dedicated component which doesn’t render anything.

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
azangru: `CustomDownloadInfoBox`?
Or perhaps Andrea can offer some naming suggestions? `Panel` is just far too broad and non-specific.

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
azangru: Ok. It's just that such a humungous and mostly empty array looks terrifying.

azangru:
A general remark: the two columns of content in Custom download are positioned using `position: absolute` in a zero-height container, which does not look like a very good layout technique choice. `Display: grid` will more robustly represent the layout intention here. Perhaps even `display: flex` could do the job.

image

image

Don't know if this should be changed specifically in this PR, but I do think this should be changed.

src/ensembl/src/content/app/custom-download/containers/content/result-loader/ResultLoader.tsx
msmanoj: Moved it to the Header component and removed ResultLoader.

msmanoj:
> A general remark: the two columns of content in Custom download are positioned using position: absolute in a zero-height container, which does not look like a very good layout technique choice. Display: grid will more robustly represent the layout intention here. Perhaps even display: flex could do the job.

I've updated it to use grid.

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
msmanoj: Andrea did mention that this might be re-used within bulk downloads so I am not sure if it we call it something more specific to custom downloads and move it outside later with a different name.

msmanoj: Spoke to Andrea regarding this and she suggested to use `CustomDownloadInfoCard`.

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
msmanoj: > So, would really, really like us to not reuse the same name for different things in different parts of our application.

That's what I am thinking as well. Once the application grows, it would become harder to keep in mind if the name is already used.

I've currently updated my vscode settings to use absolute path when I do an auto import.

src/ensembl/src/content/app/custom-download/services/custom-download-storage-service.tsx
```diff
@@ -32,104 +24,32 @@ export class CustomDownloadStorageService {
...
-
-  public saveShowPreview(shouldShowPreview: boolean) {
-    this.storageService.save(
-      StorageKeys.SHOW_PREVIEW,
-      shouldShowPreview,
+  // FIXME: Not sure why TypeScript doesn't like it if I use {[key: string]: CustomDownloadStateForGenome} here
+  public updateActiveConfigurationsForGenome(activeConfigurations: any): void {
+    this.storageService.update(
+      StorageKeys.ACTIVE_CONFIGURATIONS,
+      activeConfigurations,
```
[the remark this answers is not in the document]
msmanoj: >  explicitly declaring the type of activeConfigurations fixed it for me

Yes, that worked 👍 

> As an aside: I don't quite remember whether the types in storageService (ObjectValue, etc.) can't simply be replaced with our JSONValue type. I suspect that they can, but haven't really investigated. Perhaps should be done in a separate PR, or can be done in this one...

I have replaced `ObjectValue` with `JSONValue` and Typescript is still happy so I guess it's fine.

src/ensembl/src/content/app/custom-download/containers/header/CustomDownloadHeader.tsx
```diff
@@ -9,29 +12,51 @@ import Select, { Option } from 'src/shared/components/select/Select';
...
+type DispatchProps = {
+  togglePreFiltersPanel: (togglePreFiltersPanel: boolean) => void;
+  setShowPreview: (setShowPreview: boolean) => void;
+  setDownloadType: (setDownloadType: string) => void;
+  fetchPreviewResult: (endpointURL: string) => void;
+  clearPreviewResult: () => void;
+  setIsLoadingResult: (isLoadingResult: boolean) => void;
+};
+
 type Props = StateProps & DispatchProps;
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/preview-download/PreviewDownload.tsx
```diff
@@ -1,59 +1,55 @@
...
+  committedSpecies: CommittedItem | null;
 };
 
 type DispatchProps = {
-  setShowPreview: (setShowPreview: boolean) => void;
-  toggleTab: (toggleTab: string) => void;
+  setShowPreview: (showPreview: boolean) => void;
 };
 
 type Props = StateProps & DispatchProps;
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/custom-download/containers/content/preview-card/previewCardHelper.ts
```diff
@@ -0,0 +1,111 @@
+import JSONValue from 'src/shared/types/JSON';
+
+import {
+  flattenObject,
+  getProcessedAttributes
+} from 'src/content/app/custom-download/containers/content/customDownloadContentHelper';
+
+const formatResponseToArray = (responseData: any): any => {
```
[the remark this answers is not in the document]
msmanoj: Removed it

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -201,13 +263,26 @@ const ContentBuilder = (props: ContentBuilderProps) => {
...
+  const buildCheckboxWithTextfields = (
+    entry: Attribute,
+    path: Path,
+    key: number
+  ) => {
     const currentPath = [...path, entry.id];
 
     const values: string[] = get(props.selectedData, currentPath, '');
 
+    if (props.showOverview) {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -133,23 +171,41 @@ const ContentBuilder = (props: ContentBuilderProps) => {
...
-        onChange={(status: boolean, id: string) =>
-          onChangeHandler([...currentPath, id], status)
-        }
-        options={gridClone}
-        label={entry.label}
-        {...additionalProps}
-      />
+      <div className={styles.checkboxGridWrapper} key={key}>
+        <CheckboxGrid
+          class
```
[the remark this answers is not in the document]
msmanoj: I've removed. I guess I added it by mistake

src/ensembl/src/content/app/custom-download/CustomDownload.tsx
```diff
@@ -1,24 +1,45 @@
...
+    return null;
+  }
   return (
     <>
-      {props.showPreFiltersPanel && <PreFilterPanel />}
-      {!props.showPreFiltersPanel && (
+      <CustomDownloadAppBar onSpeciesSelect={props.setActiveGenomeId} />
+      {props.shouldShowPreFilterPanel && <PreFilterPanel />}
+      {!props.shouldShowPreFilterPanel && (
         <>
```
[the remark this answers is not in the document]
msmanoj: No, there are two.
<CustomDownloadHeader /> 
<CustomDownloadContent />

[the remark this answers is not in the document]
imran-ebi: Ah I didn't look at it properly since both of them are in the same line. Can you put the second element in the next line so that it'll be clearer at first glance.

src/ensembl/src/services/storage-service.ts
```diff
@@ -2,19 +2,17 @@ import windowService, {
...
   LOCAL_STORAGE = 'localstorage',
   SESSION_STORAGE = 'sessionstorage'
 }
 
 type PrimitiveValue = string | number | boolean | null | undefined;
-type ArrayValue = PrimitiveValue[] | ObjectValue[];
-type ObjectValue = {
-  [key: string]: PrimitiveValue | ArrayValue | ObjectValue;
-};
+type ArrayValue = PrimitiveValue[] | JSONValue[];
```
azangru: Could you import PrimitiveValue and ArrayValue from `types/JSON.ts` as well?

msmanoj: That's done

src/ensembl/src/content/app/custom-download/components/overlay/Overlay.tsx
azangru: The naming issue for this component still stands. We can choose between:
- keeping the Overlay name (or change it to something as general, such as PopupBackground), and move the component to shared components, or
- keep the component in CustomDownload, but change its name (CustomDownloadOverlay? CustomDownloadPopupBackground?)

src/ensembl/src/content/app/custom-download/components/panel/Panel.tsx
azangru: 👌

src/ensembl/src/content/app/custom-download/components/overlay/Overlay.tsx
msmanoj: I was thinking about moving it to shared components by adding `className` prop. Would that be okay?

azangru: I think so.

msmanoj: That's done

PR #178, opened 2019-10-17
Ability to turn on/off all tracks with a single message

ens-ds23:
## Type

- [ ] Bug fix
- [X] New feature

## Related JIRA Issue(s)
ENSWBSITES-353

## Importance
This allows bookmarks to work properly re tracks which are on/off when combined with Manoj's frontend code.

## Description
This allows bookmarks to work properly re tracks which are on/off when combined with Manoj's frontend code.

The key is called default and can have any value (I suggest "true"). Don't specify the key if you don't want this behaviour. Note that a value false /will/ cause it to happen, this is a known limitation which will go away at some point in the future, but isn't a quick fix, so if you don't want this behaviour omit the key entirely.

I tested this in Chrome with the following test-setup message which turns off the gc and vairant tracks:

window.postMessage({ "type": "bpane", "payload": {  "off": ["track:gc","track:variant"] } },"*");

And then giving the following which turns off the gc track but includes the "default" key so that the variant track comes back on! The new behaviour is the variant track coming on when you run this second message.

window.postMessage({ "type": "bpane", "payload": {  "off": "track:gc", "default": true } },"*");

## Views affected
Genome browser + chrome

### Other effects
None

## Possible complications
None anticipated


ens-ds23:
We've agreed to abandon this functionality for architectural reasons.

PR #180, opened 2019-10-18
[WIP] What if we try git-lfs?

azangru:
The good news is, GitLab (at least, the enterprise edition) added support for git-lfs objects during pull-mirroring of other repositories (about 5-6 months ago, in release 11.11: see [this comment](https://gitlab.com/gitlab-org/gitlab-foss/issues/61878#note_172155520) and the [issue](https://gitlab.com/gitlab-org/gitlab/issues/10871) associated with adding git-lfs support).

The rest is bad news:

1) Git lfs is a git plugin that works through git hooks. We already have a plugin that uses git hooks, namely `husky`; and it conflicts with git-lfs. There are ways of fixing that, although they are ugly. A long conversation with the comment summarizing the solution can be found here: `https://github.com/typicode/husky/issues/108`. There's also a write-up of the problem and solution on [dev.to](https://dev.to/mbelsky/pair-husky-with-git-lfs-in-your-javascript-project-2kh0). 
2) Gitlab is failing the build of this branch — it is trying to fetch git-lfs objects, but that results in an https->http redirect, which fails:

image

Seems to be a Gitlab-related issue; discussed [here](https://forum.gitlab.com/t/lfs-broke-unexpectedly/14230); no clear reason or solution suggested.


ens-ds23:
The fix in pimjansen's comment on 22 Nov 2018 looks ok to me (as modified later by  AdrienLemaire on 7 Dec 2018). It's not spetacularly neat but is basically only a few lines in a config file. On the other hand, I've not looked at the other PR yet which might be much neater.

The http/http issue seems much more worrying. I might be a go/no-go thing with this PR. Not looked into that yet, though.

ens-ds23:
If husky doesn't play well with letting things other than husky we should probably consider if something like git-hooks-plus would work. That kind of lock-in is a bit of a worry for me for a project with (hopefully) a long lifetime.

azangru:
I am frankly amazed at how hacky it becomes to use git lfs, and am terrified by the gitlab [issue](https://forum.gitlab.com/t/lfs-broke-unexpectedly/14230) which essentially says that one day everything worked fine, and the next day, out of the blue, everything broke 🙈

ens-ds23:
I think I have a fix for the LFS issue. The http/https issue was misleading, it related to our github/gitlab integration as gitlab was going to the "wrong" LFS store. I've set it to go to github and you should see that when you do a pull (new file called .gitconfig). Could you confirm if this works for you @azangru ?

ens-ds23:
@kamaldodiya is there a way of me going from a known job (eg https://gitlab.ebi.ac.uk/ensembl-web/ensembl-client/pipelines/43651 ) to a URL where I can test it? Can I find where it's deployed just by looking at the terminal output in the pipeline somewhere?

ens-ds23:
@azangru thankfully, it looks like that in this setup we won't be using gitlab's LFS storage (but githubs).

ens-ds23:
Re the git-hooks issue, I am equally unimpressed by git-lfs and husky at the moment as both seem to insist on planting their flag on the little, contested islands of .git/hooks.

I think the proposed hack in the dev.to article should be ok if it's tidied up a bit and if I can work out a way which doesn't require special actions from a user merely cloning and building (which I'm not sure whether is the case or not, yet).

azangru:
@ens-ds23 the problem is, git hooks are put in the `.git` directory, which, by its nature, sits on developer's local machine. So along comes husky and says that if you just install your npm packages (which you have to do anyway), I'll configure the local `.git` directory for you. But then git-lfs also has similar ideas. It seems it's inevitable that at least something should plant a flag on .git/hooks, because that's where git is going to look to decide whether it needs to run additional scripts or not.

ens-ds23:
#181 is a better way of doing this. git-lfs's unpredictable promiscuity with git hooks makes this a non-starter, really. Husky, by contrast, which we want to be in control of these hooks, has no way of stamping over any unasked-for mess made by git lfs.

PR #181, opened 2019-10-19
Import genome browser static files from a github repository

azangru:
### Problem
Every time the Rust code is updated and a new wasm file is generated, it is saved in our repository. It is a fairly large binary file (approx 1.5MB), and its regular saving to the repository means that the size of the repository will grow faster than we would like.

### Solution
The artifacts of Rust build (brower.js and the wasm file) are compressed into tarballs and saved to [another repository](https://github.com/Ensembl/ensembl-genome-browser-assets), from which they are pulled by npm during installation of the rest of dependencies.

Bash scripts will make sure that only relevant tarballs (i.e. with wasm files that are actually used in ensembl-client) will stay in the other repo. The rest of the tarballs will get deleted, and the wasm-containing repo will be rebased, so that the unused files can be garbage-collected by git. This should prevent the wasm-containing repo from growing too fast. 

### Checks
- works in development
- works locally with production build
- works in deployed feature branch


ens-ds23:
The production build needs updating so that:
* builds that get pushed are deployed to git and have their hashes updated.
* builds that don't get pushed are copied into node_modules.
* some kind of git hook to make sure that we never forget to do the former.

There's already approximately difference between pushed and non-pushed builds (deploy and deploy-check).

Though sometimes it's useful to do deploy builds in non-push circumstances (when you expect there to be a bug or performance hit from the debug code), it's never right to do the opposite.

Probably there needs to be a third category, "deploy-strip" which does a to-push build but deploys into node_modules.

deploy-check: [has debug symbols, unoptimised, console messages] [to node_modules]
deploy-strip: [no debug symbols, optimised, no console messages] [to node_modules]
deploy: [no debug symbols, optimised, no console messages] [to github]

Probably we can do the push to github in the build script and use the git prepush hook (via husky) to verify its presence via http. Passing the correct hash to use around this system might be fun, but we'll see.

ens-ds23:
#197 has now unblocked this PR.

ens-ds23:
OK, well that rebase messed up this PR. Still working out how to fix it.

azangru:
Sorry, our dev got broken when I updated the packages. I am trying to sort it out now; hopefully nearly there...

azangru:
Fixed tests and updated production build from the React side.

ens-ds23:
Deploy builds fixed. Issue with jest I'll look at later. May be temporary while in broken state. Should fix non-deploy builds first.

azangru:
Pushed commits that got overwritten during the last force-push. They fix the jest error and also add instructions for copying the wasm file to dist/static folder during production build.

ens-ds23:
Current status:

- [X] deploy builds
- [X] non-deploy builds
- [ ] prune script
- [ ] switch repo into ensembl-org

ens-ds23:
Pruning script finished. Ready to go.

src/ensembl/package.json
```diff
@@ -54,6 +54,7 @@
     "connected-react-router": "6.5.2",
     "core-js": "3.3.6",
     "dotenv": "8.2.0",
+    "ensembl-genome-browser": "https://github.com/Ensembl/ensembl-genome-browser-assets.git#wasm/7def28c42e83db0db5eacc70ab014add",
```
msmanoj: How does this work? Will this line get updated every time the wasm file is changed or does it always refer to the latest version of the file?

azangru: Sorry, the PR didn't have proper description when I announced it. I updated the description; hopefully it will answer this question.

azangru:
@ens-ds23 From the npm-related perspective, this looks great. Can't approve this PR, because I created it, but whenever you are happy with pruning scripts, feel free to squash-and-merge!

PR #182, opened 2019-10-21
Add textarea shared component

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-365

## Importance
Required by https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-357

## Description
Add the textarea shared component that works similar to the Inout component.

### Other effects

### DIFF
--- a/src/ensembl/src/shared/components/textarea/Textarea.scss
+++ b/src/ensembl/src/shared/components/textarea/Textarea.scss
@@ -1,11 +1,8 @@
 .textarea {
-  display: inline-block;
-  width: 100%;
   border: none;
   padding: 7px;
-  font-size: 16px;
-  border-radius: 2px;
-
+  width: 100%;
+  height: 85px; /* To display 5 lines of text */
   &:focus,
   &:active {
     outline: none;
--- a/src/ensembl/src/shared/components/textarea/Textarea.scss
+++ b/src/ensembl/src/shared/components/textarea/Textarea.scss
@@ -1,0 +1,17 @@
+.textarea {
+  display: inline-block;
+  width: 100%;
+  border: none;
+  padding: 7px;
+  font-size: 16px;
+  border-radius: 2px;
+
+  &:focus,
+  &:active {
+    outline: none;
+  }
+}
+
+.disableResize {
+  resize: none;
+}
--- a/src/ensembl/src/shared/components/textarea/Textarea.test.tsx
+++ b/src/ensembl/src/shared/components/textarea/Textarea.test.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
 import { mount, render } from 'enzyme';
-
+import faker from 'faker';
 import Textarea from './Textarea';
 
 describe('<Textarea />', () => {
   const commonTextareaProps = {
-    id: 'testId',
-    name: 'testTextareaName',
-    className: 'testTextareaClass',
+    id: faker.random.word(),
+    name: faker.random.word(),
+    className: faker.random.word(),
     onChange: jest.fn(),
     onFocus: jest.fn(),
     onBlur: jest.fn()
@@ -22,11 +22,12 @@ describe('<Textarea />', () => {
     jest.resetAllMocks();
   });
 
+  const textareaValue = faker.random.words();
+
   describe('rendering', () => {
-    const TextareaValue = 'foo';
     const props = {
       ...commonTextareaProps,
-      value: TextareaValue
+      value: textareaValue
     };
 
     test('passes relevant props to the Textarea element', () => {
@@ -50,11 +51,10 @@ describe('<Textarea />', () => {
 
   describe('responding with data', () => {
     test('passes string value to onChange', () => {
-      const TextareaValue = 'foo';
-      const changedValue = TextareaValue + '1';
+      const changedValue = faker.random.words();
       const wrappedTextarea = getWrappedTextarea({
         ...commonTextareaProps,
-        value: TextareaValue
+        value: textareaValue
       });
 
       wrappedTextarea.simulate('change', { target: { value: changedValue } });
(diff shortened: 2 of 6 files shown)


src/ensembl/src/shared/components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+        ...commonTextareaProps,
+        resizable: false
+      });
+      expect(wrappedTextarea.hasClass('disableResize')).toBe(true);
+    });
+  });
+
+  describe('responding with data', () => {
+    test('passes string value to onChange', () => {
+      const TextareaValue = 'foo';
```
azangru: Textarea -> textarea

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+      const changedValue = TextareaValue + '1';
```
azangru: faker?

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+      });
+
+      wrappedTextarea.simulate('focus');
+      expect(commonTextareaProps.onFocus).toHaveBeenLastCalledWith(
+        TextareaValue
+      );
+    });
+
+    test('passes string value to onBlur', () => {
+      const TextareaValue = 'foo';
```
imran-ebi: Camel case?

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+      });
+
+      wrappedTextarea.simulate('change', { target: { value: changedValue } });
+      expect(commonTextareaProps.onChange).toHaveBeenLastCalledWith(
+        changedValue
+      );
+    });
+
+    test('passes string value to onFocus', () => {
+      const TextareaValue = 'foo';
```
imran-ebi: Camel case?

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+      wrappedTextarea.simulate('blur');
+      expect(commonTextareaProps.onBlur).toHaveBeenLastCalledWith(
+        TextareaValue
+      );
+    });
+  });
+
+  describe('responding with events', () => {
+    test('passes event to onChange', () => {
+      const TextareaValue = 'foo';
```
imran-ebi: Same as above

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/textarea/Textarea.test.tsx
msmanoj: That's done here and in other places

src/ensembl/src/shared/components/textarea/Textarea.scss
```diff
@@ -0,0 +1,17 @@
+.textarea {
+  display: inline-block;
+  width: 100%;
```
azangru: If textarea has a width of 100%, then why is its display inline-block? Do you mean that it should behave as if it has `display:block` (and thus the two rules above are unnecessary)?

src/ensembl/src/shared/components/textarea/Textarea.scss
```diff
@@ -0,0 +1,17 @@
...
+  border: none;
+  padding: 7px;
+  font-size: 16px;
```
azangru: Default font-size of an input is 14px. Should default font-size of textarea be larger?

src/ensembl/src/shared/components/textarea/Textarea.scss
```diff
@@ -0,0 +1,17 @@
...
+  border-radius: 2px;
```
azangru: The input does not have border-radius by default. Should textarea?

src/ensembl/src/shared/components/textarea/Textarea.scss
msmanoj: These styles were taken from Input.scss. I can remove this but where there any reason as to the input was styled like this?

src/ensembl/src/shared/components/textarea/Textarea.scss
msmanoj: No, I've removed it.

src/ensembl/src/shared/components/textarea/Textarea.scss
msmanoj: No, it shouldn't. Andrea is not convinced with that so I've removed it.

src/ensembl/src/shared/components/textarea/Textarea.scss
azangru: I thought the default display property of an input element is `inline`, but now that I've checked it, it seems to be `inline-block`. So setting the `inline-block` property on the input was probably redundant.

The thinking behind the 100% width for the input was to always make it fit the width of its container. And the reason it wasn't switched to block was so that if its parent is display:inline, we won't have a block element inside of an inline element.

For textarea, I don't know whether inline-block with 100% width will offer any benefit compared to display: block.

msmanoj: I've now removed `inline-block` and left the `width:100%` as it is so that it fits the parent.

PR #156, opened 2019-09-03
Refactor ImageButtonStatus type

azangru:
## Type
- refactoring

## Description
Refactor the way we define and use the ImageButtonStatus type (currently, we use it in places where it does not communicate intent correctly).

## For discussion

[...]

PR #183, opened 2019-10-21
Extract a generic Status type, from which ImageButtonStatus is derived

azangru:
## Type
- refactoring

## Description
(alternative to https://github.com/Ensembl/ensembl-client/pull/156)

**Purpose:**
We are currently abusing ImageButtonStatus for defining statuses of things that are not directly related to the image button (e.g. browser tracks). This PR introduces a new Status type (that serves as a collection of all possible types for various things), from which other, more specific statuses, are derived.

Supersedes [this PR](https://github.com/Ensembl/ensembl-client/pull/156), which offered a different approach to solving the same problem.

**Discussion**
The unfortunate consequence of this implementation is that typescript may complain about the Status enum being too broad a type for a specific case, and thus will require some additional coaxing to type-check correctly. For example, declaring this variable:

```typescript
      const toggledTrack = {
        homo_sapiens38: {
          'Genes & transcripts': {
            'gene-pc-fwd': Status.INACTIVE
          }
        }
      };
```
and then passing it to a function that expected the 'gene-pc-fwd' field to be TrackPanelActivity type resulted in error:

image

But on the other hand, even the approach with constants mentioned in the other PR will also require the same disambiguation of the type.

### DIFF
--- a/src/ensembl/src/content/app/browser/browserActions.ts
+++ b/src/ensembl/src/content/app/browser/browserActions.ts
@@ -37,7 +37,7 @@ export type UpdateTrackStatesPayload = {
   genomeId: string;
   categoryName: string;
   trackId: string;
-  status: TrackActivityStatus; // TODO: update types so that actions do not depend on ImageButton types
+  status: TrackActivityStatus;
 };
 
 export type ParsedUrlPayload = {
--- a/src/ensembl/src/content/app/browser/browserActions.ts
+++ b/src/ensembl/src/content/app/browser/browserActions.ts
@@ -6,17 +6,13 @@ import isEqual from 'lodash/isEqual';
 
 import config from 'config';
 import * as urlFor from 'src/shared/helpers/urlHelper';
+import { getChrLocationStr } from './browserHelper';
 
 import browserMessagingService from 'src/content/app/browser/browser-messaging-service';
+import browserStorageService from './browser-storage-service';
 
 import { fetchEnsObject } from 'src/ens-object/ensObjectActions';
 
-import {
-  BrowserNavStates,
-  ChrLocation,
-  CogList,
-  ChrLocations
-} from './browserState';
 import {
   getBrowserActiveGenomeId,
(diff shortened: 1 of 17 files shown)


src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -25,18 +21,23 @@ import {
...
+import { TrackStates } from './track-panel/trackPanelConfig';
+import { RootState } from 'src/store';
+import { TrackActivityStatus } from 'src/content/app/browser/track-panel/trackPanelConfig';
+
 export type UpdateTrackStatesPayload = {
   genomeId: string;
   categoryName: string;
   trackId: string;
-  status: ImageButtonStatus; // TODO: update types so that actions do not depend on ImageButton types
+  status: TrackActivityStatus; // TODO: update types so that actions do not depend on ImageButton types
```
msmanoj: Can this comment now be removed?

src/ensembl/src/shared/types/status.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// To developer: for easier checking of which statuses are available,
+// please add new variants to the enum in the alphabetical order
+export enum Status {
+  ACTIVE = 'active',
+  DISABLED = 'disabled',
+  DEFAULT = 'default',
+  HIGHLIGHTED = 'highlighted',
+  INACTIVE = 'inactive'
+}
```
msmanoj: Do you think we should build the `LoadingState` from this enum or keep that as a separate enum on it's own?

azangru: I don’t know, to be honest. I am reluctant to put unrelated things in one place — it feels disorganized somehow. I would probably leave the loading state where it is.

src/ensembl/src/content/app/browser/browserActions.ts
azangru: Good point! :smile:

PR #185, opened 2019-10-24
Upload component

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-366

## Description
Added a shared component for file uploads that supports both click to upload as well as drag and drop to upload.

PastedGraphic-2

### Other effects

### DIFF
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -1,0 +1,18 @@
+@import 'src/styles/common';
+
+.defaultUpload {
+  border: 1px dashed $blue;
+  padding: 10px;
+  position: relative;
+  color: $blue;
+  cursor: pointer;
+}
+
+.fileInput {
+  width: 0.1px;
+  height: 0.1px;
+  opacity: 0;
+  overflow: hidden;
+  position: absolute;
+  z-index: -1;
+}
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -10,7 +10,6 @@
     display: inline-block;
     padding: 5px 10px;
     cursor: pointer;
-    border: 1px solid transparent;
   }
 }
 
@@ -19,8 +18,4 @@
   overflow: hidden;
   position: absolute;
   z-index: -1;
-
-  &:focus + label {
-    border: 1px solid $dark-grey;
-  }
 }
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -24,3 +24,7 @@
   height: 1px;
   width: 1px;
 }
+
+.defaultUpload:focus-within {
+  border: 1px dashed $black;
+}
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -1,16 +1,19 @@
 @import 'src/styles/common';
 
 .defaultUpload {
-  border: 1px dashed $blue;
+  border: 1px dashed $grey;
   color: $blue;
+  color: #33adff;
   display: inline-block;
   position: relative;
+  padding: 10px;
+  cursor: pointer;
+  transition: border 0.4s;
+}
 
-  label {
-    display: inline-block;
-    padding: 5px 10px;
-    cursor: pointer;
-  }
+.defaultUploadActive {
+  border: 1px dashed $blue;
+  transition: border 0.4s;
 }
 
 .fileInput {
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -2,17 +2,25 @@
 
 .defaultUpload {
   border: 1px dashed $blue;
-  padding: 10px;
-  position: relative;
   color: $blue;
-  cursor: pointer;
+  display: inline-block;
+  position: relative;
+
+  label {
+    display: inline-block;
+    padding: 5px 10px;
+    cursor: pointer;
+    border: 1px solid transparent;
+  }
 }
 
 .fileInput {
-  width: 0.1px;
-  height: 0.1px;
   opacity: 0;
   overflow: hidden;
   position: absolute;
   z-index: -1;
+
+  &:focus + label {
+    border: 1px solid $dark-grey;
+  }
 }
--- a/src/ensembl/src/shared/components/upload/Upload.scss
+++ b/src/ensembl/src/shared/components/upload/Upload.scss
@@ -21,4 +21,6 @@
   overflow: hidden;
   position: absolute;
   z-index: -1;
+  height: 1px;
+  width: 1px;
 }
--- a/src/ensembl/src/shared/components/upload/Upload.test.tsx
+++ b/src/ensembl/src/shared/components/upload/Upload.test.tsx
@@ -14,8 +14,7 @@ const defaultProps = {
 
 const event = { preventDefault: noop, stopPropagation: noop };
 
-const renderUpload = (props: any = defaultProps) =>
-  mount(<Upload {...props} />);
+const renderUpload = (props: any) => mount(<Upload {...props} />);
 
 const fileContents = faker.random.words();
 
@@ -23,7 +22,8 @@ const generateFile = () => {
   return new Blob([fileContents], { type: 'text/plain' });
 };
 
-const files = times(faker.random.number(10), () => generateFile());
+// const files = times(faker.random.number(10), () => generateFile());
+const files = times(10, () => generateFile());
 const addEventListener = jest.fn((_, evtHandler) => {
   evtHandler();
 });
@@ -37,12 +37,15 @@ const mockFileReader = {
 };
 
 describe('Upload', () => {
-  let wrapper: any = renderUpload();
-  windowService.getFileReader = jest.fn(() => mockFileReader) as any;
+  let wrapper: any;
+
+  beforeEach(() => {
+    wrapper = renderUpload(defaultProps);
+    windowService.getFileReader = jest.fn(() => mockFileReader) as any;
+  });
 
   afterEach(() => {
     jest.resetAllMocks();
-    windowService.getFileReader = jest.fn(() => mockFileReader) as any;
   });
 
   describe('using the input', () => {
@@ -55,18 +58,25 @@ describe('Upload', () => {
     });
 
     it('disables multiple file selection if allowMultiple is set to false', () => {
-      wrapper = renderUpload({ allowMultiple: false });
+      wrapper = renderUpload({ ...defaultProps, allowMultiple: false });
       expect(wrapper.find('input').prop('multiple')).toBeFalsy();
     });
 
-    it('calls the windowService.getFileReader when a file is selected', () => {
-      wrapper.find('input').prop('onChange')({ target: { files: [files[0]] } });
-      expect(windowService.getFileReader).toHaveBeenCalled();
-      expect(mockFileReader.readAsText).toHaveBeenCalledWith(files[0]);
+    it('calls the windowService.getFileReader for every file selected', () => {
+      wrapper.find('input').prop('onChange')({
+        ...event,
+        target: { files: files }
+      });
+
+      expect(windowService.getFileReader).toHaveBeenCalledTimes(files.length);
     });
 
     it('calls the readAsText multiple times based on the number of files', () => {
-      wrapper.find('input').prop('onChange')({ target: { files: files } });
+      wrapper.find('input').prop('onChange')({
+        ...event,
+        target: { files: files }
+      });
+
       expect(mockFileReader.readAsText).toHaveBeenCalledTimes(files.length);
     });
   });
@@ -78,16 +88,6 @@ describe('Upload', () => {
       expect(labelProps.onDragEnter).toBeTruthy();
       expect(labelProps.onDragLeave).toBeTruthy();
       expect(labelProps.onDrop).toBeTruthy();
-    });
-
-    it('calls windowService.getFileReader multiple times based on the number of files', () => {
-      const mockEvent = {
-        ...event,
-        dataTransfer: { files: files, clearData: noop }
-      };
-      wrapper.find('label').prop('onDrop')(mockEvent);
-
-      expect(windowService.getFileReader).toHaveBeenCalledTimes(files.length);
     });
 
     it('does not call the windowService.getFileReader when a file is just dragged in ', () => {
--- a/src/ensembl/src/shared/components/upload/Upload.test.tsx
+++ b/src/ensembl/src/shared/components/upload/Upload.test.tsx
@@ -4,8 +4,7 @@ import faker from 'faker';
 import times from 'lodash/times';
 import noop from 'lodash/noop';
 import Upload from './Upload';
-
-// import windowService from 'src/services/window-service';
+import windowService from 'src/services/window-service';
 
 const onChange = jest.fn();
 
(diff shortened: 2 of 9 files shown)


src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  evtHandler();
+});
+const dummyFileReader = { addEventListener, readAsText, result: fileContents };
+
+describe('Upload', () => {
+  let wrapper: any;
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    window.FileReader = jest.fn(() => dummyFileReader) as any;
```
azangru: Can we use the window-service as a proxy for window so that we don't have to mock anything on the window object itself?

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+const addEventListener = jest.fn((_, evtHandler) => {
+  evtHandler();
+});
+const dummyFileReader = { addEventListener, readAsText, result: fileContents };
+
+describe('Upload', () => {
+  let wrapper: any;
+
+  beforeEach(() => {
+    jest.resetAllMocks();
```
azangru: to afterEach?

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    window.FileReader = jest.fn(() => dummyFileReader) as any;
+    wrapper = renderUpload();
+  });
+
+  describe('using the input', () => {
+    it('always renders an input of type file', () => {
+      expect(wrapper.find('input')).toHaveLength(1);
```
azangru: description says "of type 'file'"; code isn't checking that

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    it('allows multiple files to be selected by default', () => {
+      expect(wrapper.find('input').prop('multiple')).toBeTruthy();
+    });
+
+    it('disabled multiple file selection if allowMultiple is set to false', () => {
```
azangru: disabled -> disables?

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+      wrapper = renderUpload({ allowMultiple: false });
+      expect(wrapper.find('input').prop('multiple')).toBeFalsy();
+    });
+
+    it('calls the FileReader when a file is selected', () => {
+      wrapper.find('input').prop('onChange')({ target: { files: [files[0]] } });
+      expect(FileReader).toHaveBeenCalled();
+      expect(readAsText).toHaveBeenCalledWith(files[0]);
+    });
```
azangru: what happens with multiple files?

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: I can only understand a part of it. Can you please point me to an example?

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: I guess I removed it by mistake as I remember adding it 😞

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: Added another test to check that

src/ensembl/src/shared/components/upload/Upload.test.tsx
azangru: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/services/storage-service.ts#L37-L38

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import faker from 'faker';
+import times from 'lodash/times';
+import noop from 'lodash/noop';
+import Upload from './Upload';
+
+// import windowService from 'src/services/window-service';
```
azangru: that would go to Upload.tsx

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      expect(spanProps.onDragLeave).toBeTruthy();
+      expect(spanProps.onDrop).toBeTruthy();
+    });
+
+    it('calls FileReader when a file is dropped ', () => {
+      const mockEvent = {
+        ...event,
+        dataTransfer: { files: files, clearData: noop }
+      };
+      wrapper.find('span').prop('onDrop')(mockEvent);
```
azangru: would this work?

```
const mockDropEvent = new CustomEvent("drop", {
  detail: {
    dataTransfer: ...something
  }
});
wrapper.find('span').getDOMNode().dispatchEvent(mockDropEvent)?
```

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      expect(FileReader).toHaveBeenCalled();
+    });
+
+    it('calls FileReader multiple times based on the number of files', () => {
```
azangru: with this test, the previous one is unnecessary ;-) One is the special case of many.

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -0,0 +1,167 @@
...
+      />
+      <label htmlFor={props.id}>{props.label}</label>
+    </span>
+  );
+};
+
+Upload.defaultProps = {
+  callbackWithFiles: false,
+  allowMultiple: true,
+  id: 'file',
```
azangru: erm, why would we have default id? that means we can have multiple inputs with the same id on the page, and this goes against html spec

azangru:
Let’s talk about the drag-and-drop bit.
- Is it something that Andrea actually wanted us to have? The drop area is intolerably small; usually interfaces that allow users to drag-and-drop files in the drop area will have a pretty large drop area, so that it will be difficult to miss.
- If we really do want the drag-and-drop, shouldn’t it be an option on this component (upload component as a button vs upload component as a drag-and-drop area vs some combination of the two)?

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -0,0 +1,167 @@
...
+      }
+
+      const { files } = e.dataTransfer;
+
+      forEach(files, (file) => {
+        const fileReader = new FileReader();
+        fileReaders.push(fileReader);
+        totalPendingFilesToRead++;
+        fileReader.onloadend = handleFileRead;
+        fileReader.readAsText(file);
```
azangru: That’s an assumption, isn’t it? Should the type of the file be controlled through component’s props?

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -0,0 +1,167 @@
...
+          props.onChange(e.dataTransfer.files[0]);
+          return;
+        }
+        props.onChange(e.dataTransfer.files);
+        return;
+      }
+
+      const { files } = e.dataTransfer;
+
+      forEach(files, (file) => {
```
azangru: Why is the code using lodash's forEach instead of array’s native forEach? Does the `files` object not have a forEach? If not, can it be spread in an array (`[...files].forEach`)?

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: True. Removed it 👍

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: I need a default id that can be used in the `htmlFor` attribute of the `label` tag. I guess I will have to generate a unique ID for every Upload component?

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: I've made some changes to use the window-service but can you check if that's right, please?

src/ensembl/src/shared/components/upload/Upload.tsx
azangru: You are wrapping your input in a label. I believe you don't even need the htmlFor in this case — the click on a label will toggle the input by itself.

src/ensembl/stories/shared-components/upload/Upload.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+      action(`Files: ${filesOrContent}`)();
+      return;
+    }
+
+    action(`Content: ${filesOrContent}`)();
+  };
+
+  return (
+    <div className={styles.defaultWrapper}>
+      <div></div>
```
azangru: why the empty div?

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: `[...files].forEach` works 👍

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: Made some changes to support other types.

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: Yes, that's right. I've removed it.

src/ensembl/stories/shared-components/upload/Upload.stories.tsx
msmanoj: It's not required. I've removed it.

msmanoj:
> * Is it something that Andrea actually wanted us to have? The drop area now is almost unusably small; usually interfaces that allow users to drag-and-drop files in the drop area will have a pretty large drop area, so that it will be difficult to miss.

It was not in the initial design but I suggested it to Andrea by showing an example from ant designs and she was happy to add it. 

> * If we really do want the drag-and-drop, shouldn’t it be an option on this component (upload component as a button vs upload component as a drag-and-drop area vs some combination of the two)?

I asked the same to Andrea and she mentioned that there are no places at the moment where we wouldn't be wanting to support drag and drop. I am not sure if it is worthwhile adding such option now or later when it is required.

azangru:
> I am not sure if it is worthwhile adding such option now or later when it is required.

I am a bit confused — do you mean whether it's worthwhile adding drag-and-drop or whether it's worthwhile making drag-and-drop optional?

From what I understood from Andrea, she was perfectly happy for us not to have drag-and-drop at all until we want it.

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+      const { files } = e.dataTransfer;
+
+      [...files].forEach((file) => {
+        const fileReader = windowService.getFileReader();
+        fileReaders.push(fileReader);
+        totalPendingFilesToRead++;
+        fileReader.onloadend = handleFileRead;
+        fileReader[props.fileReaderMethod](file);
+      });
```
azangru: Two questions regarding this:

1) Technical — wouldn't it be nicer, instead of having the totalPendingFilesToRead and fileReaders variables floating around and being mutated by various functions, to re-write this bit using promises:

```
const readPromises = [...files].map((file) => {
  return new Promise((resolve, reject) => {
    const fileReader = windowService.getFileReader();
    fileReaders.push(fileReader);
    totalPendingFilesToRead++;
    fileReader.onloadend = resolve(fileReader);
    fileReader[props.fileReaderMethod](file);

    // do we handle error?
  });
});

Promise.all(readPromises).then(handleFiles); // <- this will pass all filereaders to the handleFiles function
```

2) How should the code behave if not all files have been successfully read? Your implementation will probably just hang there (since `totalPendingFilesToRead` will be greater than 0; and the above implementation with Promise.all will abort all promises (meaning that for the promise implementation each individual promise going into Promise all should have an individual catch).

msmanoj:
I meant whether it's worthwhile making drag-and-drop optional for now.

> From what I understood from Andrea, she was perfectly happy for us not to have drag-and-drop at all until we want it.

I agree that we didn't think about having a drag and drop feature initially. As I started working on the upload I asked Andrea if drag and drop is something she would like to incorporate and showed her some designs. She had a look and was happy to proceed further with it.

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: > re-write this bit using promises

I will look into implementing it using promises.

> How should the code behave if not all files have been successfully read? 
 I don't think it will hang because `onloadend` will be called even if there is a failure. I am still not sure what should happen when there is a failure as there are many scenarios we need to consider. 

I am also not sure how likely this is to happen but some the possible solutions would be:
1. Display an error below the drop-zone mentioning the file name and cancel the process.
2. Proceed with the remaining files and rely on the users to validate if all the file contents are read.
3.  Add an `error` parameter to the onChange callback and get the errors handled by the parent. (I would prefer this)

msmanoj: > do we handle error?

I am starting to think if we really need error handling as the chances are very less for an error to occur while selecting a file.

Do you know any way this error can be triggered?

src/ensembl/src/shared/components/upload/Upload.scss
```diff
@@ -21,4 +21,6 @@
   overflow: hidden;
   position: absolute;
   z-index: -1;
+  height: 1px;
+  width: 1px;
```
azangru: Out of curiosity, will the simple `display: none` do the job? Then there would be no opacity/z-index/height/width nonsense. Judging by [this example on StackOverflow](https://stackoverflow.com/a/29251249/3925302), it should work (we also have an input inside a label).

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -14,16 +14,16 @@ const defaultProps = {
...
+const renderUpload = (props: any) => mount(<Upload {...props} />);
 
 const fileContents = faker.random.words();
 
 const generateFile = () => {
   return new Blob([fileContents], { type: 'text/plain' });
 };
 
-const files = times(faker.random.number(10), () => generateFile());
+// const files = times(faker.random.number(10), () => generateFile());
```
azangru: 🚫

src/ensembl/src/shared/components/upload/Upload.scss
msmanoj: I was following this article here:
https://tympanus.net/codrops/2015/09/15/styling-customizing-file-inputs-smart-way/

It says `the input will be excluded out of tab order` if we use display: none;

src/ensembl/src/shared/components/upload/Upload.test.tsx
msmanoj: Removed the commented line

src/ensembl/src/shared/components/upload/Upload.tsx
```diff
@@ -0,0 +1,187 @@
...
+      }).then(() => {
+        readFiles.push({
+          filename: file.name,
+          content: fileReader.result,
+          error: fileReader.error
+            ? getFileReaderErrorMessage(fileReader.error.code)
+            : null
+        });
+      });
+    });
```
azangru: Still would like to push further against the mutable `readFiles` array, which now seems entirely unnecessary:

```
    const promises = filesToRead.map((file) => {
      const fileReader: FileReader = windowService.getFileReader();
      return new Promise((resolve) => {
        fileReader.onload = resolve;
        fileReader.onerror = resolve;
        fileReader[props.fileReaderMethod](file);
      }).then(() => ({
          filename: file.name,
          content: fileReader.result,
          error: fileReader.error
            ? getFileReaderErrorMessage(fileReader.error.code)
            : null
        }));
      });

    const results = await Promise.all(promises);
    props.onChange(readFiles));
```

poof, no `readFiles`!

src/ensembl/src/shared/components/upload/Upload.scss
azangru: Good point!

We may probably want to signal that the button is focussed then, because I tried focussing in Storybook, and there was no visual feedback that I am now focussed on the Upload button.

There's a handy `:focus-within` pseudoclass that will target the label that has focussed input inside of it, as in our case. The support is [not perfect](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within), but is getting there.

src/ensembl/src/shared/components/upload/Upload.tsx
msmanoj: That's done

PR #186, opened 2019-10-24
Remove RobotstxtPlugin from webpack

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-376

## Description
Updated the webpack configuration to create the robots.txt file in the path `dist/` instead of `dist/static/`

## Views affected
/robots.txt & /static/robots.txt

### DIFF
--- a/src/ensembl/robots-txt.config.js
+++ b/src/ensembl/robots-txt.config.js
@@ -1,4 +1,4 @@
 module.exports = {
   policy: [{ userAgent: '*', disallow: '/' }],
-  host: 'http://2020.ensembl.org'
+  host: process.env.DEPLOYENV === 'prod' ? '2020.ensembl.org' : process.env.DEPLOYENV + '-2020.ensembl.org'
 };
--- a/src/ensembl/robots-txt.config.js
+++ b/src/ensembl/robots-txt.config.js
@@ -1,4 +1,3 @@
 module.exports = {
-  policy: [{ userAgent: '*', disallow: '/' }],
-  host: ( (!process.env.DEPLOYENV || process.env.DEPLOYENV === 'prod') ? '' : process.env.DEPLOYENV + '-') + '2020.ensembl.org'
+  policy: [{ userAgent: '*', disallow: '/' }]
 };
--- a/src/ensembl/robots-txt.config.js
+++ b/src/ensembl/robots-txt.config.js
@@ -1,4 +1,4 @@
 module.exports = {
   policy: [{ userAgent: '*', disallow: '/' }],
-  host: process.env.DEPLOYENV === 'prod' ? '2020.ensembl.org' : process.env.DEPLOYENV + '-2020.ensembl.org'
+  host: (process.env.DEPLOYENV === 'prod' ? '' : process.env.DEPLOYENV + '-') + '2020.ensembl.org'
 };
--- a/src/ensembl/robots-txt.config.js
+++ b/src/ensembl/robots-txt.config.js
@@ -1,4 +1,4 @@
 module.exports = {
   policy: [{ userAgent: '*', disallow: '/' }],
-  host: (process.env.DEPLOYENV === 'prod' ? '' : process.env.DEPLOYENV + '-') + '2020.ensembl.org'
+  host: ( (!process.env.DEPLOYENV || process.env.DEPLOYENV === 'prod') ? '' : process.env.DEPLOYENV + '-') + '2020.ensembl.org'
 };
--- a/src/ensembl/webpack/webpack.config.prod.js
+++ b/src/ensembl/webpack/webpack.config.prod.js
@@ -111,7 +111,9 @@ const plugins = [
     }]
   }),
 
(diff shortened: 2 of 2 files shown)


azangru:
Out of curiosity, what's the benefit of having it as a webpack plugin as opposed to the actual txt file?

msmanoj:
> Out of curiosity, what's the benefit of having it as a webpack plugin as opposed to the actual txt file?

@azangru Do you mean serving the txt file from the backend?

@imran-ebi, @kamaldodiya have you got anything to input on this?

azangru:
> @azangru Do you mean serving the txt file from the backend?

As discussed, I think robots.txt should be frontend's concern; I just don't know whether there is any point in generating it through webpack as opposed to just writing it as a txt file and copying it to /dist during build.

The argument **for** using webpack plugin may be that we can then have some programmatic control over the contents of robots.txt (since the plugin will execute a javascript file). The argument **against** is that we don't need any programmatic control at the moment.

msmanoj:
> I guess we can leave it as it is now, since this is the easiest. I'll take another look at the txt output it generates in a bit.

Sorry, I missed that bit. I have made some changes to webpack that just copies the robots.txt file in to the dist directory. I guess we can get the plugin added later on if we needed programatic control over it.

src/ensembl/robots.txt
```diff
@@ -0,0 +1,3 @@
+User-agent: *
+Disallow: /
+Host: 2020.ensembl.org
```
azangru: Ok, I see that we already need the programmatic control over the generation of robots.txt.

The robots.txt on the 2020 site should say that the host is `2020.ensembl.org`, while the robots.txt on the staging-2020 site should say that the host is `staging-2020.ensembl.org`.

azangru:
So I guess we'll need to add another environment variable to be able to tell staging apart from production here: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/.env.example. 
And then use it in robots-txt.config to dynamically modify the host name.

Or perhaps just read the environment variable from gitlab environment (is it going to be DEPLOYENV, as set [here](https://github.com/Ensembl/ensembl-client/blob/dev/.gitlab-ci.yml#L116)?) from within the robots-txt.config

imran-ebi:
Sorry, I thought there weren't any changes required in this PR anymore and approved it. So we'll need this one change @azangru has said and then can merge?

msmanoj:
@azangru  If I use `process.env.DEPLOYENV`, can it be tested on dev?

src/ensembl/robots-txt.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   policy: [{ userAgent: '*', disallow: '/' }],
-  host: 'http://2020.ensembl.org'
+  host: process.env.DEPLOYENV === 'prod' ? '2020.ensembl.org' : process.env.DEPLOYENV + '-2020.ensembl.org'
```
azangru: probably should account for undefined as well? ;-)

azangru:
>  tested on dev?
Sorry, do you mean on local machine, on feature branch deployment or on deployment of the dev branch?

If you mean on local machine, then I would think setting DEPLOYENV=blah in your shell environment before running the build script should do it?

src/ensembl/robots-txt.config.js
msmanoj: According to this line
  - case "${CI_COMMIT_REF_NAME}" in dev) DEPLOYENV="staging" ;; master) DEPLOYENV="prod" ;; cicd-test) DEPLOYENV="staging" ;; *) DEPLOYENV="dev" ;; esac
it is

msmanoj:
Sorry, I meant on local machine. I tried setting the DEPLOYENV but still it returns null.

src/ensembl/robots-txt.config.js
msmanoj: `probably should account for undefined as well?`

I've done that now so that the default is 2020.ensembl.org

azangru:
> I tried setting the DEPLOYENV 

`export DEPLOYENV=foo` ?

azangru:
Hmm, I see what you mean — process.env.DEPLOYENV is defined, but the output contains host which is null

azangru:
Here's what I am seeing in the console during debugging:

image

lots and lots of calls to the robots config file; the env variable is read correctly

but then the whole build ends in a segfault:

image

azangru:
Ah!

The reason you are getting null for the host seems to be because you don't specify the protocol. This worked for me:

```
module.exports = {
  policy: [{ userAgent: '*', disallow: '/' }],
  host: process.env.DEPLOYENV === 'prod' ? 'http://2020.ensembl.org' :  `http://${process.env.DEPLOYENV}-2020.ensembl.org`
};
```

msmanoj:
I've removed the host now. @azangru thanks for helping with the debugging.

PR #187, opened 2019-10-30
Change default track config state for track labels to true

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-388

## Importance
Sprint 2019-6

## Description
The initial state of the labels and names field within the track config should be set to true instead of false.

## Views affected
Browser -> Track Config

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
+++ b/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
@@ -61,8 +61,7 @@ const BrowserTrackConfig: FunctionComponent<BrowserTrackConfigProps> = (
 
   const selectedCog = props.selectedCog || '';
 
-  const shouldShowTrackName =
-    trackConfigNames[selectedCog] !== false ? true : false;
+  const shouldShowTrackName = trackConfigNames[selectedCog] || false;
   const shouldShowTrackLabels =
     trackConfigLabel[selectedCog] !== false ? true : false;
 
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -61,8 +61,10 @@ const BrowserTrackConfig: FunctionComponent<BrowserTrackConfigProps> = (
 
   const selectedCog = props.selectedCog || '';
 
-  const shouldShowTrackName = trackConfigNames[selectedCog] || false;
-  const shouldShowTrackLabels = trackConfigLabel[selectedCog] || false;
+  const shouldShowTrackName =
+    trackConfigNames[selectedCog] !== false ? true : false;
+  const shouldShowTrackLabels =
+    trackConfigLabel[selectedCog] !== false ? true : false;
```
azangru: How is this different from `trackConfigLabel[selectedCog] || false`? Is `trackConfigLabel[selectedCog]` a boolean?

msmanoj: We need to consider `undefined` as true.  `trackConfigLabel[selectedCog]` will be empty by default.

msmanoj: No we can't do that as `undefined || true`  and `false || true` will always return true.

PR #189, opened 2019-10-30
Add Chromosome navigator (mini-map)

azangru:
## Type
- New feature

## Related JIRA Issue(s)
ENSWBSITES-213

## Description
This PR is for the thing (which goes under different names: Andrea calls it chromosome idiogram; this code calls it ChromosomeNavigator; and I am toying with the idea of naming it GenomeBrowserMiniMap) that will appear in the navigation panel above the genome browser image.

**Design:** https://xd.adobe.com/view/dcb1bb36-754c-436a-5eb1-e608099ed2d9-aa80/?fullscreen

This visualisation will show:
- the "stick" (chromosome) that user is viewing, represented by a grey rectangle
- the viewport (i.e. what the BrowserImage shows), represented by a white rectangle with square brackets on both sides
- the focus object, represented by one or two orange pointers with labels indicating the start and the end positions of the object (if the object is too small, the pointers will merge into one)
- the centromere, if data is available (most likely, only for humans):

image
 
It also accounts for circular chromosomes, where the start position of a viewport (or of a focus object) can be a larger number than its end position.

Link to Storybook: https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-genome-browser-chromosomenavigator--default

## Views affected
Navigation panel in the Genome Browser page


src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.scss
```diff
@@ -0,0 +1,45 @@
...
+    fill: $light-grey;
+    stroke: $light-grey;
+  }
+
+  .centromereCentre {
+    fill: $dark-grey;
+  }
+
+  .label {
+    font-family: 'IBM Plex Mono';
```
imran-ebi: Can we add variables for the font family too? So next time the font family changes, we don't need to change them everywhere.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.scss
```diff
@@ -0,0 +1,45 @@
...
+    fill: white;
+  }
+
+  .viewportBorder {
+    fill: none;
+    stroke: $dark-grey;
+  }
+
+  .focusPointer {
+    fill: #ff9800; // FIXME: variable?
```
imran-ebi: Is this going to be added as a variable to the colour palette variables we have in settings?

src/ensembl/src/content/app/browser/chromosome-navigator/chromosomeNavigatorHelper.ts
```diff
@@ -0,0 +1,376 @@
...
+      points: arrowheadPoints
+    },
+    line: lineStyles
+  };
+};
+
+const getCentromereStyles = (params: StyleCalculatorParams, scale: Scale) => {
+  if (!params.centromere) {
+    return null;
+  } else {
```
imran-ebi: Is an `else` branch necessary here?

src/ensembl/src/content/app/browser/chromosome-navigator/chromosomeNavigatorHelper.ts
```diff
@@ -0,0 +1,376 @@
...
+  viewportEndX = scale(params.viewportEnd);
+  if (viewportEndX < containerWidth - fullBracketsWidth) {
+    // we have not reached the end of the chromosome, and can calculate positions of the brackets left-to-right
+    viewportStartX = scale(params.viewportStart);
+    viewportEndX =
+      viewportEndX >= viewportStartX
+        ? Math.max(viewportEndX, viewportStartX + fullBracketsWidth)
+        : viewportEndX;
+  } else {
+    // we are at the right end of the chromosome, and shoudl calculate positions of the brackets right-to-left
```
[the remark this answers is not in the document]
azangru: ✔️

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.scss
azangru: Added orange to our colour palette

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.scss
azangru: ✔️

src/ensembl/src/content/app/browser/chromosome-navigator/chromosomeNavigatorHelper.ts
azangru: Removed it.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx
```diff
@@ -0,0 +1,388 @@
...
+        const props = { ...minimalProps, viewportStart, viewportEnd };
+        const { openingBracket, closingBracket } = getRenderedViewport(props);
+
+        // we expect the viewport’s end to be in the end of the chromosome,
+        // and the viewport’s start to be two bracket widths to the left of viewport end
+        const expectedEndX = scalingFactor * props.length;
+        const expectedStartX =
+          expectedEndX - 2 * constants.VIEWPORT_BRACKET_BAR_WIDTH;
+        assertBracketX(openingBracket, expectedStartX);
+        assertBracketX(closingBracket, expectedEndX);
```
msmanoj: What does `x` represent in here and other places?

azangru: X-coordinate on the xy coordinate system.

msmanoj: Would it be better to use `position` instead of `X` as we do not use `Y` at all?

`assertBracketX` -> `assertBracketPosition`
`expectedStartX` -> `expectedStartPosition`
`expectedLabel1X` -> `expectedLabel1Position`
`expectedPointerX` -> `expectedPointerPosition`

azangru: 🤷‍♂ 

one might argue that position implies both x and y, and what's being checked in the test is only the x-coordinate

msmanoj: Would it not cause some confusion if we want to use the slider vertically?

azangru: Not sure what you mean. If you mean that we might want to take the whole stick and draw it vertically rather than horizontally, then the current implementation won't work (unless we first draw the svg horizontally and then do css transform: rotate(90deg) on it. But drawing the stick vertically was never a requirement.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx
```diff
@@ -0,0 +1,388 @@
...
+      const mockLabelWidth = 40;
+      const mockMeasure: any = () => ({ width: mockLabelWidth });
+      jest.spyOn(textHelpers, 'measureText').mockImplementation(mockMeasure);
+
+      const focusPosition = 20000;
+      const props = {
+        ...minimalProps,
+        focusRegion: {
+          start: focusPosition,
+          end: focusPosition
```
msmanoj: Is the position here same as `X`?

PR #190, opened 2019-10-31
Boundary marker for location focused views

ens-ds23:
## Type

- [X] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
ENSWBSITES-116

## Importance
Cosmetic but prominent.

## Description
Focus regions should not have flanks. Now scaling is done correctly, they no longer do.

NOTE: before flanks can be seen to be removed the corresponding PR will need to be merged into server. Do the client PR (ie this one) first, and verify nothing has broken (nor changed). Only then merge in the server PR. Once that's done, you will see the combined effect. Doing this in the wrong order will make the red dotted lines disappear entirely until both are merged.

## Views affected
Genome browser.

### Other effects

## Possible complications
This code is reasonably low level, affecting co-ordinates. Any likely bugs will be around things like URL not corresponding to bookmark location or to the nav location or something like that. I've checked everywhere I've thought of.


src/assets/browser/app/src/model/stage/position.rs
```diff
@@ -118,6 +123,22 @@ impl Position {
...
+        /*
+        inner centre is at self.pos.0
+        inner left is at self.pos.0 - bp_per_sc/2
+        inner right is at self.pos.0 + bp_per_sc/2
+        outer left is at self.pos.0 - bp_per_sc/2 - left_bp
+        outer right is at self.pos.0 + bp_per_sc/2 + right_bp
+        outer middle is at mean, ie (self.pos.0 - bp_per_sc/2 - left_bp + self.pos.0 + bp_per_sc/2 + right_bp)/2
+        so outer middle is at self.pos.0 + (right_bp-left_bp)/2
+        correction is to add (right_bp-left_bp)/2
+        */
```
azangru: Is the commented out code (and the XXX in the file above) left intentionally or forgotten?

PR #195, opened 2019-11-04
Browser Nav bar integration

azangru:
## Type
- New feature
- Refactoring

## Related JIRA Issue(s)
ENSWBSITES-403

### New functionality
Chromosome navigator component is included in browser navigation bar, alongside the fields that change browser location.

### Code refactoring
The composition of BrowserNavBar component has changed:

**before:**
Untitled Diagram

**after:**
Untitled Diagram (1)

### Styles
Directions from Andrea:

1) Region switcher elements should be aligned to the right, next to the close icon
Untitled Diagram

2) Chromosome idiogram should have a max-width (say, 1400px) and be positioned in the centre of the available space
Untitled Diagram (1)

## Views affected
Genome browser page

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
@@ -4,7 +4,7 @@ import { render } from 'enzyme';
 jest.mock('./BrowserNavBarControls', () => () => (
   <div>BrowserNavBarControls</div>
 ));
-jest.mock('./BrowserNavBarMain', () => () => <div>BrowserNavBarControls</div>);
+jest.mock('./BrowserNavBarMain', () => () => <div>BrowserNavBarMain</div>);
 
 import { BrowserNavBar } from './BrowserNavBar';
 
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
@@ -1,46 +1,21 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render } from 'enzyme';
+
+jest.mock('./BrowserNavBarControls', () => () => (
+  <div>BrowserNavBarControls</div>
+));
+jest.mock('./BrowserNavBarMain', () => () => <div>BrowserNavBarControls</div>);
 
 import { BrowserNavBar } from './BrowserNavBar';
-import { BrowserNavStates, ChrLocation } from '../browserState';
-
-import styles from './BrowserNavBar.scss';
-
-const browserStates = [...Array(6)].map(() => false);
-const chrLocation: ChrLocation = ['13', 1, 114364328];
-const toggleRegionEditorActive: any = jest.fn();
-const toggleRegionFieldActive: any = jest.fn();
 
 describe('<BrowserNavBar />', () => {
-  test('renders with appropriate classes', () => {
-    expect(
-      shallow(
-        <BrowserNavBar
-          browserNavStates={browserStates as BrowserNavStates}
-          chrLocation={chrLocation}
-          genomeKaryotype={[]}
-          isTrackPanelOpened={true}
-          regionEditorActive={true}
-          regionFieldActive={false}
-          toggleRegionEditorActive={toggleRegionEditorActive}
-          toggleRegionFieldActive={toggleRegionFieldActive}
-        />
-      ).hasClass(styles.browserNavBarExpanded)
-    ).toBe(false);
+  describe('rendering', () => {
+    it('correctly interprets the "expanded" prop', () => {
+      const contractedBar = render(<BrowserNavBar expanded={false} />);
+      expect(contractedBar.hasClass('browserNavBarExpanded')).toBe(false);
 
-    expect(
-      shallow(
-        <BrowserNavBar
-          browserNavStates={browserStates as BrowserNavStates}
-          chrLocation={chrLocation}
-          genomeKaryotype={[]}
-          isTrackPanelOpened={false}
-          regionEditorActive={true}
-          regionFieldActive={false}
-          toggleRegionEditorActive={toggleRegionEditorActive}
-          toggleRegionFieldActive={toggleRegionFieldActive}
-        />
-      ).hasClass(styles.browserNavBarExpanded)
-    ).toBe(true);
+      const expandedBar = render(<BrowserNavBar expanded={true} />);
+      expect(expandedBar.hasClass('browserNavBarExpanded')).toBe(true);
+    });
   });
 });
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import faker from 'faker';
+import times from 'lodash/times';
+
+import BrowserNavIcon from './BrowserNavIcon';
+
+import { BrowserNavBarControls } from './BrowserNavBarControls';
+import { BrowserNavStates } from '../browserState';
+
+jest.mock('./BrowserNavIcon', () => () => <div>BrowserNavIcon</div>);
+
+const browserNavStates = times(6, faker.random.boolean) as BrowserNavStates;
+
+describe('BrowserNavBarControls', () => {
+  it('renders without errors', () => {
+    expect(() =>
+      mount(
+        <BrowserNavBarControls
+          browserNavStates={browserNavStates}
+          disabled={faker.random.boolean()}
+        />
+      )
+    ).not.toThrow();
+  });
+
+  it('disables all buttons if the disabled prop is set to true', () => {
+    const wrapper = mount(
+      <BrowserNavBarControls
+        browserNavStates={browserNavStates}
+        disabled={true}
+      />
+    );
+    const controlButtons = wrapper.find(BrowserNavIcon);
+    expect(controlButtons.length).toEqual(browserNavStates.length);
+    controlButtons.forEach((button) => {
+      expect(button.props().enabled).toBe(false);
+    });
+  });
+
+  it('disables buttons if corresponding actions are not possible', () => {
+    // browserNavStates are an array of booleans that indicate whether the button
+    // has already caused maximum corresponding effect, and will have no further effect if pressed
+    const wrapper = mount(
+      <BrowserNavBarControls
+        browserNavStates={browserNavStates}
+        disabled={false}
+      />
+    );
+    const controlButtons = wrapper.find(BrowserNavIcon);
+    controlButtons.forEach((button, index) => {
+      const hasCausedMaximumEffect = browserNavStates[index];
+      expect(button.props().enabled).toBe(!hasCausedMaximumEffect);
+    });
+  });
+});
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
@@ -34,7 +34,7 @@ describe('BrowserNavBarControls', () => {
     const controlButtons = wrapper.find(BrowserNavIcon);
     expect(controlButtons.length).toEqual(browserNavStates.length);
     controlButtons.forEach((button) => {
-      expect(button.props().enabled).toBe(false);
+      expect(button.prop('enabled')).toBe(false);
     });
   });
 
@@ -50,7 +50,7 @@ describe('BrowserNavBarControls', () => {
     const controlButtons = wrapper.find(BrowserNavIcon);
     controlButtons.forEach((button, index) => {
       const hasCausedMaximumEffect = browserNavStates[index];
-      expect(button.props().enabled).toBe(!hasCausedMaximumEffect);
+      expect(button.prop('enabled')).toBe(!hasCausedMaximumEffect);
     });
   });
 });
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import BrowserNavIcon from './BrowserNavIcon';
+
+import { RootState } from 'src/store';
+import { browserNavConfig, BrowserNavItem } from '../browserConfig';
+import {
+  getBrowserNavStates,
+  getRegionEditorActive,
+  getRegionFieldActive
+} from '../browserSelectors';
+import { BrowserNavStates } from '../browserState';
+
+type Props = {
+  browserNavStates: BrowserNavStates;
+  disabled: boolean;
+};
+
+export const BrowserNavBarControls = (props: Props) => {
+  const shouldEnableButton = (index: number) => {
+    const { browserNavStates, disabled } = props;
+    const isAtMaximum = browserNavStates[index];
+
+    return !disabled && !isAtMaximum;
+  };
+
+  return (
+    <div>
+      {browserNavConfig.map((item: BrowserNavItem, index: number) => (
+        <BrowserNavIcon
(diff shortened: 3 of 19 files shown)


src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
```diff
@@ -1,46 +1,21 @@
...
+import { render } from 'enzyme';
 
-import { BrowserNavBar } from './BrowserNavBar';
-import { BrowserNavStates, ChrLocation } from '../browserState';
-
-import styles from './BrowserNavBar.scss';
+jest.mock('./BrowserNavBarControls', () => () => (
+  <div>BrowserNavBarControls</div>
+));
+jest.mock('./BrowserNavBarMain', () => () => <div>BrowserNavBarControls</div>);
```
msmanoj: Is this supposed to be `<div> BrowserNavBarMain </div>`?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const wrapper = mount(
+      <BrowserNavBarControls
+        browserNavStates={browserNavStates}
+        disabled={true}
+      />
+    );
+    const controlButtons = wrapper.find(BrowserNavIcon);
+    expect(controlButtons.length).toEqual(browserNavStates.length);
+    controlButtons.forEach((button) => {
+      expect(button.props().enabled).toBe(false);
```
msmanoj: Wouldn't it be better to just get the only prop we want like this:
`button.prop('enabled')`?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.scss
```diff
@@ -0,0 +1,41 @@
...
+  color: $blue;
+  cursor: pointer;
+
+  svg {
+    height: 20px;
+  }
+}
+
+.contentSwitcherClose {
+  font-size: 0;
```
msmanoj: Can we use something like `line-height: 10px;` instead of `font-size` as there is no text?

azangru: I would argue that `font-size` is better: it removes all `line-height`-based calculations from an inline-block element, and provides the developer with a container that has a predictable height, which is controlled solely by the height of its content.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+      expect(wrapper.find(BrowserNavBarRegionSwitcher).length).toBe(0);
+    });
+
+    it('renders RegionSwitcher when user clicks on Change', () => {
+      const wrapper = mount(<BrowserNavBarMain />);
+      const changeButton = wrapper.find('.contentSwitcher');
+      act(() => {
+        changeButton.simulate('click');
+      });
+      wrapper.update();
```
msmanoj: Here and below, why do we need `act` and `wrapper.update` ?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export enum Content {
+  CHROMOSOME,
+  REGION_SWITCHER
+}
+
+const BrowserNavBarMain = () => {
+  const [view, changeView] = useState<Content>(Content.CHROMOSOME);
+
+  const onChangeViewClick = (newView: Content) => {
```
msmanoj: Can this be renamed to something like `onSwitchHandler`? 
Having `Click` after `View` sounds a bit strange.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const props = {
+  toggleRegionEditorActive: jest.fn(),
+  toggleRegionFieldActive: jest.fn()
+};
+
+describe('BrowserNavBarRegionSwitcher', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
```
msmanoj: I don't think this is necessary here.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.tsx
```diff
@@ -1,12 +1,25 @@
 // TODO: think about the name; perhaps something like MiniMap is a better option?
```
msmanoj: Please check if we need to retain this comment.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.tsx
```diff
@@ -100,4 +113,34 @@ export const ChromosomeNavigator = (props: ChromosomeNavigatorProps) => {
...
+  let centromere = null;
+  if (chromosomeName && genomeId && genomeId.startsWith('homo_sapiens')) {
+    centromere = centromeres.humanCentromeres[chromosomeName] || null;
+  }
+
+  return {
+    length,
+    viewportStart: start,
+    viewportEnd: end,
+    focusRegion,
```
msmanoj: `focusRegion` is unused in this file.

src/ensembl/src/shared/data/centromeres.ts
```diff
@@ -0,0 +1,115 @@
+/*
```
msmanoj: Is this file going to be shared across apps?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
azangru: The "calls cleanup functions on unmount" test calls `props.toggleRegionEditorActive` and `props.toggleRegionFieldActive`. If at any point we add another test that checks anything related to those functions, they will have already been called, and this may interfere with test results.

Example: delete the afterEach function and add another "calls cleanup functions on unmount" test after the first one. The assertions `expect(props.toggleRegionEditorActive).not.toHaveBeenCalled();` and `expect(props.toggleRegionFieldActive).not.toHaveBeenCalled();` will now fail, because these functions will have been called in the previous test.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
azangru: A click on the changeButton will update component's state.

When react got updated to version 16.8, it actively complained when code causing state updates was not wrapped in an `act`, warning that `When testing, code that causes React state updates should be wrapped into act(...)` (example: `https://github.com/facebook/react/issues/14769`).

I have tried to remove both `act` and `wrapper.update`, and then re-ran the test. It passed, and I could see by using logging that react component managed to get correctly re-rendered. Investigating the reason for this, I found that Enzyme now wraps all calls to `simulate` in an `act`. It wasn't doing so when we started testing 🤷‍♂

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
azangru: I have never seen this written down in an influential style guide, but there are two traditions in naming of event handlers.

According to one tradition, they are named with the `on` prefix followed by the name of the event — `onClick`, `onScroll`, `onMouseEnter`, etc. This is how React names its own events, which map to the DOM event names (`onclick`, `onscroll`, `onmouseenter`, etc.). Here's someone [thinking out loud](https://jaketrent.com/post/naming-event-handlers-react/) in their blog the way they name events in a similar fashion.

According to the other tradition, they are named with the `handler` suffix, following the event name (e.g. `clickHandler`, `scrollHandler`, `submitHandler`, etc.). Here's an [old Microsoft style guide](https://docs.microsoft.com/en-us/dynamicsax-2012/developer/naming-conventions-delegates-and-event-handlers#event-handler-naming-conventions) for god knows what language that exemplifies this approach.

So I try to name my events according to either of these traditions (I think both are fine). What I emphatically **do not** want us to do is to mix them in a way that combines both the `on` prefix and the `handler` suffix (i.e. `onSwitch` is ok, `switchHandler` is ok, but `onSwitchHandler` is not).

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.tsx
azangru: That depends on whether we are ok with calling this thing ChromosomeNavigator, or whether we would prefer a name that is less tied to the concept of chromosome (since the nucleic acid strands displayed in genome browser can also be other than chromosomes).

src/ensembl/src/shared/data/centromeres.ts
azangru: Yes, any application that wants to draw human chromosomes with centromeres in them (EntityViewer being the most likely candidate) will have to use this file (until ENSWBSITES-385 is completed and we start getting this data from the backend).

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBar.test.tsx
azangru: ✔️

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
azangru: ✔️

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
azangru: ✔️

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
azangru: changed `onChangeViewClick` to `handleViewChange` (`view` being the name of the variable stored in the component state and referring to whether we are showing the chromosome visualisation or the region switcher).

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.tsx
msmanoj: I think keeping it as `ChromosomeNavigator` is fine as long as we can understand it. Any thought's @imran-ebi ?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
msmanoj: > If at any point we add another test that checks anything related to those functions, they will have already been called, and this may interfere with test results.

That is true but do we really need this to be there upfront? There are chances that we might never come back to update these tests.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
msmanoj: I agree that it is good to follow either of these traditions but my only concern was the link between the words.
For example, `mouseEnter` can be linked as `mouse -> enter`. `changeview` can be linked as `Change` -> `view`. But having `changeViewClick` felt a bit unclear whether it means `changeView -> click` or `change -> viewClick` .

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.scss
msmanoj: I think it would be nice to add what you just explained as a comment above the class just to make it very clear. I am sure this is not something we can easily infer just by looking at `font-size: 0;`.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
+import ChromosomeNavigator from 'src/content/app/browser/chromosome-navigator/ChromosomeNavigator';
+import BrowserNavBarRegionSwitcher from './BrowserNavBarRegionSwitcher';
+import { ReactComponent as CloseIcon } from 'static/img/shared/close.svg';
+
+import styles from './BrowserNavBarMain.scss';
+
+export enum Content {
```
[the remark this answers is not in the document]
azangru: This is not component though. It's 1) an enum; 2) a named export, rather than a default one.

In this particular case I don't think I actually need to export this enum. I thought of using it in tests, but then didn't have a need for it. I'll remove the export keyword.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.tsx
```diff
@@ -0,0 +1,53 @@
...
+          browserNavItem={item}
+          enabled={shouldEnableButton(index)}
+        />
+      ))}
+    </div>
+  );
+};
+
+const mapStateToProps = (state: RootState) => {
+  const disabled = getRegionEditorActive(state) || getRegionFieldActive(state);
```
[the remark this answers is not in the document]
azangru: This (and the expanded/isExpanded example below) is the case I am the most unsure about. On the one hand, we tend to name boolean flags as questions. On the other hand, html elements use such boolean flags without the question verb (cf. the `disabled` attribute for `<input>`). React components kinda pretend as if they are html elements, so I don't know which form to prefer 😞

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  () => () => <div>ChromosomeNavigator</div>
+);
+jest.mock('./BrowserNavBarRegionSwitcher', () => () => (
+  <div>BrowserNavBarRegionSwitcher</div>
+));
+
+describe('BrowserNavBarMain', () => {
+  describe('rendering', () => {
+    it('renders without errors', () => {
+      expect(() => mount(<BrowserNavBarMain />)).not.toThrow();
```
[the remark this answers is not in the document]
azangru: Removed the test.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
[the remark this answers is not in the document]
azangru: Removed the export.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.scss
azangru: Added a comment, but it's generally a fairly common CSS technique, and not for the first time used in our codebase.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
azangru: Between having tests that don't clean up after themselves (and do not cause problems because there are no further tests to take advantage of the clean environment), and cleaning up the definitely polluted testing environment even though there are no further tests to run, I think it's better to err on the side of the second. After all, what harm can cleaning up do — add a millisecond or two to the execution time of the test? Well-behaved tests, like good boy scouts, try to leave their environment clean.

imran-ebi: @azangru I agree. It is always better to have something that will avoid things breaking rather than adding it when it only breaks.

src/ensembl/src/content/app/browser/chromosome-navigator/ChromosomeNavigator.tsx
imran-ebi: I don't have a problem with the name. We can leave it as it is if we don't have a better name for it.

azangru: ✅

PR #197, opened 2019-11-06
Fix wasm build issues

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Description
When I tried building the wasm file as per Dan's instructions, the md5 check failed as the command that is used for it (md5sum) does not exist in MacOS by default. So this PR, will check if md5 exists and use it, if not use md5sum (which is in Linux operating systems).

## Possible complications
This could probably not work, if both md5 and md5sum do not exist in the system.

### DIFF
--- a/src/assets/browser/app/build/deploy.sh
+++ b/src/assets/browser/app/build/deploy.sh
@@ -34,10 +34,10 @@ WASMNAME="browser-$WASMHASH.wasm"
 JSNAME="browser.js"
 rm -f $DEST/*.js $DEST/*.wasm
 
-if [ "$1" == "check" ] ; then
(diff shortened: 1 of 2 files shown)


ens-ds23:
Looks good. There's a similar issue with the existence (or otherwise) of wasm-opt and one with env.sh and zsh. I'll add these to this PR and merge it (as builds are already broken on macs) to unblock #181 on which #188 #190 and #191 are blocked, in turn.

ens-ds23:
Actually, we really should only push optimised WASMs, so I'll make wasm-opt being missing into a failure with a nice error message, instead.

PR #198, opened 2019-11-06
Fix tooltip width for Safari

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
ENSWBSITES-423

## Importance
Bug in production.

## Description
Tooltips are broken in Safari: they sometimes simply wrap inner text:

image

and sometimes allow content to escape outside the tooltip:

image

The reason seems to be that Safari does not respect the `width: max-content;` rule (or perhaps interprets it incorrectly) if `box-sizing` is `border-box` (example: https://jsfiddle.net/rt9vuadf/1/), but renders correctly if  `box-sizing` is `content-box` (example: https://jsfiddle.net/rt9vuadf/).

Safari is the new IE6! 😠


imran-ebi:
@azangru Safari has definitely been the new IE for a while 😉

PR #200, opened 2019-11-07
Feature/question button

ens-ap5:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-292

## Importance
Not high priority; minor cosmetic improvement.

## Description
Extend the QuestionButton component to allow CSS tweaks, in this case a smaller size for use inside form input boxes, and then apply this change to SpeciesSelector Search Field.

## Views affected
SpeciesSelector initial page with search box.

### Other effects
None. Both versions (default and small) are in Storybook; no further test needed.

## Possible complications
None anticipated.


src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -7,6 +8,7 @@ import styles from './QuestionButton.scss';
 
 type Props = {
   helpText: React.ReactNode;
+  style: string;
```
azangru: Several points:

1) Do we want flexibility for this button (i.e. is there a reasonably high probability that it will have to have different sizes in different places), or can we be reasonably sure that there will be only a handful of modifications? If the latter is correct, then this approach is fine; but if we want to ensure flexible styling, then passing additional classes to modify the appearance of the question button (see [ImageButton](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/image-button/ImageButton.tsx#L24) as an example) will be a better approach.

2) If we are reasonably certain that there will be only a handful of question button variants, then we should explicitly type them here, for the benefit of future reader, so that `style` is not any old string, but only a member of an enum:

```
enum QuestionButtonVariant {
  SMALL = 'small'
}

type Props = {
  ...blah,
  style: QuestionButtonVariant
}
```

3) The `style` attribute has a special meaning in react — it can be used to pass styles to native html elements using an object of CSS properties. So it's better to name this property (if this solution is eventually chosen instead of passing modifying classes) `type`, or `kind`, or `variant`, or something of this sort.

azangru: Also, notice that this pull request created type errors:

image

ens-ap5: Good points. I did originally do it with an enum, but thought it might be overkill. I'm happy to rename the variable - I was stumped on what else to call it!

ens-ap5: Changes made as discussed, and it's compiling cleanly - could you have another look?

src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -20,9 +29,19 @@ const QuestionButton = (props: Props) => {
...
 
+  const styles = props.classNames
+    ? { ...defaultStyles, ...props.classNames }
+    : defaultStyles;
+
+  const classParams = props.styleOption
+    ? [defaultStyles.default, styles[props.styleOption]]
+    : defaultStyles.default;
+
+  const className = classNames(classParams);
```
azangru: The `classNames` library can take an arbitrary number of arguments, which can be either simple strings, or hashes, with keys being class names and values being whether to add them or not.

This will compose in a nicer statement like so:

```
 const className = classNames(
  defaultStyles.default,
  {
    [defaultStyles[props.styleOption]]: props.styleOption
  },
  props.className
);
```

src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -1,12 +1,21 @@
...
+// Extra styling options based on where the button is located
+export enum QuestionButtonOption {
+  INPUT = 'in-input-field',
+  INLINE = 'inline'
+}
 
 type Props = {
   helpText: React.ReactNode;
+  styleOption?: QuestionButtonOption;
+  classNames?: { [key in QuestionButtonOption]?: string };
```
azangru: I think there should be only one `className: string` rather than an object of strings. If ever the parent will want to pass an arbitrary class to the question button, it will likely be just a single class.

src/ensembl/src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -18,7 +18,11 @@ import SpeciesSearchMatch from '../species-search-match/SpeciesSearchMatch';
 
 import AutosuggestSearchField from 'src/shared/components/autosuggest-search-field/AutosuggestSearchField';
 import ClearButton from 'src/shared/components/clear-button/ClearButton';
-import QuestionButton from 'src/shared/components/question-button/QuestionButton';
+import QuestionButton, {
+  QuestionButtonOption
+} from 'src/shared/components/question-button/QuestionButton';
+
+console.log(QuestionButtonOption);
```
azangru: this should go

azangru:
Not that it's important, but for the fun of it you may also consider adding tests:
- question button renders with only default class if no styleOption prop is passed
- question button adds `inline` class if it receives "inline" styleOption prop
- question button adds 'in-input-field' class if it receives "in-input-field" styleOptionProp
- question button shows Tooltip if moused over

ens-ap5:
Changes made - tests not created.

src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -20,9 +29,21 @@ const QuestionButton = (props: Props) => {
     setIsHovering(false);
   };
 
+  const styles = props.className
+    ? { ...defaultStyles, ...props.className }
+    : defaultStyles;
```
azangru: now redundant const

ens-ap5: Oops, missed that! Now fixed.

imran-ebi:
Can we merge this PR?

PR #204, opened 2019-11-12
hide placeholder on region field focus

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-344

## Description
Hide placeholder when focusing on region input textbox.

## Views affected
Browser navigation bar

## Possible complications
n/a

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
+++ b/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
@@ -12,6 +12,12 @@
   &::placeholder {
     color: $grey;
   }
+  &:focus::-webkit-input-placeholder {
+    color: transparent;
+  }
+  &:focus::-moz-placeholder {
+    color: transparent;
+  }
 }
 
 .chrLocationView {
--- a/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
+++ b/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
@@ -12,11 +12,10 @@
   &::placeholder {
     color: $grey;
   }
-  &:focus::-webkit-input-placeholder {
-    color: transparent;
-  }
-  &:focus::-moz-placeholder {
-    color: transparent;
+  &:focus {
+    &::placeholder {
+      color: transparent;
+    }
   }
 }
 
--- a/src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx
+++ b/src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx
@@ -30,12 +30,9 @@ const ZmenuController = (props: Props) => {
   const [zmenus, setZmenus] = useState<StateZmenu>({});
 
   useEffect(() => {
-    const subscriptionHandler = (payload: ZmenuIncomingPayload) =>
-      handleBpaneEvent(payload);
-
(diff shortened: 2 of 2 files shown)


src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
```diff
@@ -12,6 +12,12 @@
   &::placeholder {
     color: $grey;
   }
+  &:focus::-webkit-input-placeholder {
+    color: transparent;
+  }
+  &:focus::-moz-placeholder {
+    color: transparent;
+  }
```
azangru: 1) In general, we shouldn't be writing vendor prefixes — we have postcssPresetEnv, which is supposed to do it for us (https://github.com/Ensembl/ensembl-client/blob/c4aea08e45e2d76f4f17f57c5b3437fe197da7fb/src/ensembl/.storybook/webpack.config.js#L46)

2) From the CSS perspective, would 
```
&:focus {
  &::placeholder {
    ... rules
  }
}
```
make more sense?

3) Rather than doing this with CSS, why don't we remove placeholder text in BrowserRegionField component (https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx#L143) when the input is focused?

src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx
```diff
@@ -30,12 +30,9 @@ const ZmenuController = (props: Props) => {
   const [zmenus, setZmenus] = useState<StateZmenu>({});
 
   useEffect(() => {
-    const subscriptionHandler = (payload: ZmenuIncomingPayload) =>
-      handleBpaneEvent(payload);
-
     const subscription = browserMessagingService.subscribe(
       'bpane-zmenu',
-      subscriptionHandler
+      handleBpaneEvent
```
azangru: _(note for our future selves):_ this change removes unnecessary intermediary function in ZmenuController. It is not directly related to removing the placeholder from the focused BrowserRegionField, but that's fine.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.scss
imran-ebi: > Rather than doing this with CSS, why don't we remove placeholder text in BrowserRegionField component (https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx#L143) when the input is focused?

We need the placeholder when the region field input is not focussed

azangru: onFocus: remove placeholder
onBlur: show placeholder

azangru: On the other hand, you are right, if this is easier done with CSS, let's do it with CSS.

ens-ma7: @azangru Updated css. Have a look and let me know.

azangru: Looks good.

PR #205, opened 2019-11-12
Update the browser navigation panel according to the new designs

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-398

## Description
The browser navigation panel is broken if you check for screens smaller than 1400px in the [live site](https://2020.ensembl.org/). So Andrea mocked [some designs for separate screen sizes on XD](https://xd.adobe.com/view/685c72fc-2fe0-4c95-7895-1bd5ce6de6e2-1e6d/?fullscreen). ~~This does not adhere to the same breakpoints though, but uses the closest breakpoints. The breakpoints used are depicted in the table below:~~

image

Apart from this, the functionality of both the region forms have been changed to match with that of the XD mockup.

**UPDATE:** The breakpoints have now been updated and now match the breakpoints in the styling. So the image above is now outdated.

## Views affected
Browser navigation bar


src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
```diff
@@ -14,23 +15,40 @@ jest.mock('./BrowserNavBarRegionSwitcher', () => () => (
...
+
+  beforeEach(() => {
+    wrapper = mount(<BrowserNavBarMain {...props} />);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('renders chromosome visualization by default for laptops or bigger screens', () => {
```
azangru: then there should also be a test that checks that chromosome navigator does not get rendered for smaller screens

imran-ebi: I added that test

azangru:
What frightens me a little is that our [javascript code](https://github.com/Ensembl/ensembl-client/blob/bugfix/navigation-bar-layout/src/ensembl/src/global/globalConfig.ts#L4-L10) and our [css code](https://github.com/Ensembl/ensembl-client/blob/bugfix/navigation-bar-layout/src/ensembl/src/styles/_breakpoints.scss#L10-L57) have two sets of breakpoints, that are already different from each other. We might want to review our breakpoint strategy. One possible way to bridge this gap is suggested in [this PR](https://github.com/Ensembl/ensembl-client/pull/215).

Also, out of curiosity, why did we use a CSS approach to showing/hiding things [in this PR](https://github.com/Ensembl/ensembl-client/commit/434a79696ec9bd8b26aa2ff60a7a0bf9ed0a2351), but a js approach here. Are these two different situations?

imran-ebi:
@azangru I tried to merge the `feature/window-match-media` branch with this PR's branch. However, I unfortunately can't do it and get this message:

```
imran➜ensembl-client/src/ensembl(test/match-media)» git merge feature/window-match-media                                [16:25:23]
merge: feature/window-match-media - not something we can merge
```

I think this is probably because these two branches have diverged from different branches.

azangru:
> I think this is probably because these two branches have diverged from different branches.

@imran-ebi yes, merged and pushed here: https://github.com/Ensembl/ensembl-client/tree/feature/integration-window-match-media

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
```diff
@@ -18,33 +18,19 @@ describe('BrowserNavBarControls', () => {
       mount(
         <BrowserNavBarControls
           browserNavStates={browserNavStates}
-          disabled={faker.random.boolean()}
+          shouldBeOpaque={faker.random.boolean()}
```
azangru: `shouldBeOpaque` is a very specific name describing how the disabled status of this particular component is currently shown. Why did you prefer it over `disabled`? I think `disabled` was far better.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
```diff
@@ -14,23 +15,44 @@ jest.mock('./BrowserNavBarRegionSwitcher', () => () => (
...
+
+  beforeEach(() => {
+    wrapper = mount(<BrowserNavBarMain {...props} />);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('renders chromosome visualization by default for screens smaller than laptops', () => {
```
azangru: does not render?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
```diff
@@ -4,14 +4,18 @@ import { connect } from 'react-redux';
...
 import {
   toggleRegionEditorActive,
   toggleRegionFieldActive
 } from '../browserActions';
+import { getBreakpointWidth } from 'src/global/globalSelectors';
 
 import styles from './BrowserNavBarRegionSwitcher.scss';
 
 type Props = {
+  breakpointWidth: BreakpointWidth;
```
azangru: From the point of view of this component, it probably just wants to know whether it should show both navigation components, or only one. Perhaps `display: full | compact` instead?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
```diff
@@ -31,19 +35,25 @@ export const BrowserNavBarRegionSwitcher = (props: Props) => {
...
+      {props.breakpointWidth >= BreakpointWidth.BIG_DESKTOP && (
+        <div className={styles.regionEditorWrapper}>
+          <BrowserRegionEditor />
+        </div>
+      )}
     </div>
   );
 };
 
+const mapStateToProps = (state: RootState) => ({
```
azangru: The component's immediate parent can tell it whether or not to show the second navigation component.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -34,7 +29,7 @@ describe('<BrowserRegionEditor', () => {
     chrLocation: initialChrLocation,
     genomeKaryotype: createGenomeKaryotype(),
     isActive: true,
-    isDisabled: false,
+    shouldBeOpaque: false,
```
azangru: As above: this prop name is too specific, tightly coupled with the way we represent the disabled state of this component, and not very informative for the consumer of the component.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -30,18 +29,16 @@ import {
...
-import browserStyles from '../Browser.scss';
 import browserNavBarStyles from '../browser-nav/BrowserNavBar.scss';
 
 export type BrowserRegionEditorProps = {
   activeGenomeId: string | null;
   chrLocation: ChrLocation | null;
   genomeKaryotype: GenomeKaryotypeItem[] | null;
   isActive: boolean;
-  isDisabled: boolean;
+  shouldBeOpaque: boolean;
```
azangru: as above

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -55,6 +52,7 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
   const genomeKaryotype = props.genomeKaryotype as GenomeKaryotypeItem[];
   const [stick, locationStart, locationEnd] = props.chrLocation as ChrLocation;
 
+  const formRef = useRef(null);
```
azangru: What does this ref do?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
imran-ebi: Because it is actually not disabled unlike how it was before. You can click and interact with it just like a normal input.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: It is not used anymore. I'll remove this.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: >Perhaps display: full | compact instead?
How about `showBrowserRegionEditor`?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: This clashes with the previous comment. If we are going to send the `breakpointWidth` via the parent, then should we still change the prop name to `display`?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -150,7 +144,18 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
 
-  const closeForm = () => {
+  const closeForm = (event: Event) => {
+    if (
+      !stickRef.current ||
+      !locationStartRef.current ||
+      !locationEndRef.current ||
+      stickRef.current.contains(event.target as HTMLElement) ||
+      locationStartRef.current.contains(event.target as HTMLElement) ||
+      locationEndRef.current.contains(event.target as HTMLElement)
```
azangru: I wonder if the time has come to try to use optional chaining :-)

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
azangru: If the only thing that this panel can do is show/hide the browser region editor field, I would say something like display: full/compact is simpler 🤷‍♂

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
azangru: Sure:

`<BrowserNavBarRegionSwitcher display={breakpoint satisfies condition ? 'full' : 'compact'} />`

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
azangru: > You can click and interact with it just like a normal input as there is no overlay on it.

I am confused. I went and asked Andrea, and she says the buttons at this point are effectively disabled. A click on an active button (which is now "knocked-back blue") will not execute what this button does, but will just remove the semitransparent screen that is overlaying the buttons.

imran-ebi: I just talked with Andrea about this. The region editor and the region field are clickable when either of them are active. However, the navigation icons should have an overlay on top of them.

imran-ebi: There has been a change again. The region will be clickable when the region editor is active. However, the region editor or the navigation icons won't be clickable when the region field is active. The user will need to click away from the region field in order to use them i.e. need an extra click.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: I agree 😄 I will use them here.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: For this to be cleaner, won't we need an enum instead of using `full` and `compact` as strings?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: For this to be cleaner, won't we need an enum instead of using `full` and `compact` as strings? If that is the case it wouldn't be simple as using `showBrowserRegionEditor`.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
azangru: I am a bit confused about the description. Am I correct in understanding that this is what will be happening?

Untitled Diagram

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
azangru: > For this to be cleaner, won't we need an enum instead of using full and compact as strings?

Yes, I was thinking of an enum :-)

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
azangru: We'll need another babel plugin for that, I suppose...

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
imran-ebi: Yes, that's correct.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: True, I am working on this and will create a PR with the babel plugin and probably update the other NPM packages too.

imran-ebi: @azangru There is now a PR for this: #205

imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.test.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.test.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -53,6 +54,11 @@ describe('<BrowserRegionEditor', () => {
     test('contains submit and close buttons', () => {
       expect(wrapper.find('button[type="submit"]')).toHaveLength(1);
     });
+
+    test('has an overlay on top when region field is active', () => {
```
azangru: > when region field is active
The test doesn't assert anything about whether region field is active; it just sets `isDisabled` to `true`. So "has an overlay on top when disabled" 😄

src/ensembl/src/global/globalConfig.ts
```diff
@@ -9,6 +9,11 @@ export enum BreakpointWidth {
   BIG_DESKTOP = 1800
 }
 
+export enum DisplayType {
+  FULL = 'full',
+  COMPACT = 'compact'
+}
```
azangru: That doesn't belong to the global config; it's a very local setting of `BrowserNavBarMain`

imran-ebi: It is for now, but this looks like an `enum` that we can easily reuse elsewhere. Hence, why I put it in `globalConfig`.

azangru: Oh-uh.
Is it the part of the [Status type](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/types/status.ts) then?

_(Although I still think they belong in `BrowserNavBarMain` component; they don't even leak out of it)_

imran-ebi: [`TrackPanel`](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/track-panel/TrackPanel.tsx#L45) is another example of this. And then there is the [`BrowserBar`](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx#L108) too.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imran-ebi: ✅

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imran-ebi: ✅

src/ensembl/src/global/globalConfig.ts
azangru: @msmanoj do you have any opinion about this type?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.tsx
```diff
@@ -37,16 +36,18 @@ export const BrowserNavBarControls = (props: Props) => {
           enabled={shouldEnableButton(index)}
```
msmanoj: How about just `enabled={!props.browserNavStates[index]}` ?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
```diff
@@ -12,25 +18,40 @@ enum Content {
...
+};
+
+export const BrowserNavBarMain = (props: BrowserNavBarMainProps) => {
   const [view, changeView] = useState<Content>(Content.CHROMOSOME);
 
   const handleViewChange = (newView: Content) => {
     changeView(newView);
   };
 
+  const shouldShowChromsomeNavigator =
```
msmanoj: `shouldShowChromsomeNavigator` -> `shouldShowChromosomeNavigator` ?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
```diff
@@ -16,20 +17,34 @@ jest.mock(
...
+  });
+
   afterEach(() => {
     jest.resetAllMocks();
   });
 
-  it('renders correctly', () => {
-    const wrapper = mount(<BrowserNavBarRegionSwitcher {...props} />);
+  describe('rendering', () => {
+    it('renders only region field on smaller screens', () => {
```
msmanoj: renders only region field on smaller screens -> renders region field only on smaller screens?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
```diff
@@ -16,20 +17,34 @@ jest.mock(
...
+      expect(wrapper.find(BrowserRegionField).length).toBe(1);
+    });
+
+    it('renders both region field and region editor on big desktop screens', () => {
+      wrapper.setProps({ display: DisplayType.FULL });
 
-    expect(wrapper.find(BrowserRegionEditor).length).toBe(1);
-    expect(wrapper.find(BrowserRegionField).length).toBe(1);
+      expect(wrapper.find(BrowserRegionEditor).length).toBe(1);
+      expect(wrapper.find(BrowserRegionField).length).toBe(1);
```
msmanoj: Here and in other places, would it be better to check `expect(wrapper.find(BrowserRegionField)).toHaveLength(1)`?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
```diff
@@ -101,19 +89,7 @@ describe('<BrowserRegionField', () => {
...
-        .find(Input)
-        .simulate('change', { target: { value: regionInput } });
-      wrapper.find('button[role="closeButton"]').simulate('click');
-
-      expect(wrapper.find(Input).props().value).toBe('');
-      expect(wrapper.props().toggleRegionFieldActive).toHaveBeenCalledWith(
-        false
-      );
-    });
+    // TODO: Test if the form is reset when clicked outside the form. Need to be able to mock useOutsideClick for this.
```
msmanoj: Here and in the other place, would it be possible to call the callback function passed on to the `useOutsideClick` directly and check if it resets the form?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.tsx
```diff
@@ -101,16 +96,31 @@ export const BrowserRegionField = (props: BrowserRegionFieldProps) => {
...
+      return;
+    }
+
+    resetForm();
+  };
+
+  useEffect(() => {
+    document.addEventListener('click', closeForm);
+    return () => document.removeEventListener('click', closeForm);
+  }, []);
```
msmanoj: Created a ticket for myself to check if we can update useOutsideClick to intake multiple refs.
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-452

src/ensembl/src/global/globalConfig.ts
msmanoj: As discussed, we can revert this to use the `breakpointwidth` within the child component to show/hide elements until we have a clear idea of how `DisplayType` needs to work.
Also, we agreed to change the prop name `breakpointwidth` to `viewportwidth` instead.

src/ensembl/src/global/globalHelper.ts
```diff
@@ -3,12 +3,16 @@ import { BreakpointWidth } from './globalConfig';
...
     return BreakpointWidth.DESKTOP;
   } else if (BreakpointWidth[key] >= BreakpointWidth.LAPTOP) {
     return BreakpointWidth.LAPTOP;
-  } else {
+  } else if (BreakpointWidth[key] >= BreakpointWidth.TABLET) {
     return BreakpointWidth.TABLET;
+  } else {
+    return BreakpointWidth.PHONE;
   }
 };
```
msmanoj: It looks like could be as simple as saying `return BreakpointWidth[key]`. 
@ima23 if this is too much to change in this PR, could you create a ticket to clean this up, please?

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
azangru: I think either is fine. Both formats are equally readable ("expect components array length to be X" vs "expect components array to have length X"); none is significantly more verbose than the other.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
imran-ebi: This test should also say that region editor is not rendered. So will word it is as 'renders region field and not region editor on smaller screens'.

src/ensembl/src/global/globalHelper.ts
imran-ebi: @msmanoj I'll create a ticket for this as per our discussion offline

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
imran-ebi: I tried a few different approaches to get test this, but unfortunately all of them failed. I'd prefer to defer this to a later time rather than holding this PR back.

imran-ebi: I tried a few different approaches to get test this, but unfortunately all of them failed. I'd prefer to defer this to a later time rather than holding this PR back.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.test.tsx
imran-ebi: :white_check_mark:

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarMain.tsx
imran-ebi: :white_check_mark:

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarControls.tsx
imran-ebi: :white_check_mark:

src/ensembl/src/global/globalConfig.ts
imran-ebi: :white_check_mark:

PR #209, opened 2019-11-16
Bugfix/scheduler catastrophe

ens-ds23:
## Type

- [X] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
None

## Importance
Urgent

## Description
This fixes a massive, accidental performance regression from an update merged onto dev and live on an apparently random subset of machines and browsers.

## Views affected
Genome browser

### Other effects

## Possible complications
This is quite a quick fix and should be double-checked before pushing but it is a simple reversion of an earlier change believed at the time to have no effect. It's been tested on Dan's home machine in FF and Andrey's home machine in Chrome so far. We should try it on a few more around the office before pushing it live.

### DIFF
--- a/src/assets/browser/app/src/controller/global/apprunner.rs
+++ b/src/assets/browser/app/src/controller/global/apprunner.rs
@@ -163,14 +163,16 @@ impl AppRunner {
                         }
                     }),5,false);
                 }
+                /* animate & draw */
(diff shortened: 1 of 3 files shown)


azangru:
> We should try it on a few more around the office before pushing it live.
Tested on a Macbook Air in the office; looks fine.

PR #210, opened 2019-11-16
Watch ensembl-genome-browser node modules

azangru:
## Type
- Infrastructure

### Problem
With our current config, webpack-dev-server is not watching changes in `node_modules` folder. However, after merging of https://github.com/Ensembl/ensembl-client/pull/181, every Rust rebuild changes the browser.js file in `node_modules/ensembl-genome-browser`; so without webpack watching files in this folder, Rust developer will need to re-build the whole ensembl-client project every time the wasm is recompiled. Which takes time and is annoying.

### Solution
Ignore all node_modules except for ensembl-genome-browser (regex [copied from Stack Overflow](https://stackoverflow.com/a/44166532/3925302)).

### DIFF
--- a/src/ensembl/webpack/webpack.config.dev.js
+++ b/src/ensembl/webpack/webpack.config.dev.js
@@ -106,7 +106,7 @@ const devConfig = {
   // disable webpack from watching node modules
   // this would reduce memory consumption and also should improve build times
   watchOptions: {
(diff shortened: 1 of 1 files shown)


ens-ds23:
Confirmed working following merge onto `feature/animation-improvements-2` and rebuild there in realistic setting.

PR #213, opened 2019-11-20
Create our own useResizeObserver hook

azangru:
## Type
- Refactoring

### Problem
One of our dependencies is [`use-resize-observer`](https://github.com/ZeeCoder/use-resize-observer#readme). It provides a React hook for observing dimensions of a given DOM element.
This dependency, however is:
- very simple to re-create
- _(probably because of that)_ not particularly popular with the community
- does not have type definitions (developer does not use typescript and does not want other people's contributions of type definitions that he cannot check

### Solution
Create our own useResizeObserver hook, roughly based on the updated (v.5) `use-resize-observer` library.

### Why?
- so that we have more control over the hook
- so that we have proper type definitions for this hook
- because it is simple; we don't really need to depend on another library here

### Root component
Move `breakpointWidth` as a prop from the component to the action (so as to avoid unnecessary re-rendering of this component when this prop changes).

### DIFF
--- a/src/ensembl/src/root/Root.tsx
+++ b/src/ensembl/src/root/Root.tsx
@@ -1,6 +1,10 @@
-import React, { FunctionComponent, useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { connect } from 'react-redux';
-import useResizeObserver from 'use-resize-observer';
+
+import useResizeObserver from 'src/shared/hooks/useResizeObserver';
+
+import { BreakpointWidth } from '../global/globalConfig';
+import { updateBreakpointWidth } from '../global/globalActions';
 
 import Header from '../header/Header';
 import Content from '../content/Content';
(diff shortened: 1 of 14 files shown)


src/ensembl/src/root/Root.tsx
```diff
@@ -9,34 +13,22 @@ import privacyBannerService from '../shared/components/privacy-banner/privacy-ba
...
-
-type RootProps = StateProps & DispatchProps & OwnProps;
-
-export const Root: FunctionComponent<RootProps> = (props: RootProps) => {
-  const [ref, width] = useResizeObserver();
-  const currentBreakpoint: BreakpointWidth = getBreakpoint(width);
+export const Root = (props: Props) => {
+  const elementRef = useRef<HTMLDivElement>(null);
+  // TODO: this should be properly done via window.matchMedia() rather than
+  // via ResizeObserver; will refactor in subsequent PR
```
azangru: see description of [this PR](https://github.com/Ensembl/ensembl-client/pull/215)

src/ensembl/src/shared/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,68 @@
...
+        setSize({ width: newWidth, height: newHeight });
+      }
+    });
+
+    resizeObserver.observe(element);
+
+    return () => resizeObserver.unobserve(element);
+  }, [ref]);
+
+  return useMemo(() => ({ ref, ...size }), [
```
msmanoj: Do we need `useMemo` here?

azangru: The author of the original library added `useMemo` in this commit: https://github.com/ZeeCoder/use-resize-observer/commit/5c4aa04e52f73d032f8136a59cec775ce40f7fb3#diff-1fdf421c05c1140f6d71444ea2b27638R69-R73

Hard to tell what problem he was trying to solve. The only difference is that now the hook will return the same object in certain situations rather than returning a new object all the time. Can't imagine why that should matter — building a new object is trivial, and React component shouldn't re-render If the return value from this hook is different from the previous return value. I suppose we could remove the useMemo.

msmanoj: I was not able to be sure either. It it very unlikely that the dimensions will remain unchanged after playing around with the window size. Maybe we can remove it and look into adding it later on if it starts to cause any problem?

azangru: Removed the useMemo

PR #214, opened 2019-11-20
Toggle focus track using both gene or transcript eye icon

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-400

## Description
Using a temporary hack to toggle the focus track on/off using the eye icon next to both gene and transcript track list item.

## Views affected
Browser - > TrackPanel


azangru:
Couple of things:

1) Is it ok to save `track:gene-feat-1` status for genes that don't have transcript information (as is currently done)? On the one hand, it doesn't break anything, so maybe we don't care. On the other hand, this is weird.

2) Not sure whether to consider this related to this PR, but here's a use case:
- View BRCA2
- Switch off BRCA2 track (either gene or transcript)
- Now change focus to FRY
- Notice that FRY gets invisible once its switches to the focus gene track

Is this something that needs fixing? We save track states independently for different focus objects, so it seems that we don't want other focus objects to become invisible when we switch to them. On the other hand, we might postpone solving this problem until we get proper transcripts.

msmanoj:
> 1. Is it ok to save `track:gene-feat-1` status for genes that don't have transcript information (e.g. for FRY)? On the one hand, it doesn't break anything, so maybe we don't care. On the other hand, this is weird.

I could do something like the following to prevent it:
`if(track.track_id === 'track:gene-feat-1' || track.child_tracks ){
        updateTrackStates('track:gene-feat-1', newStatus);
}`


> Is this something that needs fixing?

I try to get this fixed as it looks a bit weird.

ens-ds23:
There really shouldn't be anything that handles track:... specially in the frontend code, particularly not with "gene" in. On the other hand, this may be inevitable at the moment, as the focus track is special. Could someone explain this PR to me? At the moment we don't have multiple transcripts, so I suppose a hack is ok for these "doing the same thing". But why is "gene" in there? Is that something I need to fix on the backend?

azangru:
> Could someone explain this PR to me? At the moment we don't have multiple transcripts, so I suppose a hack is ok for these "doing the same thing". But why is "gene" in there? Is that something I need to fix on the backend?

1) When we switch to a new focus object, we ask the api for its "track list"

- If the focus object is one of the "entry points", the api will send back a meaningful response ([example](https://2020.ensembl.org/api/object/track_list?object_id=homo_sapiens_GCA_000001405_27:gene:ENSG00000139618)). Notice that the response contains the field called `track_id`, both at the top level and in the `child_tracks`. This id contains the word "gene", if that’s what you were asking about. The backend team doesn't know how to meaningfully generate these ids, so at the moment they are hard-coded. This endpoint is expected to change (see rough outline as result of a several-months-old discussion [here](https://github.com/Ensembl/ensembl-client/wiki/Modelling-of-api-responses-for-focus-objects)), at which point no-one knows how information about track ids is going to be passed around.
- _(follows from the previous point)_ if the focus object is anything other than an "entry point", the api will respond with a 400-type error.

2) The React code in this PR is doing 2 things:
- first, it tells Genome Browser that a track with a given id has a new visibility status (this is when the `updateGenomeBrowser` function is called)
- second, it remembers user’s preference to hide or show this track, and saves it in browser’s local storage for the time when user returns to this focus object (this is when `updateTrackStates` is called)

ens-ds23:
Ok, thanks @azangru . From the genome browser's point of view there is (these days) only a "feature track". It's the same track whatever is in it (gene, variant, etc), so anything that talks about trackids with "-gene" in them should probably be purged. I'll check my own code for that. (This can be a different PR). As well as being on a chromosome/species and at a position and scale the browser has a focus object, as a whole. Multiple tracks respond to that being "on a focus object" in a potential multitude of ways. Most obviously the "feature track" but also others (at the moment, the one where the feature comes from changes because it's elided).

I don't yet fully know how this will translate into the world of multiple transcripts, but I envisage there being an additional "sub-id" passed (probably in a set feature object call as an additional parameter or as a separate call) amounting to being a sub-type of the feature-object. The track, even when it has multiple things in it, would still just be the feature track.

I'm fine with this PR from my PoV but it has raised some hackery-removal issues on the backend etc. I thought I should try to convey the general direction. A lot of the existing stuff that's now "wrong" came back from when we hadn't really modelled the focus object track /at all/. Now it's pretty well modelled (with dodgy corners) except for things with complex structure (like genes with multiple transrcipts).

ens-ds23:
We need to think collectively, I think, about where and how we do the transformation between things like (as Andrea said at startup) things like visibility being per-object from the user PoV but the genome browser considering eg the feature track as a single thing.

I'd personally rather this was handled in the React because otherwise I'm worried that the genomebrowser would start creepingly having state which I've tried to avoid to keep Redux as the source of all truth, but it's always suspiciously easy to find reasons that you think that something is someone else's task, :-) ,

msmanoj:
> We need to think collectively, I think, about where and how we do the transformation between things like (as Andrea said at startup) things like visibility being per-object from the user PoV but the genome browser considering eg the feature track as a single thing.

I agree to that, Dan. I made some changes to the React app to store the track states along with the focus object when I was working on bookmarks. At that time, it was still a bit incomplete as we were only storing the states of the toggled tracks on Redux, ignoring the rest. 
In this PR I have now pushed some changes that stores the status of all the tracks on to Redux as they get drawn.

imran-ebi:
The track IDs won't have the prefix `track:` anymore, though it will be part of the track ID name. You can see that change in the latest track categories API endpoint that Kamal is working on: http://193.62.55.91:30895/api/genome/track_categories?genome_id=homo_sapiens_GCA_000001405_14

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -139,42 +177,29 @@ export const TrackPanelListItem = (props: TrackPanelListItemProps) => {
...
+      updateTrackStates('track:gene-feat', newStatus);
+      // Make sure we have a track with id 'track:gene-feat-1' before setting a status.
+      if (
+        track.track_id === 'track:gene-feat-1' ||
+        (track.child_tracks && track.child_tracks.length)
+      ) {
+        updateTrackStates('track:gene-feat-1', newStatus);
+      }
+      updateGenomeBrowser('track:gene-feat', newStatus);
+      return;
```
imransl: Think it'll be easier to understand what is going on here if this `return` was replaced by an `else` statement

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -75,13 +75,51 @@ export const TrackPanelListItem = (props: TrackPanelListItemProps) => {
...
+  const updateTrackStates = (
+    trackId: string,
+    newStatus: TrackActivityStatus
+  ) => {
+    if (!activeGenomeId || !activeEnsObjectId) {
+      return;
+    }
+
+    // FIXME: Temporary hack until we have a set of proper track names
+    if (track.track_id.startsWith('track:gene')) {
```
azangru: Ids of these tracks:

image

also start with track:gene (e.g. `track:gene-pc-fwd`), but belong in a different segment of redux state. Isn't it going to be a problem with this approach?

ens-ma7:
Need to start from scratch once multiple transcripts are available. (See jira for updates)

PR #215, opened 2019-11-20
Use window.matchMedia instead of ResizeObserver for window size changes

azangru:
## Type
- Refactoring

## Description
This is a suggestion. Not sure whether good or bad.

Currently, we use ResizeObserver to observe changes in size of the top-level React container. While there is technically nothing wrong with this, there exists a dedicated API `window.matchMedia` that does precisely that.

I am still uncertain whether ResizeObserver completely removes the need for `window.matchMedia` or not, but here's a quick assessment of using this API:

### Pros
- it is more economical — it will fire only when a specific breakpoint is crossed, in contrast to ResizeObserver that fires upon every change in dimensions of an element
- it is probably more idiomatic for answering the question "I wonder which breakpoint does my screen match"; we could thus share breakpoints between js and scss
- it allows detection of switching between landscape and portrait modes

### Cons
- it is somewhat more cumbersome than the ResizeObserver API, which simply returns width and height of the observed element.

### DIFF
--- a/src/ensembl/src/global/globalHelper.ts
+++ b/src/ensembl/src/global/globalHelper.ts
@@ -1,9 +1,11 @@
 import { BreakpointWidth } from './globalConfig';
 
-export const getBreakpoint = (width: number): BreakpointWidth => {
-  if (width > BreakpointWidth.DESKTOP) {
+export const getBreakpoint = (
+  key: keyof typeof BreakpointWidth
+): BreakpointWidth => {
+  if (BreakpointWidth[key] >= BreakpointWidth.DESKTOP) {
     return BreakpointWidth.DESKTOP;
-  } else if (width > BreakpointWidth.LAPTOP) {
+  } else if (BreakpointWidth[key] >= BreakpointWidth.LAPTOP) {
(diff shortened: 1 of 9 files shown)


src/ensembl/src/global/globalHelper.ts
```diff
@@ -1,9 +1,11 @@
...
 
-export const getBreakpoint = (width: number): BreakpointWidth => {
-  if (width > BreakpointWidth.DESKTOP) {
+export const getBreakpoint = (
+  key: keyof typeof BreakpointWidth
+): BreakpointWidth => {
+  if (BreakpointWidth[key] >= BreakpointWidth.DESKTOP) {
     return BreakpointWidth.DESKTOP;
-  } else if (width > BreakpointWidth.LAPTOP) {
+  } else if (BreakpointWidth[key] >= BreakpointWidth.LAPTOP) {
```
imran-ebi: `PHONE` and `BIG_DESKTOP` also need to be here

azangru: You have them in your branch ;-) Perhaps add them when you merge this PR/dev into your branch?

PR #216, opened 2019-11-21
Create 0014-visualisation-library.md

azangru:
## Type
- Documentation

### DIFF
--- a/src/ensembl/doc/architecture/decisions/0014-visualisation-library.md
+++ b/src/ensembl/doc/architecture/decisions/0014-visualisation-library.md
@@ -4,13 +4,13 @@ Date: 2019-11-21
 
 ## Status
 
-Pending
+Accepted
 
 ## Context
 
 While we have a powerful data visualisation tool already in the form of the new genome browser, we have many additional visualisation requirements, particularly in Entity Viewer. These will need to show not only biological objects such as transcripts but also graphs, trees and other complex interactive diagrams.
 
-Developing these types of views by extending the genome browser would be very labour-intensive, since everything would have to be written from scratch. We therefore need a separate library that will enable us to create a wide variety of custom visualisations, with or without animations and interactivity.
+Developing these types of views by extending the genome browser would be very labour-intensive, since everything would have to be written from scratch. We therefore need a separate library/codebase that will enable us to create a wide variety of custom visualisations, with or without animations and interactivity.
 
 ## Decision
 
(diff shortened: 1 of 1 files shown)


azangru:
First, I'll just point out the aspects of the ADR that I completely agree with:

1) We almost certainly don't need anything nearly as complex as genome browser for visualisations in EntityViewer, and none of us is proficient in Rust to the extent of being dangerous. The drawback to that is that we won't have access to Dan's solutions and will have to implement our own; but I think everyone agrees that it's fine.
2) If there are third-party libraries that can help us and that are likely to be re-used, we might use them. D3 is a good candidate for being such a library.

Here's where I have my concerns.

There are two distinct ways of using d3 in a React-based project (and a middle road between them), due to the fact that d3 traditionally works with the actual DOM, while React interacts with a virtual DOM (which then gets rendered to the actual DOM).

### Option 1
One way of co-ordinating these two libraries is to erect a clear-cut boundary between them, such as rendering an empty container with React and then passing it over to d3 to populate. In the extreme case, the d3-based code and the React-based code don't know anything about each other, just like our genome browser and React chrome don't, and communicate with each other over a messaging protocol. The **advantage** of such an approach is its familiarity for people who used to work with d3 rendering directly into the DOM. The **disadvantage** is that React is unaware of what d3 is doing, and there's nothing stopping React from deleting/re-rendering the element used as a container for d3; as well as there's nothing preventing d3 from keeping hold of the element even when it was discounted by React during re-rendering (a potential source of memory leaks).

### Option 2
The opposite way is to keep using React as a renderer, but integrate bits and pieces of d3 into relevant components whenever they become needed. With this approach, d3 is treated more as a collection of useful helper functions than a complete solution with its traditional select-enter-update-exit scenario. As Shirley Wu [put it](http://sxywu.com/react-d3/#/combine), "React renders, D3 calculates". There are several arguments in favour of this approach, namely: a) since version 4, d3, previously a monolith, has been split into smaller modules that can be used independently; b) there are rendering targets other then DOM (for example, canvas), for which this approach with d3 has been successfully used. The **advantage** of this approach is that we keep using idiomatic React, and don't switch renderers. The **disadvantage** is that it may require a deeper knowledge of d3.

### Option 3
There is also a middle road of having React components that, similar to approach 2, have some d3 logic in them, but, similar to approach 1, apply d3 to already mounted components.

## Examples of concrete implementations

### Option 1 (kinda):
- The [TVV widget](https://github.com/Ensembl/private-plugins/tree/release/98/js_widgets/htdocs/Widgets/widgets/TVV). React rendered some kind of a pseudo-DOM element (using [react-faux-dom](https://github.com/Olical/react-faux-dom)) that D3 then [treated](https://github.com/Ensembl/private-plugins/blob/1b505bb6a76fdd52355e0b8191bd37b576c77ed0/js_widgets/htdocs/Widgets/widgets/TVV/src/utils/tvvDiagramUtils.js#L50-L54) as if it were a real DOM. I personally feel very wary of this react-faux-dom thing.

### Option 2:
- Victory charts (e.g. using d3 interpolations: https://github.com/FormidableLabs/victory/blob/6a7b68e2f271dd2ff89ae8b6e7e6daa665f6304f/packages/victory-core/src/victory-animation/util.js)

### Option 3:
- Examples of using React with D3 from Shirley Wu's courses:
  - charts: https://github.com/sxywu/react-d3-example/tree/master/src/visualizations
  - being able to select a portion of the chart with D3 brush https://codesandbox.io/s/k2z4w991l5

imran-ebi:
@azangru Does this PR need to be reviewed by anyone?

azangru:
@imran-ebi sure, why not? This PR is an invitation for discussion about what technologies and how we should use for visualisations in EntityViewer.

ens-ap5:
@imran-ebi We need to have a face-to-face meeting so that we can make an informed decision by consensus, but while we were waiting for you to get back from holiday I thought I'd jumpstart the process by writing an ADR :)

imran-ebi:
@ens-ap5 @azangru When shall we have a meeting about this?

ens-ma7:
Just adding my bit here, i read the article that @azangru suggested and i like the "Lifecycle Methods Wrapping" option. This is exactly how we do it in widgets ensembl with the observer. To me it seems we are keeping to the react framework (no faux dom) as well as using d3. Not sure if we can use the raw d3 command this way (having dom manipulation) but if it is then thats a good way. we can discuss further in the meeting.

PR #217, opened 2019-12-06
Add typescript optional chaining support and update packages

imran-ebi:
## Type
Infrastructure

## Importance
Need this for #205

## Description
TypeScript recently got a minor update (v3.7) and it includes some nice features such as, optional chaining which we can use in our existing code. A good example for using this would be in an `if` condition that checks if an object exists:

```
const obj = {
  childObj: {
    prop: 'some property'
  }
}

// check if props exists:
// the old way
if (obj && obj.childObj && obj.childObj.prop) {}

// the new way
if (obj?.childObj?.prop) {}
```

There is more details on this in the [TypeScript 3.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining).

Also, this PR updates the NPM packages. There was one updated package, `file-loader`, that was causing issues with SVG imports. So I reverted it to its previous version.

### Checks
- `npm t` — works fine
- `npm run check-types` — works fine
- development build — works fine
- production build — works fine


imran-ebi:
There are some errors that I came across while testing again. So marking this as `in progress` till I am done fixing the issues.

imran-ebi:
Turns out, I had switched branches and run `npm start` in the other branch. So when I got back to this branch and just ran `npm run build` it failed. Nothing unexpected there, but a few minutes for me to realise 🤦‍♂

PR #218, opened 2019-12-09
Use a dedicated component for toggle switch in BrowserTrackConfig panel

azangru:
## Type
- Refactoring

## Description
This PR replaces the current element that toggles track names and labels with a dedicated component, which uses a single svg, and also animates the transition between the on and off states.

image

It also relates to @jyothishnt 's branch `replace-svg-with-ImageButton`. Jyo had problems when he replaced the svg representing the toggle — specifically, after that change, clicking on the toggle also automatically closed the config panel (because that click was interpreted as a click outside the panel). This PR makes sure that click on the toggle will not close the config panel.

## Views affected
Genome browser


imransl:
Can we merge this to dev @azangru?

PR #220, opened 2019-12-16
Add standard app layout

azangru:
## Type
- Refactoring

## Description
This PR extracts a common layout pattern into a separate layout component. The component has:
- the main content area
- a top bar
- a side bar
- a toolstrip alongside the sidebar
- (optionally) a drawer

The sidebar opens and closes using CSS transition (which _might_ perform better than react-spring on genome browser page). The drawer also opens and closes using CSS transition (which will almost certainly perform better, because it's based on animating the transform property). When opened, the drawer still leaves a narrow space through which the user can see the main content area.

The layout currently exists only as a Storybook component. If approved, it will be a separate task to integrate it into the app.

The StandardAppLayout component can be viewed in Storybook — either locally, or at https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-layout-standardapplayout--with-drawer

**Note:** the layout component, as it's implemented in this PR, is not responsible for showing the modal inside the sidebar. It is up to the parent component to decide whether it wants to put regular sidebar content or the modal content inside layout's sidebar.

### DIFF
--- a/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
+++ b/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
@@ -11,13 +11,15 @@ $drawerWindowWidth: 45px;
 }
 
 .topBar {
-  display: flex;
+  display: grid;
+  grid-template-columns: 1fr $sideBarContentWidth;
+  grid-column-gap: $sideBarToolStripWidth;
   align-items: center;
   height: $topBarHeight;
   background: $light-grey;
   box-shadow: 0 2px 3px $grey;
   position: relative;
-  padding: 0 18px;
+  padding-left: 18px;
   z-index: 200;
 }
 
@@ -101,6 +103,38 @@ $drawerWindowWidth: 45px;
 
   &Open {
     transform: rotate(180deg);
+  }
+}
+
+.sidebarTabs {
+  display: flex;
+  justify-content: space-around;
+  padding: 0 15px;
+}
+
+.sidebarTab {
+  display: inline-block;
+  color: $blue;
+  cursor: pointer;
+
+  &Active {
+    position: relative;
+    color: $black;
+    cursor: default;
+
+    &:after {
+      content: '';
+      position: absolute;
+      border: 6px solid $grey;
+      border-color: transparent transparent $light-grey $light-grey;
+      box-shadow: -2px 2px 2px 0 $grey;
+      height: 0;
+      right: calc(60% - 6px);
+      transform-origin: 0 0;
+      transform: rotate(-45deg);
+      top: 30px;
+      width: 0;
+    }
   }
 }
 
--- a/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
+++ b/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
@@ -3,6 +3,7 @@
 $sideBarContentWidth: 320px;
 $sideBarToolStripWidth: 46px;
 $topBarHeight: 40px;
+$drawerWindowWidth: 45px;
 
 .standardAppLayout {
   height: 100%;
@@ -59,7 +60,6 @@ $topBarHeight: 40px;
   width: 100%;
   height: 100%;
   display: flex;
-  background-color: rgba(255, 255, 255, 0.8);
   z-index: 1;
   transition: transform 0.3s ease-in-out;
 
@@ -70,8 +70,6 @@ $topBarHeight: 40px;
   }
 
   &DrawerOpen {
-    // DO THIS
-    // transform: translateX(45px);
     transform: translateX(0);
   }
 
@@ -86,6 +84,7 @@ $topBarHeight: 40px;
   width: $sideBarContentWidth;
   border-left: 1px solid $grey;
   padding: 15px;
+  background-color: white;
 }
 
 .sideBarToolstrip {
@@ -125,6 +124,12 @@ $topBarHeight: 40px;
   position: relative;
   padding: 15px 26px 15px 30px;
   border-left: solid $medium-light-grey 1px;
+  background-color: white;
+}
+
+.drawerWindow {
+  width: $drawerWindowWidth;
+  height: 100%;
 }
 
 .drawerClose {
--- a/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
+++ b/src/ensembl/src/shared/components/layout/StandardAppLayout.scss
@@ -10,6 +10,8 @@ $topBarHeight: 32px;
 }
 
 .topBar {
+  display: flex;
+  align-items: center;
   height: $topBarHeight;
   background: $light-grey;
   box-shadow: 0 2px 3px $grey;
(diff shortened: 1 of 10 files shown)


src/ensembl/src/shared/components/layout/StandardAppLayout.scss
```diff
@@ -0,0 +1,162 @@
...
+
+  &FullWidth {
+    margin-right: $sideBarToolStripWidth;
+  }
+}
+
+.sideBarWrapper {
+  position: absolute;
+  top: 0;
+  left: 0;
```
msmanoj: Do we need this `left: 0` here?

src/ensembl/src/shared/components/layout/StandardAppLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+};
+
+StandardAppLayout.defaultProps = {
+  isDrawerOpen: false,
+  onDrawerClose: noop
+};
+
+const SidebarModeToggle = (props: SidebarModeToggleProps) => {
+  const chevronClasses = classNames(styles.sidebarModeToggleChevron, {
+    [styles.sidebarModeToggleChevronOpen]:
```
msmanoj: This is not defined in the styles and also how about just `toggleChevronOpen` if it is required?

azangru: It is defined; otherwise the chevron wouldn't rotate :-)
https://github.com/Ensembl/ensembl-client/pull/220/files#diff-28418f7ce136b1f5ddf0c4bf3d65f653R104-R106

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
azangru: Not as such, no; I added it just for clarity, because I wasn't sure of the default value of the left property in an absolutely positioned element, and wanted to be explicit. I can remove it if you want.

src/ensembl/src/shared/components/layout/StandardAppLayout.tsx
msmanoj: Oh right! I didn't know `sidebarModeToggleChevron{ &Open..` can be used as `sidebarModeToggleChevronOpen`

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
msmanoj: I am quite confused why this is not placing the sidebar completely to the left?

azangru: It is :-)
Then the whole sidebar-cum-drawer block it transform-translated by 100% minus the width of the sidebar to the right

msmanoj: Right! I get it now.

msmanoj:
Should the folder `src/ensembl/src/shared/components/layout` be named as `layouts`  as we will be having more layouts like this?

azangru:
> Should the folder src/ensembl/src/shared/components/layout be named as layouts as we will be having more layouts like this?

I think I have a mild preference towards layout singular, much like the button singular that we already have (but which contains PrimaryButton and SecondaryButton). From the purely syntactic and conventional perspective, I would rather see `import {StandardAppLayout} from '/path/layout'` (similar to `import {BrowserRouter, StaticRouter} from 'react-router-dom'`) than `import {StandardAppLayout} from '/path/layouts'`. But if you have a strong preference of the opposite, we can change layout to layouts.

msmanoj:
I don't strongly prefer one way to the other but there are many folders that are named in plural like services, typings, stories, tests, favicons, fonts and also node_modules. I also noticed that [Ant Design](https://ant.design/components/layout/) and [Bootstrap](https://getbootstrap.com/docs/4.4/layout/overview/) calls this section as just `Layout`. 

We can either keep it consistent within our app or follow what the other frameworks are doing. I guess the former will be slightly better?

azangru:
> there are many folders that are named in plural like services, typings, stories, tests, favicons, fonts and also node_modules

That's right. I believe that's determined by how people think of stuff that goes inside those folders — whether as of different things or as of essentially the same thing. Intuitively, tests, stories, fonts, etc. are things that are very different from one another, as are the contents of the `components` folder, where one "thing" is one component and another "thing" is a different component. But in the case of, say `components/button(s)` or `components/layout(s)` we can think of the stuff inside that folder as being either different things (different buttons, different layouts), or being variations of the same thing (variations of a button, variations of a layout).

I would say, that the way of thinking of similar components as of different varieties of the same component is fairly common among developers who build components (in React or other frameworks). That's why it's common, I believe, to see folder structures that say "components/singular-component-name/different-files-related-to-component". In other traditions it is different. For example, in Jekyll projects there is a dedicated `_layouts` folder — which tells us that Jekyll developers think of layouts as of different individual things, not as variants of the same one.

The naming reflects the way we think about things. I don't believe that the fact that the `tests`, or the `stories`, or the `services` folder are named in plural should have any bearing on how we name a particular component folder. The question to ask here is  — how do we prefer to think about what goes into a given folder: as of many different things or as of a single one.

As a side note, we have a similar situation with files named "helper(s)". Some files are named in singular, some in plural. This may be easier to standardise than folder names.

@imran-ebi do you have any opinion on this?

src/ensembl/src/shared/components/layout/StandardAppLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+    { [styles.mainFullWidth]: !props.isSidebarOpen }
+  );
+
+  const sidebarWrapperClassnames = classNames(
+    styles.sideBarWrapper,
+    { [styles.sideBarWrapperOpen]: props.isSidebarOpen },
+    { [styles.sideBarWrapperClosed]: !props.isSidebarOpen },
+    {
+      [styles.sideBarWrapperDrawerOpen]:
+        'isDrawerOpen' in props ? props.isDrawerOpen : false
```
imran-ebi: Can't we instead use [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)? e.g. `props.isDrawerOpened ?? false`

azangru: 👍
Great idea. This looks as if it were specifically designed for typescript to prevent it complaining of a missing object field when you try to check this field directly 😄

azangru: image

Well, I'll just need to add a babel plugin for that :-)

imran-ebi: Yes, just like for optional chaining. Maybe we can add this with a new PR?

azangru: Added the babel plugin in this PR.

imran-ebi:
@azangru I don't have any strong opinions on whether these directory names need to be singular or plural. Both look fine to me and using them interchangeably doesn't seem to do any harm.

PR #223, opened 2019-12-20
Fix track panel opened on desktop rather than big desktop

imran-ebi:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-454

## Description
The track panel is opened on screens smaller than big desktop. This is not correct according to the design. It needs to be opened only on a big desktop screen or bigger.

## Views affected
Browser and track panel

### Other effects

**UPDATE**: As @azangru mentioned in a [comment below](https://github.com/Ensembl/ensembl-client/pull/223#discussion_r360349866), track panel needs to be closed only for screens smaller than desktop (rather than big desktop). So this PR won't be changing the way the track panel behaves. However, tests have been added to check this and also the incorrect `DESKTOP` breakpoint width value has been fixed.

### DIFF
--- a/src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
+++ b/src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
@@ -33,7 +33,7 @@ describe('<TrackPanel />', () => {
   const defaultProps: TrackPanelProps = {
     activeGenomeId: null,
     browserActivated: false,
-    breakpointWidth: BreakpointWidth.DESKTOP,
+    breakpointWidth: BreakpointWidth.LAPTOP,
     isDrawerOpened: false,
     activeEnsObject: null,
     isTrackPanelModalOpened: false,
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
```diff
@@ -41,19 +41,32 @@ describe('<TrackPanel />', () => {
...
   describe('rendering', () => {
     test('renders track panel when active genome is present', () => {
-      const wrapper = mountBrowserImageComponent({
+      const wrapper = mountTrackPanel({
         activeGenomeId: faker.lorem.words()
       });
       expect(wrapper.html()).not.toBe(null);
     });
 
+    test('shows track panel only if screen is or bigger than big desktop', () => {
```
azangru: "if screen is or bigger than big desktop" -> "if the screen width is big desktop or larger" 😃

src/ensembl/src/content/app/browser/track-panel/TrackPanel.test.tsx
```diff
@@ -41,19 +41,32 @@ describe('<TrackPanel />', () => {
...
+        breakpointWidth: BreakpointWidth.BIG_DESKTOP
+      });
+      expect(wrapper.props().toggleTrackPanel).toHaveBeenCalledWith(true);
+
+      jest.resetAllMocks();
+
+      wrapper.setProps({ breakpointWidth: BreakpointWidth.DESKTOP });
+      wrapper.update();
+      expect(wrapper.props().toggleTrackPanel).toHaveBeenCalledWith(false);
+    });
```
azangru: Sorry, I am confused. I have specifically asked Andrea about our sidebar situation, and she said that sidebar should be open by default at 1200 pixels and greater. Is this test making sure that the sidebar will be closed unless screen width >= 1800? If so, then it goes against Andrea's directions.

imran-ebi: That's not how it is in the XD. Anyway, if Andrea says so then it probably should be opened from 1200px by default.

PR #224, opened 2019-12-23
Add an option to disable opening/closing of accordion items

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-456

## Importance
Needed for the accordions used in the Entity Viewer

## Description
Added the additional prop `disabled` to the AccordionItemButton component that can be used to disable the toggling of the Accordion item.

### Other effects


src/ensembl/stories/shared-components/accordion/Accordion.stories.scss
```diff
@@ -19,3 +19,30 @@
...
+  color: $blue;
+}
+
+.entityViewerAccordionItem {
+  border-color: transparent;
+}
+
+.entityViewerAccordionItem + .entityViewerAccordionItem {
+  border-color: transparent;
+}
```
imransl: Isn't this the same as lines 35-37?

src/ensembl/src/shared/components/accordion/components/AccordionItemButton.tsx
```diff
@@ -1,30 +1,43 @@
...
+    ...rest
+  } = props;
 
   let styles = className;
 
   if (extendDefaultStyles) {
-    styles = classNames(defaultStyles.accordionButtonDefault, className);
+    styles = classNames(
+      defaultStyles.accordionButtonDefault,
+      { [defaultStyles.accordionButtonDefaultDisabled]: props.disabled },
```
azangru: Any reason why the name ends in DefaultDisabled, and not just Disabled?

azangru:
The cursor changes to a pointer when moused over a disabled accordion item. It shouldn't.

src/ensembl/src/shared/components/accordion/components/AccordionItemButton.tsx
```diff
@@ -34,7 +47,7 @@ AccordionItemButton.defaultProps = {
   extendDefaultStyles: true
 };
 
-type WrapperProps = Pick<
+type WrapperProps = { disabled?: boolean } & Pick<
   DivAttributes,
   Exclude<keyof DivAttributes, keyof InjectedButtonAttributes>
 >;
```
azangru: The line below will say `const AccordionItemButtonWrapper: React.SFC<WrapperProps>`. `React.SFC` is a deprecated type (in favour of `FunctionComponent`), and the type annotation of `AccordionItemButtonWrapper` is likely unnecessary.

azangru: P.S.: there are several examples of React.SFC in our codebase — all in accordion component. Could you please weed them out?

src/ensembl/stories/shared-components/accordion/Accordion.stories.scss
msmanoj: Yes, but this is needed to override the default styles.

src/ensembl/src/shared/components/accordion/components/AccordionItemButton.tsx
msmanoj: I was thinking that the disabled button has a default style which can be overridden. I have renamed it to be just Disabled as it sounds better.

src/ensembl/src/shared/components/accordion/components/AccordionItemButton.tsx
msmanoj: That's done

src/ensembl/stories/shared-components/accordion/Accordion.stories.scss
azangru: But is the rule on 35-37 actually doing anything?

msmanoj: No, it doesn't. I've removed it.

PR #229, opened 2020-01-14
Fix drawer alignment

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature
- [x ] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-462

## Description
This is fixing the alignment of the description in the drawer. There are multiple drawers and hence all these files. Also replacing the dl and dd with div.

## Views affected
Genome browser.

### Other effects

## Possible complications
None

### DIFF
--- a/src/ensembl/src/content/app/browser/drawer/Drawer.scss
+++ b/src/ensembl/src/content/app/browser/drawer/Drawer.scss
@@ -6,6 +6,7 @@
 
 .drawerView {
   overflow-y: auto;
+  margin-top: 10px;
 
   h2 {
     display: flex;
@@ -14,30 +15,36 @@
     text-align: center;
   }
 
-  dd {
-    width: 600px;
+  p {
+    margin: 0px;
   }
 
-  label {
+  .clearFix {
+    float: left;
+    width: 600px;
+    margin-bottom: 20px;
+    clear: both;
+  }
+
+  .label {
     float: left;
     font-weight: $light;
     text-align: right;
-    width: 150px;
-  }
-
-  p {
-    margin: 0 0 20px 0;
+    width: 90px;
   }
 
   .details {
     float: right;
     text-align: left;
-    width: 400px;
+    width: 490px;
   }
 
   .mainDetail {
     font-size: 16px;
     font-weight: $bold;
+    display: block;
+    float: left;
+    margin-top: -3px;
   }
 
   .secondaryDetail {
--- a/src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerContigs.tsx
+++ b/src/ensembl/src/content/app/browser/drawer/drawer-views/DrawerContigs.tsx
@@ -4,24 +4,24 @@ import styles from '../Drawer.scss';
 
 const DrawerContigs: FunctionComponent = () => {
   return (
-    <dl className={styles.drawerView}>
-      <dd className="clearfix">
-        <label htmlFor="">Track name</label>
+    <div className={styles.drawerView}>
+      <div className={styles.clearFix}>
+        <div className={styles.label}>Track name</div>
         <div className={styles.details}>
           <p>Contigs</p>
         </div>
-      </dd>
+      </div>
 
-      <dd className="clearfix">
-        <label htmlFor="">Description</label>
+      <div className={styles.clearFix}>
+        <div className={styles.label}>Description</div>
         <div className={styles.details}>
           <p>
             Shows a contiguous stretch of DNA sequence without gaps that has
             been assembled solely based on direct sequencing information.
           </p>
         </div>
-      </dd>
-    </dl>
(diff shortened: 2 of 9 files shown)


src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -14,30 +15,36 @@
     text-align: center;
   }
 
-  dd {
+  p {
+    margin: 0px;
+  }
+
+  .clearFix {
+    float: left;
```
azangru: Please let's not lay this out with floats? Let's use something more robust and better fit for layout, such as CSS grid (see [example on jsfiddle](https://jsfiddle.net/o7y9dw2a/1/)) or flexbox (see [example on jsfiddle](https://jsfiddle.net/rq2vxn7e/)), or even a regular html table?

ens-ma7: Sure thing, Its just that float was already there so i just went with what was there to get it working.Because it would have been a bigger change that's why i didn't use the others. I didn't know it was that bad to use float; i did try moving the layout and it was fine.

ens-ma7:
@azangru i have now replaced all the float with grid. The PR is ready to be reviewed.

src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -14,30 +16,38 @@
...
-    text-align: right;
-    width: 150px;
+  .nextLine {
+    display: block;
+    margin-bottom: 10px;
   }
 
-  p {
-    margin: 0 0 20px 0;
+  .clearFix {
```
azangru: We should rename this class name to something more meaningful. "Clearfix" is a scary name from the past, when developers used floats and made sure the float will not continue to the next line by using the technique they called the "clearfix".

In our case the element with this class is more like a "grid" or a "container" or a "table", or something like that.

src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -14,30 +16,38 @@
...
+    display: grid;
+    grid-template-columns: 100px 1fr;
+    grid-column-gap: 1em;
+    grid-row-gap: 1em;
+  }
+
+  .label {
+    font-weight: $light;
+    text-align: right;
+    width: 90px;
```
azangru: This width, that leaves this empty right corner in the left column of the grid, is confusing and likely unnecessary:
 
Untitled Diagram (1)

If its purpose is to add white space between the label and the details field, then let's do so by increasing the gap between the columns:

```
grid-column-gap: calc(1em + 10px);
```
or something to this effect

src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -14,30 +16,38 @@
...
   }
 
   .details {
-    float: right;
+    display: flex;
     text-align: left;
-    width: 400px;
+    width: 490px;
```
azangru: Why the 490px?

If we want the details field to be smaller, let's do this via the grid itself:

```
something-that-has-display-grid;
max-width: 600px;
```

or something to that effect, if that is indeed what we want.

src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -14,30 +16,38 @@
...
   }
 
   .mainDetail {
     font-size: 16px;
     font-weight: $bold;
+    margin-top: -3px;
```
azangru: This looks fishy.
I believe you are trying to fight the incorrect alignment of two lines with different font sizes. I talked to Andrea, and she suggests to reduce the font-size of this element from 16px to 14px.

ens-ma7: Done, decrease the font size to 14px and alignment is fixed but I found out why the alignment was messed up, its because of the line-height.  In the future if we do want bigger font-size we need to check the line height.

src/ensembl/src/content/app/browser/drawer/Drawer.scss
ens-ma7: Remove width completely, this was from previous commits when using float :)

src/ensembl/src/content/app/browser/drawer/Drawer.scss
ens-ma7: The width again was from previous commits when using float but i also use the width to move the label to the left. Having removed it, it looks ok. we may need to get Andrea to review it and see what she says.

ens-ma7:
@azangru made the suggested changes. can you please have a look and let me know if it is ok.

PR #230, opened 2020-01-14
Update global layout and fix opening and closing of Launchbar

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
- ENSWBSITES-266
- ENSWBSITES-451

## Description
- Updated CSS for layout of the application in general, so that opening and closing of the launchbar does not cause the vertical scrollbar to appear, even when showing/hiding of the sidebar is animated. Note that we eventually **decided to disable animation of this interaction**, partly due to performance reasons (it is an expensive animation, as evidenced by running it on GenomeBrowser page), partly because nobody was particularly excited about it. 
- Removed `react-slidedown` library. Even if we were using the slidedown animation, we can easily recreate it without a third-party library; and now, it seems, we have decided not to animate this interaction at all.
- Fixed showing/hiding of launchbar icons when we open or close the launchbar.

_Note:_ According to Andrea, opening and closing of the launchbar was not a useful interface interaction anyway.

## Views affected
All

Deployed to:
http://193.62.55.91:30150


imransl:
There is a conflict that I wasn't there when I approved this PR. This is a result of adding the Entity Viewer link to the Launchbar. Can you fix this @azangru?

ens-st3:
Looks OK

PR #231, opened 2020-01-14
Refactor image button

jyothishnt:
## Type

- Refactoring

## Related JIRA Issue(s)

https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-471

## Importance
This PR could be reviewed while I continue to work on creating the VisibilityIcon component on the same branch.

## Description
We now have new global STATUS and thus require an ImageButton component refactor.
Also updated storybook to show a single ImageButton example.

## Views affected
All ImageButtons including the eye icon in TrackPanelList.
Includes Bookmarks content and CustomDownload reset icons.

### Other effects
No functionality changes


src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -216,7 +216,7 @@ export const TrackPanelListItem = (props: TrackPanelListItemProps) => {
         </label>
         <div className={styles.ellipsisHolder}>
           <ImageButton
-            buttonStatus={Status.ACTIVE}
+            buttonStatus={Status.UNSELECTED}
```
azangru: this is a bit confusing — if the ellipsis button has only a single status, shouldn't it then be default?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -47,7 +47,10 @@ export const ExampleLinks = (props: ExampleLinksProps) => {
         });
 
         return (
-          <div key={`exampleLinks__${exampleObject.object_id}`} className={styles.linkHolder}>
+          <div
+            key={`exampleLinks__${exampleObject.object_id}`}
```
azangru: I know it’s inherited from previous commit, but I strongly suspect that the key can be simply `exampleObject.object_id`. It just needs to be unique; adding the `exampleLinks` string is not doing anything for that.

Same with `viewedLinks__${previouslyViewedObject.object_id}` below

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -152,7 +158,7 @@ export const TrackPanelBookmarks = (props: TrackPanelBookmarksProps) => {
             {props.previouslyViewedObjects.length > 20 && (
               <span className={styles.ellipsis}>
                 <ImageButton
-                  buttonStatus={Status.ACTIVE}
+                  buttonStatus={Status.UNSELECTED}
```
azangru: or default?

src/ensembl/src/content/app/custom-download/containers/content/attributes-accordion/AttributesAccordion.tsx
```diff
@@ -122,7 +122,7 @@ const AttributesAccordion = (props: Props) => {
         </span>
         <span className={styles.resetIcon} onClick={onReset}>
           <ImageButton
-            buttonStatus={Status.ACTIVE}
+            buttonStatus={Status.UNSELECTED}
```
azangru: Here and below — perhaps default is better?

Maybe the default status should also be styled in ImageButton as having a blue fill then? So that we don't have to write "UNSELECTED", which implies that something can be selected, when the button can exist only in a single state?

src/ensembl/src/header/launchbar/LaunchbarButton.tsx
```diff
@@ -28,7 +28,7 @@ const LaunchbarButton: FunctionComponent<LaunchbarButtonProps> = (
     <ImageButton
       classNames={{
         [Status.DEFAULT]: styles.launchbarButtonImage,
-        [Status.ACTIVE]: styles.launchbarButtonSelectedImage,
+        [Status.UNSELECTED]: styles.launchbarButtonSelectedImage,
```
azangru: That’s weird: why is UNSELECTED paired with the DEFAULT here, and not with SELECTED?

src/ensembl/src/shared/components/image-button/ImageButton.test.tsx
```diff
@@ -73,15 +73,15 @@ describe('<ImageButton />', () => {
 
   describe('prop classNames', () => {
     it('always has the default className applied', () => {
-      const wrapper = mount(<ImageButton buttonStatus={Status.ACTIVE} />);
+      const wrapper = mount(<ImageButton buttonStatus={Status.DEFAULT} />);
 
       expect(wrapper.find(ImageHolder).find('div.default')).toHaveLength(1);
     });
```
azangru: Uh-oh, we might want to change this behaviour

src/ensembl/stories/shared-components/image-button/ImageButton.stories.tsx
```diff
@@ -1,45 +1,116 @@
...
-  });
+const icon = {
+  image: DownloadIcon,
+  imageName: 'ImageButton',
+  imagePath: 'static/img/track-panel/download.svg',
+  imageClass: ''
+};
+
+trackPanelButtonStories.add(icon.imageName, () => {
+  const [buttonStatus, setVisible] = useState(Status.DEFAULT);
```
azangru: `setVisible` -> `setButtonStatus`?

src/ensembl/stories/shared-components/image-button/ImageButton.stories.tsx
```diff
@@ -1,45 +1,116 @@
...
+  const computedStyles = { ...styles };
```
azangru: that's the same as just `styles`

src/ensembl/stories/shared-components/image-button/ImageButton.stories.tsx
```diff
@@ -1,45 +1,116 @@
...
+        </div>
+
+        <div className={classNames(styles.imageCard)}>
+          <div className={classNames(styles.imageHolder)}>
+            <ImageButton
+              buttonStatus={Status.DISABLED}
+              description={'enable/disable'}
+              image={icon.image}
+              classNames={computedStyles}
+            />
```
azangru: Do we still have the strikethrough sign over the disabled button? Are we using it anywhere?

jyothishnt: Not for ImageButton but we have it for VisibilityIcon

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
imransl: I think I was the culprit in adding these prefixes to the key names to fix an issue that occurs in the `TrackPanelBookmarks` tests randomly. I tried to fixing them on the test side of things itself, and the issues goes away. I'll talk to @jyothishnt about this and get those fixes into this PR. As for the key names we are using in the component itself, what do you suggest we use instead @azangru?

azangru: > As for the key names we are using in the component itself

Just the `exampleObject.object_id`?

By the way, I have noticed a number of tests complain (although still passing) about non-unique keys in some arrays of components. I suspect this complaint arises from the fact that we are using `faker.lorem.word` to generate random strings, and faker repeats some words. Perhaps a better idea might be to use `faker.random.uuid` (even `faker.random.word` seems to be more random than `faker.lorem.word`)?

src/ensembl/src/shared/components/image-button/ImageButton.test.tsx
```diff
@@ -73,15 +73,15 @@ describe('<ImageButton />', () => {
...
 
     it('applies the respective className depending on the button status', () => {
-      const wrapper = mount(<ImageButton buttonStatus={Status.ACTIVE} />);
+      const wrapper = mount(<ImageButton buttonStatus={Status.SELECTED} />);
 
-      expect(wrapper.find(ImageHolder).find('div.active')).toHaveLength(1);
+      expect(wrapper.find(ImageHolder).find('div.selected')).toHaveLength(1);
```
imransl: Just `.selected` would do

src/ensembl/src/shared/types/status.ts
```diff
@@ -8,9 +8,9 @@
...
 // please add new variants to the enum in the alphabetical order
 export enum Status {
-  ACTIVE = 'active',
-  DISABLED = 'disabled',
   DEFAULT = 'default',
-  HIGHLIGHTED = 'highlighted',
-  INACTIVE = 'inactive'
+  SELECTED = 'selected',
+  UNSELECTED = 'unselected',
+  PARTIALLY_SELECTED = 'partiallySelected',
```
imransl: `partiallySelected` can also be written using a space as `partially selected`. There is no harm in it since this is a string 😃

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
imransl: Yes, that is one reason why this keeps happening, though I wouldn't rely only on `faker.random.word` to generate a unique ID. `faker.random.uuid` will be much better.

jyothishnt: changed 👍

src/ensembl/src/shared/components/image-button/ImageButton.scss
```diff
@@ -4,20 +4,23 @@
   position: relative;
   font-size: 0;
 
+  svg {
+    width: 100%;
+    height: 100%;
+    transition: opacity 0.4s;
+    -webkit-transition: opacity 0.4s;
```
azangru: In most cases, we shouldn't need vendor prefixes, because we use autoprefixer, which should automatically add them if needed. Also, the `transition` rule likely doesn't need vendor prefix at all.

src/ensembl/src/shared/components/image-button/ImageButton.scss
```diff
@@ -41,28 +44,9 @@
 }
 
 .disabled {
-  cursor: not-allowed;
+  cursor: not-allowed !important;
```
azangru: No way! `important`s are a swear word in modern CSS. Why would we need it here?

Besides, haven't we decided that the disabled class simply has `cursor: default`?

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.scss
```diff
@@ -0,0 +1,61 @@
...
+  }
+  &svg:hover {
+    opacity: 0.8;
+  }
+}
+
+.disabled {
+  cursor: not-allowed;
+  svg {
+    fill: $grey !important;
```
azangru: important — 🙅‍♂

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { ReactComponent as Eye } from 'static/img/track-panel/eye.svg';
+import { ReactComponent as EyePartial } from 'static/img/track-panel/eye-half.svg';
+
+import { Status } from 'src/shared/types/status';
+
+import styles from './VisibilityIcon.scss';
+
+type VisibilityIconStatus =
+  | Status.DEFAULT
```
azangru: Should the visibility icon even have the `default` status? Isn't is just selected-unselected-partiallySelected-disabled?

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+        [Status.DISABLED]: styles.disabled
+      }}
+      onClick={props.onClick}
+    />
+  );
+};
+
+VisibilityIcon.defaultProps = {
+  description: ''
+};
```
azangru: We can remove the default prop and instead have description a nullable property `description?: string` in VisibilityIconProps type.

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.tsx
jyothishnt: True. Its the question of using either default or unselected. Chose unselected. :-)

src/ensembl/stories/shared-components/image-button/ImageButton.stories.tsx
jyothishnt: I will leave it for now. We can decide later. It should be quick to remove it.

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.tsx
jyothishnt: I think we need the DEFAULT as we use it when PARTIALLY_SELECTED is passed as props - use it together with className passed to ImageButton component

src/ensembl/src/header/launchbar/Launchbar.scss
```diff
@@ -82,6 +82,9 @@
     fill: $white;
     background-color: $grey;
   }
+  button {
+    cursor: default;
+  }
```
azangru: What do we need this for?

src/ensembl/src/header/launchbar/LaunchbarButton.tsx
```diff
@@ -29,11 +29,11 @@ const LaunchbarButton: FunctionComponent<LaunchbarButtonProps> = (
...
     <ImageButton
       classNames={{
-        [Status.DEFAULT]: styles.launchbarButtonImage,
-        [Status.ACTIVE]: styles.launchbarButtonSelectedImage,
+        [Status.UNSELECTED]: styles.launchbarButtonImage,
+        [Status.SELECTED]: styles.launchbarButtonSelectedImage,
         [Status.DISABLED]: styles.launchbarButtonDisabledImage
       }}
-      buttonStatus={imageButtonStatus}
+      status={imageButtonStatus as ImageButtonStatus}
```
azangru: Do we know why `as ImageButtonStatus` has to be specified here? `getImageButtonStatus` function returns values of type `ImageButtonStatus` anyway; why do we have to declare ImageButtonStatus here as well?

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.scss
```diff
@@ -0,0 +1,56 @@
...
+    position: absolute;
+    content: '';
+    left: 0px;
+    border-color: #2c343a;
+    border: 2px solid red;
+    border-radius: 50%;
+    height: 100%;
+    width: 100%;
+  }
+}
```
azangru: I still think this 🚫 over the disabled visibility icon drawn by the `before` and `after` pseudoelements is entirely unnecessary.

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { VisibilityIcon } from './VisibilityIcon';
+import { Status } from 'src/shared/types/status';
+
+describe('renders', () => {
+  it('partially selected icon ', () => {
```
azangru: 1) usually, a good practice is to write the description for the test so that it reads as a sentence when read alongside the function name.
`describe('rendering')`
`it('renders partially selected icon')`
`test('button visibility')` (does the same as `it`)

2) Test description says something about the icon, but the actual code of the test checks the class. I am not sure what the test was intended to test, but either the description, or the implementation are lying.

src/ensembl/src/header/launchbar/Launchbar.scss
jyothishnt: Thats because we have pointer for disabled launchbar buttons.

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.scss
jyothishnt: Removed after confirming with Andrea. Thanks

src/ensembl/src/shared/components/visibility-icon/VisibilityIcon.test.tsx
jyothishnt: Thanks Andrey I have changed the description to match the test. Test is to check it it applies proper class when button status is PARTIALLYSELECTED.

src/ensembl/stories/shared-components/image-button/ImageButton.stories.tsx
jyothishnt: Removed.

PR #232, opened 2020-01-15
Update styles for small screens

azangru:
## Type

## Description
- as introduced in https://github.com/Ensembl/ensembl-client/pull/225, mobile devices will show the desktop version of the site, but will scale it down to fill the screen (i.e. no responsive/adaptive mobile-first design)
- html body element has a minimum width of 980 pixels (this is the typical width that mobile devices in portrait mode will scale the site to)
- scrollbar is customised in such a way that should make it always visible on Macs, in the browsers that support such customisation (these are webkit- or blink-based browsers, such as Chrome or Safari)

## Views affected
All

**Deployed to:**
http://193.62.55.91:30603/

### DIFF
--- a/src/ensembl/src/shared/components/app-bar/AppBar.scss
+++ b/src/ensembl/src/shared/components/app-bar/AppBar.scss
@@ -3,6 +3,7 @@
 .appBar {
   display: grid;
   grid-template-columns: minmax(0, 1fr) minmax(250px, auto);
+  align-content: start;
   grid-template-areas:
     'top top'
     'main right';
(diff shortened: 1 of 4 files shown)


imransl:
@azangru I can approve this PR once the conflict is resolved

src/ensembl/src/shared/components/app-bar/AppBar.scss
```diff
@@ -3,6 +3,7 @@
 .appBar {
   display: grid;
   grid-template-columns: minmax(0, 1fr) minmax(250px, auto);
+  align-content: start;
```
azangru: Curiously, this little rule makes a difference on Safari between this (bad, currently in production — notice how species elements are pressed to the grey bar below them):
 
image

and this (better):

image

PR #237, opened 2020-01-21
Add tabs for EntityViewer sidebar

azangru:
## Type
New feature

## Related Jira issue
ENSWBSITES-466

## Description
- Added EntityViewerSidebarTabs component
- Split entity viewer redux state into smaller parts, one of which is `sidebar`
- Added the `sidebar` part of entity viewer redux state, modelled as an object keyed by genome id.

## Behaviour
- Entity viewer shows two tabs ("Overview" and "Publications") above the sidebar, as in the design mockup
- The "Overview" tab is active by default
- User can change the active tab by clicking on it. Other than visual appearance of the tabs, nothing changes externally (but internally, in the app, this is all wired up so the app knows that the sidebar tab has changed)
- If sidebar is closed, then clicking on any sidebar tab will open the sidebar

**Deployed to**
http://193.62.55.91:30704/

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/components/entity-viewer-sidebar-tabs/EntityViewerSidebarTabs.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/components/entity-viewer-sidebar-tabs/EntityViewerSidebarTabs.tsx
@@ -2,9 +2,15 @@ import React from 'react';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 
-import { setSidebarTabName } from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions';
+import {
+  toggleSidebar,
+  setSidebarTabName
+} from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions';
 
-import { getEntityViewerSidebarTabName } from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSelectors';
+import {
+  isEntityViewerSidebarOpen,
+  getEntityViewerSidebarTabName
+} from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSelectors';
 
 import styles from './EntityViewerSidebarTabs.scss';
 
@@ -13,7 +19,9 @@ import { RootState } from 'src/store';
 
 type Props = {
   activeTabName: SidebarTabName | null;
+  isSidebarOpen: boolean;
   setSidebarTabName: (name: SidebarTabName) => void;
+  toggleSidebar: (isOpen?: boolean) => void;
 };
 
 const EntityViewerSidebarTabs = (props: Props) => {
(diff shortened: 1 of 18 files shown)


src/ensembl/src/content/app/entity-viewer/components/entity-viewer-sidebar-tabs/EntityViewerSidebarTabs.tsx
```diff
@@ -0,0 +1,76 @@
...
+  toggleSidebar,
+  setSidebarTabName
+} from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions';
+
+import {
+  isEntityViewerSidebarOpen,
+  getEntityViewerSidebarTabName
+} from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSelectors';
+
+import styles from './EntityViewerSidebarTabs.scss';
```
msmanoj: Styles to the end?

azangru: ✔️

src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts
```diff
@@ -0,0 +1,52 @@
...
+  Action<string>
+>> = (isOpen?: boolean) => (dispatch, getState: () => RootState) => {
+  const state = getState();
+  const genomeId = getEntityViewerActiveGenomeId(state);
+  if (!genomeId) {
+    return;
+  }
+  if (isOpen === undefined) {
+    const isCurrentlyOpen = isEntityViewerSidebarOpen(state);
+    isOpen = !isCurrentlyOpen;
```
msmanoj: This is quite confusing to read. Can `isOpen` be renamed to something like `shouldBeOpen` ?

azangru: `should` is also confusing (if `shouldBeOpen` is false, it does not mean that the thing should be closed).
We can change it to an enum value:

```
enum SidebarPosition = {
  OPEN = 'open',
  CLOSED = 'closed'
};
```
and pass that instead of a boolean. It might be a bit of an overkill though.

msmanoj: If `shouldBeOpen` is false, would it not be easy to consider it as "should not be open"?

azangru: I pushed a commit that uses the Status enum for expressing the open or closed state of the sidebar:
https://github.com/Ensembl/ensembl-client/pull/237/commits/56240d9d14858ca5ef2bf0d1da836749936df309

msmanoj: Cool, I guess you missed my previous comment?

azangru: No I saw it, but if you think that `isOpen` is confusing, then `shouldBeOpen` will probably be just as confusing.

PR #238, opened 2020-01-22
Entity viewer sidebar toolstrip

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-467](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-467)

## Importance
@azangru's work on Entity Viewer might depend on this PR

## Description
This PR adds the sidebar toolstrip to the Entity Viewer. It has also refactored the track panel bar of the browser. Also changed the following to lowercase to fix tests failing: `sideBar`, `topBar`, `toolStrip`.

This PR depends and will need to wait till #231 is merged to `dev`.

## Views affected
Entity Viewer


azangru:
I thought during one of our planning meetings for entity viewer we decided not to abbreviate its name to EV 😕

imransl:
@azangru Oh sorry, I can't recall that. So have we decided to stick with the long names then?

azangru:
> have we decided to stick with the long names then

I remember Ridwan winced when we briefly toyed with the idea of abbreviating app names. Both approaches have their advantages and disadvantages. We might want to briefly collectively discuss it again.

azangru:
> @azangru's work on Entity Viewer might depend on this PR

This PR links tightly to what @jyothishnt is doing in https://github.com/Ensembl/ensembl-client/pull/231. It looks like he is almost there, too. Our approach to ImageButton has changed.

src/ensembl/src/content/app/entity-viewer/components/ev-sidebar-toolstrip/EVSidebarToolstrip.tsx
```diff
@@ -0,0 +1,32 @@
...
+      {evSidebarToolstripConfig.map((item: EVSidebarToolstripItem) => (
+        <div className={styles.sidebarIcon} key={item.name}>
+          <ImageButton
+            key={item.name}
+            buttonStatus={Status.INACTIVE}
+            description={item.description}
+            onClick={noop}
+            image={item.icon}
+          />
+        </div>
```
azangru: 1) compare this with just defining ImageButton properties directly in this file without having a config file:

```
return (
  <>
    <ImageButton
      buttonStatus={Status.INACTIVE}
      description="Search"
      classNames={{[ Status.INACTIVE ]: styles.sidebarIcon }}
      onClick={noop}
      image={SearchIcon}
    />
    <ImageButton
      buttonStatus={Status.INACTIVE}
      description="Bookmarks"
      classNames={{[ Status.INACTIVE ]: styles.sidebarIcon }}
      onClick={noop}
      image={BookmarkIcon}
    />
    ...
  </>
)
```
won't it be easier than having a separate config file?

2) _(definitely for a different PR)_ Looking at the example above, won't you agree that our refactored ImageButton needs another property to style the ImageButton regardless of its status (like, always make it that big)? Now that I think of it, I believe that's what the DEFAULT status could be (ab)used for; but now we are using it for a different purpose.  Perhaps, instead of the `classNames` property it should have a `className` and a `statusClasses` properties?

imransl: The first one is done and agree with the second. What do you think @msmanoj and @jyothishnt?

msmanoj: Are there any other use cases other than `always make it that big`? 
Although it is annoying to have div around the ImageButton, I feel it will be simpler this way rather than having two different class related props without having many use cases.

azangru: Hard to tell whether we'll have other use cases, but I would argue that a prop is simpler than a wrapper div; and that unnecessary DOM nodes are always evil.

msmanoj: It looks like the default classname is always applied regardless of the status we use:
https://github.com/Ensembl/ensembl-client/blob/82540e9fd2a6cb20d8267ff336067b2a44f6085c/src/ensembl/src/shared/components/image-button/ImageHolder.tsx#L16

I guess that can be used to make the button always big?

azangru: > It looks like the default classname is always applied regardless of the status we use:

This is being changed in Jyo's PR: https://github.com/Ensembl/ensembl-client/pull/231

We realised that we are confused by the existing statuses, and tried to associate them with predictable semantics:

**SELECTED:** applied when ImageButton has been pressed and stays pressed
**UNSELECTED:** opposite of SELECTED; indicates that the button (which stays pressed when pressed) is released
**DEFAULT:** applied to buttons that do not keep the pressed state, but are released immediately (e.g. BrowserReset button); or to some weird edge cases (e.g. the visibility button in TrackPanelListItem can be selected, unselected, or partially selected; we use the default status for the partially selected variant)
**DISABLED:** indicates that the button cannot be interacted with

Going in that direction means that default status with its associated classes will not be automatically applied to all image buttons, but only to those that are actually in the DEFAULT status.

msmanoj: Ah right! In that case it makes more sense to introduce another prop.

imransl:
@azangru @msmanoj I have merged the latest dev code that includes Jyo's `ImageButton` refactoring code. Can you please check review this PR again.

src/ensembl/src/content/app/browser/Browser.scss
```diff
@@ -10,7 +10,7 @@
   margin-top: 50px;
   padding-left: 40px;
 
-  &__emptyTopBar {
+  &__emptyTopbar {
```
azangru: out of curiosity, why do you prefer Topbar to TopBar? Sidebar has merged to form a separate word, but I have never seen top bar do the same.

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
 import { connect } from 'react-redux';
 
-import { trackPanelBarConfig, TrackPanelBarItem } from './trackPanelBarConfig';
+import ImageButton from 'src/shared/components/image-button/ImageButton';
+
+import { Status } from 'src/shared/types/status';
```
azangru: Since it's a type, move this import down, before styles.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -52,8 +53,9 @@ const EntityViewer = (props: Props) => {
       {params.entityId ? (
         <StandardAppLayout
           mainContent={<div>Main content is coming...</div>}
-          sidebarContent={<div>Sidebar content is coming...</div>}
-          sidebarNavigation={<EntityViewerSidebarTabs />}
+          sidebarContent={<EntityViewerSidebar />}
+          sidebarNavigation={<div>Sidebar navigation goes here</div>}
```
azangru: `<EntityViewerSidebarTabs />` should stay.

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import noop from 'lodash/noop';
+
+import ImageButton from 'src/shared/components/image-button/ImageButton';
+import { Status } from 'src/shared/types/status';
```
azangru: :arrow_down:

src/ensembl/src/content/app/browser/Browser.scss
imransl: I saw it in a few places when I did a quick Google search. And then there was this for the inspiration: https://trigger.io/modules/topbar/current/docs/index.html

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: Ah good catch. Thanks!

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: ✔️

src/ensembl/src/content/app/browser/track-panel/track-panel-bar/TrackPanelBar.tsx
imransl: ✔️

PR #241, opened 2020-01-28
Make BrowserInfo a shared component

ens-ma7:
…or entity viewer top bar

## Type

- [ ] Bug fix
- [ ] New feature
- [X] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-463

## Description
The browser bar component in genome browser uses the browser info component which contain the feature type information and which can be a shared component and reuse for entity viewer top bar.
The changes made are for making the browserinfo component a shared component.

## Views affected
GenomeBrowser and EntityViewer for later

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
@@ -17,10 +17,10 @@ jest.mock(
   'src/content/app/browser/browser-location-indicator/BrowserLocationIndicator',
   () => () => <div>Browser Location Indicator</div>
 );
-jest.mock('src/shared/components/feature-summary-strip', () => ({
-  GeneSummaryStrip: () => <div>Gene Summary Strip</div>,
-  RegionSummaryStrip: () => <div>Region Summary Strip</div>
-}));
+jest.mock(
+  'src/shared/components/feature-summary-strip/FeatureSummaryStrip',
+  () => () => <div>Feature Summary Strip</div>
+);
 
 describe('<BrowserBar />', () => {
   const defaultProps = {
@@ -49,25 +49,20 @@ describe('<BrowserBar />', () => {
       expect(renderedBrowserBar.find(BrowserLocationIndicator).length).toBe(1);
     });
 
-    test('renders GeneSummaryStrip if focus object is gene', () => {
-      const renderedBrowserBar = mount(
-        renderBrowserBar({ ensObject: createEnsObject('gene') })
-      );
-      expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
-    });
-
-    test('renders RegionSummaryStrip if focus object is region', () => {
-      const renderedBrowserBar = mount(
-        renderBrowserBar({ ensObject: createEnsObject('region') })
-      );
+    test('contains FeatureSummaryStrip', () => {
       expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
     });
   });
 
   describe('behaviour', () => {
-    test('shows FeatureSummaryStrip panel by default', () => {
-      const renderedBrowserBar = mount(renderBrowserBar());
+    let renderedBrowserBar: any;
+
+    test('shows FeatureSummaryStrip when ensObject is not null', () => {
+      renderedBrowserBar = mount(renderBrowserBar());
       expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
+
+      renderedBrowserBar = mount(renderBrowserBar({ ensObject: null }));
+      expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(0);
     });
   });
 });
--- a/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
@@ -1,13 +1,10 @@
 import React from 'react';
 import { mount } from 'enzyme';
-import { BrowserBar, BrowserInfo, BrowserBarProps } from './BrowserBar';
+import { BrowserBar, BrowserBarProps } from './BrowserBar';
 
 import BrowserReset from 'src/content/app/browser/browser-reset/BrowserReset';
 import BrowserLocationIndicator from 'src/content/app/browser/browser-location-indicator/BrowserLocationIndicator';
-import {
-  GeneSummaryStrip,
-  RegionSummaryStrip
-} from 'src/shared/components/feature-summary-strip';
+import FeatureSummaryStrip from 'src/shared/components/feature-summary-strip/FeatureSummaryStrip';
 
 import { ChrLocation } from '../browserState';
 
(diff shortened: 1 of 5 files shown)


src/ensembl/src/shared/components/feature-summary-strip/FeatureSummaryStrip.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import { GeneSummaryStrip, RegionSummaryStrip } from '../feature-summary-strip';
+
+import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
+
+type FeatureSummaryStripProps = {
+  ensObject: EnsObject | null;
```
azangru: This component hardly makes sense if ensObject is `null`. Perhaps better to declare that in `FeatureSummaryStripProps` `ensObject` is always non-nullable, and check for null in the parent component that renders the `FeatureSummaryStrip`?

src/ensembl/src/shared/components/feature-summary-strip/FeatureSummaryStrip.tsx
```diff
@@ -0,0 +1,30 @@
...
+  isDrawerOpened: boolean;
```
azangru: As a component, `FeatureSummaryStrip` does not need to know anything about the drawer — it's not its responsibility. Since the children components have an `isGhosted` prop, I suggest that `FeatureSummaryStripProps` also includes `isGhosted`, which will be set by the parent component, which will be aware of the drawer.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
```diff
@@ -56,21 +53,21 @@ describe('<BrowserBar />', () => {
       const renderedBrowserBar = mount(
         renderBrowserBar({ ensObject: createEnsObject('gene') })
       );
-      expect(renderedBrowserBar.find(GeneSummaryStrip).length).toBe(1);
+      expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
```
azangru: description of the test (string in the `test` function) needs to be updated accordingly

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
```diff
@@ -56,21 +53,21 @@ describe('<BrowserBar />', () => {
...
     });
 
     test('renders RegionSummaryStrip if focus object is region', () => {
       const renderedBrowserBar = mount(
         renderBrowserBar({ ensObject: createEnsObject('region') })
       );
-      expect(renderedBrowserBar.find(RegionSummaryStrip).length).toBe(1);
+      expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
```
azangru: As mentioned in the comment above, this test no longer belongs to BrowserBar tests (but does belong to FeatureSummaryStrip tests).

src/ensembl/src/shared/components/feature-summary-strip/FeatureSummaryStrip.tsx
ens-ma7: I think its just a naming problem if i understand the code correctly; further down isDrawerOpened  is then assigned to isGhosted which whats needed as you suggested. I have just renamed the variable to isGhosted, that should be fine isn't it?

ens-ma7:
@azangru PR has been updated with the changes you suggested. Can you please have  a look. Thanks

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
```diff
@@ -45,32 +37,19 @@ export const BrowserBar = (props: BrowserBarProps) => {
...
-      <BrowserInfo
-        ensObject={props.ensObject}
-        isDrawerOpened={props.isDrawerOpened}
-      />
+      {props.ensObject ? (
+        <FeatureSummaryStrip
+          ensObject={props.ensObject}
+          isGhosted={props.isDrawerOpened}
+        />
+      ) : null}
```
azangru: sorry, here the `null` is unnecessary; we can use the `&&` operator instead of the ternary

```
      { props.ensObject && (
        <FeatureSummaryStrip
          ensObject={props.ensObject}
          isGhosted={props.isDrawerOpened}
        />
      }
```

using `nulls` is
- absolutely important when you do not want the component to render anything at all (react components can't return undefined, but can return null)
- useful when the value you are testing for can be 0, e.g.:

```
{ things.length ? <SomeComponent /> : null }
```

because if we write `{ things.length && <SomeComponent />`, then the component will render `0` if the list of `things` is empty.

Otherwise, when we are just checking whether or not to include something in the JSX depending on a certain condition, it's more customary to use the `&&` operator rather than a ternary.

ens-ma7: Thanks Andrey. From what i understand we use null when we do not want the component to render anything, is that the same as returning blank? Because thats what i want the component to do. 
Also if i use the ternary can you please check if i need to change my test when checking if the component FeatureSummaryStrip is present.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
```diff
@@ -52,25 +49,20 @@ describe('<BrowserBar />', () => {
...
-      expect(renderedBrowserBar.find(GeneSummaryStrip).length).toBe(1);
-    });
-
-    test('renders RegionSummaryStrip if focus object is region', () => {
-      const renderedBrowserBar = mount(
-        renderBrowserBar({ ensObject: createEnsObject('region') })
-      );
-      expect(renderedBrowserBar.find(RegionSummaryStrip).length).toBe(1);
+    test('contains FeatureSummaryStrip', () => {
+      expect(renderedBrowserBar.find(FeatureSummaryStrip).length).toBe(1);
```
ens-ma7: @azangru  just making sure this is still valid if i change the if check to ternary.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
imransl: I'd argue that it is always safe to use the ternary variant as then we wouldn't need to bother what the checked condition would result in i.e. `0` or a valid string.

imransl: I'd argue that it is always safe to use the ternary variant as then we wouldn't need to bother what the checked condition would result in i.e. `0` or a valid string.

azangru: Let's consider several hypothetical scenarios:

```
const Foo = () => {
  return x ? <div>Hello</div> : null;
}
```
in this case, we have to use a ternary with `null`, because if a component should not render anything at all, it must return `null`. Returning `undefined` or `false` is not allowed.

Now compare this with the following:

```
const Foo = () => {
  return (
    <div>
      <p>Hello</p>
      { x &&
        <p>world</p>
      }
    </div>
  );
}
```
in this example, the component will always return a JSX Element (div), which may or may not contain the part "world". In this situation, when we are using a condition to check whether to include a component or not, we prefer the `&&` operator to the ternary syntax.

Now the last example:

```
const Foo = () => {
  return (
    <div>
      <p>Hello</p>
      { x
        ? <p>world</p>
        : <p>universe</p>
      }
    </div>
  );
}
```
Here, the component is always returning a JSX element, but we are using a ternary to decide which of the two possible children to render inside of this component.

Your case is the same as example 2 above.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
azangru: Totally valid :-) The test would have failed otherwise.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.tsx
imransl: Touche. I didn't notice this was in the middle of the JSX code.

src/ensembl/src/content/app/browser/browser-bar/BrowserBar.test.tsx
ens-ma7: Thanks @azangru i have push the ternary change, please have a look and let me know if its all good to merge. I have another PR waiting :)

PR #242, opened 2020-01-28
Create panel and tabs shared component

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-470

## Importance
Required for Entity viewer

## Description
This PR is to create the two shared components called `Panel` and `Tabs` that are currently required for the Entity viewer.

- The `Panel` works and looks more of the way it is required by the Entity viewer.

- The `Tabs` component will look very basic by default and it will have to be customised by using CSS for specific needs. I have added two different styles to the storybook which we will be using for the panel-with-tabs and Entity viewer.

## Views affected
Storybook

### Other effects

### DIFF
--- a/src/ensembl/src/shared/components/panel/Panel.scss
+++ b/src/ensembl/src/shared/components/panel/Panel.scss
@@ -1,0 +1,45 @@
+@import 'src/styles/common';
+
+.panelDefault {
+  position: relative;
+  min-height: 100px;
+  border-radius: 5px;
+  width: 600px;
+  max-width: 100%;
+  background: $white;
+  box-shadow: 1px 1px 4px $shadow-color;
+}
+
+.headerDefault {
+  border-top-left-radius: 5px;
+  border-top-right-radius: 5px;
+  background-color: $light-grey;
+  box-shadow: 0px 1px 3px $medium-light-grey;
+  height: 34px;
+  padding: 4px 0 4px 10px;
+  width: 100%;
+  line-height: 24px;
+}
+
+.bodyDefault {
+  position: relative;
+  padding-right: 10px;
+  padding-bottom: 10px;
+  padding-left: 10px;
+  border-bottom-right-radius: 5px;
+  border-bottom-left-radius: 5px;
+  overflow: auto;
+  height: calc(100% - 54px);
+  margin-top: 10px;
+}
+
+.closeButton {
+  position: absolute;
+  right: 10px;
+  top: 40px;
+  cursor: pointer;
+  img {
+    height: 15px;
+    width: 15px;
+  }
+}
--- a/src/ensembl/src/shared/components/panel/Panel.scss
+++ b/src/ensembl/src/shared/components/panel/Panel.scss
@@ -1,6 +1,6 @@
 @import 'src/styles/common';
 
-.panelDefault {
+.panel {
   position: relative;
   min-height: 100px;
   border-radius: 5px;
@@ -10,7 +10,7 @@
   box-shadow: 1px 1px 4px $shadow-color;
 }
 
-.headerDefault {
+.header {
   border-top-left-radius: 5px;
   border-top-right-radius: 5px;
   background-color: $light-grey;
@@ -21,7 +21,7 @@
   line-height: 24px;
 }
 
-.bodyDefault {
+.body {
   position: relative;
   padding-right: 10px;
   padding-bottom: 10px;
--- a/src/ensembl/src/shared/components/panel/Panel.scss
+++ b/src/ensembl/src/shared/components/panel/Panel.scss
@@ -16,16 +16,15 @@
   background-color: $light-grey;
   box-shadow: 0px 1px 3px $medium-light-grey;
   height: 34px;
-  padding: 4px 0 4px 10px;
+  padding: 4px 10px;
   width: 100%;
   line-height: 24px;
+  overflow: auto;
 }
 
 .body {
   position: relative;
-  padding-right: 10px;
-  padding-bottom: 10px;
-  padding-left: 10px;
+  padding: 0 20px 10px 10px;
   border-bottom-right-radius: 5px;
   border-bottom-left-radius: 5px;
   overflow: auto;
--- a/src/ensembl/src/shared/components/panel/Panel.scss
+++ b/src/ensembl/src/shared/components/panel/Panel.scss
@@ -19,7 +19,6 @@
   padding: 4px 10px;
   width: 100%;
   line-height: 24px;
-  overflow: auto;
 }
 
 .body {
@@ -29,13 +28,13 @@
   border-bottom-left-radius: 5px;
   overflow: auto;
   height: calc(100% - 54px);
-  margin-top: 10px;
+  margin-top: 15px;
 }
 
 .closeButton {
   position: absolute;
   right: 10px;
-  top: 40px;
+  top: 45px;
   cursor: pointer;
   img {
     height: 15px;
--- a/src/ensembl/src/shared/components/panel/Panel.test.tsx
+++ b/src/ensembl/src/shared/components/panel/Panel.test.tsx
@@ -23,8 +23,8 @@ const defaultProps: PanelProps = {
   classNames: panelClassNames
 };
 
-const renderPanel = (props: PanelProps = defaultProps) => {
-  return <Panel {...props} />;
+const renderPanel = (props?: any) => {
+  return <Panel {...defaultProps} {...props} />;
 };
 
 describe('<Tabs />', () => {
@@ -48,7 +48,7 @@ describe('<Tabs />', () => {
   it('displays the string header', () => {
     const stringHeader = faker.lorem.words();
 
-    wrapper = mount(renderPanel({ ...defaultProps, header: stringHeader }));
+    wrapper = mount(renderPanel({ header: stringHeader }));
     expect(wrapper.find('.header').text()).toBe(stringHeader);
   });
 
@@ -58,7 +58,7 @@ describe('<Tabs />', () => {
   });
 
   it('displays the close button only if onClose is set', () => {
-    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    wrapper = mount(renderPanel({ onClose }));
     expect(wrapper.find('.closeButton')).toHaveLength(1);
   });
 
@@ -68,7 +68,7 @@ describe('<Tabs />', () => {
   });
 
   it('calls the onClose function when the close button is clicked', () => {
-    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    wrapper = mount(renderPanel({ onClose }));
     wrapper.find('.closeButton').simulate('click');
 
     expect(onClose).toHaveBeenCalled();
@@ -98,7 +98,7 @@ describe('<Tabs />', () => {
   });
 
   it('applies the passed in closeButton className', () => {
-    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    wrapper = mount(renderPanel({ onClose }));
     expect(wrapper.find('.closeButton').prop('className')).toContain(
       panelClassNames.closeButton
     );
--- a/src/ensembl/src/shared/components/panel/Panel.test.tsx
+++ b/src/ensembl/src/shared/components/panel/Panel.test.tsx
@@ -23,7 +23,7 @@ const defaultProps: PanelProps = {
   classNames: panelClassNames
 };
 
-const renderPanel = (props?: any) => {
+const renderPanel = (props?: Partial<PanelProps>) => {
   return <Panel {...defaultProps} {...props} />;
 };
 
--- a/src/ensembl/src/shared/components/panel/Panel.test.tsx
+++ b/src/ensembl/src/shared/components/panel/Panel.test.tsx
@@ -23,22 +23,20 @@ const defaultProps: PanelProps = {
   classNames: panelClassNames
 };
 
-const renderPanel = (props: PanelProps) => {
-  return <Panel {...props}></Panel>;
+const renderPanel = (props: PanelProps = defaultProps) => {
+  return <Panel {...props} />;
 };
 
 describe('<Tabs />', () => {
   let wrapper: any;
-
-  beforeEach(() => {
-    wrapper = mount(renderPanel({ ...defaultProps }));
-  });
 
   afterEach(() => {
     jest.resetAllMocks();
   });
 
   it('displays the HTML header', () => {
+    wrapper = mount(renderPanel());
+
     expect(
       wrapper
         .find('.header')
@@ -55,6 +53,7 @@ describe('<Tabs />', () => {
   });
 
   it('displays the body content', () => {
+    wrapper = mount(renderPanel());
     expect(wrapper.find('.body').contains(defaultProps.children)).toBeTruthy();
   });
 
@@ -64,6 +63,7 @@ describe('<Tabs />', () => {
   });
 
   it('does not display the close button if onClose is not set', () => {
+    wrapper = mount(renderPanel());
     expect(wrapper.find('.closeButton')).toHaveLength(0);
   });
 
@@ -74,22 +74,27 @@ describe('<Tabs />', () => {
     expect(onClose).toHaveBeenCalled();
   });
 
-  it('applies the passed in panel className', () => {
-    expect(wrapper.find('.panel').prop('className')).toContain(
-      panelClassNames.panel
-    );
-  });
+  it('applies the passed in class', () => {
+    wrapper = mount(renderPanel());
 
-  it('applies the passed in header className', () => {
-    expect(wrapper.find('.header').prop('className')).toContain(
-      panelClassNames.header
-    );
-  });
-
-  it('applies the passed in body className', () => {
-    expect(wrapper.find('.body').prop('className')).toContain(
-      panelClassNames.body
-    );
+    expect(
+      wrapper
+        .find('.panel')
+        .first()
+        .hasClass(panelClassNames.panel)
+    ).toBeTruthy();
+    expect(
+      wrapper
+        .find('.body')
+        .first()
+        .hasClass(panelClassNames.body)
+    ).toBeTruthy();
+    expect(
+      wrapper
+        .find('.header')
+        .first()
+        .hasClass(panelClassNames.header)
+    ).toBeTruthy();
   });
 
   it('applies the passed in closeButton className', () => {
--- a/src/ensembl/src/shared/components/panel/Panel.test.tsx
+++ b/src/ensembl/src/shared/components/panel/Panel.test.tsx
@@ -1,0 +1,101 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import faker from 'faker';
+
+import Panel, { PanelProps } from './Panel';
+
+const panelContent = <p>{faker.lorem.words()}</p>;
+
+const panelHeader = <p>{faker.lorem.words()}</p>;
+
+const onClose = jest.fn();
+
+const panelClassNames = {
+  panel: faker.lorem.word(),
+  header: faker.lorem.word(),
+  body: faker.lorem.word(),
+  closeButton: faker.lorem.word()
+};
+
+const defaultProps: PanelProps = {
+  header: panelHeader,
+  children: panelContent,
+  classNames: panelClassNames
+};
+
+const renderPanel = (props: PanelProps) => {
+  return <Panel {...props}></Panel>;
+};
+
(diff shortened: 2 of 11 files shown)


src/ensembl/src/shared/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+    props.selectTab(tabTitle);
+  };
+
+  return (
+    <div className={tabsContainerClassName}>
+      {Object.values(props.tabs).map((tab: Tab) => (
+        <span
+          className={getTabClassNames(tab)}
+          key={tab.title}
+          onClick={tab.isDisabled ? () => null : () => onTabSelect(tab.title)}
```
imransl: `() => null` can be `noop` from `lodash`? Or you can do an early return inside `onTabSelect` itself. I think we've done the latter in other parts of the site.

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+const tabClassNames = {
+  selected: faker.lorem.word(),
+  default: faker.lorem.word(),
+  disabled: faker.lorem.word(),
+  tabsContainer: faker.lorem.word()
+};
+
+const defaultProps = {
+  tabs: tabsData,
+  selectedTab: tabsData.find((tab) => !tab.isDisabled)?.title || '',
```
imransl: Why should the `selectedTab` be disabled at all?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+  it('adds respective class to the selected tab', () => {
+    const selectedTabIndex = tabsData.findIndex(
+      (tab) => tab.title === defaultProps.selectedTab
+    );
+    expect(
+      wrapper
+        .find('.tab')
+        .at(selectedTabIndex)
+        .prop('className')
+    ).toContain('selected');
```
imransl: Can `hasClass` be used instead of this?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  it(' applies the passed in disabled className', () => {
+    const disabledTabIndex = tabsData.findIndex((tab) => tab.isDisabled);
+    expect(
+      wrapper
+        .find('.tab')
+        .at(disabledTabIndex)
+        .prop('className')
+    ).toContain(tabClassNames.disabled);
```
imransl: Same as a comment I left above. `hasClass` instead?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  it(' applies the passed in tabsContainer className', () => {
+    expect(wrapper.find('.tabsContainer').prop('className')).toContain(
+      tabClassNames.tabsContainer
```
imransl: `hasClass` again?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const defaultProps: PanelProps = {
+  header: panelHeader,
+  children: panelContent,
+  classNames: panelClassNames
+};
+
+const renderPanel = (props: PanelProps) => {
+  return <Panel {...props}></Panel>;
```
imransl: Can this be a self closing component since it is empty?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+describe('<Tabs />', () => {
+  let wrapper: any;
+
+  beforeEach(() => {
+    wrapper = mount(renderPanel({ ...defaultProps }));
+  });
```
imransl: We shouldn't be running this for every test case, as we are remounting the component with different props in some test cases. How about using a render function that would take optional props as arguments to mount instead?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    expect(wrapper.find('.body').contains(defaultProps.children)).toBeTruthy();
+  });
+
+  it('displays the close button only if onClose is set', () => {
+    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    expect(wrapper.find('.closeButton')).toHaveLength(1);
+  });
+
+  it('does not display the close button if onClose is not set', () => {
+    expect(wrapper.find('.closeButton')).toHaveLength(0);
```
imransl: Can we move this assertion to the test case above?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  it('calls the onClose function when the close button is clicked', () => {
+    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    wrapper.find('.closeButton').simulate('click');
+
+    expect(onClose).toHaveBeenCalled();
```
imransl: Shouldn't this be the `onClose` we get as props?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+      panelClassNames.body
+    );
+  });
+
+  it('applies the passed in closeButton className', () => {
+    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    expect(wrapper.find('.closeButton').prop('className')).toContain(
+      panelClassNames.closeButton
+    );
+  });
```
imransl: Shouldn't we merge all these test cases to just one, as they are testing the same thing?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+    );
+    wrapper
+      .find('.tab')
+      .at(unselectedDisabledTabIndex)
+      .simulate('click');
+
+    expect(onTabSelect).not.toBeCalled();
+  });
+
+  it(' applies the passed in default className', () => {
```
imransl: As I've mentioned for `Panel.test.tsx`, these test cases that check if the correct class name is applied can be merged into just one.

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+  it('calls the onTabSelect function when a tab is selected', () => {
+    const unselectedTabIndex = tabsData.findIndex(
+      (tab) => tab.title !== defaultProps.selectedTab && !tab.isDisabled
+    );
+    wrapper
+      .find('.tab')
+      .at(unselectedTabIndex)
+      .simulate('click');
+
+    expect(onTabSelect).toBeCalledWith(tabsData[unselectedTabIndex].title);
```
imransl: Shouldn't we be checking if `props.selectTab` is called instead?

src/ensembl/src/shared/components/panel/Panel.scss
```diff
@@ -0,0 +1,45 @@
...
+  box-shadow: 1px 1px 4px $shadow-color;
+}
+
+.header {
+  border-top-left-radius: 5px;
+  border-top-right-radius: 5px;
+  background-color: $light-grey;
+  box-shadow: 0px 1px 3px $medium-light-grey;
+  height: 34px;
+  padding: 4px 0 4px 10px;
```
azangru: There's a 10px padding on the left, but no symmetrical padding on the right? What happens if the heading occupies the whole header space?

src/ensembl/src/shared/components/panel/Panel.scss
```diff
@@ -0,0 +1,45 @@
...
+  padding-left: 10px;
+  border-bottom-right-radius: 5px;
+  border-bottom-left-radius: 5px;
+  overflow: auto;
+  height: calc(100% - 54px);
+  margin-top: 10px;
+}
+
+.closeButton {
+  position: absolute;
```
azangru: image

I can get this ☝️

src/ensembl/src/shared/components/panel/Panel.tsx
```diff
@@ -0,0 +1,52 @@
...
+    closeButton?: string;
+  };
+  onClose?: () => void;
+};
+
+const Panel = (props: PanelProps) => {
+  const { header, onClose, classNames } = props;
+
+  const panelClassNames = classNames
+    ? classNamesMerger(classNames.panel, styles.panel)
```
azangru: Here, and in other mergers: shouldn't the order of the arguments be the opposite? Shouldn't the class passed as a prop be able to override the default class of the Panel?

src/ensembl/src/shared/components/panel/Panel.test.tsx
imransl: There's something I missed here, which is the whole description of the test cases. What was passed according to it? Can we clearly mention that in the description?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
imransl: What was passed?

src/ensembl/src/shared/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+  selectedTab: string;
+  selectTab: (selectedTab: string) => void;
+};
+
+export const Tabs = (props: TabsProps) => {
+  const [selectedTab, setSelectedTab] = useState(props.selectedTab);
+
+  const getTabClassNames = (tab: Tab) => {
+    const defaultClassNames = classNames(props.classNames?.default, styles.tab);
```
azangru: here and below — same as in comment above about the order of the arguments

src/ensembl/src/shared/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+    disabled?: string;
+    selected?: string;
+    tabsContainer?: string;
+  };
+  selectedTab: string;
+  selectTab: (selectedTab: string) => void;
+};
+
+export const Tabs = (props: TabsProps) => {
+  const [selectedTab, setSelectedTab] = useState(props.selectedTab);
```
azangru: Why would this component keep state of selected tab? The parent knows which tab is selected (and passes it to this component as a prop), and knows what to do when a tab is changed — therefore, it looks like it's the responsibility of the parent to keep state of the selected tab.

src/ensembl/src/shared/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+export type TabsProps = {
+  tabs: Tab[];
+  classNames?: {
+    default?: string;
+    disabled?: string;
+    selected?: string;
+    tabsContainer?: string;
+  };
+  selectedTab: string;
+  selectTab: (selectedTab: string) => void;
```
azangru: `onTabChange`?

src/ensembl/stories/shared-components/panel/Panel.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  selected: styles.selectedTab,
+  default: styles.defaultTab
+};
+const tabs = (
+  <Tabs
+    tabs={tabsData}
+    classNames={tabClassNames}
+    selectedTab={'Proteins'}
+    selectTab={(tab: string) => action('selected-tab')(tab)}
+  ></Tabs>
```
azangru: self-closing

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: I'm getting the first tab on the list that is not disabled and assuming it to be the selected tab.

src/ensembl/src/shared/components/tabs/Tabs.tsx
msmanoj: Done using noop

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: > hasClass again?

That's done

> What was passed?

A random classname for the tabs container

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: > Shouldn't we merge all these test cases to just one, as they are testing the same thing? And also change className in the test case description to class e.g. panel className => panel class.

That's done

>  What was passed according to it?

The class name is what is actually passed in and it is randomly generated using faker.

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/tabs/Tabs.tsx
msmanoj: That's done

src/ensembl/src/shared/components/panel/Panel.tsx
msmanoj: That's right. I had a different issue so I tried the other way around but forgot to revert it back.

src/ensembl/src/shared/components/tabs/Tabs.tsx
msmanoj: That's done

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: Will there be any difference?

src/ensembl/src/shared/components/tabs/Tabs.tsx
msmanoj: Removed the state from the tabs component.

src/ensembl/stories/shared-components/panel/Panel.stories.tsx
msmanoj: That's done

src/ensembl/src/shared/components/panel/Panel.scss
msmanoj: Change the padding to be symmetrical.

>What happens if the heading occupies the whole header space?

I am not sure what is best to do here.
Adding overflow seems to be causing the problem we had here:
https://github.com/Ensembl/ensembl-client/pull/160

src/ensembl/src/shared/components/panel/Panel.scss
msmanoj: Increased the right padding to avoid the overlapping text.

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
imransl: Even though it might not be that different, isn't it what we actually want to know?

msmanoj: Are you suggesting to replace `onTabSelect` with `defaultProps.onTabSelect`  and remove the declaration ` const onTabSelect`?

imransl: No. I am suggesting to check if the `props.selectTab` is called e.g. `expect(wrapper.props().selectTab)`

msmanoj: I doubt if we need to do that since all we care about is whether our callback function is getting called or not.

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const defaultProps: PanelProps = {
+  header: panelHeader,
+  children: panelContent,
+  classNames: panelClassNames
+};
+
+const renderPanel = (props: PanelProps = defaultProps) => {
+  return <Panel {...props} />;
+};
```
imransl: We could improvise this with making the `props` parameter optional and include the `defaultProps` in the component as:

const renderPanel = (props?: PanelProps) => {
  return <Panel {...defaultProps} {...props} />;
};

This way we won't need to pass `defaultProps` along with the `props` every time we call this function.

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  it('calls the onClose function when the close button is clicked', () => {
+    wrapper = mount(renderPanel({ ...defaultProps, onClose }));
+    wrapper.find('.closeButton').simulate('click');
+
+    expect(onClose).toHaveBeenCalled();
+  });
+
+  it('applies the passed in class', () => {
```
imransl: class to classes, since we are checking multiple classes?

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+const tabClassNames = {
+  selected: faker.lorem.word(),
+  default: faker.lorem.word(),
+  disabled: faker.lorem.word(),
+  tabsContainer: faker.lorem.word()
+};
+
+const defaultProps = {
+  tabs: tabsData,
+  selectedTab: tabsData.find((tab) => !tab.isDisabled)?.title || '',
```
azangru: The `|| ''` here is a little bit of a lie: `tabsData` guarantee that there is at least one disabled tab, so the find function will never return null. Would typescript be happy if you just indicated that the selected tab is always a string?

`selectedTab: tabsData.find((tab) => !tab.isDisabled)?.title as string` ?

or

`selectedTab: (tabsData.find(tab => !tab.isDisabled) as {title: string}).title` ?

src/ensembl/src/shared/components/tabs/Tabs.tsx
azangru: Github still shows me the reverse order of the arguments:
`classNames(props.classNames?.default, styles.tab);`

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
msmanoj: `(tabsData.find((tab) => !tab.isDisabled) as Tab).title` seems to work.

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: > This way we won't need to pass defaultProps along with the props every time we call this function.

I guess that's how it currently works as I have used:
 > props: PanelProps = defaultProps

imransl: Not really. You now need to always merge `defaultProps` with `props` before passing it to this function.

src/ensembl/src/shared/components/tabs/Tabs.tsx
msmanoj: I missed just this one place. It's done now

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: That's done

src/ensembl/src/shared/components/panel/Panel.test.tsx
msmanoj: That's updated now

PR #243, opened 2020-01-29
Add entity viewer gene view

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-484](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-484)

## Importance
This PR is high priority as it is required by many other entity viewer tickets for this sprint.

## Description
This is the gene view for entity viewer. I have used CSS grids for this and **have coloured them with background colours and added some height. This is temporary only to make it easier to see how the grid looks. It will be removed once the PR is approved**.

## Views affected
Entity viewer

**Deployed to::**
http://193.62.55.91:30849


ens-st3:
Can I see this on your sandbox IP ?

imransl:
@ens-st3 Sure, the address is: http://172.23.50.50:8080/

azangru:
What worries me is that this layout:

image

Is specific only to the gene view of the EntityViewer. Andrea has mockups of the karyotype view (very preliminary, not in XD) that does not look like this layout. So calling it EntityViewerContent without further specification is likely too broad.

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-content/EntityViewerContent.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.entityViewerTabs {
+  background-color: lightyellow;
+  height: 75px;
+  grid-area: 2 / 2 / 3 / 4;
+}
+
+.entityViewerTable {
+  background-color: lightgreen;
+  height: calc(100vw - 620px);
```
azangru: :question:

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-content/EntityViewerContent.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.viewInLinks {
+  background-color: lightblue;
+  height: 150px;
+  grid-area: 1 / 3 / 2 / 4;
+}
+
+.entityViewerTabs {
+  background-color: lightyellow;
+  height: 75px;
```
azangru: Shouldn't this height be defined entirely by the content of this row? (If so, it's unnecessary)

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-content/EntityViewerContent.scss
```diff
@@ -0,0 +1,38 @@
...
+  display: grid;
+  background-color: $black;
+  grid-template-columns: 1fr 5fr 2fr;
+  grid-template-rows: auto;
+  padding: 15px;
+  width: 100%;
+}
+
+// NOTE:
+// Remove the background colours and the heights once the PR is approved.
```
azangru: Especially heights! :smiley: I started leaving comments to them

msmanoj:
I was thinking it will be better to have a layouts directory within the Entity viewer where we can create different layouts for each views.
Also, the folder `entity-viewer-content` inside components directory, can it not just be `content`?

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-content/EntityViewerContent.scss
imransl: Ah, I mentioned that in the description too. I should have made them bold to be easily noticeable :smile:

imransl:
@azangru I checked with Andrea yesterday and she said that this will be the only layout for the Entity Viewer. I'll check with her again and ask about the layout for the karyotype.

azangru:
I'm really sceptical that there will be any benefit to layout components per se for the contents of EntityViewer. We don't even know whether the layout we are observing on the gene view page will recur in any other views, because we don't have mockups for them.

Perhaps, instead of having layout components (dumb components with slots in which they place whatever they received through different props), we can have a geneview component, a transcriptview component, etc., which will import the components that they need to display (the gene visualisation thingy, the table of transcripts, and so on), and position them accordingly?

If there truly emerges a recurring layout pattern for EntityViewer, we can always refactor...

ens-ma7:
I agree with you @azangru  if we wont have any more entities with the same layout then no point creating a layout but From the sprint planning and design reviews, i know for sure Andrea mention that transcript will be another entity with the same layout and probably variant. I think if there is more than 2 that will share the same layout then it will be good enough to create a layout and the others that divert from this then doesn't use the layout at all. I dont think we should have multiple layout unless there is a strong case for it but just one layout for entities that shared the same layout, the others will have their own layout.
Of course things can change but as agreed in the sprint and the meeting lets carry on with the layout and if needed change as we did for the genome browser and entity viewer layout, we can change it in the next sprint (maybe something to discuss in the next sprint with more people involved).

ens-st3:
@ens-ma7 We've agreed to rename this layout to be more specific for genes, and to revisit whether we actually need to this later if (or much more likely when) we need to

imransl:
@ens-st3 The deployment of this branch is at: http://193.62.55.91:30849

azangru:
One thing to ask Andrea regarding this layout:

image

is whether she wants any kind of max width for her content (i.e. what should happen to her design when the screen is really wide)

ens-ma7:
@ens-st3 as long as the changes are within the scope of the sprint and also this PR thats fine.  To me not going with the layout is basically different from what was agreed in the sprint and this change needs factoring in the next sprint or we can have a meeting to discuss this.

imransl:
I spoke to Andrea today about the entity viewer content to clear some of the doubts we have mentioned in this PR. My findings from the discussion were:

- the entity viewer is going to have many views and gene view is only one of them
- the gene view will most probably not be the same as any other view
- there will be max widths set in the future for individual components but the view area itself will not have a max width

imransl:
@azangru @ens-st3 @msmanoj I have renamed the `EntityViewerContent` component to `GeneView`

src/ensembl/src/content/app/entity-viewer/components/views/gene-view/GeneView.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
azangru: perhaps move the `views` directory to the same level as `components`?

Or do we want a more substantial reshuffling, in a separate PR, to achieve something like:

```
--entity-viewer
    |
    |--views
        |
        |--gene-view
            |
            |--components
            ...
```

msmanoj:
```
--entity-viewer
    |
    |--gene-view
        |
        |--components
```

I vote for this one.

imransl: This looks better and you both agree 😄 So I'll change the directory structure to this.

imransl:
```
--entity-viewer/
    |
    |--entity-viewer-app-bar/
    |--entity-viewer-sidebar/
    |--entity-viewer-sidebar-tabs/
    |--gene-view/
         |
         |--feature-length-axis/
    |--state/
    |--EntityViewer.tsx
```

I suggest we remove the existing `components` folder so that we can include state files inside the view folders.

azangru: image

Perhaps there can be a `components` (or maybe `shared/components`) folder as the first child of `entity-viewer` where these things can go?

imransl: Should they be shared though? These components are going to always be in EV, despite the view changing. On the other hand, if we put them in `components`, then where should we put the views, which happen to be components themselves?

imransl:
I talked with Andrea about the layout and she is happy with it. @azangru @msmanoj If we can resolve the directory structure, then maybe we can get this PR merged within this sprint itself?

src/ensembl/src/content/app/entity-viewer/components/views/gene-view/GeneView.tsx
imransl: After the discussion I had with @azangru, we decided to use the following directory structure:

```
--entity-viewer/
    |
    |--shared/
         |--components/
              |--app-bar/
              |--view-in/
         |--state
         |--helpers
    |--gene-view
         |--components/
         |--state/
         |--helpers
    |--transcript-view
         |--components/
         |--state/
         |--helpers/
```

imransl:
I'll merge the `dev` branch and refactor the directory structure.

imransl:
@azangru I have refactored the directory structure according to what we agreed in our discussion. Can you please review again.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -18,11 +18,12 @@ import {
...
 import { StandardAppLayout } from 'src/shared/components/layout';
-import EntityViewerAppBar from 'src/content/app/entity-viewer/components/entity-viewer-app-bar/EntityViewerAppBar';
-import EntityViewerSidebar from './components/entity-viewer-sidebar/EntityViewerSideBar';
-import EntityViewerSidebarTabs from './components/entity-viewer-sidebar-tabs/EntityViewerSidebarTabs';
-import EntityViewerSidebarToolstrip from './components/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip';
-import EntityViewerTopbar from './components/entity-viewer-topbar/EntityViewerTopbar';
+import GeneView from './gene-view/GeneView';
+import EntityViewerAppBar from './shared/components/entity-viewer-app-bar/EntityViewerAppBar';
+import EntityViewerSidebar from './gene-view/components/entity-viewer-sidebar/EntityViewerSideBar';
+import EntityViewerSidebarTabs from './gene-view/components/entity-viewer-sidebar-tabs/EntityViewerSidebarTabs';
```
azangru: These would probably have to be renamed (entity-viewer-sidebar -> gene-view-sidebar, etc.); but this can be done at our convenience in a subsequent PR.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -0,0 +1,38 @@
...
+  background-color: $black;
+  grid-template-columns: 1fr 5fr 2fr;
+  grid-template-rows: auto;
+  padding: 15px;
+  width: 100%;
+}
+
+// NOTE:
+// Remove the background colours and the heights once the PR is approved.
+// They are for demostration purposes only.
```
azangru: Remember! :smiley:

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: Renamed `EntityViewerSidebar` and `EntityViewerSidebarTabs` in ad59d2b

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
imransl: Done in 7b2ef4d

azangru: 👌

PR #244, opened 2020-01-29
Add BasePairsRuler

azangru:
## Type
- New feature

## Related JIRA Issue(s)
ENSWBSITES-489

## Description
This PR adds an axis (ruler?) component for genomic entities in entity viewer. Currently it’s being used in designs of the gene view of the Entity Viewer. It’ll probably be reused in the transcript view. It’s hard to tell whether other views will have a similar ruler component, and whether their requirements for such a component as regards to how many ticks it should show will be similar.

The rules governing the number of ticks and which ticks are labelled are as follows:

```
1. The ruler starts at 1 and ends at the length of the feature. Both the start and the end positions of the ruler are labelled.
2. Apart from the start and the end positions, there should be at least one label, but no greater than 5 labels
3. There may also be some unlabelled ticks. The total number of ticks (both labelled and unlabelled) between the start and the end positions should not exceed 10.
4. Last tick cannot be labelled if it is at a less than 10% distance from the end of the ruler
5. Ticks can be either:
a) multiplier of the same power of 10 as can be used to represent the length of the feature in the normalized scientific notation, or
b) half of this power of 10
 
example of (a):
for a length of 80792 bp (8.0792 * 10^4)
ticks are multiples of 10^4 (i.e. 10000, 20000, 30000, 50000, 60000, 70000, 80000)
 
example of (b):
for a length of 1000000 bp (10^6)
the respective power of 10 is the same as the length
therefore, we split the length it in half and show a tick at 500000
```

## Views affected
Entity Viewer

## Other effects
This PR introduces d3 into the code. It would be interesting to check with Webpack bundle analyzer whether this will significantly increase the size of our bundle, and whether setting babel-preset-env’s `modules` option to `false`, as described in `https://github.com/babel/babel-loader/issues/521#issue-269118972`, will have any effect of webpack’s tree-shaking capabilities.

## Note
You can run Storybook to check the component in action (in the EntityViewer section).


src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/FeatureLengthAxis.tsx
```diff
@@ -0,0 +1,82 @@
...
+      </text>
+      {ticks.map((tick) => (
+        <g key={tick} transform={`translate(${scale(tick)})`}>
+          <rect className={styles.tick} key={tick} width={1} height={6} />
+          {labelledTicks.includes(tick) && (
+            <text className={styles.label} x={0} y={20} textAnchor="middle">
+              {getCommaSeparatedNumber(tick)}
+            </text>
+          )}
+        </g>
```
msmanoj: Would it not be better to loop through and draw the labelled ticks separately instead of checking each ticks?

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
```diff
@@ -0,0 +1,58 @@
...
+  // use d3 scale to get 'approximately' 10 ticks (exact number not guaranteed)
+  // which are "human-readable" (i.e. are multiples of powers of 10)
+  // and are guaranteed to fall within the scale's domain
+  let ticks = scale.ticks();
+  const length = scale.domain()[1]; // get back the initial length value on which the scale is based
+
+  // choose only the "important" ticks for labelling
+  const exponent = Number(
+    (length.toExponential().match(/e\+(\d+)/) as string[])[1]
+  );
```
msmanoj: Is this same as writing (String(length).length-1) ?

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
```diff
@@ -0,0 +1,58 @@
...
+  const powerOfTen = 10 ** exponent; // e.g. 100, 1000, 10000, etc.
+  ticks = ticks.filter((number) => {
+    // do not add a tick in the end of the ruler (it is handled specially)
+    // and throw away all the possible 'inelegant' intermediate ticks, such as 50, etc.
+    return number !== length && number % powerOfTen === 0;
```
msmanoj: Since we are filtering out ticks such as 50:
For length 500:
Screenshot 2020-01-30 at 12 12 44


For length 1000:
Screenshot 2020-01-30 at 12 13 06

Should the one with length 1000 be having more ticks?

imransl:
>  It would be interesting to check with Webpack bundle analyzer whether this will significantly increase the size of our bundle, and whether setting babel-preset-env’s modules option to false, as described in https://github.com/babel/babel-loader/issues/521#issue-269118972, will have any effect of webpack’s tree-shaking capabilities.

Isn't this already set to `false` in our babel configuration?

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/FeatureLengthAxis.tsx
azangru: Not sure what you mean. Labelled ticks are a subset of ticks, so when we are looping through the ticks we are also checking whether a given tick should also contain the label text next to it within the same `g` group. The x-coordinate of the tick (and the label) is defined by the `g` tag surrounding them both. Not sure what other approach you have in mind and what it will win (apart from some fewer loops caused by the `includes`, but they don't matter in this situation).

msmanoj: Sorry, I thought that the labels can be drawn separately but didn't noice the wrapping `g` tag. It's good 👍

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
```diff
@@ -0,0 +1,58 @@
...
+    const newLabelledTicks = getLabelledTicks(ticks, halvedPowerOfTen, length);
+    if (newLabelledTicks.length > 0 && newLabelledTicks.length < 5) {
+      labelledTicks = newLabelledTicks;
+    }
+  } else if (!labelledTicks.length) {
+    // let's have at least one label, roughly in the middle of the ruler
+    const halvedPowerOfTen = powerOfTen / 2;
+    ticks = [...ticks, halvedPowerOfTen].sort();
+    labelledTicks = [halvedPowerOfTen];
+  }
```
msmanoj: Any chance these can be moved inside `getLabelledTicks`?

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
azangru: I guess so...

But I found a cleaner one: `Math.floor(Math.log10(length))`

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/FeatureLengthAxis.tsx
```diff
@@ -0,0 +1,82 @@
...
+const FeatureLengthAxis = (props: Props) => {
+  const domain = [1, props.length];
+  const range = [0, props.width];
+  const scale = scaleLinear()
+    .domain(domain)
+    .range(range);
+  const { ticks, labelledTicks } = getTicks(scale);
+
+  useEffect(() => {
+    props.onTicksCalculated &&
```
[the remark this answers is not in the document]
azangru: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
azangru: Updated the code so that if the length of feature falls in the the interval between a power of ten and a power of ten plus distance between two ticks (in this case, 1000-1100), the ruler shows intermediate unlabelled ticks as well.

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
azangru: Ok, done.

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/FeatureLengthAxis.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { useEffect } from 'react';
+import { scaleLinear } from 'd3';
+
+import { getTicks } from './featureLengthAxisHelper';
+
+import styles from './FeatureLengthAxis.scss';
```
msmanoj: Styles to the end, please.

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
```diff
@@ -0,0 +1,94 @@
...
+  };
+};
+
+const handleLengthAsPowerOfTen = (
+  ticks: number[],
+  powerOfTen: number,
+  step: number,
+  totalLength: number
+) => {
+  ticks = ticks.filter((number) => {
```
msmanoj: Can `number` be renamed to `tick` as it is making it confusing to read?

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (number === powerOfTen) {
```
msmanoj: Can this ever be true for any ticks other than the last one?

msmanoj:
I am not sure if this needs to be handled specifically but passing length as 1 shows 0.15 as the centre label:
Screenshot 2020-02-06 at 09 41 39

azangru:
Oh, I am sure 1 will break. But I don't think it's a valid input. The ruler, by definition, starts at 1. If you start at 1 and your length is 1, then what can the ruler even mean?

But there is a bug in the interval between 1 and 6 — an extra label appears at 1. I will fix that. Although these lengths are also improbable.

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/featureLengthAxisHelper.ts
azangru: I don't think so. It was definitely intended to check the last tick; so I updated the code to make it explicit.

src/ensembl/src/content/app/entity-viewer/components/feature-length-axis/FeatureLengthAxis.tsx
azangru: ✔️

azangru:
Renamed component to `BasePairsRuler`

PR #247, opened 2020-02-03
Add tracking for browser navigation inputs

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-379](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-379)

## Description
Add google analytics tracking for browser navigation. There were three areas that were covered:

- Browser navigation opening
- Browser region editor usage
- Browser region field usage

The questions that need to be answered by the GA events are:
- How often do the users open the navigation bar?
- How often are the users using the region field and region editor?

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
+++ b/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
@@ -29,6 +29,8 @@ import {
 } from 'src/shared/helpers/numberFormatter';
 import { validateRegion, RegionValidationErrors } from '../browserHelper';
 
+import analyticsTracking from 'src/services/analytics-service';
+
 import applyIcon from 'static/img/shared/apply.svg';
 
 import styles from './BrowserRegionEditor.scss';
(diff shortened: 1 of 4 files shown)


azangru:
@imran could you please formulate the questions that should be answered using these analytics events?

imransl:
@andrey I have added them into the description

azangru:
> How often do the users click on the 'change' link?

> Are the start and end inputs in the region editor used a lot?

How is this tracked in the PR?

imransl:
> > How often do the users click on the 'change' link?
> 
> > Are the start and end inputs in the region editor used a lot?
> 
> How is this tracked in the PR?

@azangru I am not tracking the `change` link click, since the region field and region editor are already tracked. The exception though, would be for users of smaller screens who won't see the `change` link, but in that case we will still know that they are accessing the site via a small device.

As for the second question, this is not done properly yet.

azangru:
I am confused.

If the PR is not addressing questions 2 and 5, then why include them in the description?

imransl:
These were the questions mentioned in the Google doc that we used to discuss what to find out with analytics. So I didn't change them based on what is included in this PR yet.

azangru:
Let's leave in the description only those questions that are answered by this PR (or actively ~cross out~ the ones that are not addressed). This helps the reviewer check whether the PR achieves what it set out to achieve.

ens-st3:
What questions does Andrea want data for ?

imransl:
After having a chat with Andrea today morning, I found out there actually only two things we need to find out. I have updated the description accordingly.

imransl:
@azangru @ens-st3 @msmanoj I have updated the code as well. Can you please review again.

PR #248, opened 2020-02-05
Adding content and style to scale switcher

ens-ma7:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-464](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-464)

## Description
This is to show the scale switcher which is a dummy component (no interaction at all and fixed for now). Also styling the element.

## Views affected
Entity Viewer


azangru:
Formatting issue:

image

I think our editors add an empty string to the end of the file by default. We probably should all do it.

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.tsx
```diff
@@ -5,9 +5,15 @@ import styles from './ScaleSwitcher.scss';
...
   return (
     <div className={styles.container}>
-      This is placeholder for scaleswitcher
+      <div className={styles.emptyNode}></div>
+      <div className={styles.emptyNode}></div>
+      <div className={styles.emptyNode}></div>
+      <div className={styles.selectedNode}>Gene</div>
+      <div className={styles.emptyNode}></div>
+      <div className={styles.emptyNode}></div>
+      <div className={styles.emptyNode}></div>
```
azangru: 1) I think we prefer self-closing tags if they do not have any content:

`<div className={styles.emptyNode}></div>` -> `<div className={styles.emptyNode} />`

2) If we want to be slightly smarter about the repeated elements (not sure whether we want to, since this component will be changed significantly anyway), we could do something like this:

```
const renderEmptyDots = (number: number) => {
  return [...Array(number)].map((_, index) => <div key={index} className={styles.emptyNode} />;
}
```
and then:

```
<div className={styles.container}>
  {renderEmptyDots(3)}
  <div className={styles.selectedNode}>Gene</div>
  {renderEmptyDots(3)}
</div>
```

ens-ma7:
All done, have a look @azangru

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
```diff
@@ -1,3 +1,25 @@
...
+  color: $white;
+  text-align: center;
+  cursor: default;
+}
+
+.emptyNode {
+  border-radius: 10px;
+  background-color: $grey;
+  margin: 6px 24px 0 0;
+  display: block;
```
azangru: Empty node is a div, right? So it will have display: block by default, which makes this rule redundant.

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
```diff
@@ -1,3 +1,25 @@
...
+  padding: 1px 20px 0 20px;
+  margin-right: 24px;
+  background-color: $black;
+  color: $white;
+  text-align: center;
+  cursor: default;
+}
+
+.emptyNode {
+  border-radius: 10px;
```
azangru: 100%?

src/ensembl/src/content/app/entity-viewer/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
```diff
@@ -1,3 +1,25 @@
...
+
+.selectedNode {
+  border-radius: 12px;
+  height: 24px;
+  padding: 1px 20px 0 20px;
+  margin-right: 24px;
+  background-color: $black;
+  color: $white;
+  text-align: center;
+  cursor: default;
```
azangru: isn't it the default?

ens-ma7: It becomes a text cursor when you hoover over the text within the div, if you hoover within the div but not on the text then it behaves default. We wanted it to be default on the text as well. I have pushed the changes for the above.

PR #249, opened 2020-02-07
Genome browser link component for entity viewer

jyothishnt:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-483](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-483)

## Description
Creating View in Genome Browser link component for Gene entity viewer.

## Views affected
Entity viewer


imransl:
@jyothishnt Can you add this component into the gene view?

jyothishnt:
> @jyothishnt Can you add this component into the gene view?

Hi @imransl I have split the task into 2. One to create component and another to link this component to geneview. So this work is meant to create just the component.

imransl:
@jyothishnt You can still use this PR to complete both the tasks. It's totally up to you 😉

imransl:
@jyothishnt Can you take a look at this PR again?

jyothishnt:
Ready for review. Expecting many changes :-)

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -51,7 +51,7 @@ export type EntityViewerParams = {
 };
 
 const client = new ApolloClient({
-  uri: 'http://193.62.55.158:30466'
+  uri: 'http://localhost:5000'
```
imransl, line 54: **IMPORTANT**: Don't forget to remove this before merging this PR.

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const Apps = {
+  gb: {
+    tooltip: 'Genome Browser',
+    icon: BrowserIcon
+  },
+  ev: {
+    tooltip: 'Entity Viewer',
+    icon: EntityViewerIcon
+  }
```
azangru, line 21: Ew. I would definitely prefer `genomeBrowser` and `entityViewer` to `gb` and `ev`. Would love to hear what others have to say.

src/ensembl/src/shared/components/view-in-app/ViewInApp.scss
```diff
@@ -0,0 +1,22 @@
+@import 'src/styles/common';
+
+.viewIn {
+  color: $grey;
+  width: 50px;
+}
+.entityViewerAppLinkButtons {
```
azangru, line 7: If ViewInApp is a generic component, why do its classes have "entityViewer" in their names?

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import ViewInApp from './ViewInApp';
+import * as urlFor from 'src/shared/helpers/urlHelper';
+
+jest.mock('react-router-dom', () => ({
+  Link: () => <div>Link</div>
+}));
+
+describe('<ViewInApp />', () => {
+  it('contains Link and ImageButton', () => {
```
azangru, line 13: The description says, "and ImageButton"; the body of the test does not check anything about the ImageButton.

But more to the point, what is the test actually trying to check? If I were writing the tests for this component, I would probably be interested in which buttons are rendered in response to which props, and what urls are passed to what buttons.

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const ViewInApp = (props: ViewInAppProps) => {
+  return (
+    <div className={styles.entityViewerAppLinkButtons}>
+      <span className={styles.viewIn}>View in</span>
+      {Object.keys(props.links).map((appId) => {
+        const id = appId as keyof urlObj;
+        return (
+          <AppButton
+            key={'AppButton_' + appId}
```
azangru, line 41: The appId itself is unique; the `'AppButton_'` portion of the key is redundant.

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
```diff
@@ -0,0 +1,68 @@
...
+export type urlObj = {
+  gb: string;
+  ev: string;
+};
+
+export type ViewInAppProps = {
+  links: Partial<urlObj>;
+};
+
+const ViewInApp = (props: ViewInAppProps) => {
```
azangru, line 33: should probably return null if there aren’t any keys in props.links object.

azangru:
Suggestions re types in ViewInApp.tsx

1) Add a type (I am calling it AppName; but it could just as well be called AppId) that is:

`type AppName = keyof typeof Apps;`

2) Change urlObj type to:

`export type urlObj = Record<AppName, string>;`

(which means you do not have to repeat keys from the Apps constant in your urlObj type

3) change this:

```
      {Object.keys(props.links).map((appId) => {
        const id = appId as keyof urlObj;
        return (
          <AppButton
            key={'AppButton_' + appId}
            appId={appId}
            url={props.links[id] as string}
          />
        );
```

to this:

```
      {(Object.keys(props.links) as AppName[]).map((appId) => {
        return (
          <AppButton
            key={appId}
            appId={appId}
            url={props.links[appId] as string}
          />
        );
      })}
```

reason: you won't need to define a superfluous variable id that is just a copy of appId

4) In type `AppButtonProps` now change `appId` from `string` to `AppName` (reason: appId is not just any string, but one of a closed set of strings

5) Now you can safely delete `as keyof urlObj` from the props you are passing to ImageButton.

src/ensembl/src/shared/components/view-in-app/ViewInApp.scss
jyothishnt: True. Thanks

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
jyothishnt: Yes test needs improvement. Will ping you when I start working on this.

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
```diff
@@ -1,18 +1,58 @@
...
-    const url = urlFor.browser();
-    const wrapper = mount(<ViewInApp links={{ genomeBrowser: url }} />);
-    expect(wrapper.find(Link)).toHaveLength(1);
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('checks if correct url is passed when image button is clicked', () => {
+    const wrapper = renderComponent({ links: linkUrl });
+    expect(wrapper.find(ImageButton)).toHaveLength(2);
```
azangru, line 42: `.toHaveLength(Object.keys(linkUrl).length)`?

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
```diff
@@ -1,18 +1,58 @@
...
+    ...defaultProps,
+    ...props
+  };
+
+  const renderedComponent = <ViewInApp {...completeProps} />;
+
+  return mount(renderedComponent);
+};
+
+const linkUrl = {
```
azangru, line 25: I can't help feeling that the name should be in plural (linkUrls, and linkIcons below).

azangru:
I pushed a commit to update the test file:

- Instead of redefining `linkUrls` as a completely new object in the test file, I am generating it from the `Apps` config imported from `ViewInApp` component
- I randomised the number of links passed to the `ViewInApp` component. It should correctly handle different number of links
- I am afraid we can't really test the icons, because the svgr mock file, from which they are imported, defines these mocks simply as a string `'icon-mock'`. Which means that all svg icons will be represented by the same string, `'icon-mock'`; and there is no point in making any assertions about them.

With that, I am happy to approve the PR.

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
```diff
@@ -22,37 +26,45 @@ const renderComponent = (props: Partial<ViewInAppProps>) => {
...
 
-const linkUrl = {
-  genomeBrowser: faker.internet.url(),
-  entityViewer: faker.internet.url()
-};
+const appLinkTuples = Object.keys(Apps).map(
+  (appName) => [appName as AppName, faker.internet.url()] as const
+);
+const randomSampleSize = Math.ceil(Math.random() * appLinkTuples.length);
+const tuplesSample = sampleSize(appLinkTuples, randomSampleSize);
```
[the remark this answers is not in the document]
azangru: > but is it really needed here? Why complicate? There wont be many such apps.

What an exhaustive test would prove is that, for two applications:
- when you pass just the genome browser link, the component renders only the genome browser button
- when you pass just the entity viewer link, the component renders only the entity viewer button
- when you pass both links, the component renders both buttons

The number of combinations grows when you add more buttons (let's say, we add sequence viewer):
- only genome browser link — render only the genome browser button
- only entity viewer link — render only the entity viewer button
- only sequence viewer link — render only sequence viewer button
- genome browser link and entity viewer link — render genome browser button and entity viewer button
- genome browser link and sequence viewer link — render genome browser button and sequence viewer button
- entity viewer link and sequence viewer link — render entity viewer button and sequence viewer button
- genome browser link, entity viewer link and sequence viewer link — render all three buttons

These tests are tedious to write manually. Are all of them required? I don't know, but it's certainly nice. The random function ensures that over time all combinations will be tried. The problem with the random function, of course, is that the test becomes non-deterministic; every time you run it you have a different set of buttons. We can, of course, write a function that generates all possible combinations of links and runs assertions against all of them, which will make the test run a bit longer. I don’t know which trade-off is more preferable here.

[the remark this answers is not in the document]
jyothishnt: okay let us merge this as it doesn't do any harm and come back once we decide to do more rigorous testing.

PR #250, opened 2020-02-09
Define & Use templates and improvement in CI/CD pipeline

kamaldodiya:
## Type

- [ ] Bug fix
- [ ] New feature
- [x] CI/CD Pipeline
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-505

## Importance
Define & Use template for CI/CD jobs 

## Description
The use of templates improves consistency and readability.


.gitlab-ci.yml
```diff
@@ -30,50 +126,45 @@ Test:
...
-  - cd src/ensembl
-  - npm ci
+  variables:
+    GOOGLE_ANALYTICS_KEY: ${GA_KEY}
+    ENVIRONMENT: internal
+    API_HOST: ""
 
   script:
   - npm run test
   - export NODE_ENV=development
```
azangru: This should be production; but since this job extends `.build-static`, do we even need the `script` block?

.gitlab-ci.yml
```diff
@@ -30,50 +126,45 @@ Test:
...
-  - export GOOGLE_ANALYTICS_KEY=${GA_KEY}
-  - export ENVIRONMENT=internal
-  - export API_HOST=""
   - npm run build
 
   artifacts:
-    name: static_assets_internal
+    name: static_assets
     paths:
     - src/ensembl/dist/
```
azangru: Below, in `Test_N_Build:feature`, `NODE_ENV=development` should also be replaced with `NODE_ENV=production`.

.gitlab-ci.yml
kamaldodiya: If NODE_ENV is production while building for internal deployment then yes we don't need script block in Test_N_Build:internal job.

.gitlab-ci.yml
kamaldodiya: I would be doing another major refactoring for Feature related jobs later as this will need redeployment of all the feature branch with different tags. This may hamper the development so I will be doing it later.

PR #251, opened 2020-02-10
Transcript image for entity viewer

azangru:
## Type
- New feature

## Related JIRA Issue(s)
[ENSWBSITES-490](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-490)

## Description
This PR adds visualisation for a transcript, according to the design:

image

- a transcript shows exons as coloured rectangles
- parts of the exons that lie outside the coding sequence are shown as empty rectangles; parts that are lying within the coding sequence are shown as filled rectangles

**Note:**
- The data for drawing transcripts in this PR is pulled from Ensembl's REST api. Adaptor functions are added to transform the response from REST api into the shape that will likely be returned from the next-generation api.
- We have a toy graphql server deployed; so we will likely delete the REST-related code in subsequent PRs _(probably should keep the REST-related code in this PR though, because switching to graphql will raise questions of whether to use dedicated client libraries for fetching a caching the data)_
- Types of entities with a shape approximating what is going to be returned from the next-generation api are recorded in files in the `entity-viewer/types` folder
- The code in this PR **will not work correctly** for transcripts that are on the reverse strand or on circular DNA. That's all right though, because we hope to get the server to send appropriately recalculated data so that we won't have to independently compute positions of things in the client.

## Views affected
- Storybook / Entity Viewer

The new component can be viewed in Storybook under EntityViewer/transcripts


src/ensembl/src/content/app/entity-viewer/components/transcript-visualisation/TranscriptVisualisation.scss
```diff
@@ -0,0 +1,7 @@
+.containerSvg {
+  overflow: visible;
+}
+
+.emptyBlock {
+  fill: none !important;
```
azangru: Normally, I am dead set against `!important`, but in this case it looks like the only mechanism to override fill property that might be coming from the parent (it doesn't make sense for the parent to also pass a custom `emptyBlock` class, because empty blocks will always have the same `fill: none` rule. Either use `!important` in a css file, or an inline style rule in a react component.

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-adaptors/rest-transcript-adaptor.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
+import { Strand } from 'src/content/app/entity-viewer/types/strand';
+import { Exon } from 'src/content/app/entity-viewer/types/exon';
+import { CDS } from 'src/content/app/entity-viewer/types/cds';
+
+// transform ensembl rest /overlap data into a transcript data structure
+export const restTranscriptAdaptor = (
+  transcriptId: string,
+  data: FeatureInResponse[]
+): Transcript => {
```
imransl: Can we omit mentioning the return type here as `buildTranscript` will anyway return a `Transcript`?

src/ensembl/src/content/app/entity-viewer/types/slice.ts
```diff
@@ -0,0 +1,22 @@
+import { Strand } from './strand';
+
+export type Slice = {
+  location: {
+    start: number;
+    end: number;
+    // length: number,
```
imransl: Can this be removed?

src/ensembl/stories/entity-viewer/transcripts/transcriptData.ts
```diff
@@ -0,0 +1,35 @@
...
+  id: string
+): Promise<Gene | Transcript | null> => {
+  const featureType = await getFeatureType(id);
+
+  if (featureType === FeatureType.Gene) {
+    return fetchGene(id);
+  } else if (featureType === FeatureType.Transcript) {
+    return fetchTranscript(id);
+  } else {
+    console.log(`${id} is not a valid id`); // eslint-disable-line no-console
```
imransl: Shouldn't we use `console.error` here instead? If we do so then we won't need the disable comment.

src/ensembl/src/content/app/entity-viewer/types/slice.ts
azangru: It could... At the moment, I think, Kieron’s implementation has the `length` property in the `location` object; but I feel `length` is a property of the feature itself, not of location. Eventually, after further discussions, this property will have to end up either in the feature itself or in its location field; but in the meantime we can of course delete it.

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-adaptors/rest-transcript-adaptor.ts
azangru: ✔️

src/ensembl/stories/entity-viewer/transcripts/transcriptData.ts
azangru: ✔️

src/ensembl/src/content/app/entity-viewer/types/slice.ts
azangru: removed the comment

src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+import { getFeatureCoordinates } from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
+
+import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
+import { Exon } from 'src/content/app/entity-viewer/types/exon';
+import { CDS } from 'src/content/app/entity-viewer/types/cds';
+
+import styles from './UnsplicedTranscript.scss';
+
+export type Props = {
```
msmanoj: Since we are exporting this, would it be better to rename this to `UnsplicedTranscriptProps` ?

azangru: ✔️ 

turned out these props aren't imported anywhere; so technically they don't even need to be exported; but I suppose it wouldn't hurt.

msmanoj:
Not sure if you wanted to have a look, but the empty blocks are getting slightly overlapped like shown below:
Screenshot 2020-02-24 at 11 59 39

Rest of the PR looks good.

azangru:
@msmanoj thanks for spotting this. Fix pushed; here is what it looks like now:

image

imransl:
This PR was merged without approval. Is this by accident?

imransl:
Manoj told me that this was approved in a comment above. We probably should explicitly approve the PR as well to avoid confusion like this in the future.

PR #254, opened 2020-02-18
ADR for auxiliary content

ens-ap5:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-512](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-512)

## Importance
Required fairly soon so that the above epic can be moved forward.

## Description
We need some way of managing auxiliary content (help, etc) for the MVP, and this ADR describes our requirements and proposed technical solution.

## Views affected
N/A - just documentation.

### Other effects
None.

## Possible complications
None.


azangru:
It would be so sad to give away control over our content to a third-party commercial platform-as-a-service 😞 

I wonder if the option of storing auxiliary content in markdown files in a github repo is still worth exploring. This would require splitting the problem into several:

1) Saving data to markdown files (can content managers write/be taught to write markdown? willl they need a custom graphical interface, such as Netlify CMS, or can they be taught to do it in their text editors?)
2) Retrieving data from markdown files (we would probably write our custom server for that)
3) Indexing markdown files for search (example: someone who uses Hugo as static-site generator wrote a [package](https://blog.travismclarke.com/project/hugo-elasticsearch/) for generating Elasticsearch index from source files)

I don't know whether this is a horrible idea, and how much engineering it will require; but it would ensure that we own our data.

ens-ap5:
Final draft written, based on testing and discussion within the team.

PR #255, opened 2020-02-19
Add license header to source script files

azangru:
## Type
:nauseated_face:

## Related JIRA Issue(s)
ENSWBSITES-513

## Description
- Adds a script for adding legal header to source files. Due to the used comment syntax, files are expected to be javascript or typescript.
- Script is accompanied with tests.
- Contains potential to replace or remove the license header (see license-manager.js), although no actual scripts using this functionality yet.
- Script will be run on every pre-commit hook.

I am sorry :disappointed: 

## For discussion
We could extract this script into a separate package and put it in its own git repo. The logic for this script is pretty reusable.

## Possible complications
Source files get burdened with a legal header.


imransl:
Are we going to use the American spelling of `license` instead of `licence` (British spelling)? 🤔

azangru:
Nah, we must change to British spelling 😞

azangru:
Curiously though, the header itself has the word License spelled with an "s".

imransl:
I am not sure if we need to change what is mentioned in the licence notice. The current website code used the same licence notice that was available in the Apache site. For example check the notice in this file: https://github.com/Ensembl/ensembl-webcode/blob/release/99/modules/EnsEMBL/Web/DOM.pm

azangru:
@ens-st3 @andrewyatz do we change the text of the licence header to be British?

@andrewyatz could you please confirm that [this](https://github.com/Ensembl/ensembl-client/blob/94c1b0ed790aa6c919fcd8952c63c2faa161202d/src/ensembl/scripts/legal/licence-header.txt) is the copy of the header that you would like us to append to all script files.

andrewyatz:
We should maintain the license text that Apache uses. So no changes to the text please.

@azangru that looks like the correct text to me

imransl:
@azangru Can we merge this PR?

azangru:
Two things I wanted to do before merging:

1) Decide (collectively? 😄 ) whether to use `@license` tag on the first line as JSDocs [suggests](https://jsdoc.app/tags-license.html)

Google [does this](https://github.com/Polymer/polymer/blob/master/lib/elements/array-selector.js), while Facebook [doesn't](https://github.com/facebook/react/blob/master/packages/react/index.js). I've heard some IDEs can automatically collapse the license comment — perhaps identifying is by the `@license` tag? Probably worth adding it if it's so.

2) Run the script for all our ts and tsx files (`find  ./src -name "*.ts" | xargs node scripts/legal/add-license.js`) and accept that first in a separate PR.

imransl:
@azangru The first one looks like a good one to have especially if our IDEs hides it away. As for the second, is `*.ts` enough to identify tsx files as well?

azangru:
Is superseded by https://github.com/Ensembl/ensembl-client/pull/299

PR #256, opened 2020-02-20
Transcript filter for EV

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-501

## Description
- Created an entity viewer specific component to implement the transcript filters. 
- A storybook entry has been added alongside EntityViewer -> BasePairRuler.

## Views affected
Entity viewer

### Other effects


src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  label: string;
+};
+
+export type OptionsGroup = Option[];
+
+type Props = {
+  selectedValues: string[];
+  onChange: (selectedValues: string[]) => void;
+  options: OptionsGroup[];
+  hideUnchecked?: boolean;
```
azangru: There seem to be two conflicting tendencies here.

If this is indeed only an `EntityViewerTranscriptFilter`, as the name of the component suggests, then why add the `hideUnchecked` option? Unless I missed something, the prototype doesn't show the case of disappearing unchecked checkboxes.

If, on the other hand, it is a more generic component, where unchecked checkboxes can disappear, then why call it EntityViewerTranscriptFilter?

Incidentally, I think if the api of this component were different, such as something like this:

```
<OptionsFilter className="blah">
  <OptionsGroup className="blah-blah" >
    <Option className="bleh" isSelected={boolean} onChange={(something) => void}>label</Option>
    <Option isSelected={boolean} onChange={(something) => void}>label</Option>
    ....
  </OptionsGroup>
  <OptionsGroup>
    <Option isSelected={boolean} onChange={(something) => void}>label</Option>
    <Option isSelected={boolean} onChange={(something) => void}>label</Option>
    ....
  </OptionsGroup>
</OptionsFilter>
```
then it would be parent's concern whether to show or hide selected options; and the component would also probably be very reusable. I don't know whether we want a reusable filters component — I haven't yet seen them recur in Andrea's designs, but they are almost guaranteed to reappear. Still, a question to Andrea...

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  value: string;
+  label: string;
+};
+
+export type OptionsGroup = Option[];
+
+type Props = {
+  selectedValues: string[];
+  onChange: (selectedValues: string[]) => void;
+  options: OptionsGroup[];
```
azangru: `optionsGroup` then?

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  hideUnchecked?: boolean;
+};
+
+const EntityViewerTranscriptFilter = (props: Props) => {
+  const handleChange = (status: boolean, optionIndex: [number, number]) => {
```
azangru: `status` -> `isChecked`

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+import classNames from 'classnames';
+
+import Checkbox from 'src/shared/components/checkbox/Checkbox';
+
+import styles from './EntityViewerTranscriptFilter.scss';
+
+export type Option = {
+  value: string;
+  label: string;
+};
```
azangru: as the component is written, the Option type begs to have an `isSelected` field; then the `selectedValues` prop becomes unnecessary.

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const handleChange = (status: boolean, optionIndex: [number, number]) => {
+    const [groupIndex, itemIndex] = optionIndex;
+    const selectedOption = props.options[groupIndex][itemIndex].value;
+
+    if (status) {
+      const newSelectedOptions = [...props.selectedValues, selectedOption];
+      props.onChange(newSelectedOptions);
+    } else {
+      const newSelectedOptions = [...props.selectedValues].filter(
+        (option) => option != selectedOption
```
azangru: we usually try to avoid comparison operators that will result in type casting. Probably should add this to eslint somewhere 😄

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+      );
+
+      props.onChange(newSelectedOptions);
+    }
+  };
+
+  return (
+    <div className={styles.defaultWrapper}>
```
azangru: if there is nothing to override this class, why the `default` part in the className?

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {props.options.map((options, groupIndex) => {
+        const optionGroupClassName = classNames(styles.checkboxPadding, {
+          [styles.optionGroup]: !props.hideUnchecked
```
azangru: If we remove the hideUnchecked prop, then the `checkboxPadding` class should be renamed to optionGroup as better reflecting what it is for

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: > why add the hideUnchecked option? Unless I missed something, the prototype doesn't show the case of disappearing unchecked checkboxes.

Closing the filter after making some selections will only show the selected options.

I initially started creating it as a shared component but I had a feeling that a generic shared component like this should probably work more like a Select with popup and a controller.

Particularly in this case, the controller is located on the table header (or a different element) and not attached to the filter itself which is why I though of making this EV specific.

azangru: > Closing the filter after making some selections will only show the selected options.

Oh, sorry, I didn't realise that was the behaviour. I couldn't find it anywhere in the description, but can now see it in the prototype.

Does this suggest perhaps that what the prop should actually be saying is whether the component is in an "expanded" or "collapsed" state?

msmanoj: Using "expanded" or "collapsed" wouldn't make it clear that the collapsed state is going to show all the selected options.

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: Would that mean we will have to pass the entire set of options back to the parent component onChange and let it filter for the selected values?

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: Changed it to `!==`

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: Removed `default`

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
azangru: The collapsed variant of the component will show what the collapsed variant is expected to show.
The expanded variant of the component will show what the expanded. variant is expected to show.
This is the whole point of declarative vs imperative approach.

For example, the collapsed variant of the filters component is also not showing borders between filter groups. But you did not explicitly instruct the component to do that. It just knows that it shouldn't.

msmanoj: To me collapsed would mean more like it is meant to hide everything that is inside, which is not something that happens here.  I used `hideUnchecked` because that's what it does. 
Also, when all the options are selected, there will be nothing to expand/collapse?

azangru: The intent to expand or collapse the filters is expressed in this arrowhead button though:

image

when you click on it, you are not saying to the filters component "please show unchecked filters"; what you are saying to it is "please open".

And when you click on it again:

image

you effectively say "please close".

I would say the fact that in the closed state the component shows the selected filters is just how it behaves in its closed state 🤷‍♂ 

By the way, this raises the question of whether the component should be "openable" when you have checked all checkboxes. It probably should (in which case it will draw the lines between option groups), but it is still rather confusing 😬

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.scss
```diff
@@ -0,0 +1,36 @@
+@import 'src/styles/common';
+
+$checkbox-border-color: rgba(154, 167, 177, 0.4);
+$checkbox-background-color: rgba(111, 129, 144, 0.3);
+$option-group-border-color: rgba(255, 153, 0, 0.4);
```
azangru: Are these our regular colour (apart from the opacity)?

If not — why not? If yes, then we can use SCSS variables here (`$checkbox-border-color: rgba($grey, 0.4);`).

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+  value: string;
+  label: string;
+};
+
+export type OptionsGroup = Option[];
+
+type Props = {
+  selectedValues: string[];
+  onChange: (selectedValues: string[]) => void;
+  optionsGroup: OptionsGroup[];
```
azangru: `optionsGroup` -> `optionGroups`

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+      {props.optionsGroup.map((options, groupIndex) => {
+        const optionGroupClassName = classNames(styles.checkboxPadding, {
+          [styles.optionGroup]: props.isExpanded
+        });
+
+        return (
+          <div key={groupIndex} className={optionGroupClassName}>
+            {options.map((option, itemIndex) => {
+              const checkedStatus =
+                props.selectedValues.indexOf(option.value) !== -1;
```
azangru: `indexOf !== -1;` -> `includes`

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.scss
msmanoj: I was not aware of that. It's done now. Thanks 👍

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.tsx
msmanoj: I guess this is not necessary anymore?

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const selectedValues = getRandomSelectedValues();
+
+const defaultProps = {
+  optionGroups: optionGroups,
+  selectedValues: selectedValues,
+  onChange
+};
+
+const renderTranscriptFilter = (props?: any) => {
```
azangru: Surely not `any`. Looks like it's `Partial<EntityViewerTranscriptFilterProps>`.

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+};
+
+describe('<EntityViewerTranscriptFilter />', () => {
+  let wrapper: any = renderTranscriptFilter();
+
+  beforeEach(() => {
+    wrapper = renderTranscriptFilter();
+  });
+
+  it('displays N number of Checkbox based on the number of options', () => {
```
azangru: it displays the same number of checkboxes as there are options

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(wrapper.find(Checkbox).length).toBe(totalOptions);
+  });
+
+  it('displays N number of option groups based on the number of option groups', () => {
```
azangru: it displays all option groups

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(wrapper.find('.optionGroup').length).toBe(optionGroups.length);
+  });
+
+  it('calls the onChange when a Checkbox is changed', () => {
+    wrapper
+      .find(Checkbox)
+      .at(Math.floor(Math.random() * totalOptions))
+      .find('input')
+      .simulate('change');
+    expect(onChange).toBeCalled();
```
azangru: aren't we more interested in what will be passed to onChange?

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/components/transcript-filter/EntityViewerTranscriptFilter.test.tsx
msmanoj: That's done

PR #260, opened 2020-02-25
Add default transcripts list layout to gene view of EV

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-495](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-495)

## Importance
This is needed by a lot of other components that need to go into the ~~transcripts table~~ default transcripts list layout.

## Description
~~This is the transcripts table for the EV gene view. There are two panels with mock text for demonstration purposes. Real or mock data will need to be used for panels 02 and 03. This will be done either in this or another PR.~~

~~The table has been divided into sections rather than a conventional table that has straightforward columns and rows. This is because of the middle content of the table that contains three different panels.~~

This is the default transcripts list layout for the EV gene view. 

## Views affected
EV gene view

### Other effects


src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { ReactComponent as chevronLeftIcon } from 'static/img/shared/chevron-left.svg';
+import { ReactComponent as chevronRightIcon } from 'static/img/shared/chevron-right.svg';
+
+import { Status } from 'src/shared/types/status';
+
+import styles from './TranscriptsTable.scss';
+
+const TranscriptsTable = () => {
+  const [paginationNumber, setPaginationNumber] = useState(1);
```
azangru: `paginationNumber` → `page`
`setPaginationNumber` →`setPage`

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+    if (paginationNumber > 1) {
+      setPaginationNumber(paginationNumber - 1);
+    }
+  };
+
+  const gotoNext = () => {
+    if (paginationNumber < 3) {
+      setPaginationNumber(paginationNumber + 1);
+    }
+  };
```
azangru: `gotoPrevious`, `gotoNext` → `previousPage`, `nextPage`

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
+@import 'src/styles/common';
+
+.table {
+  display: grid;
+  grid-template-columns: 1fr 3fr 1fr 1fr 2fr;
+  grid-template-rows: auto;
+  min-width: 1900px;
```
azangru: Umm. We calculated with Andrea that the minimum width for the whole main content area of the Entity Viewer should be 1600, although even that is intolerably high. 1900 is way too much!

azangru:
I’m seeing this on my home laptop:

image

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { Status } from 'src/shared/types/status';
+
+import styles from './TranscriptsTable.scss';
+
+const TranscriptsTable = () => {
+  const [paginationNumber, setPaginationNumber] = useState(1);
+
+  const gotoPrevious = () => {
+    if (paginationNumber > 1) {
```
azangru: Should the button even be clickable if this condition is not satisfied? If not, this should probably be done at the point of the styling the button / adding click handler.

Alternatively, is it possible that the table will show the last page if user clicks on the button when page is 1? What does Andrea say?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  const getCurrentTableMiddleComponent = () => {
+    switch (paginationNumber) {
+      case 1:
+        return null;
+      case 2:
+        return <MainContent />;
+      case 3:
+        return <SecondaryContent />;
```
azangru: I think instead of MainContent or SecondaryContent these should say PageTwo and PageThree, or something like that.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
+@import 'src/styles/common';
+
+.table {
+  display: grid;
+  grid-template-columns: 1fr 3fr 1fr 1fr 2fr;
```
azangru: here and elsewhere: we might be better off actually specifying in pixels the widths of individual columns

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
...
+
+.tableBody {
+  padding-top: 20px;
+}
+
+.tableHeader {
+  color: $grey;
+
+  .tableRow {
+    border-bottom: 1px solid #8d621c;
```
azangru: What color is this? Is it not our regular orange color? If not, is it a recurrent color that should be added to the color palette?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <span className={tableStyles.tableCell}>Transcript name</span>
+          <span className={tableStyles.tableCell}>TSL</span>
+        </div>
+      </div>
+      <div className={tableStyles.tableBody}>
+        <div className={tableStyles.tableRow}>
+          <span className={tableStyles.tableCell}>Protein coding</span>
+          <span className={tableStyles.tableCell}>11,986</span>
+          <span className={tableStyles.tableCell}>26 of 27</span>
+          <span className={tableStyles.tableCell}>3418aa</span>
```
azangru: notice that the design has a non-breaking space between the number and the aa (although I don't like this aa thing here at all :disappointed:  )

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
...
+  grid-template-rows: auto;
+  min-width: 1900px;
+}
+
+.tableRow {
+  border-bottom: 1px solid $dark-grey;
+}
+
+.tableCell {
+  display: inline-block;
```
azangru: why is this inline-block and not block?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+}
+
+.tableMiddle {
+  grid-area: 1 / 2 / auto / 3;
+
+  .tableRow {
+    display: inline-grid;
+    grid-template-rows: auto;
```
azangru: No-no-no, the class tells us that it’s a table row. A row should not consist of rows; that's what columns do.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.scss
```diff
@@ -0,0 +1,5 @@
+.mainContent {
+  & > div > div {
```
azangru: `& .tableRow` ?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.scss
```diff
@@ -0,0 +1,5 @@
...
+    grid-template-columns: 2fr 1.5fr 1.5fr 1fr 1.5fr 1fr;
```
azangru: As in other places of where we use CSS grid for this table, we probably want actual pixels rather than fractions of available width here

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  };
+
+  return (
+    <div className={styles.table}>
+      <div className={styles.tableLeftWing}>
+        <div className={styles.tableHeader}>
+          <div className={styles.tableRow}>
+            <span className={styles.tableCell}>
```
azangru: I think, table cells should be divs rather than spans.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: It's not exactly a page and that's why I went with pagination. But in hindsight that is still telling us about a page. How about panel instead?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: Don't those look more like variables than listeners?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: I'll change it to 1600px fixed width then

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
```diff
@@ -0,0 +1,93 @@
...
+
+.table {
+  display: grid;
+  grid-template-columns: 1fr 3fr 1fr 1fr 2fr;
+  grid-template-rows: auto;
+  min-width: 1900px;
+}
+
+.tableRow {
+  border-bottom: 1px solid $dark-grey;
```
msmanoj: By looking at the XD, it looks like the border needs to be slightly transparent.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: It should be the same orange that's used in the transcript scale

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: It still works. In fact, both seem to work fine but I agree block would be better.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.scss
imransl: It won't work as the .tableRow class comes from the `TranscriptsTable`

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: Ah I forgot to remove this one. Thanks for spotting it.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
azangru: I am ok with panels if you prefer them to pages.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
azangru: They do, yes. What I didn't like about the previous names was that they made me ask the question "previous what". `toPreviousPanel`, `toNextPanel` perhaps?

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
azangru: Weird, I was using the `orange` variable from settings.scss for them, but that's not `#8d621c`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
azangru: Block will be the default if tableCell is changed from span to div.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.scss
azangru: What happens if you try to import classes from one file to the other with `@use`? ([docs](https://sass-lang.com/documentation/at-rules/use))

I see SASS [strongly discourages](https://sass-lang.com/documentation/at-rules/import) the use of `@import` directive, and we are using it in our code all the time

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: Yes, they look much better.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.scss
imransl: I tried to use `@use` with mixins for this, but webpack keeps complaining about this. I will investigate this further and see if it can be fixed.

azangru: Oh, what I didn't realise was that Sass regards imported files as partials (which have to start with an underscore). Which means, I guess, that we can't use the `@use` directive in this case.

I think in this particular case I would prefer all styles to be defined in a single style file then. Less desirable (but still tolerable) would be the `[class*='tableRow']` trick. `& > div > div` is just super fragile!

azangru: Sorry, I was actually being stupid in both my initial and subsequent suggestion. There is a very simple solution to the problem we are discussing, and we’ve used it tons of times before :smiley: 

in the tsx file:
```javascript

import styles from './MainContent.scss';
import tableStyles from '../TranscriptsTable.scss';

// then in the component
  const rowClassNames = classNames(
    tableStyles.tableRow,
    styles.tableRow
  );

// and then during rendering
<div className={rowClassNames}>
   ...
</div>
```

imransl: I tried `[class*='tableRow'] already and it worked well, though looked hacky. Using `classNames` looks better 😀

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: :heavy_check_mark

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/main-content/MainContent.tsx
imransl: :heavy_check_mark:

imransl: I checked with Andrea about this. She said that we need to use `aa` as well as `bp` in the table. So updated accordingly.

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: It should keep rotating the panel e.g. when at panel 3 and next is clicked -> show panel 1

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.tsx
imransl: :heavy_check_mark:

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-table/TranscriptsTable.scss
azangru: Also, `inline-grid` -> `grid`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -19,15 +19,24 @@ const DefaultTranscriptslist = (props: Props) => {
 
   return (
     <div className={styles.defaultTranscriptsList}>
```
azangru: This style has been deleted from CSS?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -1,7 +1,13 @@
 @import 'src/styles/common';
 @import 'constants';
 
-.defaultTranscriptsList {
+.header {
+  border-bottom: 1px solid $orange;
+  line-height: 2;
```
azangru: What’s the purpose of the line-height here?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -1,7 +1,13 @@
...
+  width: 95%;
```
azangru: Why is the header narrower than the whole list? Couldn’t find in the XD what this might be referring to.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -19,15 +19,24 @@ const DefaultTranscriptslist = (props: Props) => {
...
-          key={index}
-          gene={gene}
-          transcript={transcript}
-          rulerTicks={props.rulerTicks}
-        />
-      ))}
-      <StripedBackground {...props} />
+      <div className={styles.header}>
+        <div className={styles.row}>
+          <div className={styles.left}>Filter &amp; sort</div>
```
azangru: Just a simple `&`?

src/ensembl/static/img/shared/chevron-left.svg
```diff
@@ -2,8 +2,5 @@
 <!-- Generator: Adobe Illustrator 23.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 28 28" style="enable-background:new 0 0 28 28;" xml:space="preserve">
-<style type="text/css">
-	.st0{fill:#0099FF;}
-</style>
-<polygon class="st0" points="19,21.3 11.7,14 19,6.7 15.3,6.7 8,14 15.3,21.3 "/>
+<polygon class="chevron" points="19,21.3 11.7,14 19,6.7 15.3,6.7 8,14 15.3,21.3 "/>
```
azangru: Why do we even need a class in this and the other svg?

imransl: This must have come from the merge but I wonder why it is appearing as if I added 🤔

imransl: I was using it before when the panels were there. Don't think it is needed anymore.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: Removed it

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
imransl: Changed this to padding

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
imransl: Removed this

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✔️

src/ensembl/static/img/shared/chevron-left.svg
azangru: > as we can target it to change the checkmark colour.
We can do it by targeting the parent, the way it's currently done in the `layout/StandardAppLayout.scss` file above. If we aren't targeting a part of the svg shape (as we did with half-blue half-grey eye), then there's no point in having the class attribute. But I don't have particularly strong feelings about it either way.

PR #262, opened 2020-02-27
Refactor image button classes

imransl:
## Type
- Refactoring

## Related JIRA Issue(s)
[ENSWBSITES-527](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-527)

## Description
This PR adds an extra prop to `ImageButton` to receive a class to style the button itself. This will get rid of the need to use a wrapper `div` to style it.

## Views affected
Wherever `ImageButton` is used

### DIFF
--- a/src/ensembl/src/header/launchbar/LaunchbarButton.tsx
+++ b/src/ensembl/src/header/launchbar/LaunchbarButton.tsx
@@ -1,6 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import { NavLink } from 'react-router-dom';
 import { withRouter, RouteComponentProps } from 'react-router';
+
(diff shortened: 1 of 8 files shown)


src/ensembl/src/header/launchbar/LaunchbarButton.tsx
```diff
@@ -28,7 +29,8 @@ const LaunchbarButton: FunctionComponent<LaunchbarButtonProps> = (
 
   const imageButton = (
     <ImageButton
-      classNames={{
+      className={props.enabled ? '' : styles.launchbarButton}
```
azangru: `''` -> `undefined` ?

PR #268, opened 2020-03-10
Pluralisation formatter

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-525](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-525)

## Description
Adds a pluralisation formatter, which checks if the count is more than 1 and adds an `s` in case it is. It can also accommodate special cases using a dictionary that will keep growing with time. Also, created a formatter to display the strand name.

## Views affected
Wherever you find some count e.g. 1 transcript, 2 genes


imransl:
@ens-ap5 @azangru Can you let me know if this is what we want. I just copy pasted what we had in the comments of the JIRA ticket. Once you are happy with this, I'll add the unit tests.

azangru:
Perharps it's a good PR to move code that's responsible for various string formatting into the formatters folder inside the helpers, as suggested [here](https://app.slack.com/client/T0F48FDPE/CKXFYCK97)?

src/ensembl/src/shared/data/pluralisedNouns.ts
```diff
@@ -0,0 +1,9 @@
+type PluralDictionary = {
+  [name: string]: string;
+};
+
+const pluralDictionary: PluralDictionary = {
+  species: 'species'
+};
+
+export default pluralDictionary;
```
azangru: I think there will be so few exceptions to the general rule of pluralisation that the dictionary can safely live in the same file as the pluralise function.

imransl: ✔️

azangru:
image

?

imransl:
@azangru I moved them to `/src/shared/helpers/formatters`

azangru:
Looks good! Let's also use the pluralisationFormatter in GeneOverviewImage [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-overview-image/GeneOverviewImage.tsx#L121)?

With the EHK cluster unavailable, you'll need to clone the toy server from [here](https://gitlab.ebi.ac.uk/andrey/ensembl-graphql-server) and run it locally to see the EntityViewer page.

imransl:
@azangru That's done now

PR #269, opened 2020-03-11
EntityViewer - Gene view tabs and panels

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-469

## Description
- Added the Gene view tab and the corresponding panels that gets displayed when the tabs are selected.
- Created the state related files for the gene view. Currently Gene tab, Gene function panel tab & Gene relationships panel tab states are maintained.

## Views affected
Entity Viewer -> Gene view


src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -4,10 +4,12 @@ import { useQuery } from '@apollo/react-hooks';
 import { gql } from 'apollo-boost';
 
 import { getEntityViewerActiveEnsObject } from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSelectors';
-
 import GeneOverviewImage from './components/gene-overview-image/GeneOverviewImage';
-
 import DefaultTranscriptslist from './components/default-transcripts-list/DefaultTranscriptsList';
+import GeneViewTabs from './components/gene-view-tabs/GeneViewTabs';
+import { getEntityViewerActiveGeneTab } from 'src/content/app/entity-viewer/state/gene-view/entityViewerGeneSelectors';
```
azangru: _stylistic:_ move ⬆️  to selector imports?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -90,14 +99,24 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
...
       </div>
-      <div className={styles.transcriptsListWrapper}>
-        {basePairsRulerTicks && (
-          <DefaultTranscriptslist
-            gene={props.gene}
-            rulerTicks={basePairsRulerTicks}
-          />
+      <div className={styles.geneViewTabbedContent}>
+        {props.selectedGeneTabName === 'Transcripts' && (
+          <div className={styles.geneViewTable}>
```
azangru: what does this class do?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneState.ts
```diff
@@ -0,0 +1,34 @@
...
+  geneRelationships: EntityViewerGeneRelationshipsState;
+};
+
+export type EntityViewerGeneState = Readonly<{
+  [activeGenomeId: string]: {
+    [activeObjectId: string]: EntityViewerGeneObjectState;
+  };
+}>;
+
+export const defaultEntityViewerGeneObjectState = {
```
azangru: > GeneObject

Do you mean `GeneView`?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -17,10 +19,12 @@ import styles from './GeneView.scss';
 
 type GeneViewProps = {
   geneId: string | null;
+  selectedGeneTabName: string | null;
```
azangru: Does the selector ever return null?

Also, a string or an enum?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -5,7 +5,7 @@
   display: grid;
   background-color: $black;
   grid-template-columns: 228px 902px 1fr;
-  grid-template-rows: min-content 150px 1fr;
+  grid-template-rows: min-content 31px 1fr;
```
azangru: The distance between the gene above and the buttons that I am seeing in the PR:

image

is smaller than in the XD:

image

Are you sure about this change? Isn't `align-self: end` on `geneViewTabs` better suited for this role?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.scss
```diff
@@ -0,0 +1,42 @@
...
+    box-shadow: none;
+    box-sizing: border-box;
+    content: '';
+    position: absolute;
+    left: calc(50% - 8px);
+    transform-origin: 0 0;
+    transform: rotate(-45deg);
+    top: 28px;
+  }
+}
```
azangru: Apparently, the arrow drops a shadow in the mockup 😨 

image

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships.scss
```diff
@@ -0,0 +1,35 @@
+@import 'src/styles/common';
+
+.fullWidthPanel {
+  width: 100%;
+  height: 100%;
+}
+
+.shrinkedPanel {
```
azangru: 1) ~shrinked~ -> shrunken
2) even better, shrunken -> narrow?
3) however, I don't think this is the effect Andrea was after:

image

I expect she would want the sidebar to overlap the panel as well, seeing that it started overlapping the rest of the content of entity viewer.
I have no idea whether it's possible to achieve this with CSS. Perhaps Andrea could give us the min width for the panel? I really don't understand how she wants us to draw it.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneActions.ts
```diff
@@ -0,0 +1,104 @@
...
+  'entity-viewer/update-state-for-active-genome',
+  (payload: {
+    activeGenomeId: string;
+    activeObjectId: string;
+    data: EntityViewerGeneObjectState;
+  }) => {
+    const { activeGenomeId, activeObjectId, data } = payload;
+    return { activeGenomeId, activeObjectId, data: cloneDeep(data) };
+  }
+)();
```
azangru: I feel that I **strongly** prefer the opposite approach: instead of updating the whole state in an action and then replacing this state in the reducer, I would create a fragment of the state in an action, and then replace that fragment in the reducer ([example](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts#L16-L36)). Philosophically, it's reducers' job to update the state.

@imransl do you have any thoughts on this one?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneActions.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  if (!activeGenomeId || !activeObjectId) {
+    return;
+  }
+
+  dispatch(
+    updateActiveGeneObjectState({
+      activeGenomeId,
+      activeObjectId,
+      data: set(
```
azangru: I don't remember whether getState returns the actual state or the copy of the state (I assume the actual state, making a copy would be expensive). If it's the actual state, then you have just mutated the state, which is a big no-no. If it's the copy of the state (unlikely), then `cloneDeep` in `updateActiveGeneObjectState` has no purpose.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: Removed it

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneState.ts
msmanoj: Updated it all places

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
msmanoj: Changed it to 90px as 150px creates a big gap.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships.scss
msmanoj: I did speak to her about this and she preferred it to work this way.

I've renamed it to narrow.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: Changed it to enum

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.scss
msmanoj: Added the shadow 👍

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneActions.ts
msmanoj: I guess this no longer applies after addressing the previous comment

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneActions.ts
msmanoj: I have made this change but it seems to have made it more complex than it was before. Would it be better to use `set` and `cloneDeep` within the the reducer instead of the actionCreator?

azangru: I pushed a commit with the updated reducer to demonstrate what I had in mind.

msmanoj: Thanks @azangru. I was thinking something like the same using `set` and `cloneDeep` but this looks cleaner.

msmanoj:
@azangru regarding the discussion about the tabs behaviour on Slack, I think it can be done when we start adding the contents to the tabs instead of this PR.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -51,7 +51,7 @@ export type EntityViewerParams = {
 };
 
 const client = new ApolloClient({
-  uri: 'http://193.62.55.158:30466'
+  uri: 'http://localhost:5000'
```
azangru: probably shouldn't commit this?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Tabs, { Tab } from 'src/shared/components/tabs/Tabs';
+import Panel from 'src/shared/components/panel/Panel';
```
azangru: Here, in GeneRelationships, and GeneTabs — let’s order the imports?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type EntityViewerGeneViewObjectState = {
+  selectedGeneTabName: GeneViewTabName;
+  geneFunction: EntityViewerGeneFunctionState;
+  geneRelationships: EntityViewerGeneRelationshipsState;
+};
+
+export type EntityViewerGeneViewState = Readonly<{
+  [activeGenomeId: string]: {
+    [activeObjectId: string]: EntityViewerGeneViewObjectState;
```
azangru: `activeGenomeId` → `genomeId`
`activeObjectId` → `geneId`

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -0,0 +1,55 @@
+export type EntityViewerGeneFunctionState = {
+  selectedTabName: GeneFunctionTabName;
+};
+
+export type EntityViewerGeneRelationshipsState = {
+  selectedTabName: GeneRelationshipsTabName;
+};
+
+export type EntityViewerGeneViewObjectState = {
```
azangru: `GeneViewObject` doesn’t tell us much

`EntityViewerGeneState`?
or perhaps `EntityViewerGeneUIState`? or `EntityViewerGeneViewUIState`?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -21,8 +21,10 @@
...
 .geneViewTabs {
   grid-area: 2 / 2 / 3 / 4;
+  z-index: 10;
+  align-self: end;
 }
 
-.transcriptsListWrapper {
+.geneViewTabbedContent {
   grid-area: 3 / 1 / 4 / 4;
 }
```
imransl: `tabbed` to `tab`?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -90,14 +100,23 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
...
-        These are the Entity Viewer tabs...
+        <GeneViewTabs />
       </div>
-      <div className={styles.transcriptsListWrapper}>
-        {basePairsRulerTicks && (
-          <DefaultTranscriptslist
-            gene={props.gene}
-            rulerTicks={basePairsRulerTicks}
-          />
+      <div className={styles.geneViewTabbedContent}>
```
imransl: Same as above: `tabbed` -> `tab`

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
msmanoj: Changed it to `EntityViewerGeneViewUIState` 👍

azangru:
> @azangru regarding the discussion about the tabs behaviour on Slack, I think it can be done when we start adding the contents to the tabs instead of this PR.

But why? This behaviour does not depend on the content of the tabs in any way. It just concerns opening and closing the panel upon click on the tabs.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
msmanoj: Yes, I will revert it back before merging. Let's leave this conversation unresolved. Thanks

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: Related to the previous comment. That's done

msmanoj:
> > @azangru regarding the discussion about the tabs behaviour on Slack, I think it can be done when we start adding the contents to the tabs instead of this PR.
> 
> But why? This behaviour does not depend on the content of the tabs in any way. It just concerns opening and closing the panel upon click on the tabs.

Please ignore my previous comment. That's done now.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+    default: styles.geneTab,
+    selected: styles.selectedGeneTab,
+    disabled: styles.disabledGeneTab,
+    tabsContainer: styles.geneViewTabs
+  };
+
+  const onTabChange = (selectedTabName: string) => {
+    if (selectedTabName === props.selectedGeneTabName) {
+      props.setActiveGeneTab(DEFAULT_TAB);
+      return;
```
imransl: Won't it be clearer if we use an else statement rather than return from here?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
```diff
@@ -0,0 +1,127 @@
...
+  null,
+  Action<string>
+>> = (selectedTabName: GeneFunctionTabName) => (
+  dispatch,
+  getState: () => RootState
+) => {
+  const activeGenomeId = getEntityViewerActiveGenomeId(getState());
+  const activeObjectId = getEntityViewerActiveEnsObjectId(getState());
+
+  if (!activeGenomeId || !activeObjectId) {
```
imransl: We can also add the `dispatch` statement inside the `if` clause by changing the condition to `(activeGenomeId && activeObjectId)`.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
```diff
@@ -0,0 +1,127 @@
...
+      fragment: {
+        geneFunction: {
+          selectedTabName
+        }
+      }
+    })
+  );
+};
+
+export const updateActiveGeneViewObjectGeneRelationshipsState = createAction(
```
imransl: Isn't this a very long variable name?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
```diff
@@ -0,0 +1,127 @@
...
+  void,
+  any,
+  null,
+  Action<string>
+>> = (selectedTabName: GeneRelationshipsTabName) => (
+  dispatch,
+  getState: () => RootState
+) => {
+  const activeGenomeId = getEntityViewerActiveGenomeId(getState());
+  const activeObjectId = getEntityViewerActiveEnsObjectId(getState());
```
imransl: Too long again? I am sure @ens-ma7 and @ens-ap5 might shudder at this.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  if (!activeGenomeId || !activeObjectId) {
```
imransl: Again like above, how about reversing the condition and including `dispatch` inside the `if` clause?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewReducer.ts
```diff
@@ -0,0 +1,35 @@
...
+  state: EntityViewerGeneViewState = initialEntityViewerGeneViewState,
+  action: ActionType<typeof actions>
+) {
+  switch (action.type) {
+    case getType(actions.updateActiveGeneViewObjectState):
+      const { activeGenomeId, activeObjectId, fragment } = action.payload;
+      const oldStateFragment = get(
+        state,
+        `${activeGenomeId}.${activeObjectId}`,
+        defaultEntityViewerGeneViewUIStateState
```
imransl: Why `State` twice?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors.ts
```diff
@@ -0,0 +1,30 @@
+import { RootState } from 'src/store';
+import {
+  getEntityViewerActiveGenomeId,
+  getEntityViewerActiveEnsObjectId
+} from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSelectors';
+import { defaultEntityViewerGeneViewUIStateState } from 'src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState';
```
imransl: Same as above about `State`

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type EntityViewerGeneViewUIStateState = {
+  selectedGeneTabName: GeneViewTabName;
+  geneFunction: EntityViewerGeneFunctionState;
+  geneRelationships: EntityViewerGeneRelationshipsState;
+};
+
+export type EntityViewerGeneViewState = Readonly<{
+  [genomeId: string]: {
+    [objectId: string]: EntityViewerGeneViewUIStateState;
```
imransl: Again two `State`s

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.tsx
msmanoj: I always feel this approach is slightly cleaner that using else blocks. What's your opinion @azangru ?
Also, there is a discussion on stackexchange about this:
https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: At the moment it would be easier to understand that this function does nothing if either `activeGenomeId` or `activeObjectId` is missing. There is nothing complex going on here that has made it harder to understand anyways.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: Yes, it is. I am not sure how to shorten it. Any suggestions?

azangru: Oh gosh, yes.
`updateGeneRelationshipsUI`?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.tsx
azangru: There is value in an early return from an if statement when the body of the function is bulky and you feel that putting it all in an else block (and indenting it to the right) would decrease readability.

In this particular case though, I agree with Imran. Both branches of the if-condition are equally short; it is more readable to put the alternative branch inside the else block.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
```diff
@@ -0,0 +1,72 @@
...
+  { title: 'Gene ontology', isDisabled: true },
+  { title: 'Gene pathways' }
+];
+
+const tabClassNames = {
+  selected: styles.selectedTabName
+};
+
+type Props = {
+  isSidebarOpen: boolean;
```
azangru: Here and in GeneRelstionship, I think that these components do not need to know about the existence of such a thing as sidebar somewhere in the other part of the interface. What they need to know is only whether their shape is"wide" or "narrow". Perhaps better call this prop `isNarrow`?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
imransl: Negated conditions are considered taboo by some who usually say to test for only `true` and not `false`. However, I get your point about letting someone reading the code to know on first glance that the function does nothing without the two ID values. What do you think about this @azangru?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewReducer.ts
msmanoj: Renamed it

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors.ts
msmanoj: Renamed it

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
msmanoj: Renamed it

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
msmanoj: Good point. I've renamed it to `isNarrow`.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: Just realised that this actionCreator is no longer needed. I've removed it. Sorry!

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: I don't think it is a good to rename this to `getActiveGenomeId` as it will make it ambiguous.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
azangru: > Negated conditions are considered taboo by some who usually say to test for only true and not false.

We have definitely not been following this advice in our codebase 😄 I don't think testing for nulls, undefined or falses has impacted readability of our code.

As for whether to prefer:

```
if (condition_not_met) {
  return
}

do something
```

vs

```
if (condition_met) {
  do something
}
```

I personally would probably have written it in the second form (as more natural), but do not particularly object to reading the first one either.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { RootState } from 'src/store';
```
azangru: this is a type; we generally place types just above styles

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type EntityViewerGeneViewUIState = {
+  selectedGeneTabName: GeneViewTabName;
+  geneFunction: EntityViewerGeneFunctionState;
+  geneRelationships: EntityViewerGeneRelationshipsState;
+};
+
+export type EntityViewerGeneViewState = Readonly<{
+  [genomeId: string]: {
+    [objectId: string]: EntityViewerGeneViewUIState;
```
azangru: I suggest we use the word "object" as little as possible. Here we know what the objects are — they are genes.

msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.tsx
msmanoj: Cool, I've updated it.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: I've updated it to:
```
if (condition_met) {
  do something
}
```

I guess we should strictly follow this way going forward in order to avoid this confusion again?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions.ts
imransl: Maybe we should, but we'll all need to agree to this 😉What do you think @azangru?

msmanoj: I would prefer to allow both but if we strictly need to follow one way, we may have to consider the following scenario as well:
https://github.com/Ensembl/ensembl-client/blob/efaf4e8dd434381b8024b515c52d3e77abece7fb/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar-tabs/GeneViewSidebarTabs.tsx#L27

PR #273, opened 2020-03-24
Default transcripts list content

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-499](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-499)

## Description
This pull request adds the content for the individual transcript list items in the gene view.

**TODO:**

1. Fix styling issues
2. Fix amino acid length
3. Write tests

## Views affected
Gene view of the Entity Viewer


src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -51,7 +51,8 @@ export type EntityViewerParams = {
 };
 
 const client = new ApolloClient({
-  uri: 'http://193.62.55.158:30466'
+  // uri: 'http://193.62.55.158:30466'
+  uri: 'http://localhost:5000'
```
imransl: **NOTE TO SELF**: Remove this before merging this PR

azangru:
**General considerations:**

- I am seeing long delay between when I click on a transcript and when the transcript info panel opens. The delay is due to an extra ajax request sent to get transcript info. I suggest that we either include this info the initial payload for the list of all transcript (I think that will be preferable), or open an empty info panel and show a spinner while the data is being loaded.
- When we mouse over the clickable area that expands or collapses transcript info panel, the cursor should change to pointer
- Transcript info panels should not be transparent (see designs)
- The Right column should show transcript id, in blue (the data is already available)
- Need to make sure with Andrea what happens when the content of the main area of StandardAppLayout is taller than the viewport of the window. Currently, all page scrolls down. I suspect what should happen is that the page should have max height of 100%, and the main area should be able to scroll independently of the sidebar. Probably not for this PR, but we should figure this out and do this soon.
- Ask Andrea about vertical distances between transcripts. They look cramped up together at the moment

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -29,22 +30,38 @@ const DefaultTranscriptListItem = (props: Props) => {
...
+          onClick={toggleListItemInfo}
+        >
+          Right
         </div>
       </div>
-      <div className={transcriptsListStyles.right}>Right</div>
-    </div>
+      {shouldShowInfo ? (
+        <TranscriptsListItemInfo transcriptId={props.transcript.id} />
+      ) : null}
```
azangru: image

image

image

I thought the info thing would be the child of DefaultTranscriptListItem, not its sibling

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -0,0 +1,214 @@
...
+          }
+        }
+      }
+      cds {
+        start
+        end
+      }
+    }
+  }
+`;
```
azangru: Most of this has likely already been queried and is accessible from GeneView component.

src/ensembl/src/content/app/entity-viewer/types/slice.ts
```diff
@@ -9,6 +9,7 @@ export type Slice = {
     name: string;
     strand: {
       code: Strand;
+      value?: number;
```
azangru: why?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
imransl: Sorry, I got that one wrong yesterday. I made it as its sibling since it can reuse the same grid styling of the default transcript list row.

src/ensembl/src/content/app/entity-viewer/types/slice.ts
imransl: `value` is in the GraphQL schema so I added that here

azangru: If we aren't using it, I don't think we need it in our types.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
azangru: I am not sure what you mean about the grid.

If instead of having a single `row` element, you put its left part in the `left__DefaultTranscriptList` div, the right part in the `right__DefaultTranscriptList` div, and the actual panel in the `middle__DefaultTranscriptList` div, then the grid from the parent component should still work.

imransl: But shouldn't this be more like an accordion? If we put these columns into the same divs of `DefaultTranscriptListItem` then we need to do three checks to hide/show on toggle.

azangru: > If we put these columns into the same divs of DefaultTranscriptListItem then we need to do three checks to hide/show on toggle.

You are right about the three checks, but I pictured an accordion item as working like this:

image

rather than like this:

image

I am not quite sure whether it matters. Personally, I think I would have gone for the first design. @msmanoj what's your take?

imransl:
> Transcript info panels should not be transparent (see designs)
This is not due to the styling of this PR. It is like that by default, so I will investigate this further to see why it is appearing like that.

src/ensembl/src/content/app/entity-viewer/types/slice.ts
imransl: Shouldn't our types align with the GraphQL schema?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: The transcript data now comes from the `GeneView` component. So the loading problem should be resolved too.

src/ensembl/src/content/app/entity-viewer/types/slice.ts
azangru: They would align with the graphql schema if they were automatically generated from it, but since we are writing them manually, and since we are the only consumer of our own types, there is no point in our describing fields that we do not use and thus do not really care about.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/types/slice.ts
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
msmanoj: Sorry, missed this one. The accordion shared component we currently have works like the first design. To me, keeping the Head separate feels like it would become something like a separate toggle button that shows/hides something.

imransl: @andrey @msmanoj After some meddling I ended up putting the accordion head and body as two rows into a wrapper `div` as:

```
<div className={styles.defaultTranscriptListItem}>
  <div className={styles.row}>accordion head</div>
  <div className={styles.row}>accordion body</div>
</div>
```

I ended up with this taking the following into account:

1. This solves the problem of having the head and body separate from each other.
2. It can use the same grid styling that we already have for the default transcripts list
3. It is neater than putting the body columns alongside the head columns (as mentioned in [this comment](https://github.com/Ensembl/ensembl-client/pull/273#discussion_r397761859))
4. No need to have three separate checks to toggle the accordion body

imransl:
@msmanoj @azangru Can you please re-review this PR?

azangru:
I forgot about exons in the untranslated region (because in my mock, the CDS was the same length as transcript); so in this PR they started looking like holes in the transcript:

image

I pushed a [fix](https://github.com/Ensembl/ensembl-client/pull/273/commits/4f4995f2977d3d9341edb12d6c27db0216d9b6a3 ) to this PR directly. Now the exons look as they should:

image

azangru:
And another commit to turn this:

image

into this:

image

azangru:
@imransl there is something wrong with the way the number of coding exons is calculated:

image

Multiple non-coding exons visible on the visualisation, but the info block says that 15 of 15 are coding 😕

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -29,6 +29,7 @@ const DefaultTranscriptslist = (props: Props) => {
         </div>
       </div>
       <div className={styles.content}>
+        <StripedBackground {...props} />
         {sortedTranscripts.map((transcript, index) => (
```
azangru, line 33: Sorry, we let the transcript-sorter into dev; but shouldn't the sorting rule be the opposite to the [accepted one](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/shared/helpers/transcripts-sorter.ts#L14-L20), i.e.:

```
  if (transcriptOneLength > transcriptTwoLength) {
    return -1;
  }

  if (transcriptOneLength < transcriptTwoLength) {
    return 1;
  }
```

to get this:

image

rather than this:

image

I seem to recall that that was Andrea's intent.

src/ensembl/tests/fixtures/entity-viewer/gene.ts
```diff
@@ -0,0 +1,39 @@
...
+    symbol: faker.lorem.word(),
+    so_term: faker.lorem.word(),
+    slice: geneSlice,
+    transcripts: [transcript]
+  };
+};
+
+const getScale = () => {
+  const domain = [1, faker.random.number()];
+  const range = [0, faker.random.number()];
```
azangru, line 27: probably safer to give these numbers at least slightly meaningful values. For example, if domain starts at 1, faker.random.number() should never be 0; and also it should probably be between a 1000 and 100000. The range's max number shouldn't be outrageously huge — let's say in between 100 and 600.

src/ensembl/tests/fixtures/entity-viewer/gene.ts
```diff
@@ -0,0 +1,39 @@
...
+  const scale = scaleLinear()
+    .domain(domain)
+    .range(range);
+
+  return scale;
+};
+
+export const createRulerTicks = (): TicksAndScale => ({
+  labelledTicks: times(faker.random.number(), Number),
+  ticks: times(faker.random.number(), Number),
```
azangru, line 37: faker.random.number() here should reasonably be between 1 and 10

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: Yes, that's true. I'll fix it in this PR itself.

imransl:
> @imransl there is something wrong with the way the number of coding exons is calculated:
> 
> image
> 
> Multiple non-coding exons visible on the visualisation, but the info block says that 15 of 15 are coding 😕

Thanks for spotting this. I'll check what is wrong with the code.

src/ensembl/tests/fixtures/entity-viewer/gene.ts
imransl: Should the `labelledTicks` also be less than 10?

azangru: labelledTicks should be between 1 and 5

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✔️

src/ensembl/tests/fixtures/entity-viewer/gene.ts
imransl: ✔️

src/ensembl/tests/fixtures/entity-viewer/gene.ts
imransl: ✔️

imransl:
@azangru The coding exon calculation error is fixed now

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+const ItemInfo = (props: ItemInfoProps) => {
+  const { transcript } = props;
+
+  const getTranscriptLocation = () => {
+    const { start, end } = getFeatureCoordinates(transcript);
+    const startStr = getCommaSeparatedNumber(start);
+    const endStr = getCommaSeparatedNumber(end);
+    const regionName = getRegionName(transcript);
+    return `${regionName}:${startStr}-${endStr}`;
```
azangru, line 27: We definitely have a formatter for that 😄

imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  const getRNALength = () => {
```
azangru, line 34: `getRNALength` -> `getSplicedRNALength`

(the current name may mean that you are getting the whole length of the transcript)

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      for (let index = exons.length - 1; index > -1; index -= 1) {
+        const exonLocation = getFeatureCoordinates(exons[index]);
+
+        if (exonLocation.start <= cds.end && exonLocation.end >= cds.end) {
+          lastCodingExonIndex = index;
+          break;
+        }
+      }
```
azangru, line 65: Aren't these two for-loops better expressed as Array.prototype.find?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+  };
+
+  const getNumberOfCodingExons = () => {
+    const {
+      firstCodingExonIndex,
+      lastCodingExonIndex
+    } = getFirstAndLastCodingExonIndexes();
+    return getCommaSeparatedNumber(
+      lastCodingExonIndex - firstCodingExonIndex + 1
```
azangru, line 120: Why the +1?

imransl: Because this calculation would give the difference rather than all the coding exons without padding with 1. For example:

```
All exons = 27
First coding exon = 1
Last coding exon = 26
Coding difference = Last coding exon - First coding exon = 25
```

azangru: I am confused. Why does first coding exon have an index of 1? I thought it will have an index of 0, as follows from this for-loop:

```
      for (let index = 0; index < exons.length; index += 1) {
        const exonLocation = getFeatureCoordinates(exons[index]);

        if (exonLocation.start <= cds.start && exonLocation.end >= cds.start) {
          firstCodingExonIndex = index;
          break;
        }
      }
```

azangru: Sorry, I now see that you are right.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: Merged `dev` with the correct uri

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
imransl: ✔️

imransl:
@azangru I have fixed the fixture that creates the exons and removed the first/last exon index checks. The tests are passing all the time now.

PR #274, opened 2020-03-25
Default sorting for transcripts list in EV gene view

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-537](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-537)

## Description
This default transcript list needs to be sorted according to the following criteria:

1. Transcripts with CCDS come before the rest, and are sorted by length.
2. Within the non-CCDS group, the transcripts are sorted by biotype (see point 3, below) and then by length.
3. The biotype sort is "Protein coding comes first, then everything else is sorted alphabetically".

This is going to change in the future and we are not certain how it will. So for the time being, the above criteria is what we will consider for sorting.

## Views affected
Gene view in entity viewer

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -1,11 +1,16 @@
 import React from 'react';
+import sortBy from 'lodash/sortBy';
 
-import { getFeatureCoordinates } from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
+import {
+  getFeatureCoordinates,
+  getFeatureLength
+} from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
 
 import DefaultTranscriptsListItem from './default-transcripts-list-item/DefaultTranscriptListItem';
 
 import { TicksAndScale } from 'src/content/app/entity-viewer/gene-view/components/base-pairs-ruler/BasePairsRuler';
 import { Gene } from 'src/content/app/entity-viewer/types/gene';
+import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
 
 import styles from './DefaultTranscriptsList.scss';
 
@@ -14,8 +19,54 @@ type Props = {
   rulerTicks: TicksAndScale;
 };
 
+const compareTranscriptLength = (
+  transcriptOne: Transcript,
+  transcriptTwo: Transcript
+) => {
+  const transcriptOneLength = getFeatureLength(transcriptOne);
+  const transcriptTwoLength = getFeatureLength(transcriptTwo);
+
+  if (transcriptOneLength < transcriptTwoLength) {
+    return -1;
+  }
+
+  if (transcriptOneLength > transcriptTwoLength) {
+    return 1;
+  }
+
+  return 0;
+};
+
+const sortTranscripts = (transcripts: Transcript[]) => {
+  const transcriptsWithCds = transcripts
+    .filter((transcript) => transcript.cds)
+    .sort(compareTranscriptLength);
+
+  const transcriptsWithoutCds = transcripts.filter(
+    (transcript) => !transcript.cds
+  );
+
+  const proteinCodingTranscripts = transcriptsWithoutCds
+    .filter((transcript) => transcript.biotype === 'protein_coding')
+    .sort(compareTranscriptLength);
+
+  const nonProteinCodingTranscripts = sortBy(
+    transcriptsWithoutCds.filter(
+      (transcript) => transcript.biotype !== 'protein_coding'
+    ),
+    ['biotype']
+  );
+
+  return [
+    ...transcriptsWithCds,
+    ...proteinCodingTranscripts,
+    ...nonProteinCodingTranscripts
+  ];
+};
+
 const DefaultTranscriptslist = (props: Props) => {
   const { gene } = props;
+  const sortedTranscripts = sortTranscripts(gene.transcripts);
 
   return (
     <div>
@@ -27,7 +78,7 @@ const DefaultTranscriptslist = (props: Props) => {
         </div>
       </div>
       <div className={styles.content}>
-        {gene.transcripts.map((transcript, index) => (
+        {sortedTranscripts.map((transcript, index) => (
           <DefaultTranscriptsListItem
             key={index}
             gene={gene}
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -1,16 +1,12 @@
 import React from 'react';
-import sortBy from 'lodash/sortBy';
 
-import {
-  getFeatureCoordinates,
-  getFeatureLength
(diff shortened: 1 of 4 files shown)


src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -1,8 +1,20 @@
...
 
 export const getFeatureCoordinates = (feature: {
   slice: SliceWithLocationOnly;
 }) => {
   const { start, end } = feature.slice.location;
   return { start, end };
 };
+
+export const getFeatureStrand = (feature: { slice: Slice }) =>
+  feature.slice.region.strand;
```
azangru: Perhaps return feature.slice.region.strand.code? There's nothing else we will be interested in regarding strand.

The consumer of the function can call the return value ("forward"/"reverse") "strand" (as in `const strand = getFeatureStrand(feature);`), to simplify the code slightly.

src/ensembl/src/content/app/entity-viewer/types/transcript.ts
```diff
@@ -6,7 +6,8 @@ export type Transcript = {
   type: 'Transcript';
   id: string;
   symbol: string;
-  so_term: string; // is there a better name for it?
+  so_term?: string; // is there a better name for it?
+  biotype?: string;
```
azangru: Can you leave a fixme comment for us to remove one of these fields (likely the `biotype`)?

imransl: There is already a comment in the #273. So I didn't want to add a comment here that would conflict with it when merging.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -14,8 +19,54 @@ type Props = {
...
+  }
+
+  if (transcriptOneLength > transcriptTwoLength) {
+    return 1;
+  }
+
+  return 0;
+};
+
+const sortTranscripts = (transcripts: Transcript[]) => {
```
azangru: According to design, there will be different sorting options. Perhaps this calls for moving the sorting code in a separate file (`sortTranscripts.ts`? `transcriptsSorter.ts`? don't know what name will be best), and calling this sorting function `defaultSort` (it won't even need to have the word "transcripts" in its name)?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -14,8 +19,54 @@ type Props = {
...
+  const proteinCodingTranscripts = transcriptsWithoutCds
+    .filter((transcript) => transcript.biotype === 'protein_coding')
+    .sort(compareTranscriptLength);
+
+  const nonProteinCodingTranscripts = sortBy(
+    transcriptsWithoutCds.filter(
+      (transcript) => transcript.biotype !== 'protein_coding'
+    ),
+    ['biotype']
+  );
```
azangru: is this a sort in alphabetic order? is this what Andrea wants?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -14,8 +19,54 @@ type Props = {
...
+    .filter((transcript) => transcript.cds)
+    .sort(compareTranscriptLength);
+
+  const transcriptsWithoutCds = transcripts.filter(
+    (transcript) => !transcript.cds
+  );
+
+  const proteinCodingTranscripts = transcriptsWithoutCds
+    .filter((transcript) => transcript.biotype === 'protein_coding')
+    .sort(compareTranscriptLength);
```
azangru: I am confused. Do such transcripts exist? Why would a transcript have a cds if it is not protein-coding?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: Yes, sorting based on the alphabetical order of the biotype for the non protein coding transcripts. This is criteria number 3.

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -1,8 +1,20 @@
...
+
+export const getFeatureLength = (feature: { slice: Slice }) => {
```
azangru: Can you add a FIXME here so that we use feature.slice.location.length property when it becomes available and not calculate it ourselves (we will have problems with circular chromosomes if we try to do it ourselves)?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: I am not sure and it's probably we'll need to ask during the stand-up today. I did this since it is in the criteria assuming that they can exist.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { getFeatureCoordinates } from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
+import { defaultSort } from '../../../shared/helpers/transcripts-sorter';
```
azangru: `'../../../` -> `src/shared`?

src/ensembl/src/content/app/entity-viewer/shared/helpers/transcripts-sorter.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function defaultSort(transcripts: Transcript[]) {
+  const transcriptsWithCds = transcripts
+    .filter((transcript) => transcript.cds)
+    .sort(compareTranscriptLengths);
+
+  const transcriptsWithoutCds = transcripts.filter(
+    (transcript) => !transcript.cds
+  );
```
azangru: since cds is not the same as ccds anyway, let's remove the with/without cds split from the code.

PR #270, opened 2020-03-11
Refactor tooltip and zmenu (to PointerBox, Tooltip, and Toolbox)

azangru:
## Type
- Refactor
- Bug fix
- Improvement to existing feature

## Related JIRA Issue(s)
ENSWBSITES-518

## Description
1) Refactor tooltip and zmenu to use a common base component: PointerBox

image

2) Current tooltip incorrectly calculates its position if the page has been scrolled. You can see this in our Storybook (Tooltip/Positioning) if you squeeze the window vertically until the vertical scrollbar appears, and then scroll down:

image

3) The term "Zmenu" remains to describe the floating thing in genome browser (unless we come up with a better term). However, Zmenu has been refactored to:

- use the reusable Toolbox component for rendering
- show and hide extra content in the footer

4) Added the `useRefWithRerender` custom hook to contain custom logic that makes sure that component updates after `ref.current` has been set (useful when rendering of component's children depends on existence of refs to other DOM elements).

## What to check/see
- Storybook stories for Pointerbox (the generic component used by Toolbox and Tooltip), Toolbox, and Tooltip
- Zmenu on GenomeBrowser page can now open a footer with extra content
- Tooltip should still be working properly (i.e. not get cropped when it's shown under launchbar icons or under sidebar icons)

PR #277, opened 2020-04-08
Bring back the tooltip timeout

azangru:
## Type
- Bug fix

## Description
Before https://github.com/Ensembl/ensembl-client/pull/270, the default delay for the Tooltip used to be greater than 0 (specifically, 800ms)

image

but in the course of refactoring, it was changed to 0, which makes the Tooltip appear uncomfortably quickly:

image

This PR changes this back.

## Views affected
All

### DIFF
--- a/src/ensembl/src/shared/components/image-button/ImageButton.tsx
+++ b/src/ensembl/src/shared/components/image-button/ImageButton.tsx
@@ -57,11 +57,7 @@ export const ImageButton = (props: Props) => {
         )}
       </button>
       {shouldShowTooltip && (
-        <Tooltip
-          anchor={hoverRef.current}
-          autoAdjust={true}
-          delay={TOOLTIP_TIMEOUT}
-        >
+        <Tooltip anchor={hoverRef.current} autoAdjust={true}>
           {props.description}
         </Tooltip>
       )}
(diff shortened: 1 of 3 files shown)


imransl:
It still appears uncomfortably slowly. Can't the timeout that we have given be reduced from `800` to a smaller figure?

azangru:
> It now appears uncomfortably slowly

Really? Doesn't It have the same delay as on 2020 site?

imransl:
> > It now appears uncomfortably slowly
> 
> Really? Doesn't It have the same delay as on 2020 production site?

True albeit very slow as in the production site.

azangru:
> It is as if the tooltip doesn't exist when I hover over an icon and then it finally comes up.

I think this should indeed be the case. Most of the times when you interact with an element you shouldn't see its tooltip. It's when you hover over an element for a while — which means that you are unsure about what the element does — that the interface tries to give you a hint.

Try interacting with elements on other sites / in other apps. For example, pause the cursor over the Star or the Fork button on Github:

image

you will see a similar delay before a native tooltip appears.

imransl:
> > It is as if the tooltip doesn't exist when I hover over an icon and then it finally comes up.
> 
> I think this should indeed be the case. Most of the times when you interact with an element you shouldn't see its tooltip. It's when you hover over an element for a while — which means that you are unsure about what the element does — that the interface tries to give you a hint.
> 
> Try interacting with elements on other sites / in other apps. For example, pause the cursor over the Star or the Fork button on Github:
> 
> image
> 
> you will see a similar delay before a native tooltip appears.

Didn't think of it in that way. I was for some reason thinking about tooltips as additional info that I would want to know immediately. Anyway let's leave it like this, since that's what others are also following 😄

PR #282, opened 2020-04-11
Refactoring to support changed feature ids

azangru:
## Type
- Refactoring

## Related JIRA Issue(s)
ENSWBSITES-531

## Important!
This PR relies on the updated endpoints of the `ensembl-2020-genome-search` api, which are currently committed to [this branch](https://github.com/Ensembl/ensembl-2020-genome-search/tree/refactor/feature-id). Merge only together with the backend branch! 

## Background
Our feature id (which we used to call ensObject id) is fragmenting into several distinctly different identifiers:
- feature id in the url
- feature id for backend apis
- feature id for genome browser
- feature id for use as keys in in-app storage (e.g., in redux)

### Feature id in the url 
Previous genome browser urls were formed according to the following pattern:

`/browser/:genome_id?focus=:feature_id`, where `feature_id` was `<genome_id>:<type>:<stable_id>`

The problems with this were that:
- `genome_id` as part of focus id was redundant, because it was already present in another part of url
- this redundancy created two sources of truth for `genome_id` — it could be read either from the url path, or directly from the feature_id

The decision was to change feature ids in the url to be just `<type>:<stable_id>`, without the genome_id part.

### Feature id for backend apis
When we built Species Selector and Genome Browser, we worked under assumption that we will have our dedicated backend that will understand the feature id format of `<genome_id>:<type>:<stable_id>`. Since then, the consensus seems to be growing that the client will talk with various apis that do not understand our feature id format. Therefore, when we send api requests, we rely on three independent parameters: genome_id, type, and stable_id to fetch information about entities.

### Feature id for genome browser
The feature id for genome browser remains the same, and has a format of `<genome_id>:<type>:<stable_id>`

### Feature id for use as keys to in-app storage
For storing information about the features in our client-side application, we need unique keys. Neither the stable_id by itself, nor the combination of `<type>:<stable_id>` are sufficiently unique; therefore, we will use the old format of `<genome_id>:<type>:<stable_id>` for keys.

## Description
The PR contains code changes to support the updated feature id format described above.

The corresponding backend changes are:
- in the genome info endpoint — the shape of the example objects field has changed from array of srtings to array of objects ([example](http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32014/api/genome/info?genome_id=homo_sapiens_GCA_000001405_27)):

```js
{
  example_objects: [
    {
      id: "ENSG00000139618",
      type: "gene",
    },
    {
      id: "17:63992802-64038237",
      type: "region",
    },
  ]
}
```
- `api/object/info` and `api/object/track_list` endpoints now expect independent `genome_id`, `type` and `stable_id` parameters (examples [1](http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32014/api/object/info?genome_id=homo_sapiens_GCA_000001405_27&type=gene&stable_id=ENSG00000139618), [2](http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32014/api/object/track_list?genome_id=homo_sapiens_GCA_000001405_27&type=gene&stable_id=ENSG00000139618))

**ALSO:**
- moved routing logic from Browser.tsx to a custom hook (`useBrowserRouting`)

**NOTE:**
Previous use of the ActionCreator type from the redux library when typing our action creators turned out to be a very bad idea, because the ActionCreator type is [defined](https://github.com/reduxjs/redux/blob/9d3273846aa8906d38890c410b62fb09a4992018/src/types/actions.ts#L52-L54) as follows:

```
export interface ActionCreator<A> {
  (...args: any[]): A
}
```
which means that all the type information about the arguments passed to the action creator is lost. I have changed the code of some of the action creators to avoid using this type.

## Views affected
- GenomeBrowser page
- EntityViewer page

## Deployed to
http://feature-feature-id.review.ensembl.org


src/ensembl/webpack/environments/webpack.config.dev.js
```diff
@@ -32,7 +32,7 @@ const devServerConfig = {
   // rules to proxy requests to the backend server in development
   proxy: {
     '/api': {
-      target: 'https://staging-2020.ensembl.org',
+      target: 'http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32014',
```
azangru, line 35: **NOTE TO SELF:** remember to change back

src/ensembl/src/content/app/browser/browserHelper.ts
```diff
@@ -23,6 +31,13 @@ export function getChrLocationStr(
   return `${chrCode}:${startBp}-${endBp}`;
 }
 
+export const stringifyGenomeBrowserFocusId = (params: GenomeBrowserFocusIdConstituents) => 
+  buildEnsObjectId(params);
```
msmanoj, line 35: Why do we need both?

src/ensembl/src/content/app/browser/zmenu/zmenuGenomeBrowserAdaptor.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * This is a temporary file to adapt data sent from genome browser with bpane-zmenu event
+ * to the needs of browser chrome (primarily as regards to the id format)
+ */
+
+import { parseFeatureId } from 'src/content/app/browser/browserHelper';
```
msmanoj, line 6: I guess this can either be removed or commented out.

src/ensembl/src/content/app/browser/browserHelper.ts
azangru: You mean why do we need both genome browser focus id and our internal id for use as a key when storing ensObject in redux?

The reason for this is to decouple genome browser ids from our internal ids. The format genome_id:type:stable_id should be enough for us to keep track of ensObjects in redux store; but we cannot predict whether genome browser's requirements for focus ids will change. Genome browser is essentially a different application from browser chrome; there is no point in coupling these two ids together. They are identical at the moment, but may not stay so in the future.

src/ensembl/src/content/app/entity-viewer/components/example-links/ExampleLinks.tsx
```diff
@@ -0,0 +1,78 @@
...
+    gene(byId: { id: $id }) {
+      id
+      symbol
+    }
+  }`;
+
+// NOTE: the component currently handles only example gene
+const ExampleLinks = (props: ExampleLinksProps) => {
+  const exampleGeneId = props.exampleEntities.find(({ type }) => type === 'gene')?.id;
+  const { loading, data } = useQuery<{ gene: ExampleGene }>(QUERY, {
```
msmanoj, line 39: `loading` is unused.

src/ensembl/src/shared/state/ens-object/ensObjectHelpers.ts
```diff
@@ -1,4 +1,97 @@
...
+    type,
+    objectId
+  };
+};
+
+export const buildRegionObject = (
+  payload: EnsObjectIdConstituents
+): EnsObjectRegion => {
+  const { genomeId, objectId: regionId } = payload;
+  const [chromosomeName, start, end] = getChrLocationFromStr(regionId);
```
msmanoj, line 81: `chromosomeName`  -> `chromosome`?

azangru: ✔️

src/ensembl/src/content/app/entity-viewer/components/example-links/ExampleLinks.tsx
azangru: Added a spinner to show for the `loading` state.

src/ensembl/src/content/app/browser/zmenu/zmenuGenomeBrowserAdaptor.ts
azangru: Yep, the whole `zmenuGenomeBrowserAdaptor.ts` can actually be removed. Thanks for reminding!

azangru: ✔️

PR #283, opened 2020-04-14
Add radio button to storybook

ens-ma7:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-543](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-543)

## Description
This is to extend the current radio button shared component to make it a styled radio button and also to add it to storybook.

## Views affected
Storybook

## Possible complications
n/a


src/ensembl/src/shared/components/radio/Radio.scss
```diff
@@ -1,7 +1,42 @@
...
+  width: 9px;
+  height: 9px;
+  border-radius: 50%;
+  background: $green;
+  display: block;
+  content: '';
+  position: absolute;
+}
+
+.radioLabel {
```
azangru, line 39: When hovering over the label, the cursor should indicate that the label is clickable (see [conversation on Slack](https://genomes-ebi.slack.com/archives/C67KM6TD0/p1585687915016400) and [discussion in another PR](https://github.com/Ensembl/ensembl-client/pull/275#discussion_r404806722)).

azangru:
According to the DOM representation, there are several problems with how the radiobutton element is modelled:

image

1. The most important problem is that the input is using index for id. This works if there is only one group of radio buttons on the page, but if there are several, they will start having the same ids  (whereas an id has to be unique per page).

The id is being used to associate the radio button with the label. We can remove both the id from the input and the htmlFor from the label, and still have the desired behaviour, by either of the following approaches:
- wrap the input inside the label tag, or
- add the onClick handler to the label

2. There is a `div` element inside a `span` element, which, although admissible, is generally frowned upon, because spans are inline elements by default, and divs are block elements by default. Either change the span to the div (so that there is div inside a div), or the div to the span (so that there is span inside a span). Probably changing span to div is better.

src/ensembl/src/shared/components/radio/Radio.scss
```diff
@@ -1,7 +1,42 @@
...
+  margin-left: 30px;
+  color: $medium-light-grey;
```
azangru, line 41: Default label color should probably be black (compare to checkbox)

src/ensembl/src/shared/components/radio/Radio.scss
```diff
@@ -1,7 +1,42 @@
...
+  height: 9px;
+  border-radius: 50%;
+  background: $green;
+  display: block;
+  content: '';
+  position: absolute;
+}
+
+.radioLabel {
+  margin-left: 30px;
```
azangru, line 40: Question to discuss with Andrea: should the distance between a checkbox and a label:

image

be the same as the distance between a radio button and a label:

image

In Storybook, it seems that they are different

ens-ma7: Yep thats something i was thinking of confirming with Andrea once we add it to 2020 (filter and sort by component in EV). For now just setting some margin and i will confirm later if thats ok.

src/ensembl/src/shared/components/radio/Radio.tsx
```diff
@@ -14,6 +16,9 @@ type Props = {
   classNames?: {
     radioInput?: string;
```
azangru, line 17: I understand that this is from a previous commit, but we don't want to allow parent components to style the native radio input element at all — we always want to hide it. I think this field should be removed.

src/ensembl/src/shared/components/radio/Radio.tsx
```diff
@@ -14,6 +16,9 @@ type Props = {
...
     radioLabel?: string;
+    radioWrapper?: string;
+    defaultRadio?: string;
```
azangru, line 20: Notice the repeating word "radio"? We could remove some of the repetition it by renaming the properties to:

- radio
- checked (or checkedRadio)
- label
- wrapper

src/ensembl/src/shared/components/radio/Radio.tsx
```diff
@@ -14,6 +16,9 @@ type Props = {
...
+    checked?: any;
```
azangru, line 21: Why `any`? Surely it's `string`

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [selectedRadio, setselectedRadio] = useState('default');
+
+  const handleOnchange = (value: string) => {
+    setselectedRadio(value);
+    action('radio-clicked')(value);
+  };
+
+  return (
+    <div>
+      <Radio
```
azangru, line 24: One thing I now realise I missed in the initial commit of this component is its name. I thought the name `Radio` describes a single radiobutton with the label; but now I see it describes the whole group of radio buttons. Does anyone other than me find this confusing? Is it not too late to rename the component to RadioGroup?

Examples in other design systems:
- Ant Design — RadioGroup ([link](https://ant.design/components/radio/#components-radio-demo-radiogroup-options))
- Material UI — RadioGroup ([link](https://material-ui.com/components/radio-buttons/))

src/ensembl/src/shared/components/radio/Radio.scss
ens-ma7: That wont work for the entity viewer filter and sort where the background is dark. Lets leave it as it is for now and we can change it if needed and if it is used in other components with white background.

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+      ></Radio>
+    </div>
+  );
+};
+
+storiesOf('Components|Shared Components/Radio', module)
+  .add('default', () => {
+    return <Wrapper />;
+  })
+  .add('Entity Viewer style', () => {
```
azangru, line 38: I am not sure there is much point of having a dedicated story for EntityViewer background — it doesn't show us anything new except for how the radio element can accept custom classes from the parent. For example, we don't have such stories for our checkboxes.

src/ensembl/src/shared/components/radio/Radio.scss
azangru: > That wont work for the entity viewer filter and sort where the background is dark.

That's what custom classes are for (see checkboxes).

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
ens-ma7: I thought the same as well but then it clicked in my head that a radio on it own will never exist, they are always a group, so the name sort of mean a group of radio since one of its own will never exist and wont cause any confusion.

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
ens-ma7: Yep i wasn't sure about this one. So just the default one should be enough, is that right?

azangru: > So just the default one should be enough, is that right?

Yes, I think so.

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
azangru: @jyo @imransl @msmanoj what are your opinions?

Personally, I would strongly prefer RadioGroup.

src/ensembl/src/shared/components/radio/Radio.scss
ens-ma7: Agreed with to keep it same as checkbox, will change it the default label colours.

src/ensembl/stories/shared-components/radio/Radio.stories.tsx
imransl: RadioGroup more sense to me than Radio if it is a group of radio buttons.

ens-ma7: ok going with RadioGroup

src/ensembl/src/shared/components/radio/Radio.tsx
ens-ma7: i didnt quite understand this as well, when i put string, it throws a warning at the following line:
`{ [styles.checked]: option.value === props.selectedOption}`
And the error is 
`
(property) checked?: string | undefined
A computed property name must be of type 'string', 'number', 'symbol', or 'any'.ts(2464)
`
I didn't quite understand. Any idea why?

ens-ma7:
@azangru  Implemented the suggested changes except for the margin between the label (will confirm with Andrea when adding to Entity viewer). Please have a look and let me know. Thanks

src/ensembl/src/shared/components/radio/Radio.tsx
azangru: Yes. With `checked` as an optional property, what it is saying is that checked can be `undefined`, and it doesn't want to set an object key as undefined. 

you can change this:
```
const radioClass = classNamesMeger(styles.defaultRadio, {
  [styles.checked]: option.value === props.selectedOption
});
```
to this:
```
const radioClass = classNamesMeger(styles.defaultRadio, 
  option.value === props.selectedOption ? styles.checked : undefined
);
```
then it won't complain.

Also, please change `any` in `props.options.map((option: any, index: number)` to `RadioOption`. Or, better yet, just remove the types from both option and index parameters — they are trivially inferred, and Typescript can figure them out by itself.

azangru:
Checks are failing:

```
stories/shared-components/radio-group/RadioGroup.stories.tsx(5,8): error TS2307: Cannot find module 'src/shared/components/radio-group/RadioGroup'.

 ERROR: Job failed: exit code 1
```

src/ensembl/src/shared/components/radio/Radio.tsx
ens-ma7: @azangru thats all done now. Thanks for your help. can you please have a look and let me know if its all good (Dont forget to accept the PR if its all good :) ).
The check should also now be fixed, i have forgotten to push one file.

azangru:
I made some small changes and pushed them to this branch as separate commits, not to delay the PR further:

- renamed class names: `default` -> `radio`; `checked` -> `radioChecked`
- made sure radio button does not call the `onChange` handler when it is already selected (it was doing so before)
- removed the `applies the style "default" by default` test. It was incorrect anyway (ran `expect(wrapper.find('.default')).toBeTruthy()` instead of `expect(wrapper.find('.default').length).toBeTruthy()`), and so was passing even after I renamed the `default` class to `radio`.

With that, I am approving the PR.

src/ensembl/src/shared/components/radio-group/RadioGroup.tsx
```diff
@@ -1,26 +1,26 @@
 import React from 'react';
+import classNames from 'classnames';
```
ens-ma7, line 2: I get confused with this classNames and the classNames inside props. is that ok? dont they conflict or do you guys not find it confusing, which one is which. Thats why i prefer classNamesMerger.

ens-ma7:
Thanks for the changes @azangru it makes sense just one tiny thing i asked about the classNames import.

src/ensembl/src/shared/components/radio-group/RadioGroup.tsx
azangru: > dont they conflict or do you guys not find it confusing, which one is which

Personally, I don't find them confusing, because the `classNames` that comes from props sits in the `props` object and can easily be accessed with `props.classNames`. So they don't really conflict, and the moment they start conflicting, typescript will yell at you. I think @msmanoj used `classNamesMerger` in some cases where he found having the same name confusing. One thing to keep in mind is that in most of our files we call this imported function classNames, so we kinda got used to the name. Don't really have particularly strong feelings (apart from instead of `classNamesMerger`, which is a noun, I would have used `mergeClassNames` which is a verb and as such better suited to be the name of a function).

ens-ma7: Thanks for this, i guess for me as i was trying to understand the code and i know for sure specially this file when i was trying to figure out each line, i did get confuse with classNames but as you pointed out and i didnt realise it the props one will always accompany props. and i guess its just a matter getting use with the react code. I am fine with that, thanks for the explanation, it is helpful.

PR #286, opened 2020-04-17
Integrate instant download into default transcripts list

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-557

## Description
This PR integrates InstantDownloadTranscripts component into the DefaultTranscriptsList of the EntityViewer.

Steps to see the component:
- go to the gene view of EntityViewer
- click on one of the transcripts to open a transcript info block 
- click on the Download link to see the InstantDownload component

## Views affected
Entity viewer page

## Feature branch deployed to:
http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32015/

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
@@ -3,21 +3,24 @@ import { mount } from 'enzyme';
 
 import TranscriptsListItemInfo from './TranscriptsListItemInfo';
 
+import { createGene } from 'tests/fixtures/entity-viewer/gene';
 import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
 
 describe('<TranscriptsListItemInfo /', () => {
   let wrapper: any;
+  const transcript = createTranscript();
+  const gene = createGene({ transcripts: [transcript] });
+  const props = {
+    gene,
+    transcript
+  };
 
   beforeEach(() => {
-    const props = {
-      transcript: createTranscript()
-    };
-
     wrapper = mount(<TranscriptsListItemInfo {...props} />);
   });
 
   it('displays amino acid length when transcript has CDS', () => {
-    expect(wrapper.find('#amino-acid-length')).toHaveLength(1);
+    expect(wrapper.find('.topMiddle strong')).toHaveLength(1);
   });
 
   it('contains the download link', () => {
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
@@ -19,8 +19,19 @@ describe('<TranscriptsListItemInfo /', () => {
     wrapper = mount(<TranscriptsListItemInfo {...props} />);
   });
 
+  /*
+   * FIXME: the test below will have to change when api payload updates:
+   * 1) we will use protein length from api response instead of calculating it ourselves
+   * 2) we will check that protein product is present on a transcript instead of looking at CDS
+   */
   it('displays amino acid length when transcript has CDS', () => {
-    expect(wrapper.find('.topMiddle strong')).toHaveLength(1);
+    const totalExonsLength = props.transcript.exons.reduce((sum, exon) => {
+      return sum + exon.slice.location.end - exon.slice.location.start + 1;
+    }, 0);
+    const expectedProteinLength = Math.floor(totalExonsLength / 3);
+    expect(wrapper.find('.topMiddle strong').text()).toMatch(
+      `${expectedProteinLength}`
+    );
   });
 
   it('contains the download link', () => {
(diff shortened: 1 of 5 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -8,18 +8,28 @@ import {
...
 import transcriptsListStyles from '../DefaultTranscriptsList.scss';
 import styles from './TranscriptsListItemInfo.scss';
 
 type ItemInfoProps = {
+  gene: Gene;
   transcript: Transcript;
 };
 
 const ItemInfo = (props: ItemInfoProps) => {
+  const [isDownloadShown, setIsDownloadShown] = useState(false);
```
imransl, line 27: How about `isDownloadOpened` instead of `isDownloadShown` since that's what we are using `setIsDownloadShown` to update the state?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -3,21 +3,24 @@ import { mount } from 'enzyme';
...
-    const props = {
-      transcript: createTranscript()
-    };
-
     wrapper = mount(<TranscriptsListItemInfo {...props} />);
   });
 
   it('displays amino acid length when transcript has CDS', () => {
-    expect(wrapper.find('#amino-acid-length')).toHaveLength(1);
+    expect(wrapper.find('.topMiddle strong')).toHaveLength(1);
```
msmanoj, line 23: Should the `strong` tag have some className specified? 
I am asking because if we change the content to be something else in future instead of `amino acid length`, this test will still pass.

azangru: There are several layers to this question:
1) What to do when we do not have a sufficiently unique identifier
2) Are CSS selectors good enough at all
3) Is a test that just looks at the presence of a particular element on the page useful at all.

Your question "what if we change the content to be something else" refers to all three of them. Even if an element has a class name associated with it, when we change the content of this element to be something else, without changing the class of the container element, the test would still pass.

I agree that `.topMiddle strong` is rather non-specific. What I'd like to avoid (going back to sub-question 1) is having classes that are not associated with any styling. What we **definitely** should avoid is adding actual ids to non-unique elements, because ids should be unique per page. This is why I changed the selector in this test.

Some people solve this problem by adding the `data-test-id` attribute to respective elements, and some even clean up these otherwise useless attributes from the production build with something like [this plugin](https://www.npmjs.com/package/babel-plugin-jsx-remove-data-test-id). I don't know if we want to move in this direction.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
azangru: >  since that's what we are using setIsDownloadShown to update the state

I am not sure I quite understand what you mean, or that the download is a thing that "opens" (I thought what was opening/closing was more like the footer of the info panel; and the download element is its content). What is it about the "shown" that you do not like?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
msmanoj: I agree that there is no point in adding a classname when they are not associated with any styling.
I don't have a strong opinion to which way to go for but the only thing I can think of is, in future, if a new developer wants to investigate a wrong value displayed here, it will be more easier to search where `#amino-acid-length` is used rather than `topMiddle` or `strong`. It will be straightforward for those who clearly knows the directory structure but for others, I guess not. This is something I do for the current site to find where certain values are used.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: Apologies. I got them mixed up. Ignore my suggestion 😃

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
azangru: I've updated the test to be more rigorous. I kept the selector the same, but I am now checking the actual content of the element showing amino acids count.

Incidentally, when I updated the test to actually check the content of the element, I found that the code in the component had an error — it was returning an incorrect number of amino acids if a transcript has a single exon. It's a perfect demonstration of how checking simply for the presence of a DOM element is not a particularly useful test in many circumstances 😄 Given that it's temporary code anyway, I added a quick-and-dirty fix to the component.

The conversation in this thread is very important, but I suggest we discuss our attitude towards selectors in testing separately and maybe create a separate PR with appropriate changes. I am leaning towards admitting that in some circumstances we might need to resort to `data-test-id`s.

PR #288, opened 2020-04-29
Integrate InstantDownloadTranscript in zmenu

azangru:
## Type
- New feature
- Bugfix

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-556

## Description
This PR adds the InstantDownloadTranscript component to the Zmenu in GenomeBrowser

Testing scenario:
- visit Genome Browser page
- click on a feature inside the Genome Browser canvas
- in the zmenu that will pop up, click on "Download"
- observe the InstantDownload component

Note: non-protein-coding features should not show checkboxes related to CDS or protein sequence.

**Also in this PR:**
- added `useApiService` hook to use the api service from inside a component (in the future, might consider switching to [react-query](https://github.com/tannerlinsley/react-query))
- fixed a bug in InstantDownloadTranscript component (it was not hiding protein-related checkboxes for non-protein-coding features)
- updated styling (including that of the Loader) after consultation with Andrea

## Views affected
Genome Browser page

## Feature deployed to
http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:31372


src/ensembl/src/shared/hooks/tests/useApiService.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    expect(wrapper.find('.success').length).toBe(0);
+    expect(wrapper.find('.error').text()).toBe(mockErrorData.message);
+  });
+
+  it('aborts request on unmount if passed isAbortable option', () => {
+    const wrapper = mount(<TestingComponent isAbortable={true} />);
+    wrapper.unmount();
+
+    expect(onAbort).toHaveBeenCalled();
```
imransl, line 95: Shouldn't we check the inverse of this as well i.e. when isAbortable is not passed then `onAbort` is not called?

src/ensembl/src/content/app/browser/zmenu/ZmenuInstantDownload.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  if (loadingState === LoadingState.LOADING) {
+    return (
+      <div className={styles.zmenuInstantDowloadLoading}>
+        <CircleLoader />
+      </div>
+    );
+  }
+
+  if (error) {
```
imransl, line 35: Should there be a todo comment here to revisit this later?

src/ensembl/src/shared/hooks/useApiService.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import { LoadingState } from 'src/shared/types/loading-state';
+
+type Params = FetchOptions & {
+  endpoint: string;
+  isAbortable?: boolean;
+};
+
+type StateAtLoading = {
+  loadingState: LoadingState.LOADING;
```
msmanoj, line 13: Just out of curiosity, are there any advantages in using the `LoadingState` enum for the loadingState instead of `boolean`?

imransl: I don't think `boolean` is suited for this as we also have types for success and error which will be of different `loadingState` type.

azangru: It's a difficult question, because it can have several answers.

On a purely practical level, the answer would be that the `isLoading` boolean in `useQuery` clearly works. When writing a components, one can first check `isLoading`, then, if it's false, assume that either the `data` or the `error` will have a value, and so check the `error` field and the `data` field. It works.

The answer on a more abstract level, though, is that the `isLoading` boolean does not tell you anything about whether your request has succeeded or failed. It does not allow you to infer the values of your `data` and error fields:

success:
```js
{
  isLoading: false,
  data: { foo: 'bar' },
  error: null
}
```

failure:
```js
{
  isLoading: false,
  data: null,
  error: { message: 'oh no!' }
}
```

🤷‍♂️ 
```js
{
  isLoading: false,
  data: null,
  error: null
}
```

🙅‍♂️ (but no guarantee in code to guard against this)
```js
{
  isLoading: false,
  data: { foo: 'bar' },
  error: { message: 'oh no!' }
}
```

The last combination is clearly invalid; the one before last is confusing (either invalid or initial).

Using an enum field as a flag allows you to type different combinations of fields and to eliminate impossisble combinations; so from the perspective of type guarantees, it's a definite win.

A more robust and formal alternative is using a state machine; but I don't know whether we are ready to go quite as far.

There are a number of blog posts on this topic (e.g. [one by Kent C. Dodds](https://kentcdodds.com/blog/stop-using-isloading-booleans), or [one by ThoughtBot](https://thoughtbot.com/blog/booleans-and-enums))

src/ensembl/src/shared/hooks/tests/useApiService.test.tsx
azangru: Bit hard to prove the negative 😄 , but I added the check in the `fetches data` test

src/ensembl/src/shared/hooks/useApiService.ts
msmanoj: Cool. Thanks for that.

Do you think we need an option to `skip`?

src/ensembl/src/content/app/browser/zmenu/ZmenuInstantDownload.tsx
azangru: 🤷‍♂️ but ✔️

src/ensembl/src/shared/hooks/useApiService.ts
azangru: > Do you think we need an option to `skip`?

Good question! `skip` will complicate the matters slightly, because it will introduce a new loading state possibility — where no request has been made:

```
{
  loadingState: LoadingState.NOT_REQUESTED;
  data: null;
  error: null;
}
```
I didn't have a need for it, but probably a good thing to add it just for completeness' sake.

PR #290, opened 2020-05-12
Add cache to api-service

azangru:
## Type
- New feature

## Related JIRA Issue(s)
ENSWBSITES-608

## Description
**Purpose:** Some user interactions resulting in http requests can occur repeatedly (e.g. user opening and closing an element that retrieves its content from the network). We would prefer that if the data has been received by the client it were used in subsequent requests. 

This PR:
- Adds a cache class implementing the Least Recently Used mechanism for ensuring maximum size
- Adds the cache to the api service
- Adds tests

Also:
- update the way we are getting random values in Tabs.test to avoid duplicate keys in lists, such as happened here:

Screenshot 2020-05-11 at 23 05 28

## Possible alternatives
There is a library that's gaining traction in the React community called [react-query](https://github.com/tannerlinsley/react-query), which does caching and other cool things such as request cancellation, request deduplications, or retries. It can't be used from within redux though, as opposed to our api-service; so if we decide to use it, we might want to reconsider how we fetch stuff.

### DIFF
--- a/src/ensembl/src/services/api-service.ts
+++ b/src/ensembl/src/services/api-service.ts
@@ -22,7 +22,7 @@ type FetchOptions = {
   headers?: { [key: string]: string };
   body?: string; // stringified json
   preserveEndpoint?: boolean;
-  skipCache?: boolean;
+  noCache?: boolean;
 };
 
 const defaultMethod = HTTPMethod.GET;
@@ -68,7 +68,7 @@ class ApiService {
     const fetch = this.getFetch();
     const url = options.preserveEndpoint ? endpoint : `${host}${endpoint}`;
 
-    if (!options.skipCache) {
+    if (!options.noCache) {
       const cachedItem = this.cache.get(url);
       if (cachedItem) {
         return cachedItem;
@@ -86,7 +86,7 @@ class ApiService {
         response,
         fetchOptions
       );
-      if (!options.skipCache) {
+      if (!options.noCache) {
         this.cache.set(url, processedResponse);
       }
       return processedResponse;
--- a/src/ensembl/src/services/api-service.ts
+++ b/src/ensembl/src/services/api-service.ts
@@ -2,6 +2,7 @@
 // we will modify it to use a library once we decide which one to choose
 
 import config from 'config';
+import LRUCache from 'src/shared/utils/lruCache';
 
 export enum HTTPMethod {
   GET = 'GET',
@@ -21,6 +22,7 @@ type FetchOptions = {
   headers?: { [key: string]: string };
   body?: string; // stringified json
   preserveEndpoint?: boolean;
+  skipCache?: boolean;
 };
 
 const defaultMethod = HTTPMethod.GET;
@@ -28,9 +30,11 @@ const defaultHeaders = { Accept: 'application/json' };
 
 class ApiService {
   private host: string;
+  private cache: LRUCache;
 
   public constructor(config: ApiServiceConfig) {
     this.host = config.host || '';
+    this.cache = new LRUCache({ maxAge: 60 * 60 * 1000 });
   }
 
   // temporary method, for easy testing, until we choose a library
@@ -63,6 +67,14 @@ class ApiService {
     const host = options.host || this.host;
     const fetch = this.getFetch();
     const url = options.preserveEndpoint ? endpoint : `${host}${endpoint}`;
+
+    if (!options.skipCache) {
+      const cachedItem = this.cache.get(url);
+      if (cachedItem) {
+        return cachedItem;
+      }
+    }
+
     const fetchOptions = this.buildFetchOptions(options);
 
     try {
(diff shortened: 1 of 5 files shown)


src/ensembl/src/services/api-service.ts
```diff
@@ -21,16 +22,19 @@ type FetchOptions = {
   headers?: { [key: string]: string };
   body?: string; // stringified json
   preserveEndpoint?: boolean;
+  skipCache?: boolean;
```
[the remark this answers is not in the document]
azangru: preserveEndpoint above is also a verb. Weirdly, it feels fine to me to have verbs/commands in config options.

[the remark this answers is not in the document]
azangru: `noCache` ? 😉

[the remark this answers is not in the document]
imransl: Yes, that would sound even better.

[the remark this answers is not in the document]
azangru: Option renamed.

src/ensembl/src/shared/utils/lruCache.ts
```diff
@@ -0,0 +1,157 @@
...
+      }
+      this.moveNodeToHead(cachedNode);
+      cachedNode.timestamp = Date.now();
+      return cachedNode.value;
+    }
+  }
+
+  public set(key: string, value: any) {
+    const savedNode = this.listNodesMap.get(key);
+    if (savedNode) {
```
msmanoj, line 72: Will this ever be true?

If I understand correctly, we `set` only when `get` returns nothing.

azangru: > If I understand correctly, we set only when get returns nothing.
That's how the api service is using the cache. But if you just take LRUCache as a standalone unit, there is nothing preventing you from calling the method `set` on it without first calling the `get`. I think it should be capable of handling this kind of misuse gracefully.

src/ensembl/src/shared/utils/lruCache.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (previousNode) {
+      previousNode.next = nextNode;
+    }
+    if (nextNode) {
+      nextNode.previous = previousNode;
+    }
+    const formerHeadNode = this.head;
+    this.head = node;
+    node.next = formerHeadNode;
+    node.previous = null;
```
msmanoj, line 106: Shouldn't it be the other way?
```javascript
node.previous = formerHeadNode
node.next = null
```

azangru: The list is growing towards its tail. The most recently accessed node is the head; the least recently accessed node is the tail.

If you access a node in the middle of the list, you remove it from its place and put it in the head. The head does not have previous nodes, so node.previous is null. The one that used to be head before that becomes less recently accessed, and thus it gets moved towards the tail, which means that the new head node will reference it via node.next.

msmanoj: > The list is growing towards its tail.

In the following example, is the node `[null,1,2]` considered to be the head with previous as `null` and next as `2`?
[null,1,2] <-> [1,2,3] <-> [2,3,null]

azangru: Yes, if I understand your notation correctly.

So when we move node 2 to head, it should be [null,2,1] <-> [2,1,3] <-> [1,3,null].

Uh-oh, did I forget to link node 3 with node 1? I did, didn't I.

msmanoj: > Uh-oh, did I forget to link node 3 with node 1? I did, didn't I.

Why will that be required?

azangru: Well, my list is broken without it:

Given:
[null,1,2] <-> [1,2,3] <-> [2,3,null]

Expected after node switching:
[null,2,1] <-> [2,1,3] <-> [1,3,null]

But am likely getting:
```
[null,2,1] <-> [2,1,3] -> [2,3,null]
           <-----------
```

azangru: Sorry, now that I am reading through the code of the method again, the only thing that I missed is assigning previous to formerHeadNode.

Starting with:
[null,1,2] <-> [1,2,3] <-> [2,3,null]

If we want to move [1,2,3] to head:

```
    const previousNode = node.previous;
    const nextNode = node.next;
    if (previousNode) {
      previousNode.next = nextNode;
    }
    if (nextNode) {
      nextNode.previous = previousNode;
    }
```
this should establish the mutual links between node 1 and node 3:

```[null,1,3] <-> [1,3,null]```

And the next set of instructions:

```
    const formerHeadNode = this.head;
    this.head = node;
    node.next = formerHeadNode;
    node.previous = null;
```

will move node 2 to head:

```[null,2,1] <-> [null,1,3] <-> [1,3,null]```

So what remains to be done is to set formerHeadNode.previous to node (i.e. to 2):

```[null,2,1] <-> [2,1,3] <-> [1,3,null]```

msmanoj: Cool, that sounds right 👍

azangru: I updated the tests, and found and fixed another bug in the process 😅

PR #292, opened 2020-05-12
EV sticky tabs and filter

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-578

## Description
Updated the tabs and table filter to be sticky.

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
@@ -26,7 +26,7 @@
 
 .geneViewTabs {
   grid-area: 2 / 1 / 3 / 4;
-  z-index: 1;
+  z-index: 2;
   position: sticky;
   top: -105px;
   height: 100%;
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
@@ -9,8 +9,6 @@
   padding: 75px 20px 10px 20px;
   height: 100%; // may need to change this later
   width: 100%;
-  max-width: 100vw;
-  position: absolute;
   overflow-y: auto;
   overflow-x: hidden;
 }
@@ -29,7 +27,6 @@
 .geneViewTabs {
   grid-area: 2 / 1 / 3 / 4;
   z-index: 1;
-  align-self: end;
   position: sticky;
   top: -105px;
   height: 100%;
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
@@ -9,6 +9,10 @@
   padding: 75px 20px 10px 20px;
   height: 100%; // may need to change this later
   width: 100%;
+  max-width: 100vw;
+  position: absolute;
+  overflow-y: auto;
+  overflow-x: hidden;
 }
 
 .featureImage {
@@ -23,9 +27,14 @@
 }
 
 .geneViewTabs {
-  grid-area: 2 / 2 / 3 / 4;
+  grid-area: 2 / 1 / 3 / 4;
   z-index: 1;
   align-self: end;
+  position: sticky;
+  top: -105px;
+  height: 100%;
+  background-color: $black;
+  padding-left: 228px;
 }
 
 .geneViewTabContent {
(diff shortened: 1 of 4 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -9,6 +9,10 @@
   padding: 75px 20px 10px 20px;
   height: 100%; // may need to change this later
   width: 100%;
+  max-width: 100vw;
```
azangru, line 12: This (and the previous) lines are probably useless. geneView is just a div whose width is controlled by the width of the main container of StandardAppLayout

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -9,6 +9,10 @@
...
+  position: absolute;
```
azangru, line 13: I don't like this idea. Rather, let's fix the height on the `.standardAppLayout` class:

```
.standardAppLayout {
  height: calc(100vh - 159px);
}
```
where 159px is header height plus appBar height; and then geneView will take 100% of the allotted height.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
```diff
@@ -23,9 +27,14 @@
 }
 
 .geneViewTabs {
-  grid-area: 2 / 2 / 3 / 4;
+  grid-area: 2 / 1 / 3 / 4;
   z-index: 1;
   align-self: end;
```
azangru, line 32: no need for align-self: end then, if the position of the buttons is controlled through position: absolute

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
```diff
@@ -38,8 +38,9 @@ $drawerWindowWidth: 45px;
 }
 
 .mainWrapper {
-  height: calc(100% - #{$topbarHeight});
+  max-height: calc(100% - #{$topbarHeight});
   position: relative;
+  height: 100%;
```
azangru, line 43: these two changed lines are probably unnecessary and should behave exactly the same as the deleted line.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -4,6 +4,10 @@
 .header {
   border-bottom: 1px solid $orange;
   padding: 8px 0;
+  position: sticky;
+  top: -15px;
+  background-color: $black;
+  z-index: 1;
```
azangru, line 10: probably should raise `.geneViewTabs` z-index to 2 then; so that in the unlikely case the header scrolls past the tab buttons it will go behind them

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
msmanoj: Removed it

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
msmanoj: Removed it

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
msmanoj: Reverted it

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
msmanoj: That's done

azangru:
Do we know how EntityViewer is expected to behave when the panel is opened (when you click on GeneFunction or GeneRelationships button)?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
azangru: To `.geneViewTabs`, not to `.header` 😄 To avoid this:

image

image
 
image

msmanoj: Ah! My bad. That's done.

msmanoj:
> Do we know how EntityViewer is expected to behave when the panel is opened (when you click on GeneFunction or GeneRelationships button)?

I would assume the scrolling will be inside the panel as otherwise we will have to think about making the panel grow as we scroll 😱

PR #289, opened 2020-05-06
Protein list in entity viewer

imransl:
## Type
- New feature and refactoring

## Related JIRA Issue(s)
[ENSWBSITES-553](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-553)
## Importance
The protein view depends on this. Also other features that area a part of the protein view list item will be easier to test and implemented after this gets merged to dev. Once this PR is approved then we will have a protein view.

## Description
This PR adds the protein list and protein list item components. There was also some refactoring that had to be done for the REST fetching, so that the code is not too messy. I wanted to avoid it, but it was very difficult to progress without doing it.

Some parts of the code might get thrown away when we get protein data from GraphQL. So I made changes to the data types and components keeping this in mind.

Deployed to: http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32137

## Views affected
Entity viewer

### Other effects
- [ ] I have checked any requirements for Google Analytics
- [ ] I have created any required tests
- [ ] The PR has as its final commit a WASM/JS update commit if any Rust files were updated.

## Possible complications
The changes I have made to the REST fetching code might affect the Storybook components.

PR #294, opened 2020-05-15
Add view mode for entity viewer

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-577](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-577)

## Description
We need to let users directly navigate to the protein view in entity viewer from Google search results. This PR adds a view query param that will allow just that. It works only in one direction though i.e. when the view query param is available the user will be shown the proteins list. However, the URL doesn't change if the user manually navigates to the proteins list: entity viewer -> gene function -> proteins

So to test this you will need to access it through the URL itself. You can check it here:
http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32391/app/entity-viewer/homo_sapiens_GCA_000001405_27/homo_sapiens_GCA_000001405_27:gene:ENSG00000139618?view=protein

**NOTE**: Currently you'll see a lot of commits in this PR and that is because I branched from `feature/protein-list-item` which is still open (#289). These commits should disappear, once the other PR has been merged to `dev`.

## Views affected
Entity viewer


azangru:
```feature
Given I navigate to EntityViewer page (no view parameter in the url)
When I click on the "Gene Function" button
Then the panel with proteins opens
And the "view=protein" query parameter is appended to the url
```
👆 last step doesn't work

```feature
Given I navigate to EntityViewer page with the "view=protein" query in the url
Then I should see open "Gene Function" panel with proteins
When I click on the "Transcripts" button
Then the panel with proteins should close
And I should see the list of transcripts
And the "view=protein" query parameter should be removed from the url
```
👆 last step doesn't work

Without reading through the code, I believe you are still keeping the variable responsible for whether we are showing the panel or not, and which tab in the panel is active, in redux. I suspect we'll want to move that variable to the url.

imransl:
@azangru I working on fixing these issues atm. They are still a WIP, but I'll be able to finish it today. Will let you know once it's done.

imransl:
@azangru I've updated the PR with the fixes

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -133,10 +141,23 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
...
+  useEffect(() => {
+    if (props.entityViewerQueryParams.view) {
+      props.setGeneViewMode(props.entityViewerQueryParams.view);
+    }
+  }, [props.entityViewerQueryParams.view]);
+
+  const changeViewMode = (tab?: string) => {
+    const view = getViewModeName(props.selectedGeneTabName, tab);
+    const newPath = urlFor.entityViewer({ genomeId, entityId, view });
+    history.push(newPath);
```
azangru, line 158: We usually do this through connected react router.

src/ensembl/src/content/app/entity-viewer/gene-view/shared/views-helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const getViewModeName = (
+  geneViewTab: GeneViewTabName,
+  childTab?: string
+) => {
+  const allGeneViewModes = Object.assign(
+    {},
+    geneFunctionViews,
+    geneRelationshipsViews
+  );
```
azangru, line 46: just:
```
{
  ...geneFunctionViews,
  ...geneRelationshipsViews
}
```
?

src/ensembl/src/content/app/entity-viewer/gene-view/shared/views-helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const geneRelationshipsViews: { [key: string]: string } = {
+  orthologues: GeneRelationshipsTabName.ORTHOLOGUES,
+  gene_families: GeneRelationshipsTabName.GENE_FAMILIES,
+  gene_panels: GeneRelationshipsTabName.GENE_PANELS
+};
+
+export const getViewModeName = (
+  geneViewTab: GeneViewTabName,
+  childTab?: string
```
azangru, line 40: why is it nullable? according to the code below, it's used in the `find` method

azangru:
General consideration: I think the proliferation of `changeViewMode` into multiple components might suggest that there is something wrong with the approach.

From what I can see from the code, the source of truth for the view mode is currently redux. I'd have thought that we would go in the other direction, where the source of truth is the url, controlled by ReactRouter:

image

Except for the edge-case scenarios where genomeId changes, in which case we would first look in the redux store, check whether any particular view has been saved for respective species, update the url with the saved view if yes, and go from there as before:

```
// pseudocode
useEffect(() => {
  if (viewInRedux && viewInRedux !== viewInUrl) {
    replace(urlFor(entityViewer({ genomeId, entityId, view: viewInRedux })))
  }
}, [genomeId])
``` 

The tabs and the buttons then, instead of dispatching redux actions, would be ReactRouter's `Link`s, which would just change the url.

I believe this will simplify the routing logic. It can also be encapsulated then into a custom hook, something like `useGeneViewRouting`, which is used by GeneView, and returns genomeId, entityId, and the actual view mode.

src/ensembl/src/content/app/entity-viewer/gene-view/shared/views-helpers.ts
imransl: It is optional in this case as it is a param. When using this function to get the transcript view there will be no child tab, hence why it is optional.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
imransl: ✔️

imransl:
@azangru The main reason why I didn't use `Link` is because it is difficult to use it with the shared `Tabs` component which changes tabs using the `onTabChange` function. If we are to use `Link`, then we will have to write a different `Tabs` component.

imransl:
@azangru I have now made the view mode to change inside the gene view tab component (e.g. `GeneFunction`) rather than in `GeneView`. I am still using the `connected-react-router` to change the URL. I've also fixed the types so the deployment should go ahead without breaking.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -106,7 +109,8 @@ const mapStateToProps = (state: RootState) => {
 const mapDispatchToProps = {
   setDataFromUrl,
   fetchGenomeData,
-  toggleSidebar
+  toggleSidebar,
+  updateEnsObject
```
azangru, line 113: does `updateEnsObject ` do anything?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -14,18 +14,24 @@
  * limitations under the License.
  */
 
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
+import { push, Push } from 'connected-react-router';
```
azangru, line 19: where is this used?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -14,9 +14,13 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useEffect } from 'react';
+import { connect } from 'react-redux';
+import { useParams } from 'react-router-dom';
+import { push, Push } from 'connected-react-router';
```
azangru, line 20: Do we know whether we prefer a `push` or a `replace` here? I.e. when user clicks on the `back` button in the browser, do we want to cycle through the tabs in the gene function panel, or do we want to transport them to a different screen?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -29,12 +33,18 @@ import styles from './DefaultTranscriptsList.scss';
...
 };
 
 const DefaultTranscriptslist = (props: Props) => {
+  const params: { [key: string]: string } = useParams();
   const { gene } = props;
   const sortedTranscripts = defaultSort(gene.transcripts);
 
+  useEffect(() => {
+    props.push(getGeneViewPath(params, null));
+  }, []);
```
azangru, line 46: Sorry, I'm confused.
1) Why are you doing this in multiple components? Surely it should be enough to just have one useEffect responsible for reconciling redux state with the url
2) What are the situations when the url has to be reconciled with the redux state? I thought it's only when the species changes.
3) I have a suspicion the current code not work when we start switching between species. It's working now; but the reason may be that every time we switch between species we are unmounting and re-mounting components.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: Ah that's a remnant from the merge with dev. I'll remove it.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: I am not sure what we would prefer. I'll check this with Andrea.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: > What does this code do? You are getting params from the router — and then you update the router with the same params? Why?

How else can we generate the URL that includes those params using `urlHelper`?

> Why does the reconciliation code exist in multiple components? Surely it should be sufficient to just have one useEffect responsible for reconciling redux state with the url

This is because we aren't using `Link` for the tabs and instead using the `onTabChange` function, which we pass onto `Tabs`, make the changes.

> What are the situations when the url has to be reconciled with the redux state? I thought it's only when the species changes.

We also need to update redux when the view mode is changed e.g. protein -> variation

> I have a suspicion the current code not work when we start switching between species. It's working now; but the reason may be that every time we switch between species we are unmounting and re-mounting components.

I have not looked into this scenario properly and should definitely check it.

azangru: > This is because we aren't using Link for the tabs and instead using the onTabChange function, which we pass onto Tabs, make the changes.

I am lost. What you are saying does not explain what is going on in the `useEffect` hook. The hook doesn't call `onTabChange`.

imransl: > I am lost. What you are saying does not explain what is going on in the useEffect hook. The hook doesn't call onTabChange.

Sorry for the confusion. I actually need to update the code to reflect this in GeneRelationships and GeneFunction. I'll point it out once I've pushed the code.

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
imransl: This has been removed now ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: @azangru I have changed the code to make the `onTabChange` function to be reused in `useEffect` for GeneRelationships and GeneFunction.

The DefaultTranscriptsList functions in a different way as it it doesn't have any tabs. However, the view param needs to change when navigated to it from a different view mode such as protein or variation.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
msmanoj: I would prefer cycling through the tabs as the user still can get back to a different screen where they came from if needed.

src/ensembl/src/shared/state/ens-object/ensObjectActions.ts
```diff
@@ -86,7 +86,7 @@ export const fetchEnsObject = (
     dispatch(
       fetchEnsObjectAsyncActions.success({
         id: ensObjectId,
-        data: response,
+        data: response
```
msmanoj, line 89: @imransl @azangru it looks like we all have different settings here. Should we have a discussion about keeping it consistent?

azangru: Prettier seems to be doing the right thing now, following the settings in prettier.config.json. If you ever see commits that contain trailing commas, it would be interesting to investigate why.

There was a brief period when we had different versions of prettier, and the latest version defaulted to placing trailing commas everywhere. I think we are dealing with the legacy of that period now.

imransl: I suppose this is something we can't avoid now till all trailing commas are erased by prettier. It is annoying that they enabled trailing commas by default. This probably makes me wonder whether we should not rely too much on their defaults and override all of them in our prettier config.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: > Wouldn't it be better if the url replacement is done where tab change is triggered?

It is a tricky one as only the transcript tab of GeneViewTabs has this behaviour. The other two (GeneRelationships and GeneFunction) have child tabs that change the URL.

imransl:
Superseded by #309

PR #300, opened 2020-05-29
Collapsed exons image

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-555](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-555)

## Description
The collapsed exons image displays all the coding exons of a transcript in one image.

## Views affected
Protein view of entity viewer

### Deployed site link
http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:32043/app/entity-viewer/homo_sapiens_GCA_000001405_27/gene:ENSG00000139618


imransl:
[](url)

imransl:
@azangru @msmanoj @ens-st3 I've made the changes according to Andrea's new design. Please note that the protein list item design is still the old one in this PR. I'm working on it in a different branch at the moment.

azangru:
Diagrams like this, where protein domains continue to the right of the spliced RNA, should be impossible:

image

It means that:
- either the calculation of the width of the spliced RNA should happen in the common parent for both spliced RNA component and protein domains component, or
- the spliced RNA component should report the width of the spliced RNA to its parent, which should then pass this width to the protein domains component

imransl:
@azangru My understanding of the redesign was that the protein domain image will take the exon-intron boundaries into account. However, after reading your explanation I feel this is probably wrong. I'll check with Andrea and see what she says.

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -32,6 +32,7 @@ const TRACK_HEIGHT = 24;
 
 export type ProteinDomainImageProps = {
   transcriptId: string;
+  refCDSLength: number;
```
azangru, line 35: let's remove the `ref` part of the name so that it doesn't imply that you are passing a React ref. `cdsLength` should be sufficient

imransl: `cdsLength` is not accurate as it is not the length of the protein's CDS. We need to an alternative name for this.

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -117,8 +118,11 @@ export const ProteinDomainImageWithData = (
   const proteinDomainsResources = getDomainsByResourceGroups(
     protein.protein_domains_resources
   );
+
+  const refProteinLength = refCDSLength / 3;
+
   const scale = scaleLinear()
-    .domain([1, protein.length])
+    .domain([0, refProteinLength])
```
azangru, line 125: I believe this is wrong? If I understand what's going on correctly:
- your domain is the total number of amino acids in your protein; this should remain the same
- what's changing is the output range into which you have to fit your protein.

Here's a diagram:

image

azangru: In fact, the width that is to be used as a range in the protein domain component can probably be passed from the exons image component like so:

image

azangru: On the other hand, after a second thought, your implementation should work too. It is probably easier than what I am suggesting as well. It may be a bit less intuitive in terms of what is being calculated here.

src/ensembl/.eslintrc.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     '@typescript-eslint/interface-name-prefix': 0,
     '@typescript-eslint/prefer-interface': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
+    "@typescript-eslint/explicit-module-boundary-types": 0,
```
azangru, line 24: Out of curiosity: do you remember the place where eslint complained otherwise?

imransl: This started happening recently. It was complaining about no explicit return type for default exports.

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -15,27 +15,22 @@
  */
```
azangru, line 15: Regarding the naming.

1) If we change this component to represent the length of the actual protein, it should no longer be called CollapsedExonsImage.
2) I still think that it's better to use names that make sense _within a component_. Neither this component nor the ProteinDomainImage component need to be concerned about the concept of the "longest protein" (which becomes a thing only at the level of their parent component, which can compare proteins with each other). What this and the ProteinDomainImage component are interested in is the length of the whole track in amino acids. Could be called something like `trackLength`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -15,27 +15,22 @@
...
-  getFeatureCoordinates
-} from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
-import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
 
 import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
 
 import transcriptsListStyles from 'src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss';
 import styles from './CollapsedExonsImage.scss';
 
 const IMAGE_HEIGHT = 24;
```
azangru, line 28: TrackHeight?

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -76,28 +71,22 @@ export const ExonsImage = (props: ExonsImageProps) => {
 const ExonsImageWithData = (props: ExonsImageWithDataProps) => {
   const { transcript } = props;
 
-  if (!transcript.cds) {
+  if (!transcript.product) {
```
azangru, line 74: We may have discussed this, but aren't you making sure that this component will be created only for transcripts that do have products?

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -107,53 +96,22 @@ const ExonsImageWithData = (props: ExonsImageWithDataProps) => {
...
-                  exon={exon}
-                  className={props.className}
-                  scale={scale}
-                />
-              ))}
+              <rect
+                className={styles.protein}
+                y="8"
+                height={PROTEIN_HEIGHT}
+                width={scale(transcript.product?.length)}
```
azangru, line 107: Same as comment above about availability of the product; but in this case, if we imagine that the product didn't exist, the value of `transcript.product?.length` would be `undefined`. We don't want to pass any undefineds as rect attributes, do we?

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -76,28 +71,22 @@ export const ExonsImage = (props: ExonsImageProps) => {
...
-      const { start, end } = getFeatureCoordinates(exon);
-      return length + (end - start + 1);
-    }, 0);
-
-    return getCommaSeparatedNumber(rnaLength);
-  };
-
+  // If we consider the image's starting point as A and end point as B the scale can be sketched as:
+  // A--  longest protein's length  --B
+  // A--      width in pixels       --B
```
azangru, line 80: This comment isn't really explaining what's going on. It's just repeating what we see two lines below in:

```
.domain([0, props.longestProteinLength])
.range([0, props.width])
```

Maybe something like this:

> Create a scale where the domain is the total length of the track, in amino acids.
> The track is as wide as the longest protein generated from the gene;
> therefore, it is guaranteed that the length of the protein drawn by this component
> will fall within this domain

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -76,28 +71,22 @@ export const ExonsImage = (props: ExonsImageProps) => {
...
   const scale = scaleLinear()
-    .domain([0, props.refCDSLength])
+    .domain([0, props.longestProteinLength])
     .range([0, props.width])
     .clamp(true);
 
   const midStyles = classNames(transcriptsListStyles.middle, styles.middle);
 
+  // The scale/longest protein is displayed using the first <rect> from point A to point B
+  // The protein is displayed from point A (check comment above) using the second <rect>
```
azangru, line 89: > The scale/longest protein is displayed

Didn't we agree that we call this light-grey rectangle over which we draw darker (or blue) rectangles a track (or a lane, whatever)? Calling it "the longes protein" in this comment is confusing.

Also, it should be fairly obvious what each rectangle represents from the names of their classes. Is this comment adding anything useful?

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
```diff
@@ -107,53 +96,22 @@ const ExonsImageWithData = (props: ExonsImageWithDataProps) => {
           height={IMAGE_HEIGHT}
         >
           <g>
-            <g className={styles.splicedRNABlock}>
+            <g className={styles.longestProtein}>
```
azangru, line 99: `longestProtein` -> `track`

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -32,7 +32,7 @@ const TRACK_HEIGHT = 24;
 
 export type ProteinDomainImageProps = {
   transcriptId: string;
-  refCDSLength: number;
+  longestProteinLength: number;
```
azangru, line 35: Same as comment regarding names in the file that's currently called `CollapsedExonsImage.tsx`

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -74,3 +75,88 @@ export const getNumberOfCodingExons = (transcript: Transcript) => {
...
+    if (transcript.cds) {
+      const currentProteinLength = transcript.cds?.protein_length as number;
+
+      if (currentProteinLength > longestProteinLength) {
+        longestProteinLength = currentProteinLength;
+      }
+    }
+  });
+
+  return longestProteinLength;
```
azangru, line 98:
```
const proteinLengths = gene.transcripts.map(transcript => transcript.cds?.protein_length || 0);
return Math.max(proteinLengths);
```

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -74,3 +75,88 @@ export const getNumberOfCodingExons = (transcript: Transcript) => {
...
+};
+
+export const getCodingExonsForImage = (transcript: Transcript) => {
```
azangru, line 101: Is this function still needed?

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: > Also, it should be fairly obvious what each rectangle represents from the names of their classes. Is this comment adding anything useful?

It probably doesn't anymore, since the comment will be changed for the scale code. I'll remove this.

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: Not anymore

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: This is a lot neater 😀

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: Yes sorry, I missed this one out. Will make the change.

imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: I have changed the component's name to `ProteinImage`. Also, have changed `longestProteinLength` to `trackLength`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: ✔️

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
imransl: I've removed this

src/ensembl/src/content/app/entity-viewer/gene-view/components/collapsed-exons-image/CollapsedExonsImage.tsx
imransl: I've changed it here and also in other places where I came across the same issue

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -102,26 +103,31 @@ const ProteinDomainImage = (props: ProteinDomainImageProps) => {
...
-    return null;
-  }
+  const proteinDomainsResources = getDomainsByResourceGroups(proteinDomains);
 
-  const proteinDomainsResources = getDomainsByResourceGroups(
-    protein.protein_domains_resources
-  );
+  // If we consider the image's starting point as A and end point as B the scale can be sketched as:
+  // A--  longest protein's length  --B
+  // A--      width in pixels       --B
```
azangru, line 123: same comment as in the protein component

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -102,26 +103,31 @@ const ProteinDomainImage = (props: ProteinDomainImageProps) => {
...
   const scale = scaleLinear()
-    .domain([1, protein.length])
+    .domain([0, trackLength])
     .range([0, props.width])
     .clamp(true);
 
+  // The <Track> represents the longest protein length
+  // The <DomainBlock> is where the domain is position in relation to the protein
```
azangru, line 130: remove?

imransl: ✔️

src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
imransl: ✔️

PR #301, opened 2020-06-02
Fix sidebar overview content and topbar tabs pointer

msmanoj:
NOTE: The trailing commas were automatically removed by prettier since we are using `trailingComma: "none"`. 

## Type

- [x] Bug fix
- [ ] New feature

## Description
This PR is to fix the following two issues:
- Missing sidebar overview tab content
- Missing pointer on topbar tabs

## Views affected
Entity viewer -> Sidebar

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
+++ b/src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
@@ -21,7 +21,6 @@ import { push, replace } from 'connected-react-router';
 import { ThunkAction } from 'redux-thunk';
 
 import * as urlHelper from 'src/shared/helpers/urlHelper';
-import { buildEnsObjectId, parseFocusIdFromUrl } from 'src/shared/state/ens-object/ensObjectHelpers';
 
 import { getCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
 import {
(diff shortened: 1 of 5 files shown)


src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -67,12 +66,8 @@ export const setDataFromUrl: ActionCreator<ThunkAction<
     }
   }
 
-  const entityId = params.entityId
-    ? buildEnsObjectId({
-        genomeId: genomeId as string,
-        ...parseFocusIdFromUrl(params.entityId)
-      })
-    : null;
+  const { entityId } = params;
```
azangru, line 69: That can't be right.
`entityId` here is `gene:ENSG...`. The ids that we use for keys in redux store are `<genome_id>:<type>:<stable_id>`.

If you open the console, you will probably see the error:

image

I believe the above lines, with `buildEnsObjectId`, should stay as they are.

msmanoj: Okay, I thought the redux store should use the format `<type>:<stable_id>` since the sampleData was updated here:
https://github.com/Ensembl/ensembl-client/pull/282/files#diff-39dfc9125867911e956e3f1d875adf3eR873

azangru: > I thought the redux store should use the format <type>:<stable_id> 

That won't be sufficient to distinguish between same types of entities with same ids but from different genomes (e.g. BRCA2 for grch37 and grch38); that's why I added genome id as a part of redux keys.

azangru: > since the sampleData was updated here

That must have been my mistake. That PR went through several phases — at first I thought that we are changing everything to `<type>:<stable_id>`, but then realised that that won't work.

Sorry for forgetting to change the sampleData back.

msmanoj: Thanks. I get it now. I have reverted the changes and updated the keys used in sampleData.

PR #302, opened 2020-06-03
Gene view  function / relationships tabs availability

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-548

## Description
- the isDisabled flag is hardcoded to false for the tabs that does not have any data
- the first available tab is selected by default
- A standard message is displayed when no tabs are available to select

## Views affected
Entity viewer -> Gene view -> Gene function & Gene Relationships

### Other effects
I am not quire sure about what I can test here.

## Possible complications
At the moment there is no easy way to identify which tabs needs to be disabled. I have created a ticket for us to look into a way of getting the availability data from Thoas:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-636

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -162,7 +162,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
           )}
 
         {props.selectedGeneTabName === GeneViewTabName.GENE_FUNCTION && (
-          <GeneFunction geneId={props.gene.id} />
+          <GeneFunction geneId={props.gene.id} gene={props.gene} />
         )}
 
         {props.selectedGeneTabName === GeneViewTabName.GENE_RELATIONSHIPS && (
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -162,7 +162,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
           )}
 
         {props.selectedGeneTabName === GeneViewTabName.GENE_FUNCTION && (
-          <GeneFunction geneId={props.gene.id} gene={props.gene} />
+          <GeneFunction gene={props.gene} />
         )}
 
         {props.selectedGeneTabName === GeneViewTabName.GENE_RELATIONSHIPS && (
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
@@ -48,7 +48,6 @@ const tabClassNames = {
 };
 
 type Props = {
-  geneId: string;
   gene: Gene;
   isNarrow: boolean;
   selectedTabName: GeneFunctionTabName | null;
@@ -99,7 +98,7 @@ const GeneFunction = (props: Props) => {
   const getCurrentTabContent = () => {
     switch (selectedTabName) {
       case GeneFunctionTabName.PROTEINS:
-        return <ProteinsList geneId={props.geneId} />;
+        return <ProteinsList geneId={props.gene.id} />;
       default:
         return <>Data for these views will be available soon...</>;
     }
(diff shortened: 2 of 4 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
```diff
@@ -46,12 +49,38 @@ const tabClassNames = {
...
+  // TODO: We need a better logic to disable tabs once we have the data available for other tabs
+  if (!transcripts || !transcripts.length) {
+    const proteinTabIndex = tabsData.findIndex(
+      (tab) => tab.title === GeneFunctionTabName.PROTEINS
+    );
+
+    tabsData[proteinTabIndex].isDisabled = true;
+  }
+
+  // If the selectedTab is disabled or if there is no selectedtab, pick the first available tab
```
imransl, line 74: I wonder if we need to let the user in some way know that the tab is disabled. This is probably a UX question we'll need to ask from Anderea.

msmanoj: The tab links are greyed out and will not be clickable when they are disabled.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -162,7 +162,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
           )}
 
         {props.selectedGeneTabName === GeneViewTabName.GENE_FUNCTION && (
-          <GeneFunction geneId={props.gene.id} />
+          <GeneFunction geneId={props.gene.id} gene={props.gene} />
```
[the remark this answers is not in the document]
msmanoj: Removed `geneId`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
imransl: True, but I am wondering about the confusion that the user might be in when something else is shown instead of what he/she wanted to see.

msmanoj: That's a fair point but I don't think the user will end up having a URL where the particular view is disabled.

azangru: I think I am with Imran on this one. if there is no selected tab, I would expect us to either show a spinner (if the client is making network requests to find out which content is available) or to show a message that data is not available.

Also, I don't remember whether the Gene Function tab should be enabled if no content is available.

msmanoj: >if there is no selected tab, I would expect us to either show a spinner (if the client is making network requests to find out which content is available) or to show a message that no data is available.

If no tab is selected, I think it will default to Transcripts tab. I agree that we can make a request and show a spinner when the endpoint is available for that but I am not sure what can be done now.

Also, how will the user end up having a URL for an entity that will say `?view=protein`, if there was no data for it?

> Also, I don't remember whether the Gene Function tab should be enabled if no content is available.

I think it should. Please check the following image from Andrea:
https://www.ebi.ac.uk/panda/jira/secure/attachment/34840/ev%20non-coding%20-%20gene%20function.png

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.tsx
```diff
@@ -45,13 +48,38 @@ const tabClassNames = {
...
   setActiveGeneFunctionTab: (tab: string) => void;
 };
 
 const GeneFunction = (props: Props) => {
+  const {
+    gene: { transcripts }
+  } = props;
+  let { selectedTabName } = props;
+
+  // Disable the Proteins tab if there are no transcripts data
```
azangru, line 63: I am confused. Are there any genes without transcripts? Isn't transcripts in this case an empty array, which is never falsy?

msmanoj: Yes, that's right. I was meant to do this check but I missed it. Thanks for spotting it.

msmanoj: That's now updated 👍

PR #303, opened 2020-06-09
Adding filter and sort component to entity viewer

ens-ma7:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-497

## Description
Adding the filter and sort by component to Entity viewer. The checkboxes for the filter part are just dummies, these will need to be retrieved from the backend which is in another ticket.

## Views affected
Entity viewer

### Other effects


ens-ma7:
Not sure why the checks are failing, whats causing it to fail?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [isChecked, setChecked] = useState(false);
+
+  const filterBoxClassnames = classNames(
+    styles.filterBox,
+    { [styles.filterBoxFullWidth]: !props.isSidebarOpen }
+  );
+
+  const [selectedRadio, setselectedRadio] = useState('default');
+
+  const radioChange = (value: string) => {
```
imransl, line 38: @ens-ma7 The issues is here. The type of `value` should be `OptionValue` instead of `string`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -50,3 +50,22 @@
...
+
+  img {
+    height: 12px;
+    width: 12px;
+  }
+}
+
+.hidden {
+  display: none;
+}
```
azangru, line 71: image

Notice this red sign? I believe it says that there is no empty line in the end of the file. Not that it's particularly important, but I believe our editors are set up in a way that they automatically add an empty line in the end of the file when you save it.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -34,15 +37,26 @@ type Props = {
...
+            <button onClick={toggleFilter} className={styles.expandBtn}>
+              <img
+                src={chevronDownIcon}
+                alt={'collapse'}
+              />
+            </button>            
+          </div>          
           <div className={styles.right}>Transcript ID</div>
-        </div>
+        </div>
```
azangru, line 59: We don't want trailing empty space. Is it possible that `prettier` is not working for you? I thought it would fix these things automatically.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -34,15 +37,26 @@ type Props = {
...
         <div className={styles.row}>
-          <div className={styles.left}>Filter & sort</div>
-          <div className={styles.middle}></div>
+          <div className={isFilterOpen ?  styles.hidden : styles.filterLabel}>
+            Filter & sort
+            <button onClick={toggleFilter} className={styles.expandBtn}>
+              <img
+                src={chevronDownIcon}
+                alt={'collapse'}
+              />
```
azangru, line 55: We usually use svg elements, because then we can style their `fill` property via CSS. If you run:

```
import {ReactComponent as ChevronDown} from 'static/img/shared/chevron-down.svg';
```
you'll get a react component called `ChevronDown`, which will be an `svg` element.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.scss
```diff
@@ -0,0 +1,123 @@
...
+      position: absolute;
+      top: 2px;
+      left: 2px;
+      width: 6px;
+      height: 6px;
+      border-radius: 50%;
+      background: $green;
+    }
+  }
+}
```
azangru, line 98: Umm... Are all these styles for the radio button necessary? Surely some (`position: relative; cursor: pointer; content: ''; position: absolute;`) are already defined in your RadioButton component?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { connect } from 'react-redux';
+import { RootState } from 'src/store';
+import { isEntityViewerSidebarOpen } from '../../../state/sidebar/entityViewerSidebarSelectors';
```
azangru, line 5: when a relative import has too many "back up one directory" instructions (and two directories up already feels too many), we start from the src directory:

```
import blah from `src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSelectors`
```

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
ens-ma7: Thanks @imransl i copied this code from the storybook example for radiogroup:
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/stories/shared-components/radio-group/RadioGroup.stories.tsx#L34

How is it working there but not in my code.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
ens-ma7: It wasn't on in my editor, i have enabled it.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
ens-ma7: It wasn't installed. I just realised i had to uninstall and install vs code again to make it work with wsl properly. It did fix the issue with wsl but i guess i forgot to enable some settings and extensions :)

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
imransl: It probably should've highlighted the error there too. However, in this case the deployment requires to check the types in the `src` code. So it caught this error, which resulted in the deployment stopping half way.

ens-ma7:
@azangru i am aware of this issue (the screen size) and i had a chat with Andrea about this. She said we need to come back to this once we have the scroll bar that imran is working on. In fact you reminded that i need to create a ticket about this (ENSWBSITES-644).  we will have a better understanding what to do.

About the checkboxes being all ticked this is because of the isselected is the same for all, it was just a quick way to show them. i will put in a proper fix once we are retrieving them from the backend.

azangru:
>  once we have the scroll bar that imran is working on

Which scrollbar is that?

I remember there was a ticket for vertical scrolling of the GeneView area, which @msmanoj has done. I don't remember any conversation about the horizontal space.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -34,13 +37,19 @@ type Props = {
...
+  const toggleFilter = () => { isFilterOpen ? setFilterOpen(false) : setFilterOpen(true); }
 
   return (
     <div>
       <div className={styles.header}>
+        {isFilterOpen && <GeneFilter toggleFilter={toggleFilter} />}
         <div className={styles.row}>
-          <div className={styles.left}>Filter & sort</div>
-          <div className={styles.middle}></div>
+          <div className={isFilterOpen ?  styles.hidden : styles.filterLabel} onClick={toggleFilter}>
```
azangru, line 49: Not that it particularly matters, but why use styles.hidden when you can just conditionally render or not render this element:

```
{ !isFilterOpen &&
  <div className={styles.filterLabel} onClick={toggleFilter}>
    Stuff
  </div>
}
```

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -34,13 +37,19 @@ type Props = {
 const DefaultTranscriptslist = (props: Props) => {
   const { gene } = props;
   const sortedTranscripts = defaultSort(gene.transcripts);
+  const [isFilterOpen, setFilterOpen] = useState(false);
+
+  const toggleFilter = () => { isFilterOpen ? setFilterOpen(false) : setFilterOpen(true); }
```
azangru, line 42: `isFilterOpen ? setFilterOpen(false) : setFilterOpen(true);` -> `setFilterOpen(!isFilterOpen)`

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.scss
```diff
@@ -0,0 +1,113 @@
...
+
+.closeIcon {
+  cursor: pointer;
+  height: 16px;
+}
+
+.filterContent {
+  display: grid;
+  grid-template-columns: 245px 245px 245px;
+  height: auto;
```
azangru, line 50: this is already height's default value

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+type OptionValue = string | number | boolean;
+
+const radioData: RadioOptions = [
+  { value: 'default', label: 'Default' },
+  { value: 'length_longest', label: 'Spliced length: longest - shortest' },
+  { value: 'length_shortest', label: 'Spliced length: shortest - longest' }
+];
+
+const GeneFilter = (props: Props) => {
```
azangru, line 29: The sorting and filtering options are applied to the transcripts of a gene, not to the gene itself. Doesn't this make the name GeneFilter misleading? Wouldn't TranscriptsSortAndFilter be more appropriate?

imransl:
> Which scrollbar is that?
> 
> I remember there was a ticket for vertical scrolling of the GeneView area, which @msmanoj has done. I don't remember any conversation about the horizontal space.

That's probably the one @ens-ma7 is referring to. Maybe Andrea has some new ideas for the horizontal scrolling?

azangru:
I played a little with the styles:

13 inch, sidebar open:

image

13 inch, sidebar closed:

image

Isn't this what we are after? No scrollbar necessary.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-filter/GeneFilter.tsx
ens-ma7: The way i was looking at it is the component is being used in the Gene View hence Gene filter.  I guess it does make sense the other way round where the filters are applied to the transcript, i wasn't looking it at this way yet but makes sense more. But i think i will keep it short to TranscriptsFilter

ens-ma7:
> > once we have the scroll bar that imran is working on
> 
> Which scrollbar is that?
> 
> I remember there was a ticket for vertical scrolling of the GeneView area, which @msmanoj has done. I don't remember any conversation about the horizontal space.

This is what i was told by Andrea but maybe that will be something coming up soon not sure if you got any clarification from the sprint planning or this was discussed.

ens-ma7:
> I played a little with the styles:
> 
> 13 inch, sidebar open:
> 
> image
> 
> 13 inch, sidebar closed:
> 
> image
> 
> Isn't this what we are after? No scrollbar necessary.

@azangru i dont think that the spec of the design, if you look at the jira tickets the column width is set to 245px for each filters and in your solution you are collapsing them. what should happen is any overflow should go behind/hidden and there should be a vertical scrollbar to see the rest. what i will do is make any overflow to go behind the Right hand panel.

azangru:
We set the minimum width of the `body` element to 980px.
Currently, in dev, if you reduce the window width in Entity Viewer, the horizontal scrollbar won't appear until you go down to 980 px.
With the proposed dimensions of the filter, its minimum width exceeds 980px.
I wonder if we can convince Andrea to make the filter box properly scalable down to 980px in width. That should be quite a lot of space.

ens-ma7:
Yep saw the scrollbar you are talking about but i dont think thats the one. This is the default browser scrollbar, i think we might need another scrollbar just for the entity viewer content. I know what you are thinking now, double scrollbar or customised scrollbar :) we dont know yet whats the best solution for this, looks like we will definitely need some kind of scrollbar. 

Andrea also mentioned about making the box scalable but suggest it will be better to wait and see how other species filters appear as well and also we will have a better idea with other components in place of the need of a scrollbar.

I have put a fix for now to make the filter box go behind the panel for smaller screen size which is the behavior agreed with Andrea (for now). I know you cant see the rest which is not ideal but lets leave it like this so that it reminds us that it needs fixing.

Just push the changes, please have a look and let me know.

ens-ma7:
It seems one of the test has failed; something to do with chromosome navigator. can anyone help why its throwing this error:

`FAIL src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx
   ● Chromosome Navigator › viewport position › in the middle of chromosome › draws two viewport areas if start position is greater than end position
     expect(received).toBe(expected) // Object.is equality
     Expected: 94
     Received: 93
       129 |         }
       130 |         if (width !== undefined) {
     > 131 |           expect(areas.at(index).prop('width')).toBe(width);
           |                                                 ^
       132 |         }
       133 |       });
       134 |     };
       at forEach (src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx:131:49)
           at Array.forEach (<anonymous>)
       at assertViewportAreas (src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx:125:18)
       at Object.<anonymous> (src/content/app/browser/chromosome-navigator/ChromosomeNavigator.test.tsx:201:9)`

azangru:
> can anyone help why its throwing this error:

That must be a defect of my test — it failed with a particular set of randomly generated values (but we don't know which). I'll try to reproduce the error somehow; but meanwhile, if you re-run the test suite, the error should go away).

azangru:
It fails about once in a 1000 times 😄 😞

ens-ma7:
> It fails about once in a 1000 times 😄 😞

Lucky me, i think i need to buy the lottery today :)

ens-ma7:
@azangru i have implemented all the changes together with the reduce filter box sizes, please have a look and let me know if there is anything else.

azangru:
What's surprising to me is that if we remove additional styles for radio buttons and checkboxes from the parent component (which in this case is TranscriptsFilter), both radio buttons and checkboxes fall apart:

image

Notice that the wrapper class in Checkbox at least has display: flex on it; but the wrapper class in the RadioGroup doesn't have anything.

I would have expected styles that hold the label to the right of the radio/checkbox to be defined in the RadioGroup or Checkbox components. It must be horribly annoying to repeat them over and over in components that use these elements.

@imransl @msmanoj what's your opinion?

azangru:
Also, your component is now superseding the EntityViewerTranscriptFilter file (tsx, scss, story files TranscriptFilter.stories.tsx, TranscriptFilter.stories.scss). Please delete them.

imransl:
@azangru Should we then add the common styling into RadioGroup and CheckBox stylesheets?

azangru:
> Should we then add the common styling into RadioGroup and CheckBox stylesheets?

I would think so. I don't know how much the Checkbox is going to break in other places because of this.

azangru:
There are of course two possible arrangements of radios and checkboxes:

1) as block elements:

image

2) as inline elements:

image

I am not sure where the styles should be that regulate these arrangements — whether the radio/checkbox element should know itself that it's inline/block; or whether its parent should know it.

msmanoj:
>I would have expected styles that hold the label to the right of the radio/checkbox to be defined in the RadioGroup or Checkbox components. It must be horribly annoying to repeat them over and over in components that use these elements.

@azangru checkbox already has the styles for the label defined within the component itself:
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/checkbox/Checkbox.scss#L40

azangru:
> @azangru checkbox already has the styles for the label defined within the component itself:

@msmanoj yes, I've seen that; but the screenshot above shows what the checkboxes with the labels will look like if placed, with no additional styling., in the filter box.

ens-ma7:
> Also, your component is now superseding the EntityViewerTranscriptFilter file (tsx, scss, test, story files TranscriptFilter.stories.tsx, TranscriptFilter.stories.scss). Please delete them (or use them as an inspiration for the subsequent development of your component).

@azangru i didn't quite understand this one? which component of mine? TranscriptsFilter? I didn't even realise there was another component with similar naming. i saw on slack that @msmanoj said we can delete EntityViewerTranscriptFilter, is that correct @msmanoj can we delete the whole directory transcript-filter? if we can delete it, then there isn't any issue with my component?

ens-ma7:
@azangru the question about the checkboxes alignment (either on the same line or one on each line), i guess depends when and how it was originally used for. I have a feeling its the same way in custom download that is on the same line, is that why we have this float: left @msmanoj ? if its not used in custom downloads i can remove the float and check where else its being used to make sure its not broken.

I will also apply the same styling to the radio groups, is that ok @azangru @imransl ?

ens-ma7:
scrap my comment about removing the float left from the checkboxes, i can see they are used in the instant download. I think the checkboxes styling are ok as it is, i can match the radio group styling.

azangru:
> if its not used in custom downloads i can remove the float and check where else its being used to make sure its not broken.

It's definitely used in Custom Download; my worry is that, as we have seen, the current styles on Checkbox+label are not sufficient for it to display properly in other contexts, such as the filter box. This suggests that the styles for Checkbox may need to be revised. Not sure whether in this PR or in a dedicated one; but checkboxes with labels and radio buttons with labels should behave similarly; and ideally, there shouldn't have to be much extra styling for them from parent components.

ens-ma7:
what about we have both options, a flag to make the checkboxes or radio buttons align in one row or by column, is that not good @azangru ? if it is more complicated then i guess it needs a separate ticket/PR.

azangru:
There are two pattern libraries with different approaches to this problem: [MaterialUI](https://material-ui.com/components/checkboxes/) and [AntDesign](https://ant.design/components/radio/). It might be worth exploring how they are doing it.

ens-ma7:
Thanks @azangru i can see how its being done with MaterialUI where there is a theming option to display the checkboxes vertically. By default it is horizontal which is what the current checkboxes behavior. what i will do is make the radio group behaves (style wise) the same as the checkboxes and then if time permit add the option to make them display differently as a separate PR/ticket.

ens-ma7:
@imransl @msmanoj @azangru i have made the changes we discuss in the last meeting (about the grid layout and also making radio and checkbox same styles), please have a look and let me know of any issue.

azangru:
This happens to a single checkbox element when you select a checkbox (the number of classes keeps growing with every ticking/unticking of a checkbox). I know you were planning to change how checkboxes work; I'm just wondering whether this is something that we should fix now:

image

azangru:
Will you be up for adding another story in the Checkbox section; something like "grid with long labels"?

image

Something to show how the text of the label will behave if it has to wrap over several lines? Ideally with two columns of checkboxes, so we can also see how the lines of the text will be positioned relative to the checkboxes in the neighbouring column? Something like this (doesn't have to be on the dark background):

image

ens-ma7:
@azangru PR is ready for review. Thanks

src/ensembl/src/shared/components/checkbox/Checkbox.tsx
```diff
@@ -55,7 +55,7 @@ const Checkbox = (props: CheckboxProps) => {
     props.onChange(!props.checked);
   };
 
-  const styles = { ...defaultStyles };
+  const styles = props.classNames ? { ...defaultStyles } : defaultStyles; // this fix the test cases and also if there is no classNames passed as props
```
azangru, line 58: The reason the tests failed, as you might have learned from @imransl,  is because Jest has been set up to use `identity-object-proxy` when it's importing `css` or `scss` files (https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/jest.config.js#L10). I don't know if that was a good idea, but as a result the `defaultStyles` that gets imported in the test is an ES6 Proxy rather than a good old plain object. And ES6 Proxies cannot be iterated using a `for ... in` loop.

The code above is just a duct tape around the problem; the actual problem still remains, but we just happen not to have a test revealing it. I suspect that if you had a test similar to the [one we are running for Checkbox label](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/checkbox/Checkbox.test.tsx#L46-L53), but that would test the classes of the checkbox element instead, this test would fail.

ens-ma7: @azangru your commit is good and make the code more understandable, thanks. Just one thing i wanted to confirm is the class checkboxHolder in the component is something that cannot be overwritten from props, is that the correct behavior we want/expect.

azangru: Thanks, @ens-ma7, that's my mistake. I'll fix that and add a couple of tests to make sure that classes are applied correctly.

PR #309, opened 2020-06-21
Support view query parameter in EntityViewer url

azangru:
Supersedes https://github.com/Ensembl/ensembl-client/pull/294

## Type
- New feature

## Related JIRA Issue(s)
[ENSWBSITES-577](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-577)

## Description
This PR adds the `view` query parameter to the url of the entity viewer, which will determine which tab/panel is shown in GeneView

### Notable changes:
1) The view from the url is stored in redux in fields `view` and `selectedTabViews` of `EntityViewerGeneViewUIState`
2) These values are used by `GeneFunction` component and `GeneRelationships` component to determine which content is shown.
3) The `view` parameter from the url is synchronised with the the mentioned redux fields in a single place — namely, in the GeneView component.

## Views affected
Entity viewer

## Deployed to
http://proposal-protein-view-url.review.ensembl.org/


azangru:
Notice that all tabs in the panels have now been enabled. We may want to disable them back before this PR is merged.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -153,30 +147,58 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
...
-  selectedGeneTabName: getEntityViewerActiveGeneTab(state)
-});
+const useGeneViewRouting = () => {
+  const dispatch = useDispatch();
+  const params: { [key: string]: string } = useParams();
+  const { genomeId, entityId } = params;
+  const { objectId: geneId } = parseFocusIdFromUrl(entityId);
+  const { search } = useLocation();
+  // TODO: discuss – is using URLSearchParams better than using the querystring package?
+  const view = new URLSearchParams(search).get('view');
```
msmanoj, line 177: This looks simple and easy to me.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -51,21 +59,75 @@ export enum GeneFunctionTabName {
...
+  secondaryTab: null
+};
+
+// using Map to guarantee the order of the inserted elements
+export const GeneViewTabMap: Map<View, GeneViewTabData> = new Map();
+GeneViewTabMap.set(View.PROTEIN, {
+  view: View.PROTEIN,
+  primaryTab: GeneViewTabName.GENE_FUNCTION,
+  secondaryTab: GeneFunctionTabName.PROTEINS
+});
```
msmanoj, line 84: Any chance we can include the tab display name here?
By that we can simplify this:
https://github.com/Ensembl/ensembl-client/pull/309/files#diff-a8e10f9cafccfd34f324bb947c8c9edfR63-R69

azangru: My idea for symplifying [this](https://github.com/Ensembl/ensembl-client/blob/b1a6903bed2c43e7d36c096cc207aa6fd4f4ecd1/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships.tsx#L63-L68) was to separate the `title` property of the Tab ([link](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/tabs/Tabs.tsx#L24)) into a label and a value, like we talked about. In which case the `onClick` callback can be called with the actual value of the `view` rather than with the display name of the tab. But I thought we agreed to do this in a separate PR.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
azangru: We should review whether we want to continue using the `query-string` module then. The only significant differences seem to be that the `query-string` module is more flexible with the handling of array query parameters, and it also handles empty values nicer. On the one hand, URLSearchParams is native web api; but on the other, `query-string` is a fairly small library, 5kB minified.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
msmanoj: Yes, I was just suggesting this as a temporary solution but let's leave it as it is for now.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: We could probably continue to use `URLSearchParams` and then go for `query-string` when we come across some complex queries involving arrays and empty values.

imransl:
@azangru The first time I switch species in Entity Viewer the page is broken. However, it works when I try it a second time. I tried this twice after clearing storage, and it happened both times.

azangru:
@msmanoj I was able to replicate the issue together with @imransl . The problem already exists in the dev branch and seems to be unrelated to this PR; and the steps to replicate it are as follows:

- select Wheat
- navigate to the EntityViewer with Wheat as your active selected species
- you'll see an empty GeneView screen

Something wrong with EntityViewer initialization code. Definitely calls for a separate PR.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: I vaguely remember us having similar discussions in one of the PRs where we decided to go for `query-string`. I think we can leave it as it is for now and try to use `URLSearchParams` wherever we can and use `query-string` to handle complex queries.

msmanoj:
> Something wrong with EntityViewer initialization code. Definitely calls for a separate PR.

Thanks @azangru. Are you okay to create a ticket for it?

azangru:
> Are you okay to create a ticket for it?

@msmanoj https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-652

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
azangru: I guess our thinking should be:
- if we want to use query-string anyway, we probably should use it consistently throughout
- if URLSearchParams meets our needs and we don't really need query-string, we should get rid of it and of those extra 5kB of minified javascript that it contributes to our js bundle

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -153,30 +147,58 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
...
+  const viewInRedux = useSelector(getGeneViewName);
+  const previousGenomeId = usePrevious(genomeId); // genomeId during previous render
+  const selectedTabs = useSelector(getSelectedGeneViewTabs);
```
imransl, line 180: I thought we didn't want to use these redux hooks 🤔

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships.tsx
```diff
@@ -48,34 +52,48 @@ const tabClassNames = {
 type Props = {
   isSidebarOpen: boolean;
   selectedTabName: GeneRelationshipsTabName | null;
-  setActiveGeneRelationshipsTab: (tab: string) => void;
+  push: Push;
 };
 
 const GeneRelationships = (props: Props) => {
+  const { genomeId, entityId } = useParams() as { [key: string]: string };
```
imransl, line 59: Seems we are using this type (`{ [key: string]: string }`) in quite a few places. I wonder if we should have a shared type for this.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: I noticed that it is currently used in the following places:
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/species-selector/state/speciesSelectorEpics.ts#L66
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/helpers/urlHelper.ts

@azangru since you have worked on these files, if you think it can be easily replaced with `URLSearchParams`, I guess we can get rid of it in a separate PR.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
```diff
@@ -51,21 +59,75 @@ export enum GeneFunctionTabName {
...
+GeneViewTabMap.set(View.GENE_FAMILIES, {
+  view: View.GENE_FAMILIES,
+  primaryTab: GeneViewTabName.GENE_RELATIONSHIPS,
+  secondaryTab: GeneRelationshipsTabName.GENE_FAMILIES
+});
+GeneViewTabMap.set(View.GENE_PANELS, {
+  view: View.GENE_PANELS,
+  primaryTab: GeneViewTabName.GENE_RELATIONSHIPS,
+  secondaryTab: GeneRelationshipsTabName.GENE_PANELS
+});
```
imransl, line 124: `view` and `secondaryTab` always have the same values in this list of maps. Can it however be different in some instances? If not then can we merge them into one property?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
azangru: We are flexible :-)

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState.ts
azangru: What do you mean same values, Imran?
They come from different enums, so the actual string value is different.

imransl: Ah I failed to notice the different values 🤦 This makes sense now.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships.tsx
azangru: Not sure how best to approach it. I am definitely being lazy here; it might be best to describe the params object for different applications, so that typescript know  that `genomeId`, and `entityId` are actually the expected properties of the params object for EntityViewer.

PR #310, opened 2020-06-24
Fix browser navigation bugs

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-649

## Description
This PR fixes the following bugs:

1. Retained focus ID
Add two species -> Genome browser -> Select gene -> Switch to the other species (Notice that the focus ID is still retained from the previously active species)

2. Browser and Region Editor not updating when switching between species.

3. BrowserRegionEditor changes keeps getting reset when we click outside the form

## Views affected
Browser

Deployment:
http://fix-browser-species-switch.review.ensembl.org/


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -118,16 +122,13 @@ export const Browser = (props: BrowserProps) => {
...
-      browserQueryParams: { location }
-    } = props;
     const { genomeId } = params;
     const chrLocation = location ? getChrLocationFromStr(location) : null;
 
     if (props.browserActivated && genomeId && chrLocation) {
       props.changeBrowserLocation({ genomeId, chrLocation, ensObjectId: null });
     }
-  }, [props.browserActivated]);
+  }, [props.browserActivated, location]);
```
azangru, line 131: Wondering whether the logic of this useEffect callback belongs in useBrowserRouting.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -133,12 +131,15 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
-    });
+  const changeLocation = (newChrLocation: ChrLocation) => {
+    props.replace(
+      urlFor.browser({
+        genomeId: props.activeGenomeId as string,
+        focus: null,
+        location: newChrLocation ? getChrLocationStr(newChrLocation) : null
+      })
+    );
+  };
```
azangru, line 142: What is wrong with the previous implementation? The previous one should have sent a message to genome browser telling it to go to a different location (see the implementation in browserActions), and when the browser sends a message back, the url should have been updated.

msmanoj: Okay, I wasn't aware that the browser was going to send a message to update the URL. With the previous implementation, the RegionEditor kept getting reset which I need to find out why.

azangru: The loop goes like this:

- we send a bpane event to genome browser: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browserActions.ts#L342-L347
- browser changes the content of the canvas and reports back a new location
- we listen to the message from the browser and change the location here: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browserActions.ts#L342-L347
- and change the url here: https://github.com/Ensembl/ensembl-client/blob/500e9a4f1969020d9e8290617850202421747c05/src/ensembl/src/content/app/browser/browserActions.ts#L310

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: I have moved it to `useBrowserRouting`. Please check if it looks right.

azangru:
**Bug:**
- select a species (e.g. human grch38)
- navigate to genome browser and select example gene
- you will see the gene in the genome browser
- now refresh the page
- _(bug)_ genome browser remains empty

msmanoj:
> (bug) genome browser remains empty

The bug should now be fixed.

>It's possible that we do not need this check in the code at all. I've commented it out; and nothing horrible seems to be happening.

I checked it as well and it seems to be working fine. I have removed it now. 


>I also commented out the body of the setBrowserLocation, and nothing horrible seems to be happening either. We may not need this function at all;

Same for this.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.tsx
```diff
@@ -79,6 +79,22 @@ export const BrowserRegionEditor = (props: BrowserRegionEditorProps) => {
...
+      stickInput !== stick ||
+      getNumberWithoutCommas(locationStartInput) !== locationStart ||
+      getNumberWithoutCommas(locationEndInput) !== locationEnd
+    ) {
+      showSubmitButton(true);
+      return;
+    }
+
+    showSubmitButton(false);
+  }, [stickInput, locationStartInput, locationEndInput]);
```
azangru, line 96: You may want to add the `stick` to the list of dependencies.

And change the code to:

```
const shouldShowButton = (
      stickInput !== stick ||
      getNumberWithoutCommas(locationStartInput) !== locationStart ||
      getNumberWithoutCommas(locationEndInput) !== locationEnd
    );
showSubmitButton(shouldShowButton);
```

msmanoj: > You may want to add the stick to the list of dependencies.

Why do we need this?

azangru: We don't strictly need this, but it's used within the useEffect callback, so if we were using linter rules for react hooks, the linter would complain.  🤷‍♂️

msmanoj: Cool, that's done.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/GeneViewSideBar.tsx
```diff
@@ -32,7 +32,7 @@ type Props = {
 };
 
 const client = new ApolloClient({
-  uri: 'http://hx-rke-wp-webadmin-13-worker-1.caas.ebi.ac.uk:31497'
+  uri: 'http://web-thoas.review.ensembl.org/'
```
azangru, line 35: sneaky 😄

msmanoj: Ah! No wonder why my commit was missing from the other branch! 
I will get this reverted in this PR and do this update separetely.

PR #313, opened 2020-06-30
Update useOutsideClick hook to support multiple refs

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-452

## Description
- Updated the useOutsideClick hook to support multiple refs to be passed in.
- I have renamed the `ref` prop to `refs` and changes it's type to an array. We need to decide if this is okay.

## Views affected
- Pointer Box
- Browser Region Editor
- Select options panel
- Browser track config

## Review deployment
http://use-outside-click.review.ensembl.org/

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
@@ -53,7 +53,7 @@ export const BrowserNavBarRegionSwitcher = (props: Props) => {
       <div className={styles.regionFieldWrapper}>
         <BrowserRegionField />
       </div>
-      {props.viewportWidth >= BreakpointWidth.BIG_DESKTOP && (
+      {props.viewportWidth >= BreakpointWidth.LAPTOP && (
         <div className={styles.regionEditorWrapper}>
           <BrowserRegionEditor />
         </div>
--- a/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
+++ b/src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
@@ -53,7 +53,7 @@ export const BrowserNavBarRegionSwitcher = (props: Props) => {
       <div className={styles.regionFieldWrapper}>
         <BrowserRegionField />
       </div>
-      {props.viewportWidth >= BreakpointWidth.LAPTOP && (
+      {props.viewportWidth >= BreakpointWidth.BIG_DESKTOP && (
         <div className={styles.regionEditorWrapper}>
           <BrowserRegionEditor />
         </div>
--- a/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
+++ b/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
@@ -74,7 +74,7 @@ export const BrowserTrackConfig = (props: BrowserTrackConfigProps) => {
     selectedCog in trackConfigLabel ? trackConfigLabel[selectedCog] : true;
 
   const ref = useRef(null);
-  useOutsideClick(ref, props.onClose);
+  useOutsideClick([ref], props.onClose);
 
   const toggleName = useCallback(() => {
     if (applyToAll) {
--- a/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
+++ b/src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
@@ -74,7 +74,7 @@ export const BrowserTrackConfig = (props: BrowserTrackConfigProps) => {
     selectedCog in trackConfigLabel ? trackConfigLabel[selectedCog] : true;
 
   const ref = useRef(null);
-  useOutsideClick([ref], props.onClose);
+  useOutsideClick(ref, props.onClose);
 
   const toggleName = useCallback(() => {
     if (applyToAll) {
--- a/src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
+++ b/src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
@@ -93,7 +93,7 @@ const HomepageAppLinksRow = (props: HomepageAppLinksRowProps) => {
     toggleExpand();
   };
 
-  useOutsideClick(elementRef, onOutsideClick);
+  useOutsideClick([elementRef], onOutsideClick);
 
   const rowClasses = classNames(styles.homepageAppLinksRow, {
     [styles.homepageAppLinksRowExpanded]: isExpanded
@@ -126,8 +126,8 @@ const HomepageAppLinksRow = (props: HomepageAppLinksRowProps) => {
       <div className={styles.homepageAppLinkButtons}>
         <ViewInApp
           links={{
-            genomeBrowser: urlFor.browser({genomeId: species.genome_id}),
-            entityViewer: urlFor.entityViewer({genomeId: species.genome_id})
+            genomeBrowser: urlFor.browser({ genomeId: species.genome_id }),
+            entityViewer: urlFor.entityViewer({ genomeId: species.genome_id })
           }}
         />
       </div>
--- a/src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
+++ b/src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
@@ -93,7 +93,7 @@ const HomepageAppLinksRow = (props: HomepageAppLinksRowProps) => {
     toggleExpand();
   };
 
-  useOutsideClick([elementRef], onOutsideClick);
+  useOutsideClick(elementRef, onOutsideClick);
 
   const rowClasses = classNames(styles.homepageAppLinksRow, {
     [styles.homepageAppLinksRowExpanded]: isExpanded
--- a/src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
+++ b/src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
@@ -87,7 +87,7 @@ const PointerBox = (props: PointerBoxProps) => {
   });
   const pointerBoxRef = useRef<HTMLDivElement>(null);
 
-  useOutsideClick(pointerBoxRef, props.onOutsideClick);
+  useOutsideClick([pointerBoxRef], props.onOutsideClick);
 
   useEffect(() => {
     const pointerBoxElement = pointerBoxRef.current as HTMLDivElement;
--- a/src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
+++ b/src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
@@ -87,7 +87,7 @@ const PointerBox = (props: PointerBoxProps) => {
   });
   const pointerBoxRef = useRef<HTMLDivElement>(null);
 
-  useOutsideClick([pointerBoxRef], props.onOutsideClick);
+  useOutsideClick(pointerBoxRef, props.onOutsideClick);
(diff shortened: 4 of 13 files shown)


src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
```diff
@@ -53,7 +53,7 @@ export const BrowserNavBarRegionSwitcher = (props: Props) => {
       <div className={styles.regionFieldWrapper}>
         <BrowserRegionField />
       </div>
-      {props.viewportWidth >= BreakpointWidth.BIG_DESKTOP && (
+      {props.viewportWidth >= BreakpointWidth.LAPTOP && (
```
msmanoj, line 56: TODO: Revert this before merging

azangru:
Don't you want to create a playground in Storybook to explore how this hook is going to deal with the following situations:

1) Both the child and the parent get re-rendered during click (this may break the check based on the parent/child DOM hierarchy, which we are currently using with the `contains` method; so in this case it may be more desirable to pass to the hook an option to check the actual geometric coordinates of both parent and child on the screen, unless of course re-rendering of the parent will break the geometric coordinates as well)

image

2) A child continues beyond the parent. In this case, only the test based on the DOM hierarchy can work.

image

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -17,11 +17,23 @@
 import { useEffect } from 'react';
 
 export default function useOutsideClick<T extends HTMLElement>(
-  ref: React.RefObject<T>,
+  refs: React.RefObject<T>[],
```
azangru, line 20: Why did you decide against allowing both a single ref and an array of refs?

msmanoj: I thought about allowing both but this way I felt it will be consistent throughout. I also feel `refs` is slightly better than `refOrRefs`.

azangru: >  it will be consistent throughout

But will it be convenient?

There are lots of libraries that allow passing different types of arguments. `classnames` is a good example — you can pass a single class name to it, or you can pass several class names, or you can pass an array of class names, or you can pass a configuration object. It's not forcing the user to be consistent — it's allowing them to pass whatever is more convenient at the moment.

I think passing a single ref if you only care about one ref, and passing an array of refs if you care about many refs is more convenient than having to always pass an array of refs, even when it's a single ref.

msmanoj: Fair enough. Shall we simply call it `refs`?

azangru: this check:

`const refs = Array.isArray(refOrRefs) ? refOrRefs : [refOrRefs];` will rename the parameter to refs, so that it can then conveniently be called `refs` within the body of the function :-)

azangru: you can call `refOrRefs` `input` if you don't like the name `refOrRefs`  😄 

I just wanted to avoid calling something that may be singular with a name that is plural

msmanoj:
> Don't you want to create a playground in Storybook to explore how this hook is going to deal with the following situations:
> 
> 1. Both the child and the parent get re-rendered during click (this may break the check based on the parent/child DOM hierarchy, which we are currently using with the `contains` method; so in this case it may be more desirable to pass to the hook an option to check the actual geometric coordinates of both parent and child on the screen, unless of course re-rendering of the parent will break the geometric coordinates as well)
> 
> image
> 
> 1. A child continues beyond the parent. In this case, only the test based on the DOM hierarchy can work.
> 
> 

I have added an entry to the storybook and also update the logic to handle overflowing and removed elements (Seems to be working fine for me).

src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -74,7 +74,7 @@ export const BrowserTrackConfig = (props: BrowserTrackConfigProps) => {
     selectedCog in trackConfigLabel ? trackConfigLabel[selectedCog] : true;
 
   const ref = useRef(null);
-  useOutsideClick(ref, props.onClose);
+  useOutsideClick([ref], props.onClose);
```
azangru, line 77: no need for the array now?

src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
```diff
@@ -93,7 +93,7 @@ const HomepageAppLinksRow = (props: HomepageAppLinksRowProps) => {
     toggleExpand();
   };
 
-  useOutsideClick(elementRef, onOutsideClick);
+  useOutsideClick([elementRef], onOutsideClick);
```
azangru, line 96: no need for the array now?

src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
```diff
@@ -87,7 +87,7 @@ const PointerBox = (props: PointerBoxProps) => {
   });
   const pointerBoxRef = useRef<HTMLDivElement>(null);
 
-  useOutsideClick(pointerBoxRef, props.onOutsideClick);
+  useOutsideClick([pointerBoxRef], props.onOutsideClick);
```
azangru, line 90: [] => 🙅‍♂️

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -17,17 +17,47 @@
...
+      // Reset the clickedInside flag to false
+      clickedInside = false;
+      return;
+    }
+
+    refs.forEach((ref) => {
+      if (ref.current && ref.current.contains(event.target as HTMLElement)) {
+        clickedOutside = false;
+      }
+    });
```
azangru, line 40: You can probably do this even without having to introduce the `clickedOutside` variable.

```
for (const ref of refs) {
  if (ref.current && ref.current.contains(event.target as HTMLElement)) {
    callback();
    break;
  }
}
```

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -17,17 +17,47 @@
...
+      all the clicks to any element within the ref to update the clickedInside flag.
+    */
+    refs.forEach((ref) => {
+      ref?.current?.addEventListener('click', () => {
+        clickedInside = true;
+      });
+    });
+
     document.addEventListener('click', handleClickOutside);
     return () => document.removeEventListener('click', handleClickOutside);
```
azangru, line 61: What about all the event listeners that you added in `refs.forEach`? We should clean them up as well.

src/ensembl/stories/hooks/use-outside-click/useOutsideClick.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import React, { useState, useRef } from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import useOutsideClick from 'src/shared/hooks/useOutsideClick';
+
+import styles from './useOutsideClick.stories.scss';
+
+storiesOf('Components|Hooks/useOutsideClick', module).add('default', () => {
```
azangru, line 24: > Components|Hooks/useOutsideClick

Hooks aren't components 😄 They probably deserve a section of their own.

src/ensembl/stories/hooks/use-outside-click/useOutsideClick.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const [shouldShowChild, showChild] = useState(true);
+
+  const parentRef = useRef<HTMLDivElement>(null);
+
+  const callback = () => {
+    action('clicked-outside')();
+  };
+
+  useOutsideClick([parentRef], callback);
```
azangru, line 33: [] => 🙅‍♂️

src/ensembl/src/shared/hooks/useOutsideClick.tsx
msmanoj: Would it not get cleaned up automatically?
https://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory/12528067

azangru: That's assuming the refs passed to this hooks get unmounted when the component that called the hook is unmounted. Which they probably should; but I wonder if there can be any weird edge cases. I we just remove the listeners we won't even have to think about this.

src/ensembl/stories/hooks/use-outside-click/useOutsideClick.stories.tsx
msmanoj: That's done

src/ensembl/stories/hooks/use-outside-click/useOutsideClick.stories.tsx
msmanoj: That's done

src/ensembl/src/shared/hooks/useOutsideClick.tsx
msmanoj: That's done

src/ensembl/src/shared/components/pointer-box/PointerBox.tsx
msmanoj: That's done

src/ensembl/src/content/home/components/homepage-app-links/HomepageAppLinks.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/browser-track-config/BrowserTrackConfig.tsx
msmanoj: That's done

src/ensembl/src/shared/hooks/useOutsideClick.tsx
msmanoj: That's done

src/ensembl/src/shared/hooks/useOutsideClick.tsx
```diff
@@ -58,6 +58,13 @@ export default function useOutsideClick<T extends HTMLElement>(
     });
 
     document.addEventListener('click', handleClickOutside);
-    return () => document.removeEventListener('click', handleClickOutside);
+    return () => {
+      document.removeEventListener('click', handleClickOutside);
+      refs.forEach((ref) => {
+        ref?.current?.removeEventListener('click', () => {
+          clickedInside = true;
+        });
```
azangru, line 62: wow, wow, I don't think that's how it works. You have to asssign `() => { clickedInside = true; }` to a variable, and then pass it to removeEventListener; like document.removeEventListener is doing.

Every 

```
() => {
  clickedInside = true;
}
```
is a declaration of a new anonymous function; while you need to remove the exact same function you passed when you added a listener.

msmanoj: Ah yes! That's done now. I wonder what happens with the previous approach?

azangru:
Pushed an update to the story and a couple of fixes for the hook. The tests seem to be passing now.

src/ensembl/src/content/app/browser/browser-nav/BrowserNavBarRegionSwitcher.tsx
azangru: Don't forget!

msmanoj: This change is now reverted.

PR #315, opened 2020-07-03
Retain EntityViewer state

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-516

## Description

The following states are now maintained:
- View
- Expanded transcripts
- Expanded transcript downloads
- Expanded proteins

I have added `?view=transcripts` as we won't know if the intended view when revisiting is the default `transcripts` or the view stored in redux.

Currently the state is not retained during browser refresh as the state management spreadsheet is not updated for entity viewer. I have created a separate ticket for it: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-689

## Views affected
EntityViewer -> GeneView
EntityViewer -> TranscriptsList
EntityViewer -> ProteinsList

### Other effects


src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -47,6 +48,7 @@ import styles from './EntityViewer.scss';
 type Props = {
   isSidebarOpen: boolean;
   viewportWidth: BreakpointWidth;
+  activeGenomeId: string | null;
```
azangru, line 51: How is it used here?

msmanoj: It's unused. I missed to clean it up.

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -180,17 +180,16 @@ const useGeneViewRouting = () => {
...
+    if (view && viewInRedux !== view) {
       dispatch(setGeneViewName(view));
+    } else {
+      const url = urlFor.entityViewer({
+        genomeId,
+        entityId,
+        view: viewInRedux
+      });
+
+      dispatch(replace(url));
```
azangru, line 192: Could you explain the change please?

My thinking for the deleted version of the code was:
- View can be falsey (empty string or undefined, don't remember); and it actually is this way for the default transcripts list. After your change (`view && viewInRedux !== view`) I suspect that the view in redux will not be properly updated when switching to default transcripts view
- I also thought that the only case when the view in the url should be dictated by redux rather than by the url is when the user is switching the species.

If the reason for your change is to support a scenario such as this:
- switch to a non-default view in EntityViewer
- navigate to genome browser
- refresh the page so that redux state is reset
- navigate back to EntityViewer and see the view that you visited last time

then I am not even sure whether we want this behaviour (will need to consult with Andrea, of course)

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -23,17 +23,27 @@ import DefaultTranscriptsListItem from './default-transcripts-list-item/DefaultT
 
 import { TicksAndScale } from 'src/content/app/entity-viewer/gene-view/components/base-pairs-ruler/BasePairsRuler';
 import { Gene } from 'src/content/app/entity-viewer/types/gene';
+import { EntityViewerGeneViewTranscriptsUI } from 'src/content/app/entity-viewer/state/gene-view/transcripts/entityViewerGeneViewTranscriptsState';
+
+import { getTranscriptsUI } from 'src/content/app/entity-viewer/state/gene-view/transcripts/entityViewerGeneViewTranscriptsSelectors';
 
 import styles from './DefaultTranscriptsList.scss';
+import { connect } from 'react-redux';
```
azangru, line 31: ⬆️ the import

azangru:
I am very anxious in general by the amount of information that this PR is attempting to store.

There are some bits of information that we definitely need to store and that even now we feel are lacking from the EntityViewer. For example, we definitely need to preserve what entity is active for what species, so that we are not faced with the same blank screen with a single link to a default entity when we switch between species or between EntityViewer and other applications.

But it feels horrifying to store all the UI details about which transcript is currently expanded, and in which of them the Download section has been expanded. This information is so transient and so local, shat it should have no place in redux. It is even more terrifying to think of keeping all that in browser storage without any mechanism for deleting this information.

I know that you've put a lot of work in this PR, but is it possible that we focus in it only on the active entity and do not try to store transcript UI details?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
msmanoj: This change is just to handle the following scenario:

- switch to a non-default view in EntityViewer
- navigate to genome browser
- navigate back to EntityViewer and see the view that you visited last time

Also, with the previous implementation, when we revisit entity viewer without the view parameter, we do not know if the intended view is the default  `transcripts` or the view stored in redux. To handle this, I have added `?view=transcripts` to be the default view instead of an empty value.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -67,4 +72,18 @@ describe('<TranscriptsListItemInfo /', () => {
...
+
+  it('shows Download component by default if expandDownload is true', () => {
+    wrapper = mount(
+      <MemoryRouter>
+        <TranscriptsListItemInfo {...props} expandDownload={true} />
+      </MemoryRouter>
+    );
+
+    expect(wrapper.find(InstantDownloadTranscript)).toHaveLength(1);
+  });
```
azangru, line 88: It's a little surprising that the two neighbouring tests initialise the wrapper component differently. Maybe add a minor refactor to remove the creation of the wrapper from beforeEach into a helper function that accepts an object of partial props, mixes them together with default props, mounts the component and returns the wrapper? Such as [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx#L33-L43)?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -40,20 +41,19 @@ import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
 
 import transcriptsListStyles from '../DefaultTranscriptsList.scss';
 import styles from './TranscriptsListItemInfo.scss';
+import { connect } from 'react-redux';
```
azangru, line 44: ⬆️ the import

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
azangru: two things:

- doesn't it need the MemoryRouter anymore?
- why do you call it mount**Default**TranscriptListItem? It no longer has default set of props in the last test.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -52,17 +53,35 @@ export const setDataFromUrl: ActionCreator<ThunkAction<
...
+    dispatch(setActiveGenomeId(genomeIdFromUrl));
+    dispatch(fetchGenomeData(genomeIdFromUrl));
     // TODO: when backend is ready, entity info may also need fetching
-  } else {
+  } else if (activeGenomeId && entityIdForUrl) {
+    const newUrl = urlHelper.entityViewer({
+      genomeId: activeGenomeId,
+      entityId: entityIdForUrl
+    });
+    dispatch(replace(newUrl));
```
azangru, line 83: This is very scary, because this is the routing concern that GeneView component is handling; and now there are two places in code that can send conflicting commands to the router. This code branch shouldn't be necessary (and in fact was causing bugs in my PR https://github.com/Ensembl/ensembl-client/pull/324. I deleted it in that PR.

msmanoj: I tried removing this but it doesn't seem to retain the state when we switch between species.

Can you please try the following:
- Select BRCA2 for humans
- Switch to a different species
- Go back to Human

The example links are shown again.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
msmanoj: > doesn't it need the MemoryRouter anymore?

The test fails if I remove it?

>why do you call it mountDefaultTranscriptListItem? It no longer has default set of props in the last test. And mountComponent/renderComponent would be even shorter than mountTranscriptListItem.

Changed it to be `renderComponent`.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
azangru: @msmanoj  I have pushed a commit addressing this issue to your branch.

Briefly, it removes the redirect logic from `entityViewerGeneralActions` and puts it in the `EntityViewer` component. With this change, I expect that the routing hook in GeneView is not going to compete with the routing logic in the action. Before this change, the `replace` function both in the action and in the GeneView could fire; but with this logic moved to EntityViewer component, the GeneView will not even be mounted unless the `replace` function in the EntityViewer fires; and if  GeneView has mounted, the `replace` function in EntityViewer will not fire.

azangru:
General remarks:

1) Conventionally, actions, selectors, and reducers should be thought about together — e.g. in redux-toolkit, to which we may want to try to switch, if we have time, the combination of the three is describes as a "slice" of redux state. So I am a bit nervous that you are introducing actions files and selectors files that do not have their own reducer, but do something with the parent reducer and parent portion of the redux state.

2) We need to find out whether the expanded state of transcript download components even has to be preserved when switching between species or applications. Downloading a transcript is such a momentary concern that we may not want to keep these components expanded at all.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -32,11 +38,16 @@ import styles from './DefaultTranscriptsList.scss';
 type Props = {
   gene: Gene;
   rulerTicks: TicksAndScale;
+  transcriptscontentUI: EntityViewerGeneViewTranscriptsUI;
```
azangru, line 41: Not sure about the name (what does the word "content" add?), but if we keep it, the `c` should be upper-cased.

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsActions.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import { getProteinsUI } from 'src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsSelectors';
+import { getGeneViewContentUI } from 'src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors';
+
+import { RootState } from 'src/store';
+import { View } from 'src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewState';
+import { updateActiveGeneViewUIState } from 'src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewActions';
+import { EntityViewerGeneViewProteinsUI } from 'src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsState';
+
+export const updateGeneViewProteinsUIState: ActionCreator<ThunkAction<
```
azangru, line 33: ActionCreator has a horrible type definition that loses type information (see Note in https://github.com/Ensembl/ensembl-client/pull/282). Let's not use it.

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsActions.ts
```diff
@@ -0,0 +1,101 @@
...
+  const contentUI = getGeneViewContentUI(state);
+
+  dispatch(
+    updateActiveGeneViewUIState({
+      activeGenomeId,
+      activeObjectId,
+      fragment: {
+        contentUI: {
+          ...contentUI,
+          [View.PROTEIN]: proteinsContentUI
```
azangru, line 60: It's a bit of a philosophical question, but I think that spreading the whole slice of the current state over and applying the update to it (as done here with `...contentUI`) is the task of the reducer, not of the action. The action should just dispatch the bit that gets updated.

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsActions.ts
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
msmanoj: I missed to remove it. It's done now.

msmanoj:
>Conventionally, actions, selectors, and reducers should be thought about together — e.g. in redux-toolkit, to which we may want to try to switch, if we have time, the combination of the three is describes as a "slice" of redux state. So I am a bit nervous that you are introducing actions files and selectors files that do not have their own reducer, but do something with the parent reducer and parent portion of the redux state.

I have made some changes to have separate reducers for proteins and transcripts. I am not sure it's any better now but please have a look and let me know if it can be done differently.

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/entityViewerGeneViewProteinsActions.ts
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
```diff
@@ -36,28 +39,40 @@ jest.mock(
...
-  it('toggles transcript item info', () => {
+  it('toggles transcript item info onClick', () => {
+    wrapper = renderComponent();
     wrapper.find('.middle').simulate('click');
-    expect(wrapper.exists(TranscriptsListItemInfo)).toBe(true);
+    expect(toggleTranscriptInfo).toBeCalled();
 
     wrapper.find('.right').simulate('click');
-    expect(wrapper.exists(TranscriptsListItemInfo)).toBe(false);
+    expect(toggleTranscriptInfo).toBeCalled();
```
azangru, line 70: this line doesn't really check anything. Even if the `click` on the previous line did not trigger `toggleTranscriptInfo`, it would still test positive with `toBeCalled`, because we have not reset the mock since line 67.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
```diff
@@ -36,28 +39,40 @@ jest.mock(
...
+  });
+
+  it('displays transcript info by default if expandTranscript is true', () => {
+    wrapper = renderComponent({ expandTranscript: true });
+
+    expect(wrapper.exists(TranscriptsListItemInfo)).toBe(true);
```
azangru, line 76: you may want to make sure that TranscriptsListItemInfo wasn't there if expandTranscript is false.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors.ts
```diff
@@ -20,28 +20,28 @@ import {
...
   if (!activeGenomeId || !activeObjectId) {
-    return defaultEntityViewerGeneViewUIState;
+    return;
   }
-  return (
-    state.entityViewer.geneView?.[activeGenomeId]?.[activeObjectId] ||
-    defaultEntityViewerGeneViewUIState
-  );
+
+  return state.entityViewer.geneView?.[activeGenomeId]?.[activeObjectId];
```
azangru, line 40: 1) I am confused about the question mark after geneView (I understand that it was inherited from the previous code). `state.entityViewer.geneView` can never be null or undefined, can it?

2) I am a bit confused, too, by why you decided against returning the default state if geneView[activeGenomeId]?.[activeObjectId] is undefined. Is it better?

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/transcriptData.ts
```diff
@@ -77,18 +79,21 @@ export const fetchTranscript = async (
   signal?: AbortSignal
 ): Promise<Transcript> => {
   const transcriptUrl = `https://rest.ensembl.org/lookup/id/${id}?expand=1;content-type=application/json`;
-  const transcript: TranscriptInResponse = (await fetch(transcriptUrl, {
-    signal
-  }).then((response) => response.json())) as TranscriptInResponse;
+  const transcript: TranscriptInResponse = (await apiService.fetch(
```
azangru, line 82: `: TranscriptInResponse` is redundant?

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/transcriptData.ts
```diff
@@ -77,18 +79,21 @@ export const fetchTranscript = async (
...
+    transcriptUrl,
+    {
+      signal
+    }
+  )) as TranscriptInResponse;
 
   let proteinFeatures;
 
-  if (transcript.Translation) {
+  if (transcript?.Translation) {
```
azangru, line 91: don't you promise on line 87 that transcript is of type `TranscriptInResponse`? Why the question mark then?

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewReducer.ts
```diff
@@ -45,7 +51,45 @@ export default function entityViewerGeneViewReducer(
...
+          ...state[action.payload.activeGenomeId],
+          [action.payload.activeObjectId]: {
+            ...state[action.payload.activeGenomeId][
+              action.payload.activeObjectId
+            ],
+            contentUI: {
+              ...state[action.payload.activeGenomeId][
+                action.payload.activeObjectId
+              ].contentUI,
+              protein: proteinsReducer(state, action)
```
azangru, line 67: No, what I meant was we would split gene view reducer into smaller reducers and combine them with combineReducers, as in `entityViewerReducer`

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/transcriptData.ts
msmanoj: That's right. I temporarily added it as the rest endpoint kept breaking for me (most likely to be connection issue on my side). I've removed it.

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/transcriptData.ts
msmanoj: Yes, removed it.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
msmanoj: Added the block to reset the mock

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors.ts
msmanoj: > I am confused about the question mark after geneView (I understand that it was inherited from the previous code). state.entityViewer.geneView can never be null or undefined, can it?

That's right. It cannot be undefined or null.

> I am a bit confused, too, by why you decided against returning the default state if geneView[activeGenomeId]?.[activeObjectId] is undefined. Is it better?

When we retrieve the value for `viewInRedux` during the initial render in `GeneView.tsx`, we were returning the default value which is actually not from redux. 
As a result, the state does not get updated here:
https://github.com/Ensembl/ensembl-client/pull/315/files#diff-6d9c2d66030239cb4b481357e47aa671R187

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewReducer.ts
msmanoj: I tried that to use `combineReducers` but I had no luck getting it to work without changing the current structure of the state.

azangru: Ok, as discussed at the code review, we will try to move to `redux-toolkit` anyway, which will make us reconsider this piece of state later. It's fine if it works for now.

src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewSelectors.ts
azangru: > That's right. It cannot be undefined or null.

Let's delete the question mark after `state.entityViewer.geneView?` then?

msmanoj: Sorry, I thought I had removed it. That's done now.

PR #316, opened 2020-07-03
[ENSWBSITES-647] Entity viewer design review changes

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-647

## Description
Review changes by Andrea based on this https://docs.google.com/spreadsheets/d/1y3ThUJbbW2_7tjIchmrvNn3v3AK539d5aGyARY7jNeA/edit#gid=1244768304

## Views affected
Entity Viewer

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -24,6 +24,8 @@ import DefaultTranscriptsListItem from './default-transcripts-list-item/DefaultT
 import { TicksAndScale } from 'src/content/app/entity-viewer/gene-view/components/base-pairs-ruler/BasePairsRuler';
 import { Gene } from 'src/content/app/entity-viewer/types/gene';
 
+import { ReactComponent as Chevron } from 'static/img/shared/chevron-down.svg';
+
 import styles from './DefaultTranscriptsList.scss';
 
 type Props = {
@@ -39,7 +41,10 @@ const DefaultTranscriptslist = (props: Props) => {
     <div>
       <div className={styles.header}>
         <div className={styles.row}>
-          <div className={styles.left}>Filter & sort</div>
+          <div className={styles.left}>
+            Filter & sort
+            <Chevron className={styles.headerChevron} />
+          </div>
           <div className={styles.middle}></div>
           <div className={styles.right}>Transcript ID</div>
         </div>
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar-tabs/GeneViewSidebarTabs.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar-tabs/GeneViewSidebarTabs.scss
@@ -16,7 +16,7 @@
 .defaultTab {
   background-color: transparent;
   color: $blue;
-  font-weight: 700;
+  font-weight: 400;
   border-radius: 3px;
   margin-right: 3px;
   height: 30px;
--- a/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
+++ b/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
@@ -5,20 +5,12 @@
 }
 
 .selectedNode {
-  border-radius: 12px;
-  height: 24px;
-  padding: 1px 20px 0 20px;
+  border-radius: 30px;
+  height: 21px;
+  padding: 1px 16px 0;
   margin-right: 24px;
   background-color: $black;
   color: $white;
   text-align: center;
   cursor: default;
-}
-
-.emptyNode {
-  border-radius: 100%;
-  background-color: $grey;
-  margin: 6px 24px 0 0;
-  width: 10px;
-  height: 10px;
 }
--- a/src/ensembl/src/shared/components/tabs/Tabs.scss
+++ b/src/ensembl/src/shared/components/tabs/Tabs.scss
@@ -2,7 +2,7 @@
 
 .tabsContainer {
   display: flex;
-  font-size: 14px;
+  font-size: 13px;
   overflow-y: hidden;
   height: 40px;
   scrollbar-width: none;
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss
@@ -16,8 +16,8 @@ $backgroundColor: $black;
   display: grid;
   background-color: $backgroundColor;
   grid-template-columns: 178px 902px 1fr;
-  grid-template-rows: min-content 90px 1fr;
-  padding: 75px 20px 10px 20px;
+  grid-template-rows: min-content 75px 1fr;
+  padding: 60px 20px 10px;
   height: 100%; // may need to change this later
   width: 100%;
   overflow-y: auto;
@@ -42,7 +42,7 @@ $backgroundColor: $black;
   top: -105px;
   height: 100%;
   background-color: $backgroundColor;
-  padding-left: 228px;
+  padding-left: 180px;
 }
 
 .geneViewTabContent {
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -54,6 +54,7 @@ const QUERY = gql`
   query Gene($id: String!) {
     gene(byId: { id: $id }) {
       id
+      version
       slice {
         location {
           start
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -3,15 +3,26 @@
 
 .header {
   border-bottom: 1px solid $orange;
-  padding: 8px 0;
+  padding: 0;
   position: sticky;
   top: -15px;
   background-color: $black;
   z-index: 1;
+  font-size: 12px;
+}
+
+.headerChevron {
+  width: 16px;
+  height: 16px;
+  cursor: pointer;
+  margin-left: 0.5em;
+  top: 5px;
+  position: relative;
 }
 
 .content {
   position: relative;
+  padding-top: 36px;
 }
 
 .stripedBackground {
@@ -33,7 +44,8 @@
 .row {
   display: grid;
   grid-template-columns: $left_column $middle_column $middle_column_right_gap $right_column;
-  padding: 3px 0;
+  padding: 9px;
+  font-size: 12px;
 }
 
 .left {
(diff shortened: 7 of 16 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -25,7 +25,7 @@ import TranscriptsFilter from 'src/content/app/entity-viewer/gene-view/component
 import { TicksAndScale } from 'src/content/app/entity-viewer/gene-view/components/base-pairs-ruler/BasePairsRuler';
 import { Gene } from 'src/content/app/entity-viewer/types/gene';
 
-import { ReactComponent as Chevron } from 'static/img/shared/chevron-down.svg';
+import { ReactComponent as ChevronDown } from 'static/img/shared/chevron-down.svg';
```
azangru, line 28: I am a little confused about how we handle the chevrons. We can have a single svg that we then rotate 360 degrees using CSS. We might have 2 svgs — one for up/down chevron and one for left/right chevron that we rotate 180 degrees using CSS, or we might have 4 different svgs (seems a bit wasteful, but whatever). At the moment, I don't think we even have a clear strategy. What you guys think? @imransl @msmanoj ?

azangru:
You may notice that as you open and close the filters box, the Filter and sort label moves from one side to the other slightly. Not sure whether we want to fix this, but we might.

image

azangru:
You broke something with the paddings. Notice how the transcript images and the transcript info box continue to the right of the last (right) orange vertical line? That shouldn't happen (compare internal deployment). The vertical lines are basically the ruler; all the features should stay within the boundaries of that ruler.

image

src/ensembl/src/shared/components/tabs/Tabs.scss
```diff
@@ -2,7 +2,7 @@
 
 .tabsContainer {
   display: flex;
-  font-size: 14px;
+  font-size: 13px;
```
azangru, line 5: If `$global-font-size` has been changed to 13px, this class should just inherit this font-size from the body making this rule unnecessary.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
ens-ma7: Just to clarify this file wasn't changed by jyo and is not related to that PR, i added this change for the Filter and sort PR and it is coming up in this PR because of the git commit prehook to fix the styling. I think having one chevron and rotating is a good idea, its just one file and i know the images/icons will increase and become a maze. But i suggest creating another ticket for this.

ens-ma7:
> You may notice that as you open and close the filters box, the Filter and sort label moves from one side to the other slightly. Not sure whether we want to fix this, but we might.
> 
> image

I will fix this as part of the ticket that i am working on for filter and sort.

ens-ma7:
> The paddings broke something in the default transcripts list. Notice how the transcript images and the transcript info box continue to the right of the last (right) orange vertical line? That shouldn't happen (compare internal deployment). The vertical lines are basically the ruler; all the features should stay within the boundaries of that ruler.
> 
> image

@jyothishnt related to this, the label next to each transcripts (left, stable id) has also gone down a bit when you reduced the font size, it does look like we need to move them up a bit.

ens-ma7:
image

when you changed the font size, The text in the button tabs are not centered, looks like some padding top is needed.

ens-ma7:
image

Proteins list styling has also changed. Not sure if it was expected or will be on a different ticket but something to note.

src/ensembl/src/shared/components/tabs/Tabs.scss
azangru: global-font-size describes the body font size (see _settings.scss and _normalize.scss). Tabs container will have inherited body font size; so specifying font-size here is redundant. In other places, such as sidebar, we may have overridden the font size with something different.

ens-ma7: I just noticed that in the google doc it does say body font size to be 13px so updating global-font-size is the correct way and then removing font-size in the class above. That will mean by default everywhere the font-size will be 13px unless overwritten.

jyothishnt: Yes thats right @ens-ma7 Its the global font size that needed change. Yes I will remove this redundant one.

jyothishnt:
> image
> 
> Proteins list styling has also changed. Not sure if it was expected or will be on a different ticket but something to note.

Yes Ridwan I have noticed it. This is because of the defaultTranscriptList padding changes. May be this was intended just for the transcript image tab

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
jyothishnt: It is better create a new ticket for the chevron icon functionality

jyothishnt:
> > You may notice that as you open and close the filters box, the Filter and sort label moves from one side to the other slightly. Not sure whether we want to fix this, but we might.
> > image
> 
> I will fix this as part of the ticket that i am working on for filter and sort.

Dont worry about this. I have the fix already. :-)

jyothishnt:
Updated changes. Please have a look @ens-ma7 , @azangru

azangru:
> Updated changes. Please have a look

@jyothishnt you may want to commit them 😉 The latest commit in this branch was 11 days ago:

image

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.scss
```diff
@@ -7,6 +7,7 @@
   border-radius: 3px;
   margin-right: 3px;
   height: 30px;
+  padding-top: 5px;
```
azangru, line 10: Let's do vertical centering using `display: flex; align-items: center`? It won't care if button size changes in the future.

azangru:
OK, we have a problem.

Scroll the transcripts list down until the tab buttons stick below the info bars. You will notice that the transcripts are now visible between the tab buttons and the Filter&sort row:

image

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
```diff
@@ -5,20 +5,12 @@
 }
```
azangru, line 5: We probably want to rename this component entirely — it's no longer a ScaleSwitcher; and it also doesn't have a "selected node" anymore. Let's ask what people think it should be called?

src/ensembl/src/shared/components/radio-group/RadioGroup.tsx
```diff
@@ -79,7 +71,10 @@ const RadioGroup = (props: Props) => {
     <div>
       {props.options.map((option, index) => (
         <div key={index} className={wrapperClass}>
-          <div onClick={() => handleOnChange(option.value)} className={getRadioClass(option)} />
+          <div
+            onClick={() => handleOnChange(option.value)}
```
azangru, line 75: I missed this in Ridwan's PR. Is't perfectly fine to call a function `handleChange`; it's also fine to call it `onChange`; but `handleOnChange` sounds awkward. Will it be ok to rename this function?

ens-ma7: @azangru it looks like handleOnChange is doing something else and then calling onChange, have a look at line 44.  can we still change this to onChange?

azangru: I think so. It's component's own onChange callback that does something and then calls the onChange that came with the props.

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-topbar/scale-switcher/ScaleSwitcher.scss
ens-ma7: chosen name is entityTypeIndicator

jyothishnt: Decided to have **EntityTypeIndicator** during 29/7/2020 code-review

ens-ma7:
The scrolling issue still needs fixing

jyothishnt:
> The scrolling issue still needs fixing

Scrolling issue is fixed in my last commit. For some reason it is not reflected in the test url. Do I need to update anything @kamaldodiya ? I thought it would automatically build on a push.

jyothishnt:
http://design-review-changes.review.ensembl.org/app/entity-viewer/homo_sapiens_GCA_000001405_14/gene:ENSG00000139618 
Please test here

jyothishnt:
Also it says there are no conflicts. So not merging dev branch on to this. Hope this is ok.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar-tabs/GeneViewSidebarTabs.scss
```diff
@@ -16,7 +16,7 @@
 .defaultTab {
   background-color: transparent;
   color: $blue;
-  font-weight: 700;
+  font-weight: 400;
```
azangru, line 19: 400 is the default font weight; so this rule is redundant.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.scss
```diff
@@ -7,6 +7,8 @@
   border-radius: 3px;
```
azangru, line 7: to the `background-color` rule several lines above. I know this comes from a different PR, but `#33adff;` is our `$blue` colour; so we should use that variable instead.

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-topbar/entity-viewer-indicator/EntityViewerIndicator.tsx
```diff
@@ -16,22 +16,14 @@
 
 import React from 'react';
 
-import styles from './ScaleSwitcher.scss';
+import styles from './EntityViewerIndicator.scss';
 
-const ScaleSwitcher = () => {
+const EntityViewerIndicator = () => {
   return (
     <div className={styles.container}>
```
azangru, line 23: Several things about this component:

1) Now that it's no longer a scale switcher, there is no point in having a `container` class. It should probably be just a single DOM element.
2) We may want to add a prop (let's say `{entity: string}`), which will be used to pass to the component the type of entity this is. So that we do not hard-code the word "Gene" inside the component, but make the component reusable (we already know it's going to be used in species selector)
3)  Class name `selectedNode` no longer makes sense. We should probably name the class of this single DOM element `entityIndicator`.
4) Could you please rename the component (and its scss counterpart) from EntityViewerIndicator to EntityIndicator? We already know this component is going to be used in species selector. We might even move this component to the `src/shared/components` folder.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-tabs/GeneViewTabs.scss
jyothishnt: I didnt understand. I am using $blue right?

azangru: Take a look at line 4 of this file. It says `background-color: #33adff;`. Let's change it to `background-color: $blue;`

jyothishnt: Aah got it. Thanks

PR #322, opened 2020-07-06
Add the popup for contextual help

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-575

## Description
- This PR adds the popup for contextual help
- _Note:_ contextual help button is shown only in Development and Internal environments

Also included in the PR:
- Modal component
- CloseButton component (we keep re-implementing it in multiple other components)

## Views affected
Species selector

## Deployment url
http://feature-contextual-help.review.ensembl.org

### DIFF
--- a/src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
+++ b/src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
@@ -24,15 +24,23 @@ export type HelpVideo = {
   url: string;
 };
 
+type ArticleSummary = {
+  title: string;
+  slug: string;
+  path: string;
+};
+
 export type HelpArticle = {
   path: string;
   slug: string;
   body: string;
   videos: HelpVideo[];
+  related_articles: ArticleSummary[];
 };
 
 type Props = {
   article: HelpArticle;
+  onArticleChange: (slug: string) => void;
 };
 
 const HelpPopupBody = (props: Props) => {
@@ -44,12 +52,32 @@ const HelpPopupBody = (props: Props) => {
     </div>
   ));
 
+  const relatedArticles = article.related_articles.map((relatedArticle) => (
+    <span
+      key={relatedArticle.slug}
+      className={styles.relatedArticle}
+      onClick={() => props.onArticleChange(relatedArticle.slug)}
+    >
+      {relatedArticle.title}
+    </span>
+  ));
+
   return (
     <div className={styles.grid}>
       <div className={styles.text}>
         <div dangerouslySetInnerHTML={{ __html: article.body }} />
       </div>
       <div className={styles.video}>{videos}</div>
+      <div className={styles.aside}>
+        {Boolean(relatedArticles.length) && (
+          <>
+            <h2>Related...</h2>
+            <div className={styles.relatedArticlesContainer}>
+              {relatedArticles}
+            </div>
+          </>
+        )}
+      </div>
     </div>
   );
 };
--- a/src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
+++ b/src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
@@ -16,12 +16,16 @@
 
 import React from 'react';
 
+import { CircleLoader } from 'src/shared/components/loader/Loader';
+
+import { ReactComponent as VideoIcon } from 'static/img/shared/video.svg';
+
 import styles from './HelpPopupBody.scss';
 
 export type HelpVideo = {
   title: string;
   description: string;
-  url: string;
+  youtube_id: string;
 };
 
 type ArticleSummary = {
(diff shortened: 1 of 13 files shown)


src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
```diff
@@ -0,0 +1,85 @@
...
+};
+
+type ArticleSummary = {
+  title: string;
+  slug: string;
+  path: string;
+};
+
+export type HelpArticle = {
+  path: string;
```
msmanoj, line 34: Is path mandatory for all the articles?

src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type Props = {
+  article: HelpArticle;
+  onArticleChange: (slug: string) => void;
+};
+
+const HelpPopupBody = (props: Props) => {
+  const { article } = props;
+
+  const videos = article.videos.map((video) => (
```
msmanoj, line 49: Are we displaying multiple videos in a single popup?

src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
azangru: If you are asking whether the `path` field will always be in the article returned from the api, then yes.

If you are asking whether HelpPopupBody component needs to know about the `path` field, then no.

But currently HelpPopupBody is the place that is describing the shape of the article returned by the api. We may move the type definition to a dedicated types file if you would prefer that.

src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
azangru: I assumed that yes. Need to ask Andrea. I hope she will want to update the design of the popup anyway, because currently the video is unusably small — user would always have to switch it to full screen if they want to see what's going on.

azangru: @msmanoj I discussed the popup with Andrea, and we came to the following conclusions:

- Andrea needs to rethink the design of the contextual popup — specifically, what should happen when user clicks on a related article or a related video link.
- Until she does, there isn't much point in trying to style the popup window one way or the other; because we don't know what the final design is going to look like.
- there have already been changes, e.g. this design, which you likely used when you worked on contextual help:



has been changed to this (what I based my implementation on):



But even the second one is not final.

So I wouldn't worry about the number of allowed videos in the middle column. But I will try to avoid the flickering of the popup window when changing the article.

azangru:
@msmanoj I've updated the PR with the following:

 - if there are multiple videos related to an article, only the first one will show up in the middle column; while others will be shown as links in the right column
- click on a video link in the right column changes the video in the middle column
- the popup window no longer closes/reopens when a related article is selected

Changes can be seen on the feature branch.

msmanoj:
@azangru what happens when we click a related article/video that is linked with a different article/video? Does it update both?

azangru:
> what happens when we click a related article/video that is linked with a different article/video? Does it update both?

@msmanoj The current model posits that an article may have zero to many related articles, and zero to many related videos (see example[ json response](http://hx-rke-wp-webadmin-14-worker-1.caas.ebi.ac.uk:30853/api/article?slug=selecting-a-species) from the api).

In the current implementation, when user clicks on a related article, the whole content of the popup changes, as if the help button were linked to a different article. So if a related article has its own related videos and related articles, they would show in the middle and the right column. However, a click on a related video keeps the same article in the popup and only changes the video in the middle column of the popup.

I do not know whether it is the correct behaviour; I just implemented the behaviour that makes the most sense to me, and we'll have to wait for Andrea to describe to us what she thinks the actual behaviour should be.

src/ensembl/src/shared/components/help-popup/HelpPopupBody.tsx
msmanoj: Thanks. Happy to leave the type definition here.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const [slug, setSlug] = useState<string | null>(null);
+  const [selectedVideoId, setSelectedVideoId] = useState<string | null>(null);
+  const [shouldShowModal, setShouldShowModal] = useState(false);
+  const { helpApiHost } = config;
+
+  const query = slug ? getQuery({ slug }) : getQuery(props);
+
+  const url = `${helpApiHost}/api/article?${query}`;
+
+  // TODO: decide whether we want to show spinner while article content is loaded
```
msmanoj, line 64: I did see the spinner for a couple of seconds and it could be something to do with the speed of my connection here.

msmanoj (approved):

PR #324, opened 2020-07-08
Proposal for url/redux sync

azangru:
Proposal for PR https://github.com/Ensembl/ensembl-client/pull/315 that should allow both not having the `view` query parameter in the url by default and keeping the view in the redux state.


azangru:
It was eventually decided to go ahead with https://github.com/Ensembl/ensembl-client/pull/315 ; which makes this PR irrelevant.

PR #325, opened 2020-07-09
Restore track states on browser refresh

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-303
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-288

## Description
This PR fixes the issues with restoring the track states in the following scenarios:

- Restoring after browser refresh
- Switching between species
- Switching between apps

**Known bug**: The track states seems to be getting restored properly on the React side. However, sometimes the track will still be visible on different zoom levels where it's supposed to be hidden. This is something that needs further investigation. 

## Deployment:
http://browser-track-state.review.ensembl.org

## Views affected
Browser -> Tracks

**Steps to reproduce the hidden track bug**: (@ens-ds23)
- Go to http://browser-track-state.review.ensembl.org
- Add human and wheat
- Go to Genome browser -> Wheat -> Turn off `contig` track
- Switch to human -> Select BRCA2 -> Switch back to wheat

Now notice that the `contig` track appears to be visible but gets hidden when we zoom in or out.

Created a ticket to investigate this:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-706

### DIFF
--- a/src/ensembl/src/content/app/browser/browserActions.ts
+++ b/src/ensembl/src/content/app/browser/browserActions.ts
@@ -216,43 +216,6 @@ export const restoreBrowserTrackStates: ActionCreator<ThunkAction<
   });
 };
 
-export const resetBrowserTrackStates: ActionCreator<ThunkAction<
-  void,
-  any,
-  null,
-  Action<string>
->> = () => (_, getState: () => RootState) => {
-  const state = getState();
-  const activeGenomeId = getBrowserActiveGenomeId(state);
-  const activeEnsObjectId = getBrowserActiveEnsObjectId(state);
-
-  if (!activeGenomeId || !activeEnsObjectId) {
-    return;
-  }
-
-  const trackStates = getBrowserTrackStates(state)[activeGenomeId];
-  // const trackStatesFromStorage = browserStorageService.getTrackStates();
-  const mergedTrackStates = {
-    ...get(trackStates, `objectTracks`),
-    ...get(trackStates, `commonTracks`)
-  } as TrackStates;
-
-  const tracksToTurnOff: string[] = [];
-  const tracksToTurnOn: string[] = [];
-
-  Object.values(mergedTrackStates).forEach((trackStates) => {
-    Object.keys(trackStates).forEach((trackId) => {
-      trackStates[trackId] === Status.UNSELECTED &&
-        tracksToTurnOn.push(trackId);
-    });
-  });
-
-  browserMessagingService.send('bpane', {
-    off: tracksToTurnOff,
-    on: tracksToTurnOn
-  });
-};
-
 export const openBrowserNav = createAction(
   'browser/open-browser-navigation',
   () => {
--- a/src/ensembl/src/content/app/browser/browserActions.ts
+++ b/src/ensembl/src/content/app/browser/browserActions.ts
@@ -216,6 +216,43 @@ export const restoreBrowserTrackStates: ActionCreator<ThunkAction<
   });
 };
 
+export const resetBrowserTrackStates: ActionCreator<ThunkAction<
+  void,
+  any,
+  null,
+  Action<string>
+>> = () => (_, getState: () => RootState) => {
+  const state = getState();
+  const activeGenomeId = getBrowserActiveGenomeId(state);
+  const activeEnsObjectId = getBrowserActiveEnsObjectId(state);
+
+  if (!activeGenomeId || !activeEnsObjectId) {
+    return;
+  }
+
+  const trackStates = getBrowserTrackStates(state)[activeGenomeId];
+  // const trackStatesFromStorage = browserStorageService.getTrackStates();
+  const mergedTrackStates = {
+    ...get(trackStates, `objectTracks`),
+    ...get(trackStates, `commonTracks`)
+  } as TrackStates;
+
+  const tracksToTurnOff: string[] = [];
+  const tracksToTurnOn: string[] = [];
+
+  Object.values(mergedTrackStates).forEach((trackStates) => {
+    Object.keys(trackStates).forEach((trackId) => {
+      trackStates[trackId] === Status.UNSELECTED &&
+        tracksToTurnOn.push(trackId);
+    });
+  });
+
+  browserMessagingService.send('bpane', {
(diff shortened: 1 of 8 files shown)


azangru:
Several questions/comments:

1) `resetBrowserTrackStates` and `restoreBrowserTrackStates` seem to be doing something very similar. What's the difference between them?
2) `resetBrowserTrackStates` is called from `useBrowserRouting` hook, while `restoreBrowserTrackStates` is called from `TrackPanel`. Why are they distributed in different places? They will also dispatch their actions independently, which will send `bpane` events immediately one after another. Is this a good thing?
3) Neither `resetBrowserTrackStates` nor `restoreBrowserTrackStates` is modifying redux state; so they don't even need to be redux actions. What if we remove both of them from `browserActions` and combine their logic in a single custom hook, which will get required information about redux state via redux hooks?

azangru:
Actions arising from the discussion during code review:

- Check whether the React app knows about the full list of genome browser tracks (I believe it should). Explicitly tell genome browser to show certain tracks and to hide others when switching between the species
- Confirm that if a tracks-related message to genome browser is sent from Browser.tsx (make sure that it actually gets sent over `window.postMessage`) genome browser does not respond to it properly. If genome browser does not correctly respond to the message that we send to it, this should be raised as bug with @ens-ds23 
- Possibly move the code for restoring browser tracks from redux into a custom React hook.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -216,6 +216,43 @@ export const restoreBrowserTrackStates: ActionCreator<ThunkAction<
...
+  const state = getState();
+  const activeGenomeId = getBrowserActiveGenomeId(state);
+  const activeEnsObjectId = getBrowserActiveEnsObjectId(state);
+
+  if (!activeGenomeId || !activeEnsObjectId) {
+    return;
+  }
+
+  const trackStates = getBrowserTrackStates(state)[activeGenomeId];
+  // const trackStatesFromStorage = browserStorageService.getTrackStates();
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
```diff
@@ -140,6 +141,8 @@ const useBrowserRouting = () => {
         location: chrLocation ? getChrLocationStr(chrLocation) : null
       };
 
+      dispatch(resetBrowserTrackStates());
+
       dispatch(replace(urlFor.browser(params)));
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -216,6 +216,43 @@ export const restoreBrowserTrackStates: ActionCreator<ThunkAction<
...
+
+  Object.values(mergedTrackStates).forEach((trackStates) => {
+    Object.keys(trackStates).forEach((trackId) => {
+      trackStates[trackId] === Status.UNSELECTED &&
+        tracksToTurnOn.push(trackId);
+    });
+  });
+
+  browserMessagingService.send('bpane', {
+    off: tracksToTurnOff,
```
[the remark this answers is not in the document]
msmanoj: @imransl @azangru I have removed `resetBrowserTrackStates` entirely and now on/off instructions are sent for all the tracks every time we switch between species. 
This change seems to work fine. The only problem I noticed is the browser track visibility caching issue which is something else.

src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
```diff
@@ -141,9 +140,9 @@ const useBrowserRouting = () => {
         location: chrLocation ? getChrLocationStr(chrLocation) : null
       };
 
-      dispatch(resetBrowserTrackStates());
-
-      dispatch(replace(urlFor.browser(params)));
+      batch(() => {
+        dispatch(replace(urlFor.browser(params)));
+      });
```
azangru, line 145: Couple of things about batch:
1) it only makes sense with multiple dispatches in a row
2) even then, according to the docs, it only makes sense outside of React ("You can use it to ensure that multiple actions dispatched **outside of React** only result in a single render update", [link](https://react-redux.js.org/api/batch))

Bottom line: no need for the batch here.

msmanoj: Yes, I just removed one dispatch but forgot to remove batch. That's done now

azangru: Umm, it's still in the diff 😅

msmanoj: Just pushed it. Sorry!

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.tsx
```diff
@@ -95,10 +95,7 @@ export const TrackPanelListItem = (props: TrackPanelListItemProps) => {
   } = props;
 
   useEffect(() => {
-    const { defaultTrackStatus } = props;
-    if (trackStatus !== defaultTrackStatus) {
-      updateGenomeBrowser(trackStatus);
-    }
+    updateGenomeBrowser(trackStatus);
```
azangru, line 98: I completely forgot this code, so am rather confused here. Are we saying that every list item in the track panel sends individual message to update genome browser when it mounts? Isn't it something that the TrackPanel component has already done?

msmanoj: That's true! In which case this is completely redundant. It seems to work okay without this block.

azangru:
Well, the tests need to be fixed, but other than that — 👌

msmanoj:
>Well, the tests need to be fixed, but other than that — 👌

That's done

PR #331, opened 2020-07-24
Fix the view param

msmanoj:
This PR is to fix a small bug that was introduced in the following PR:
https://github.com/Ensembl/ensembl-client/pull/315

The view was always getting set as transcripts as the default value was defined in the wrong place.


msmanoj:
Thanks @azangru

PR #332, opened 2020-07-30
Refactor gene view state by splitting it into smaller slices

azangru:
## Type

- Bug fix
- Refactoring

## Related JIRA Issue(s)
None

## Description
There's a bug in dev (can be seen on internal deployment) which presents as follows if you click on a transcript in the transcripts view:

image

The problem is with [this piece of code](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/gene-view/entityViewerGeneViewReducer.ts#L59-L62): we are trying to access a slice of state indexing it first by activeGenomeId and then by activeObjectId; but the state at that point is an empty object, and trying to access state[activeGenomeId] returns undefined.

I used this opportunity to refactor entity viewer gene view state to the following shape, using `redux-toolkit`:

```
{
  view: {
    [genomeId]: {
      [geneId]: {
        current,
        selectedTabViews
      }
    }
  },
  transcripts: {
    [genomeId]: {
      [geneId]: {
        expandedIds: [],
        expandedDownloadIds: []
      }
    }
  },
  proteins: {
    [genomeId]: {
      [geneId]: {
        expandedTranscriptIds: []
      }
    }
  }
}
```

I hate the complexity that indexing by genome id and gene id brings; but don't know how to model the state better.

## Deployment URL
http://fix-gene-view-state.review.ensembl.org/

## Views affected
EntityViewer -> GeneView


src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
```diff
@@ -14,60 +14,82 @@
...
 
   dispatch(
-    updateGeneViewProteinsUIState({
+    proteinsSlice.actions.updateExpandedProteins({
       activeGenomeId,
       activeObjectId,
-      fragment: {
-        expandedProteinIds
-      }
+      expandedIds: [...expandedIds.values()]
```
msmanoj, line 67: Do we need to use spread here as I thought it will be taken care by Immer?

src/ensembl/src/content/app/entity-viewer/state/gene-view/view/geneViewViewSlice.ts
```diff
@@ -144,19 +118,93 @@ GeneViewTabMap.set(View.GENE_PANELS, {
...
 
-export const defaultEntityViewerGeneViewUIState: EntityViewerGeneViewUIState = {
-  view: View.TRANSCRIPTS,
+export type SelectedTabViews = Record<
+  'geneFunctionTab' | 'geneRelationshipsTab',
+  View | null
+>;
+
+export type ViewStatePerGene = {
+  current: View;
```
msmanoj, line 127: Why do you prefer `current` here instead of just `view`?

azangru: There are already so many `view`s in this model :smile: 

I named the field that holds this slice in `entityViewerGeneViewReducer` `view`:

```
export default combineReducers({
  view: geneViewViewReducer,
  transcripts: geneViewTranscriptsReducer,
  proteins: geneViewProteinsReducer
});
```

So when faced with the prospect of having this structure: `state.entityViewer.geneView.view[genomeId][geneId].view`, I thought that `state.entityViewer.geneView.view[genomeId][geneId].current` will be slightly better.

As you can see, I was unable to come up with a good name for what's now called `geneViewViewReducer` :disappointed:

ens-ap5: geneViewGeneReducer? That would be more consistent with the other names. Unless I've missed something and you're already using this name somewhere else.

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
azangru: This particular spread is to convert a set (used to ensure the uniqueness of ids) back to an array, which is then used in a reducer :smile: 

Also, there's no immer inside actions.

You'll find a more questionable spread in a reducer in `geneViewViewSlice.ts`:

```
      state[activeGenomeId][activeObjectId] = {
        ...state[activeGenomeId][activeObjectId],
        ...fragment
      };
```
Now, here Immer is available. The purpose of this spread is to ensure that the state at `state[activeGenomeId][activeObjectId]` is not replaced with the fragment from the action, but that the fragment is merged into the state. It's possible that `Object.assign(state[activeGenomeId][activeObjectId], fragment)` or lodash's `merge(state[activeGenomeId][activeObjectId], fragment)` would have worked here as well.

src/ensembl/src/content/app/entity-viewer/state/gene-view/view/geneViewViewSlice.ts
azangru: But this particular reducer isn't responsible for gene at all. It holds a very small piece of state that is currently responsible for which tab to show to the user: whether it's Transcripts, Gene Function or GeneRelationships, and whether within GeneFunction it's Proteins, Variants, Phenotypes, etc.

It's almost as if it is `geneViewTabsReducer`, but I was hesitant to call it this way. `geneViewUIReducer` would be too broad, because the `transcripts` and the `proteins` slices of this state are also responsible for the UI — they keep track of which transcripts/proteins are expanded or collapsed.

src/ensembl/src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSelectors.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!activeGenomeId || !activeObjectId) {
+    return;
+  }
+  return state.entityViewer.geneView.transcripts[activeGenomeId]?.[
+    activeObjectId
+  ];
+};
+
+export const getExpandedTranscriptIds = (state: RootState): string[] => {
+  const proteinsSlice = getSliceForGene(state);
```
msmanoj, line 39: `transcriptsSlice`?

azangru: 👍  and ✔️

PR #334, opened 2020-08-04
Improvement to Browser Messaging Service

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-707
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-708
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-709

## Description
This PR is the outcome of the discussions during the team meeting on 22 July 2020 regarding the browser messaging service.

## Deployment URL
TODO

### Other effects

## Possible complications
TODO

### DIFF
--- a/src/assets/browser/app/src/controller/global/activate.rs
+++ b/src/assets/browser/app/src/controller/global/activate.rs
@@ -14,9 +14,7 @@ impl EventListener<()> for ReadyPingPongListener {
             EventData::MessageEvent(_,_,c) => {
                 let data = &c.data().unwrap();
                 if data["type"] == "bpane-ready-query" {
-                    domutil::send_post_message("bpane-ready",&json!{{
-                        "action": "genome_browser_ready"
-                    }});
+                    domutil::send_post_message("bpane-ready",&json!{});
                 }
             },
             _ => ()
@@ -33,7 +31,5 @@ pub fn activate() {
     let pingpong = ReadyPingPongListener{};
     let mut ec = EventControl::new(Box::new(pingpong),());
     ec.add_event(EventType::MessageEvent);
-    domutil::send_post_message("bpane-ready",&json!{{{
-        "action": "genome_browser_ready"
-    }}});
+    domutil::send_post_message("bpane-ready",&json!{{}});
 }
--- a/src/assets/browser/app/src/controller/global/activate.rs
+++ b/src/assets/browser/app/src/controller/global/activate.rs
@@ -14,7 +14,9 @@ impl EventListener<()> for ReadyPingPongListener {
             EventData::MessageEvent(_,_,c) => {
                 let data = &c.data().unwrap();
                 if data["type"] == "bpane-ready-query" {
-                    domutil::send_post_message("bpane-ready",&json!{{}});
+                    domutil::send_post_message("bpane-ready",&json!{{
+                        "action": "genome_browser_ready"
+                    }});
                 }
             },
             _ => ()
--- a/src/assets/browser/app/src/controller/global/activate.rs
+++ b/src/assets/browser/app/src/controller/global/activate.rs
@@ -14,7 +14,9 @@ impl EventListener<()> for ReadyPingPongListener {
             EventData::MessageEvent(_,_,c) => {
                 let data = &c.data().unwrap();
                 if data["type"] == "bpane-ready-query" {
-                    domutil::send_post_message("bpane-ready",&json!{});
+                    domutil::send_post_message("bpane-ready",&json!{{
+                        "action": "genome_browser_ready"
+                    }});
                 }
             },
             _ => ()
@@ -31,5 +33,7 @@ pub fn activate() {
     let pingpong = ReadyPingPongListener{};
     let mut ec = EventControl::new(Box::new(pingpong),());
     ec.add_event(EventType::MessageEvent);
-    domutil::send_post_message("bpane-ready",&json!{{}});
+    domutil::send_post_message("bpane-ready",&json!{{
+        "action": "genome_browser_ready"
+    }});
 }
--- a/src/assets/browser/app/src/controller/global/activate.rs
+++ b/src/assets/browser/app/src/controller/global/activate.rs
@@ -33,5 +33,7 @@ pub fn activate() {
     let pingpong = ReadyPingPongListener{};
     let mut ec = EventControl::new(Box::new(pingpong),());
     ec.add_event(EventType::MessageEvent);
-    domutil::send_post_message("bpane-ready",&json!{{}});
+    domutil::send_post_message("bpane-ready",&json!{{{
+        "action": "genome_browser_ready"
+    }}});
 }
--- a/src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
@@ -21,7 +21,7 @@ import faker from 'faker';
 import { BrowserCogList } from './BrowserCogList';
 import BrowserCog from './BrowserCog';
 
-import browserMessagingService from 'src/content/app/browser/browser-messaging-service';
+import browserMessagingService from 'src/content/app/browser/services/browser-messaging-service/browser-messaging-service';
 
 describe('<BrowserCogList />', () => {
   afterEach(() => {
--- a/src/ensembl/src/content/app/browser/browser-image/BrowserImage.tsx
+++ b/src/ensembl/src/content/app/browser/browser-image/BrowserImage.tsx
@@ -46,7 +46,7 @@ import { changeHighlightedTrackId } from 'src/content/app/browser/track-panel/tr
 
 import {
   BrowserLocationUpdateMessage,
-  BrowserToChromeMessagingActions
+  BrowserToChromeMessagingAction
 } from 'src/content/app/browser/services/browser-messaging-service/browser-incoming-message-types';
 import { BrowserNavStates, ChrLocation, CogList } from '../browserState';
 import { RootState } from 'src/store';
@@ -106,7 +106,7 @@ export const BrowserImage = (props: BrowserImageProps) => {
 
   useEffect(() => {
     const subscription = browserMessagingService.subscribe(
-      BrowserToChromeMessagingActions.UPDATE_LOCATION,
+      BrowserToChromeMessagingAction.UPDATE_LOCATION,
       listenBpaneOut
     );
 
--- a/src/ensembl/src/content/app/browser/browser-image/BrowserImage.tsx
+++ b/src/ensembl/src/content/app/browser/browser-image/BrowserImage.tsx
@@ -25,8 +25,6 @@ import { CircleLoader } from 'src/shared/components/loader/Loader';
 import Overlay from 'src/shared/components/overlay/Overlay';
 
 import browserMessagingService from 'src/content/app/browser/services/browser-messaging-service/browser-messaging-service';
-import { parseFeatureId } from 'src/content/app/browser/browserHelper';
-import { buildEnsObjectId } from 'src/shared/state/ens-object/ensObjectHelpers';
 import {
   getBrowserCogTrackList,
   getBrowserNavOpened,
@@ -47,6 +45,7 @@ import {
 
 import { changeHighlightedTrackId } from 'src/content/app/browser/track-panel/trackPanelActions';
 
+import { BrowserToChromeMessagingActions } from 'src/content/app/browser/services/browser-messaging-service/browser-message-creator';
 import { BrowserNavStates, ChrLocation, CogList } from '../browserState';
 import { RootState } from 'src/store';
 import { BROWSER_CONTAINER_ID } from '../browser-constants';
@@ -70,9 +69,9 @@ export type BrowserImageProps = {
 };
 
 type BpaneOutPayload = {
+  action: BrowserToChromeMessagingActions.UPDATE_LOCATION;
   bumper?: BrowserNavStates;
-  focus?: string;
-  'message-counter'?: number;
+  _outgoing?: boolean;
   'intended-location'?: ChrLocation;
   'actual-location'?: ChrLocation;
   'is-focus-position'?: boolean;
@@ -88,11 +87,9 @@ const parseLocation = (location: ChrLocation) => {
 export const BrowserImage = (props: BrowserImageProps) => {
   const browserRef = useRef<HTMLDivElement>(null);
   const listenBpaneOut = useCallback((payload: BpaneOutPayload) => {
-    const ensObjectId = payload.focus;
     const navIconStates = payload.bumper as BrowserNavStates;
     const intendedLocation = payload['intended-location'];
     const actualLocation = payload['actual-location'] || intendedLocation;
-    const messageCount = payload['message-counter'];
     const isFocusObjectInDefaultPosition = payload['is-focus-position'];
 
     if (navIconStates) {
@@ -105,15 +102,6 @@ export const BrowserImage = (props: BrowserImageProps) => {
 
     if (actualLocation) {
       props.setActualChrLocation(parseLocation(actualLocation));
-    }
-
-    if (ensObjectId) {
-      const parsedId = parseFeatureId(ensObjectId);
-      props.updateBrowserActiveEnsObject(buildEnsObjectId(parsedId));
-    }
-
-    if (messageCount) {
-      props.updateMessageCounter(messageCount);
     }
 
     if (typeof isFocusObjectInDefaultPosition === 'boolean') {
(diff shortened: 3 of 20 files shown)


src/ensembl/src/content/app/browser/browser-messaging-service.ts
```diff
@@ -18,17 +18,103 @@ import windowService, {
...
+  MOVE_LEFT = 'move_left',
+  MOVE_RIGHT = 'move_right',
+  MOVE_DOWN = 'move_down',
+  MOVE_UP = 'move_up',
+  ZOOM_BY = 'zoom_by',
+  PING = 'ping',
+  ZMENU_ENTER = 'zmenu-enter',
+  ZMENU_LEAVE = 'zmenu-leave',
+  ZMENU_ACTIVITY_OUTSIDE = 'zmenu-activity-outside' // TODO: sometime later, unify underscores vs hyphens (together with Genome Browser)
+}
```
azangru, line 56: Let's not do all that in the browser-messaging-service file? Let's separate this into a separate `browser-message-creator` (or something) file that will be concerned only with the shape of the payload sent to the genome browser. Meanwhile the `browser-message-service` will only be concerned with actually sending and receiving messages.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -339,11 +366,14 @@ export const changeBrowserLocation: ActionCreator<ThunkAction<
...
-      stick: `${locationData.genomeId}:${chrCode}`,
-      goto: `${startBp}-${endBp}`,
-      'message-counter': messageCount,
-      ...focusInstruction
+    browserMessagingService.send({
+      action: ChromeToBrowserMessagingActions.SET_FOCUS_LOCATION,
+      payload: {
+        stick: `${locationData.genomeId}:${chrCode}`,
+        goto: `${startBp}-${endBp}`,
+        'message-counter': messageCount,
```
azangru, line 374: no need for message counter after https://github.com/Ensembl/ensembl-client/pull/333

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -210,9 +234,12 @@ export const restoreBrowserTrackStates: ActionCreator<ThunkAction<
...
 
-  browserMessagingService.send('bpane', {
-    off: tracksToTurnOff,
-    on: tracksToTurnOn
+  browserMessagingService.send({
+    action: ChromeToBrowserMessagingActions.TOGGLE_TRACKS,
+    payload: {
+      off: tracksToTurnOff,
+      on: tracksToTurnOn
+    }
```
azangru, line 242: As a possible api:

```
browserMessagingService.send(toggleTracksMessage({
  off: tracksToTurnOff,
  on: tracksToTurnOn
});
```

where `toggleTracksMessage` would be an action creator that will know that it wants to add `{action: ChromeToBrowserMessagingActions.TOGGLE_TRACKS}` and put whatever came in the arguments in the `payload` field.

Wouldn't it be nicer?

src/ensembl/src/content/app/browser/browser-messaging-service.ts
msmanoj: Yes, that's the actual plan. I'm just temporarily stacking it up in `browser-messaging-service`.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: Sounds good! Do you think it will be better to define these within `browser-message-creator` itself?

msmanoj: I noticed that most of the actions are used only once so I'm wondering if it would be simpler to change `browserMessagingService.send({action, payload})` to `browserMessagingService.send(action, payload)`?

azangru: > Sounds good! Do you think it will be better to define these within browser-message-creator itself?

As opposed to defining them in redux- or react-related files? Yes, I think so.

> wondering if it would be simpler to change browserMessagingService.send({action, payload}) to browserMessagingService.send(action, payload)

With `browserMessagingService.send({action, payload})` you'll be able to simply pass to `browserMessagingService.send` the output of a browser-message-creator function. With `browserMessagingService.send(action, payload)` you'll need to unwrap that output into two variables and pass them individually. I think  `browserMessagingService.send({action, payload})` looks more ergonomic. The `.send` is acting almost like redux's `dispatch`, which takes an action object rather than individual components of that object.

msmanoj: No, I was asking if we could just use `browserMessagingService.send(action, payload)` and avoid message-creator function. I am not sure if it is worth creating a message-creator function for each and every actions.

azangru: I may not have the proper appreciation of which way is going to be more convenient; but from looking at this PR it feels that encapsulating all browser message-related stuff in their own file/files might be tidier than creating action objects from scratch in redux files on react components.

I may be wrong though. Would love others' feedback during code review.

src/assets/browser/app/src/controller/global/activate.rs
```diff
@@ -14,7 +14,9 @@ impl EventListener<()> for ReadyPingPongListener {
             EventData::MessageEvent(_,_,c) => {
                 let data = &c.data().unwrap();
                 if data["type"] == "bpane-ready-query" {
-                    domutil::send_post_message("bpane-ready",&json!{{}});
+                    domutil::send_post_message("bpane-ready",&json!{{
```
azangru, line 17: Wouldn't it be `"bpane-out"` rather than `"bpane-ready"`?

msmanoj: I tried to change it to `bpane-out` but the problem is that every subscriber will now be notified about this message and they will all have to process the action to find out. For example, a message with type `bpane-ready` is sent only once so that `bpane-ready` subscriber gets notified only once. But it we change this to `bpane-out`, for every single message we receive from the browser, the `bpane-ready` subscriber will also be notified which is unnecessary. 

Considering this, I was wondering if it would be better to keep these types separated which gets used only once.

azangru: > every subscriber will now be notified

Can't browserMessagingService have a special case for bpane-out with action "genome_browser_ready" and just keep this message to itself without notifying subscribers when it sees it?

azangru: I mean, `bpane-ready` with action `"genome_browser_ready"` kinda doesn't make sense. We can leave the message type to be bpane-ready, but then there's no point adding a payload with an action to it. Or we could inspect the payload of bpane-out to decide whether to notify subscribers or not.

src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.test.tsx
```diff
@@ -21,7 +21,7 @@ import faker from 'faker';
 import { BrowserCogList } from './BrowserCogList';
 import BrowserCog from './BrowserCog';
 
-import browserMessagingService from 'src/content/app/browser/browser-messaging-service';
+import browserMessagingService from 'src/content/app/browser/services/browser-messaging-service/browser-messaging-service';
```
azangru, line 24: `/browser-messaging-service/browser-messaging-service` looks excessive. What if we add an `index.ts` file to the `browser-messaging-service` directory that would import and immediately export the `browserMessagingService` from the `browser-messaging-service` file? Then consumer modules would be able to import the service from `'src/content/app/browser/services/browser-messaging-service` (no repeated browser-messaging-service).

src/ensembl/src/content/app/browser/browser-image/BrowserImage.tsx
```diff
@@ -70,9 +69,9 @@ export type BrowserImageProps = {
 };
 
 type BpaneOutPayload = {
+  action: BrowserToChromeMessagingActions.UPDATE_LOCATION;
   bumper?: BrowserNavStates;
-  focus?: string;
-  'message-counter'?: number;
+  _outgoing?: boolean;
```
azangru, line 74: Do we know whether we need this property?

azangru:
Don't you want to add a types file describing the shape of incoming messages from genome browser? 



Currently, the types of incoming messages seem to be defined in the modules that consume them (e.g. in `BrowserCogList.tsx`).

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,210 @@
...
+
+import {
+  ZmenuLeavePayload,
+  ZmenuOutsideActivityPayload,
+  ZmenuEnterPayload
+} from 'src/content/app/browser/zmenu/zmenu-types';
+
+/*
+This is a service for communicating between genome browser and React wrapper.
+*/
```
azangru, line 25: Wrong file?

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,210 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ZmenuLeavePayload,
+  ZmenuOutsideActivityPayload,
+  ZmenuEnterPayload
+} from 'src/content/app/browser/zmenu/zmenu-types';
```
azangru, line 21: These types may want to migrate to the broweser-messaging-service folder now...

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,210 @@
...
+  ZMENU_ACTIVITY_OUTSIDE = 'zmenu-activity-outside' // TODO: sometime later, unify underscores vs hyphens (together with Genome Browser)
+}
+
+export type BrowserToggleTracksPayload = {
+  on?: string | string[];
+  off?: string | string[];
+};
+
+export type BrowserSetFocusPayload = {
+  'message-counter': number;
```
azangru, line 56: No more?

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,210 @@
...
+  focus?: string | undefined;
+};
+
+export type BrowserSetFocusLocationPayload = {
+  stick: string;
+  goto: string;
+  'message-counter': number;
```
azangru, line 63: No more?

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -364,10 +380,12 @@ export const changeFocusObject = (
...
   dispatch(updatePreviouslyViewedObjectsAndSave());
 
-  browserMessagingService.send('bpane', {
-    focus: objectId,
-    'message-counter': messageCount
-  });
+  browserMessagingService.send(
+    setFocusMessage({
+      focus: objectId,
+      'message-counter': messageCount
```
azangru, line 386: no need?

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,210 @@
...
+      action: ChromeToBrowserMessagingActions.ZOOM_BY;
+      payload: { zoom_by: number };
+    };
+
+export const toggleTracksMessage = (
+  payload: BrowserToggleTracksPayload
+): OutgoingPayload => {
+  return {
+    action: ChromeToBrowserMessagingActions.TOGGLE_TRACKS,
+    payload: { ...payload }
```
azangru, line 118: why the spread?

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-messaging-service.test.ts
```diff
@@ -15,16 +15,20 @@
  */
 
 import { BrowserMessagingService } from './browser-messaging-service';
+import { zmenuEnterMessage as dummyMessageCreator } from './browser-message-creator';
```
azangru, line 18: Aren't we abusing the `zmenuEnterMessage` here? Let's write the dummyMessageCreator in the test file itself.

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-messaging-service.ts
```diff
@@ -17,18 +17,21 @@
 import windowService, {
   WindowServiceInterface
 } from 'src/services/window-service';
+import { OutgoingPayload } from './browser-message-creator';
 
-import JSONValue from 'src/shared/types/JSON';
+import { ActivateBrowserPayload } from 'src/content/app/browser/browserActions';
```
azangru, line 22: Probably want to move the type definition to browser-messaging-service folder.

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-messaging-service.ts
```diff
@@ -17,18 +17,21 @@
...
 
-/*
-  This is a service for communicating between genome browser and React wrapper.
-*/
```
azangru, line 25: 🤷‍♂️

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-messaging-service.ts
```diff
@@ -70,11 +89,16 @@ export class BrowserMessagingService {
...
+      subscribers.forEach((subscriber: (payload: any) => void) =>
+        subscriber(payload)
+      );
     }
   };
 
-  public subscribe = (eventName: string, callback: Function) => {
+  public subscribe = (
+    eventName: string,
+    callback: (...args: any[]) => void
```
azangru, line 100: Sorry for the late realisation, @msmanoj, but, compared to what we have in dev (https://github.com/Ensembl/ensembl-client/blob/500e9a4f1969020d9e8290617850202421747c05/src/ensembl/src/content/app/browser/zmenu/ZmenuController.tsx#L49-L53), we are of course losing the power of placing subscribers only in those arrays that concern them, and, as you rightly pointed out, we'll be notifying all subscribers now.

I think we can do better. I think we can possibly use the `action` part of the payload as a key for `this.subscribers`. The question then becomes how to handle cases when a single function wants to subscribe to multiple actions — what the signature of the `subscribe` method should be. One possibility is to make it capable of handling both a string and an array of strings as the first argument. An additional possibility is to make it capable of handling wildcards, as done [here](https://sahadar.github.io/pubsub/index.html#subscribe-wildcard-a-name-subscribe_wildcard-a-); but that may be too much. Will sure be hell of a fun to type-check!

msmanoj: The required changes are made in the following commit:
https://github.com/Ensembl/ensembl-client/pull/334/commits/126cdc88fd33a53b48d985ba59dbad364dd36f3b

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-message-creator.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export type ZmenuOutsideActivityPayload = {
+  id: string;
+};
+
+export type ZmenuLeavePayload = {
+  id: string;
+};
+
+export type OutgoingPayload =
```
azangru, line 64: OutgoingMessage?

src/ensembl/src/content/app/browser/services/browser-messaging-service/browser-incoming-message-types.ts
```diff
@@ -0,0 +1,79 @@
...
+      track_y: CogList;
+    };
+
+export type BrowserLocationUpdatePayload = {
+  action: BrowserToChromeMessagingActions.UPDATE_LOCATION;
+  bumper?: BrowserNavStates;
+  'intended-location'?: ChrLocation;
+  'actual-location'?: ChrLocation;
+  'is-focus-position'?: boolean;
+};
```
azangru, line 53: Aha! We are seeing now that outgoing messages conform to the shape `{ action: string, payload: JSON }`, but the incoming messages don't. Isn't it something we'd want to fix?

msmanoj: I agree. I am not quite sure how easy it would be for me to make this change to the Rust code. I can give it a go.

ens-ma7:
waiting on https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-542

msmanoj:
Superseded by https://github.com/Ensembl/ensembl-client/pull/507

PR #335, opened 2020-08-13
Remove default text from transcript left column

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-693

## Description
Remove default text from the left column.
At the moment I am making the first one default. We may have to test/fix it once we have the sorting feature in place

http://remove-lhs-placeholder.review.ensembl.org/app/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000139618?view=transcripts

## Views affected
Transcript view

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
@@ -18,7 +18,3 @@
   justify-content: flex-end;
   align-items: baseline;
 }
-
-.helpText {
-  max-width: 300px;
-}
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
@@ -12,3 +12,13 @@
   color: $blue;
   cursor: pointer;
 }
+
+.defaultTranscriptLabel {
+  display: flex;
+  justify-content: flex-end;
+  align-items: baseline;
+}
+
+.helpText {
+  max-width: 300px;
+}
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
@@ -87,6 +87,6 @@ describe('<DefaultTranscriptListItem />', () => {
 
   it('displays selected transcript', () => {
     wrapper = renderComponent({ isDefault: true });
-    expect(wrapper.find('.defaultTranscriptLabel')).toHaveLength(1);
+    expect(wrapper.find('.defaultTranscriptLabel').text()).toBe('Selected');
   });
 });
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
@@ -84,4 +84,9 @@ describe('<DefaultTranscriptListItem />', () => {
 
     expect(wrapper.exists(TranscriptsListItemInfo)).toBe(true);
   });
+
+  it('displays selected transcript', () => {
+    wrapper = renderComponent({ isDefault: true });
+    expect(wrapper.find('.defaultTranscriptLabel')).toHaveLength(1);
+  });
 });
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
@@ -68,25 +68,19 @@ export const DefaultTranscriptListItem = (
   };
 
   const defaultTranscriptLabelMap = {
-    select: {
+    selected: {
       label: 'Selected',
-      helpText:
-        'The selected transcript is a default single transcript per protein coding gene that is representative of biology, well-supported, expressed and highly conserved'
-    },
-    plus: {
-      label: 'MANE Plus',
       helpText:
         'The selected transcript is a default single transcript per protein coding gene that is representative of biology, well-supported, expressed and highly conserved'
     }
   };
 
-  const isCanonical = props.isDefault || false;
-  const canonicalType = 'select'; // TODO Change this to transcript.mane/plus etc when available
+  const canonicalType = 'selected'; // TODO Change this to transcript.mane/plus etc when available
 
   return (
     <div className={styles.defaultTranscriptListItem}>
       <div className={transcriptsListStyles.row}>
-        {isCanonical && (
+        {props.isDefault && (
           <div className={styles.defaultTranscriptLabel}>
             {defaultTranscriptLabelMap[canonicalType]?.label}
             <QuestionButton
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
@@ -27,13 +27,22 @@ import { toggleTranscriptInfo } from 'src/content/app/entity-viewer/state/gene-v
 import { Gene } from 'src/content/app/entity-viewer/types/gene';
 import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
 import { TicksAndScale } from 'src/content/app/entity-viewer/gene-view/components/base-pairs-ruler/BasePairsRuler';
+import QuestionButton, {
+  QuestionButtonOption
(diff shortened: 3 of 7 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -33,6 +33,7 @@ import styles from './DefaultTranscriptListItem.scss';
 
 export type DefaultTranscriptListItemProps = {
   gene: Gene;
+  label: string;
```
azangru, line 36: Wouldn't `isDefault: boolean` be a better choice?

jyothishnt: Label can be Canonical/Mane Select etc. coming from Thoas

azangru: If so, won't you be able to get the actual label from the transcript that is being passed to the TranscriptListItem component?

If a TranscriptListItem component knows that it is the default one, it can look into the transcript object and find the relevant field to show the label.

jyothishnt: yes thats the end goal. but can we add in fields to the transcript object atm?

azangru: > yes thats the end goal

There are two ways I can see for showing the label:

1) Something outside the `DefaultTranscriptListItem` component decides why this particular item is the default one and what text it is going to show. This something will also have to decide what to show in the question button here.

image

2) Alternatively, the `DefaultTranscriptListItem` component has only to be notified from the outside that it happens to be the default component, and then it will itself decide what text to show in the label, and what text to show in the question button.

Option 2 looks simpler at the moment; so I suggest we pass the `isDefault: boolean` prop to the `DefaultTranscriptListItem`, which will show the hard-coded `"Selected"` string as a label if it is a default transcript.

Would love to hear what @msmanoj or @imransl think.

jyothishnt: Can there be a Selected and Canonical in the list of transcripts for a gene?

azangru: According to the [CDM](https://www.ebi.ac.uk/seqdb/confluence/display/ENS/Mini+hackathon+-+2020-08-21) (super rough draft, nothing certain), each transcript will have a `canonical: boolean` field; and one of the transcripts of a gene might also have a `mane: 'select'` field. Mane is available only for several species though.

ens-st3: 'Selected' is just a placeholder until we have real data. When we do then each gene will have one transcript that is 'Ensembl canonical', which as Andrey mentions will always be the 'MANE select' if there is one (human 38 only).

In 102 there are genes that have both a 'MANE Select' and a 'MANE Plus' transcript. I think we should be showing both of these on the transcript image. Do we need clarification of that @jyothishnt ?

jyothishnt: I think we need to code to show one or more titles based on what comes from the payload. Could be MANE Select or MANE Plus, or anything that comes in the future.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
```diff
@@ -12,3 +12,13 @@
...
 }
+
+.defaultTranscriptLabel {
+  display: flex;
+  justify-content: flex-end;
+  align-items: baseline;
+}
+
+.helpText {
+  max-width: 300px;
```
jyothishnt, line 23: I need to change this. But need help to override the help text css

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -129,7 +129,7 @@ export const TranscriptsListItemInfo = (
   };
   return (
     <div className={mainStyles}>
-      <div className={transcriptsListStyles.left}>bottom left</div>
+      <div className={transcriptsListStyles.left}></div>
```
azangru, line 132: What's happening in this div? What's its purpose?

jyothishnt: This div will be fed with data (see screenshot) once available.
Screenshot 2020-09-07 at 16 44 27

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
jyothishnt: this is no more required.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -57,10 +67,34 @@ export const DefaultTranscriptListItem = (
...
+    },
+    plus: {
+      label: 'MANE Plus',
+      helpText:
+        'The selected transcript is a default single transcript per protein coding gene that is representative of biology, well-supported, expressed and highly conserved'
+    }
+  };
+
+  const isCanonical = props.isDefault || false;
+  const canonicalType = 'select'; // TODO Change this to transcript.mane/plus etc when available
```
azangru, line 84: From code review:

1) We agreed that isCanonical is misleading, because only one transcript among all gene transcripts will be canonical, but several transcripts can have a label in the left-hand side of the component
2) We agreed not to try to predict the future, and to remove the reference to MANE Plus from this component. We will have to revisit this component anyway when we start getting real data from the backend.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
azangru: Delete it then? :wink:

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
```diff
@@ -84,4 +84,9 @@ describe('<DefaultTranscriptListItem />', () => {
 
     expect(wrapper.exists(TranscriptsListItemInfo)).toBe(true);
   });
+
+  it('displays selected transcript', () => {
+    wrapper = renderComponent({ isDefault: true });
+    expect(wrapper.find('.defaultTranscriptLabel')).toHaveLength(1);
```
azangru, line 90: Let's make this test slightly more interesting and make sure that the text content of `.defaultTranscriptLabel` is`Selected`.

jyothishnt: Does this mean we need to wrap text inside the div or is there another way in enzyme?

ens-st3: MANE Plus is here for 102, but we don't have a description for it yet.
We do have a description for MANE in OLS: https://www.ebi.ac.uk/ols/ontologies/ensemblglossary/terms?iri=http%3A%2F%2Fensembl.org%2Fglossary%2FENSGLOSSARY_0000365.
To not delay things should we 1) Create a JIRA to remind us to add MANE Plus when we get the data 2) Update the hardcoded description to that in the Glossary and create a JIRA to use OLS in the future.
Note there is potential for other attributes to be used here so we will eventually need to pick these up dynamically. How - still discussing that with HAVANA but CDM probably next step

jyothishnt: OK. But even if we add Mane Select and hardcode the description, we wont be able to make use of it until Thoas give us those. So yes, good idea to create JIRAs. I will do it.

azangru: > Does this mean we need to wrap text inside the div or is there another way in enzyme to check this?

Given the following markup (from DefaultTranscriptListItem.tsx):

```
          <div className={styles.defaultTranscriptLabel}>
            {defaultTranscriptLabelMap[canonicalType]?.label}
            <QuestionButton
              helpText={defaultTranscriptLabelMap[canonicalType]?.helpText}
              styleOption={QuestionButtonOption.INLINE}
            />
          </div>
```

you can do `expect(wrapper.find('.defaultTranscriptLabel').text()).toBe('Select');`

azangru (approved):

PR #336, opened 2020-08-16
Update typing of the browser messaging service

azangru:
## Type
- Proposal
- Bug fixes

## Description
This PR types the callback to the `subscribe` function properly, and makes sure that the callback will receive the payload associated with the appropriate action from the possible action/payload combinations.

The proper typing of the `subscribe` function helped find bugs in several subscribers (e.g. they were expecting the `action` to be a part of payload).

**Note:** we'll have to be less ambitious with the subscription function, which won't be able to subscribe a listener to an array of actions, but only to a single action at once. Given the way we are using the subscription, this should not be a problem for us.

## Demo
See [typescript playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nACgQxAGzsgJlAvFAbygDMEAuKAIlLkqgF8BuAKFEigCFkAnFdTHPiIAjHhUqjudJs1bho8OAFkIAZ1XIA5tCFRkAY2ABLOADsKAchoXGUMKgzYKivo5wy20LtxXqtOwj1DE3MoC0kbOwcBCm9XAQYWTyhfDW08WARU-ygAH04ebO1ZZKKIABV5DLKAbUoDYzNKAF0kqvjsJWQwDIJmKAGoGoBBKCNTFLU0ivlmigBRAA9gbgaAHjKAGkCGkIpR+gA+Ovt+bBaWehKqgGFkNDRRfQBrNdGIZYhTLFVJv21Kux8GVARBDhkABSnNwUDpYLpgEbNACUeHBADc4EYsCxmPozKpgFBVABXYSqfTcIzCCDcVSwnjGe5rABKEHx3CwGym-lB2wAyhBgGs7g8ns9DpLejJmMQSaZgmZiWSKVSaW8oB9gF8fn9pqDDhDdmZ9tt9PdHgZnhRRZaXm9Dqi+oMoPjTITiUKMqTyZTqbTVDVjaZmlAAPxhqCmCAAdygguAEORLBdqiFADpsFgIeaxVa9L9kKYQMmoAB6MtQcoACyMvzreig1attmEJKJyCg6NpICgAFtxnBuHpvmNTKY4PovkSzBB+oMfar-XSg4qQ96hZdZG6PdrCYpIdCYpkkNFsKjcOCd3A0BB0xhNFCz1h0zRk3iCUS98BvIfn7EeDhC8rwJG87wfJ8zhfSR32YRc-RpCErAQCxtm-RR33gtUICQiI0LUH8eEwlUEJw5C4FQqBv28UsK1dZBgH0as1Co5jNW4bghwAQjgkjsNwnhKPQhBaMrc1GOY35gDY2lOO4HigA) demonstrating a minimal working example of the `subscribe` function that is aware of the association between an action and the payload that is passed to the listener function.

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.tsx
+++ b/src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.tsx
@@ -21,7 +21,8 @@ import browserMessagingService from 'src/content/app/browser/services/browser-me
 import { toggleTracksMessage } from 'src/content/app/browser/services/browser-messaging-service/browser-message-creator';
 import {
   BrowserToChromeMessagingAction,
-  BrowserScrollPayload
+  CogScrollPayload,
+  CogTrackScrollPayload
 } from 'src/content/app/browser/services/browser-messaging-service/browser-incoming-message-types';
 
 import BrowserCog from './BrowserCog';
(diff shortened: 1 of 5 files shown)


src/ensembl/src/content/app/browser/browser-cog/BrowserCogList.tsx
```diff
@@ -57,29 +58,29 @@ type BrowserCogListProps = {
...
-      payload.action === BrowserToChromeMessagingAction.UPDATE_TRACK_POSITION
-    ) {
-      props.updateCogTrackList(payload.track_y);
-    }
+
+  const updateCogsPosition = (payload: CogScrollPayload) => {
+    props.updateCogList(payload.delta_y);
+  };
+
+  const updateCogTracksPosition = (payload: CogTrackScrollPayload) => {
```
msmanoj, line 66: Do we need to use `Cog` here?

azangru: Don't know. For whatever reason, the action that is triggered by this message is called `updateCogTrackList` 🤷‍♂️ 

I am equally unsure whether we want the word `Cog` in `CogScrollPayload` type. We are using it to move the cogs; but perhaps the better name for it would be `VerticalScrollPayload`. We need to find out exactly what these messages do, and what was the reason for using the word `cog` in the name of the `updateCogTrackList` action.

msmanoj: Cool, I see if I can get this changed in the branch I'm working on.

msmanoj:
@azangru the typescript playground link seems broken and also the tests are failing. Apart from these two, the PR looks good.

azangru:
@msmanoj I think I fixed the typescript playground link.

As for the tests, they are broken both in this branch and in  bms-improvement branch. I think we could fix them in the bms-improvement branch once we merge this branch in?

msmanoj (approved):

PR #338, opened 2020-08-23
Set up species page

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-742

## Description
- Add the skeleton species page
- The page can be accessed at `/species/:genome_id`
- If species has not yet been selected, request genome info for this species and add it to the list of selected species
- Remove `/app` from the url
- Remove the `Content` component. No point.
- Set up very basic redux state for the sidebar (open/close)

## Views affected
Species page


src/ensembl/src/content/app/species/state/speciesSidebarSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+  reducers: {
+    toggleSidebar(state) {
+      state.isOpen = !state.isOpen;
+    }
+  }
+});
+
+export const { toggleSidebar } = speciesPageSidebarSlice.actions;
+
+export const isSidebarOpen = (state: RootState) =>
```
msmanoj, line 41: Do we not put these in a separate selector file?

src/ensembl/src/content/app/App.tsx
```diff
@@ -14,99 +14,80 @@
...
-    // remove /app/ from url to get app name
-    let appName = props.location.pathname.replace('/app/', '');
-
-    // check if app name still has forward slash (/) to be sure the app name is extracted
-    // if it isn't then remove rest of the URL and extract the app name
-    if (appName.indexOf('/') > -1) {
-      const matches = appName.match(/^[^\/]*/);
-      appName = matches ? matches[0] : '';
-    }
+    const appName: string = location.pathname.split('/').filter(Boolean)[0];
```
msmanoj, line 52: Do we need `filter(Boolean)` here?

azangru: `"/foo/bar".split('/')` returns `["", "foo", "bar"]`. I have not checked whether `location.pathname` will start with a slash, but I expect it will.

On a different note, I doubt we need to keep the app name in redux at all — I don't remember us doing anything useful with it. If needed, the app name can trivially be derived from the url. It's probably worth exploring whether we want to keep this function at all — but not in this PR.

src/ensembl/src/content/app/species/state/speciesSidebarSlice.ts
azangru: I don't have a particularly strong opinion on this. Redux-toolkit [says](https://redux-toolkit.js.org/usage/usage-guide#creating-slices-of-state) that it favours the "duck" pattern when all redux logic is co-located in a single file:

image

but at the same time, they don't include selectors in their example. I thought, if the slice is simple, selectors could be put in the same file; but I can move them in a separate file if it helps.

The sidebar slice is very simple at the moment. In EntityViewer, sidebar reducer is keyed by species id. I didn't do it here, hoping that we won't need it, but I suspect that we will...

msmanoj: I agree that it is simple enough to leave it here but surely we will be moving it later on when there are more selectors.
I am just thinking that it will be simpler at that point rather than having to deal the move and updating the paths later on.

azangru: Moved the selector to a separate file.

msmanoj (approved):

PR #339, opened 2020-08-24
Update dependencies

azangru:
## Type
- maintenance

## Description
Routine update of dependencies. Required a lot of mechanical changes of storybook files.

### List of notable changes
- Major version change for Typescript (v3 -> v4)
- Major version change for Storybook (v5 -> v6)
  - removed typescript from custom webpack config (Storybook now has zero-config typescript support)
  - changed stories from using the `storiesOf` function to the "component story format" (considered best practice now)
  - changed the way we were using `addon-actions` to the one recommended by the docs (i.e. using the `args` api)
  - removed `addon-notes` (it's deprecated)
- Major version change for `postcss-loader`
  - `postcss` now has to be an independent peer dependency
  - syntax for postcss-loader options [has changed](https://github.com/webpack-contrib/postcss-loader/blob/master/CHANGELOG.md#notes)
- Major version change for D3 (v5 -> v6) — upgrade went smoothly and didn't require any code changes

## Checks
- `npm run check-types` — ✔️ 
- `npm test` — ✔️
- `npm audit fix` — 0 vulnerabilities
- development build — 👌 
- production build — 👌


msmanoj:
I'm not sure if it is related to this PR but UnsplicedTranscripts looks a bit broken:
Screenshot 2020-09-10 at 14 36 10

src/ensembl/stories/shared-components/autosuggest-search-field/AutosuggestSearchField.stories.tsx
```diff
@@ -88,67 +91,80 @@ const Wrapper = (props: any) => {
...
+};
+
+export default {
+  title: 'Components/Shared Components/AutosuggestSearchField',
+  parameters: {
+    notes
+  },
+  argTypes: {
+    onSelect: { action: 'selected' },
+    onSubmit: { action: 'submitted' }
```
msmanoj, line 168: I get JS error when I try to use this component. I guess these should be `onSelected` & `onSubmitted`.

imransl (commented):
Found out some issues with links to apps like species selector and download manager. I can't recall if we opted to stop using `/app` for apps. It seems to be the problem. Also there are a dozen lint errors, though don't think this is breaking anything.

Other than these issues, I am happy with PR as the updated packages aren't breaking anything.

azangru:
> Found out some issues with links to apps like species selector and download manager. I can't recall if we opted to stop using /app for apps.

@imransl We did. Can you remember where the broken links are?

Doesn't look like it's related to this PR, so we can fix them in a separate one. I'll soon create one that changes the `/browser` path to `/genome-browser`; we can address the broken links problems there.

msmanoj:
> > I'm not sure if it is related to this PR but UnsplicedTranscripts looks a bit broken:
> 
> @msmanoj yes, it already is this way in the [current deployment](https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-entityviewer-transcripts--unspliced) of the storybook, so this regression wasn't introduced in this PR.

If you notice closely, the component seems to expand beyond the wrapper:

Screenshot 2020-09-10 at 15 40 03

Currently on live:
Screenshot 2020-09-10 at 15 41 28

src/ensembl/stories/shared-components/autosuggest-search-field/AutosuggestSearchField.stories.tsx
azangru: Thanks for spotting this! 👍 Should be fixed now.

azangru:
@imransl I think I have addressed your concern regarding routes still containing the `/app/` segment in this PR: https://github.com/Ensembl/ensembl-client/pull/345

msmanoj:
>I actually wonder whether we need this story anymore (this, and the one about proteins). It helped us develop the respective components in isolation, before any backend was ready; but since we have already moved these components into EntityViewer, and since they are non-repeating, maybe there isn't any point for keeping them in the storybook.

I don't have a strong opinion on this but it is nice to see all these components in an isolated view.

msmanoj (approved):

imransl:
> @imransl I think I have addressed your concern regarding routes still containing the `/app/` segment in this PR: #345

I've already approved that PR :simple_smile:

PR #340, opened 2020-08-25
Add link to protein id in the transcript accordion

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-723

## Description
Clicking on the protein id in the transcript accordion will now open the protein view, automatically expands the selected protein and scroll it into the view.
- Added loaders to both ProteinList & ProteinListItemInfo
- Added the functionality to expand the first transcript by default in both transcript and protein view
- Added the functionality to retain the scroll position when revisiting the transcript and protein view.
- Created `useRestoreScrollPosition` hook

## Deployment URL
http://link-protein-id.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000139618?view=transcripts

## Views affected
Entity viewer -> Transcripts, Proteins

### Other effects


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -123,6 +128,20 @@ export const TranscriptsListItemInfo = (
...
+    props.expandProtein(transcriptId);
+
+    const url = urlFor.entityViewer({
+      genomeId,
+      entityId,
+      view: View.PROTEIN
+    });
+
+    props.push(url);
+  };
```
azangru, line 143: Looks like we could just use the `Link` component?

msmanoj: >Regarding the expandProtein part, do we possibly want to add the focus protein to the url?

I thought the same but I am not sure when we need to update the focus protein. Do we do it on mouseover?  
Also, if we don't update it and if the user revisits the proteins list, the focus will always be the same.

azangru: We may set just a unidirectional logic:

- if there is a focus-protein parameter in the url when the view is protein,
- then expand corresponding protein (if it exists)
- and scroll to the expanded protein

Not sure what should happen when user clicks on another protein. They should be allowed to expand multiple proteins.

Hmm... Maybe having focus protein id in the url is not such a good idea after all.

msmanoj: I did think about this approach of having the focus parameter in the urls but I was wondering if we could simply scroll to the id that was most recently added to `expandedTranscriptIds`. With this approach, when the user clicks on a protein id, it will get added to the `expandedTranscriptIds` and will become the most recently added id. We will of course need to handle clicking on an already expanded id in which case we will have to pop it and add to the end.

azangru: > if we could simply scroll to the id that was most recently added to expandedTranscriptIds

There are potential problems with this approach:
- I am not sure we even need to apply programmatic scrolling when the user is already on the protein page and is clicking on proteins to expand or collapse them. In this case, the user already sees the protein he is interacting with.
- How are we planning to handle page refreshes? Also, do we plan to be able to share the url with a particular expanded protein, or do we give up on this idea?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.scss
```diff
@@ -1,4 +1,13 @@
 .proteinsList {
   font-size: 13px;
   font-weight: 300;
+  overflow: auto;
+}
+
+.proteinsListLoadingContainer {
+  height: 100%;
```
azangru, line 8: Given this:

image

what's the purpose of height: 100% and flex-direction: column?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -123,6 +126,29 @@ export const TranscriptsListItemInfo = (
     objectId: props.gene.id
   });
 
+  const buildUrlForProteinView = (transcriptId: string) => {
```
azangru, line 129: the function says "buildUrl", but what it's building is a link element.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.scss
```diff
@@ -1,4 +1,13 @@
 .proteinsList {
   font-size: 13px;
   font-weight: 300;
+  overflow: auto;
```
azangru, line 4: why?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -37,3 +37,11 @@
 .proteinFeaturesCountWrapper {
   display: inline-block;
 }
+
+.loadingContainer {
+  height: 100%;
```
azangru, line 42: Given this:

image

what's the purpose of height: 100% and flex-direction: column?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
```diff
@@ -39,13 +43,47 @@ type Props = {
...
   const midStyles = classNames(transcriptsListStyles.middle, styles.middle);
 
+  const itemRef = useRef<HTMLDivElement | null>(null);
+  useEffect(() => {
+    if (transcript.id === transcriptIdToFocus) {
+      // give it some time for other expanded proteiens to load
+      setTimeout(() => {
+        itemRef.current?.scrollIntoView({
+          behavior: 'smooth',
+          block: 'start'
```
azangru, line 73: "start" is the default

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
```diff
@@ -39,13 +43,47 @@ type Props = {
...
+    props.toggleExpandedProtein(transcript.id);
+  };
 
   const midStyles = classNames(transcriptsListStyles.middle, styles.middle);
 
+  const itemRef = useRef<HTMLDivElement | null>(null);
+  useEffect(() => {
+    if (transcript.id === transcriptIdToFocus) {
+      // give it some time for other expanded proteiens to load
+      setTimeout(() => {
```
azangru, line 70: I tried running it without the timeout, and it worked fine. At the point when `ProteinsListItem` has mounted, other `ProteinsListItem`s should exist as well. Is the timeout really necessary for anything?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.scss
```diff
@@ -16,3 +16,9 @@
 .transcriptId {
   font-weight: 400;
 }
+
+
+.scrollRef{
+  position: absolute;
+  top: -55px;
+}
```
azangru, line 24: CSS rule `scroll-margin: 55px;` (or `scroll-margin-top: 55px`) applied to the ProteinsListItem div should make this span unnecessary. It currently has problems on Safari, but Safari will catch up sooner or later.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.scss
msmanoj: I copied it from here:
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.scss#L5

I have removed height and flex-direction as they don't make any difference.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
msmanoj: Renamed it to `getLinkToProteinView`

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.scss
msmanoj: I was playing around with overflow before the layout update PR got merged and I missed to clean this up.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
msmanoj: Removed them

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
msmanoj: Removed it

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
msmanoj: This was causing some issue when the newly focussed protein needs to expand along with the proteins that were already expanded previously. Now that we expand only the focussed protein, it will not be an issue anymore. I'll get it removed.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.scss
msmanoj: That's done

azangru:
Note from code review:
- return the setTimeout (prob 100ms?) before scrolling the transcript into view
- `transcriptId` -> `transcript_id` in url query

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -179,6 +191,7 @@ const useGeneViewRouting = () => {
   const { search } = useLocation();
   // TODO: discuss – is using URLSearchParams better than using the querystring package?
   const view = new URLSearchParams(search).get('view');
+  const transcriptId = new URLSearchParams(search).get('transcriptId');
```
azangru, line 194: shouldn't need to create `new URLSearchParams(search)` twice.  `urlSearchParams = new URLSearchParams(search); const view = urlSearchParams.get('view'); const transcriptId = urlSearchParams.get('transcript_id)`

msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -83,25 +83,20 @@ export const TranscriptsListItemInfo = (
     objectId: props.gene.unversioned_stable_id
   });
 
-  const getLinkToProteinView = (transcriptId: string) => {
-    if (transcript.so_term !== 'protein_coding') {
-      // TODO: Use Protein ID
-      return transcriptId;
-    }
+  const getLinkToProteinView = (stableId: string) => {
```
azangru, line 86: `stableId` => `proteinStableId` (just to avoid confusion)?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -83,25 +83,20 @@ export const TranscriptsListItemInfo = (
...
     const proteinViewUrl = urlFor.entityViewer({
       genomeId,
       entityId,
       view: View.PROTEIN,
-      transcriptId
+      transcriptId: stableId
```
azangru, line 91: So, here, for example, I am already confused. Why is the `transcriptId` field receiving the protein id?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
```diff
@@ -60,22 +61,21 @@ const ProteinsListItem = (props: Props) => {
...
+    if (product.stable_id === transcriptIdToFocus) {
       setTimeout(() => {
         itemRef.current?.scrollIntoView({
           behavior: 'smooth'
         });
       }, 100);
-
       if (!props.expandedTranscriptIds.includes(transcriptIdToFocus)) {
-        props.toggleExpandedProtein(transcript.id);
+        props.toggleExpandedProtein(product.stable_id);
```
azangru, line 78: Here, again, something that the code calls transcript id is compared to protein id.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
msmanoj: That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
msmanoj: Here and below, I have changed it all to proteinId. I was planning to update it today as it was getting late yesterday.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -35,27 +35,30 @@ import { buildFocusIdForUrl } from 'src/shared/state/ens-object/ensObjectHelpers
...
 import transcriptsListStyles from '../DefaultTranscriptsList.scss';
 import styles from './TranscriptsListItemInfo.scss';
+import { View } from 'src/content/app/entity-viewer/state/gene-view/view/geneViewViewSlice';
 
 export type TranscriptsListItemInfoProps = {
   gene: Gene;
   transcript: Transcript;
   expandDownload: boolean;
   toggleTranscriptDownload: (id: string) => void;
+  clearExpandedProteins: (id: string) => void;
```
azangru, line 53: `onProteinLinkClick` perhaps? It just looks so wrong that TranscriptListItemInfo should know something about proteins that are expanded in a completely different view.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.tsx
```diff
@@ -24,33 +27,68 @@ import {
...
   return (
     <div className={styles.proteinsList}>
-      {proteinCodingTranscripts.map((transcript) => (
-        <ProteinsListItem
-          key={transcript.stable_id}
-          transcript={transcript}
-          trackLength={longestProteinLength}
-        />
-      ))}
+      {proteinCodingTranscripts ? (
```
azangru, line 69: According to line 49 above (`proteinCodingTranscripts = sortedTranscripts.filter ...`), `proteinCodingTranscripts` are guaranteed to be at very least an empty array, which makes them always truthy; so the CircleLoader below isn't going to be rendered.

azangru:
@msmanoj I pushed a commit that updated the function name (I didn't mean to rename the redux action, sorry for the confusion), fixed type errors (there were some) and added a test. Please let me know if you are happy with it.

msmanoj:
> I didn't mean to rename the redux action, sorry for the confusion

Oops! sorry, it's just me being silly there by changing the redux action. Thanks for putting it back. The changes looks good 👍

azangru:
@msmanoj I am seeing both vertical and horizontal scroll when I navigate from transcripts to proteins. is this intentional?

scroll

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -64,19 +65,26 @@ const ProteinsListItemInfo = (props: Props) => {
...
+        }
+      }
+    );
 
-    Promise.all([
-      fetchProteinDomains(proteinId, abortController.signal),
-      fetchProteinSummary(proteinId, abortController.signal)
-    ]).then(([proteinDomains, proteinSummaryData]) => {
+    fetchProteinSummary(
+      transcript.unversioned_stable_id,
```
azangru, line 84: `fetchProteinSummary` function now expects protein id. Changed in this PR: https://github.com/Ensembl/ensembl-client/commit/29b8f1cd5e83ac0bf63879b8d2ce3f1f86809bd7. Please check how the id should be passed to this function.

msmanoj: That's done

msmanoj:
> @msmanoj I am seeing both vertical and horizontal scroll when I navigate from transcripts to proteins. is this intentional?
> 
> scroll

No, it's not intentional. `scrollIntoView` seems to be doing it automatically. I've fixed it now.

msmanoj:
> We have a problem: `fetchProteinSummary` function can return null (it is actually returning null in my dev environment, ~I wonder why~ I left a comment explaining why). In this case, the spinner will remain visible indefinitely. We either need to hide it if the response is empty, or show the error/warning message. Don't know how much of this logic belongs in this PR, but eternal spinner is probably bad.

I've put a temporary fix for this now. Will look into it further when I work on https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-727.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -60,54 +61,66 @@ const ProteinsListItemInfo = (props: Props) => {
...
+        <>
+          <ProteinDomainImage
+            proteinDomains={product.protein_domains}
+            trackLength={trackLength}
+            width={695}
+          />
+          <ProteinImage
+            product={product}
+            trackLength={trackLength}
+            width={695}
```
azangru, line 115: As a future todo: we have a lot of those magic `695`s floating around, both in typescript files and in scss files. Would be great if we could just use a single constant importable both in the script and the style files, like we did with colours for our palette.

Doesn't block this PR, but we should clean this up.

src/ensembl/src/shared/hooks/useRestoreScrollPosition.ts
```diff
@@ -0,0 +1,66 @@
...
+  const scrollPosition = useSelector(getScrollPosition)[referenceId] || {
+    scrollTop: 0,
+    scrollLeft: 0
+  };
+
+  const targetElementRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const targetElement = targetElementRef.current as HTMLDivElement;
+    // TODO: Need to find out why it doesn't work without the setTimeout
```
azangru, line 42: Is this comment still relevant?

azangru (approved):

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
msmanoj: Created a ticket to clean it up:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-800

PR #342, opened 2020-08-27
Add generic species stats component

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-745

## Description
This PR is to create a generic component to display the species stats in the species page.

XD:
https://xd.adobe.com/view/8d08bee6-405e-41d6-4a7f-0496e3f486f8-fb39/screen/92677847-0fa3-472c-a451-f462a7cb791e?fullscreen

## Deployment URL
Please checkout to this branch and run storybook locally to view it.

## Views affected
Storybook

### DIFF
--- a/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
+++ b/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
@@ -8,7 +8,7 @@ $secondary-text-color: $dark-grey;
   vertical-align: top;
 }
 
-.title {
+.preLabel {
   color: $secondary-text-color;
   font-size: 11px;
 }
@@ -17,12 +17,6 @@ $secondary-text-color: $dark-grey;
   font-size: 12px;
   margin-top: -3px;
   font-weight: $bold;
-}
-
-.labelHint {
-  color: $secondary-text-color;
-  font-size: 11px;
-  padding-left: 3px;
 }
 
 .primaryValue {
--- a/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
+++ b/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
@@ -1,0 +1,51 @@
+@import 'src/styles/common';
+
+$secondary-text-color: $dark-grey;
+
+.wrapper {
+  width: 250px;
+  display: inline-block;
+  vertical-align: top;
+}
+
+.title {
+  color: $secondary-text-color;
+  font-size: 11px;
+}
+
+.label {
+  font-size: 12px;
+  margin-top: -3px;
+  font-weight: $bold;
+}
+
+.labelHint {
+  color: $secondary-text-color;
+  font-size: 11px;
+  padding-left: 3px;
+}
+
+.primaryValue {
+  font-size: 22px;
+  padding-left: 30px;
+}
+.primaryUnit {
+  color: $secondary-text-color;
+  font-size: 13px;
+  padding-left: 5px;
+}
+.secondaryValue {
+  font-size: 12px;
+  padding-left: 30px;
+}
+.secondaryUnit {
+  color: $secondary-text-color;
+  font-size: 11px;
+  padding-left: 3px;
+}
+.link {
+  text-align: center;
+  padding-left: 30px;
+  color: $secondary-text-color;
+  font-size: 11px;
+}
--- a/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
+++ b/src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
@@ -19,20 +19,26 @@ import classNames from 'classnames';
 
 import defaultStyles from './SpeciesStats.scss';
 
-export type SpeciesStatsProps = {
-  title?: string;
-  label: string;
-  labelHint?: string;
+type PrimaryDataProps = {
   primaryValue: string;
   primaryUnit?: string;
-  secondaryValue?: string;
+};
+
+type PropsWithSecondaryData = {
+  secondaryValue: string;
   secondaryUnit?: string;
-  link?: React.ReactElement;
+};
+
+type PropsWithoutSecondaryData = {
+  secondaryValue?: never;
+  secondaryUnit?: never;
+};
+
+type ClassNamesProps = {
   classNames?: {
     wrapper?: string;
-    title?: string;
+    preLabel?: string;
     label?: string;
-    labelHint?: string;
     primaryValue?: string;
     primaryUnit?: string;
     secondaryValue?: string;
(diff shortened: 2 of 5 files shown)


src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
```diff
@@ -0,0 +1,102 @@
...
+    title?: string;
+    label?: string;
+    labelHint?: string;
+    primaryValue?: string;
+    primaryUnit?: string;
+    secondaryValue?: string;
+    secondaryUnit?: string;
+    link?: string;
+  };
+};
```
azangru, line 42: I have an alternative proposal. What if instead of passing all the props to a single component and making it figure it all out, we split the component into several smaller components.

For example, here are components with three types of labels (the second type Andrea is likely to retire, but notice, too, how weird it is, with the label growing upwards instead of pushing the main content down):

image

```jsx
// focusing just on the label

// left example
<SpeciesStats>
  <SpeciesStatsLabel>
    Coding genes
  </SpeciesStatsLabel>
</SpeciesStats>

// middle example
<SpeciesStats>
  <SpeciesStatsLabel2
    line1="No. genes in"
    line2="Gene families"
  />
</SpeciesStats>

// or possibly:
<SpeciesStats>
  <SpeciesStatsLabel>
    <span className="light">No. genes in</span><br/>
    <span className="bold">Gene families</span><br/>
  </SpeciesStatsLabel>
</SpeciesStats>

// right example
<SpeciesStats>
  <SpeciesStatsLabel>
    SNVs <span className="small">single nucleotide variants<span>
  </SpeciesStatsLabel>
</SpeciesStats>
```

The main area:

image

```jsx

// left
<SpeciesStats>
  <SpeciesStatsLabel>
    Total sequence length
  </SpeciesStatsLabel>
  <SpeciesStatsMain>
    4,537,931,177 <span className="small">bp</span>
  </SpeciesStatsLabel>
</SpeciesStats>

// middle
<SpeciesStats>
  <SpeciesStatsLabel>
    Repeat regions
  </SpeciesStatsLabel>
  <SpeciesStatsMain>
    30%
  </SpeciesStatsLabel>
</SpeciesStats>

// right
<SpeciesStats>
  <SpeciesStatsMain>
    23 <span className="small">chromosomes or plasmids</span>
  </SpeciesStatsLabel>
</SpeciesStats>
```

and all together now:

image

```jsx

// left
<SpeciesStats>
  <SpeciesStatsLabel>
    Coding genes
  </SpeciesStatsLabel>
  <SpeciesStatsMain>
    20,438
  </SpeciesStatsLabel>
  <SpeciesStatsSecondary>
    671 <span className="small">readthrough</span>
  </SpeciesStatsSecondary>
</SpeciesStats>

// right
<SpeciesStats>
  <SpeciesStatsLabel>
    Homology data
  </SpeciesStatsLabel>
  <SpeciesStatsMain>
    96%
  </SpeciesStatsLabel>
  <SpeciesStatsSecondary>
    <Link to="/some/path">view</Link>
  </SpeciesStatsSecondary>
</SpeciesStats>
```

I think this offers a higher flexibility and less confusion that a single component with numerous optional props might cause.

msmanoj: The main reason that convinced me to create a single component that all the stats share a similar structure. I am sure how breaking it down to smaller components like in the example above is going to be different from the original plan of just having the divs and classnames. 

Also, with the above example we have to import all the components and then manually build each stat based on it's structure.

azangru: > all these stats share a similar structure

But notice how many nulls/undefineds/optional_fields you have in the props type for this component. Notice also how many different props there are. It suggests that the components may not have such a similar structure after all.

There are several talks addressing these issues. One (although very imperfect, with metaphors that do not quite work), is [React is Fiction](https://www.youtube.com/watch?v=kqh4lz2Lkzs); another, much better, is [AHA programming](https://youtu.be/wuVy7rwkCfc), by Kent C. Dodds. The general feeling of the community is that there's something not quite right with this approach:

image

> I am not sure how breaking it down to smaller components like in the example above is going to be different from the original plan of just having the divs and classnames.

It will capture common styling in the stats component and its subcomponents. Compared to the option of just using classnames of the parent component, it offers some degree of encapsulation of the styles.

msmanoj: Should we really consider 9 props as too many? We already have input and textarea component that takes more than 9 pros. Almost all the components that are available in Ant design has several optional props.

If we go with your suggestion, how do we represent an array of individual json stats without having to hardcode the structure for each stat?

azangru: Decisions after code review:

- go ahead with the component structure proposed in the PR
- rename some of the props:
  - `title` -> `preLabel`
  - `labelHint` - remove

msmanoj: That's done

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <div className={styles.label}>{props.label}</div>
+
+      <div>
+        <span className={styles.primaryValue}>{props.primaryValue}</span>
+        {props.primaryUnit && (
+          <span className={styles.primaryUnit}>{props.primaryUnit}</span>
+        )}
+      </div>
+
+      {(props.secondaryValue || props.secondaryUnit) && (
```
azangru, line 95: Why the `|| props.secondaryUnit` part of the condition? According to the `PropsWithSecondaryData` type, we can't have `secondaryUnit` without `secondaryValue`; so check for just `secondaryValue` should be sufficient. Unless our assumptions and the type are wrong.

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
```diff
@@ -0,0 +1,45 @@
+@import 'src/styles/common';
+
+$secondary-text-color: $dark-grey;
+
+.wrapper {
+  width: 250px;
```
azangru, line 6: Wondering whether there is any benefit to setting the width and the display: inline-block on these elements, or whether the grid of the parent element will provide those. I would probably just go with the grid of the parent. But can't think of a convincing argument one way or the other.

azangru (approved):

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
msmanoj: As discussed during the sprint review, we will look into this when we start using this component and change it to control the the width using grid of the parent if necessary.

PR #343, opened 2020-09-03
Design review changes

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature
- [x ] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-741

## Description
Based on design review changes reported here https://docs.google.com/spreadsheets/d/1y3ThUJbbW2_7tjIchmrvNn3v3AK539d5aGyARY7jNeA/edit#gid=1244768304

## Deployment URL
Will update soon

## Views affected
Entity Viewer

## Possible complications
_constants.css have been moved to another location. This is used by the sorting @ens-ma7 . You may want to check if everything is alright.

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/_constants.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/_constants.scss
@@ -1,4 +1,5 @@
-$left_column: 178px;
+$left_column: 160px;
+$middle_column_left_gap: 18px;
 $middle_column: 695px;
 $middle_column_right_gap: 27px;
 $right_column: 180px;
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -1,14 +1,15 @@
 @import 'src/styles/common';
-@import 'constants';
+@import '../../constants';
 
 .header {
   border-bottom: 1px solid $orange;
   padding: 0;
   position: sticky;
-  top: -15px;
+  top: -5px;
   background-color: $black;
   z-index: 1;
   font-size: 12px;
+  font-weight: 300;
 }
 
 .headerChevron {
@@ -23,13 +24,14 @@
 .content {
   position: relative;
   padding-top: 36px;
+  padding-bottom: 120px;
 }
 
 .stripedBackground {
   position: absolute;
   top: 0;
   bottom: 0;
-  left: $left_column;
+  left: $left_column + $middle_column_left_gap;
 
   .stripe {
     position: absolute;
@@ -43,7 +45,7 @@
 
 .row {
   display: grid;
-  grid-template-columns: $left_column $middle_column $middle_column_right_gap $right_column;
+  grid-template-columns: $left_column $middle_column_left_gap $middle_column $middle_column_right_gap $right_column;
   padding: 9px 0px;
   font-size: 12px;
 }
@@ -52,16 +54,17 @@
   grid-column: 1/2;
   text-align: right;
   padding-right: 1em;
+  color: $dark-blue;
 }
 
 .middle {
-  grid-column: 2/3;
+  grid-column: 3/4;
   position: relative;
   top: 2px;
 }
 
 .right {
-  grid-column: 4/5;
+  grid-column: 5/6;
 }
 
 .filterLabel {
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
@@ -1,5 +1,5 @@
 @import 'src/styles/settings';
-@import '../constants';
+@import '../../../constants';
 
 .defaultTranscriptListItem {
   svg rect {
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
@@ -1,9 +1,19 @@
+@import 'src/styles/common';
+
 .proteinSummary {
   margin-left: 178px;
   width: 695px;
 
   > div {
     padding: 10px 0 10px 0;
+  }
+
+  .keyline {
+    border-bottom: 1px solid #ff9900;
+    width: 695px;
+    padding: 0;
+    margin-top: 40px;
+    margin-bottom: 20px;
   }
 }
 
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -1,9 +1,9 @@
 @import 'src/styles/common';
-@import '../default-transcripts-list/constants';
+@import '../../constants';
 
 .container {
   display: grid;
-  grid-template-columns: $left_column auto;
+  grid-template-columns: $left_column $middle_column_left_gap auto;
   padding: 9px 0;
 }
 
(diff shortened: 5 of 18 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/_constants.scss
```diff
@@ -1,4 +1,5 @@
-$left_column: 178px;
```
azangru, line 1: Maybe create a `styles` folder at this level to put the constants file in?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -1,14 +1,15 @@
 @import 'src/styles/common';
-@import 'constants';
+@import '../../constants';
```
azangru, line 2: maybe change to `src/content/app/entity-viewer/gene-view/styles/constants`? This way, files can be more easily moved around

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.scss
```diff
@@ -1,5 +1,5 @@
 @import 'src/styles/settings';
-@import '../constants';
+@import '../../../constants';
```
azangru, line 2: here, the desirability of the change to `src/content/app/entity-viewer/gene-view/styles/constants` becomes even more pressing

src/ensembl/src/content/app/entity-viewer/gene-view/_constants.scss
azangru: However, why did you decide to move these constants to the top level of the gene view? Aren't they tightly related specifically to the grid of the transcript list?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -52,16 +54,17 @@
   grid-column: 1/2;
   text-align: right;
   padding-right: 1em;
+  color: $dark-blue;
```
azangru, line 57: revert?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -43,7 +45,7 @@
 
 .row {
   display: grid;
-  grid-template-columns: $left_column $middle_column $middle_column_right_gap $right_column;
+  grid-template-columns: $left_column $middle_column_left_gap $middle_column $middle_column_right_gap $right_column;
```
azangru, line 48: I may be missing something, but why did you decide to add another column to the grid:

image


rather than increase the right padding of the "left" section (it already has some amount of right padding anyway):

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -1,10 +1,20 @@
...
 .proteinSummary {
   margin-left: 178px;
   width: 695px;
 
   > div {
     padding: 10px 0 10px 0;
   }
+
+  .keyline {
+    border-bottom: 1px solid #ff9900;
```
azangru, line 12: We have this colour in the list of variables:

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -1,10 +1,20 @@
...
+
 .proteinSummary {
   margin-left: 178px;
   width: 695px;
 
   > div {
     padding: 10px 0 10px 0;
   }
+
+  .keyline {
```
azangru, line 11: probably no need to nest it inside the `.proteinSummary` class? Nesting will result in the `.proteinSummary .keyline` CSS rule; while the simple `.keyline` rule should be specific enough

src/ensembl/src/shared/components/panel/Panel.scss
```diff
@@ -14,11 +14,14 @@
   border-top-left-radius: 5px;
   border-top-right-radius: 5px;
   background-color: $light-grey;
-  box-shadow: 0px 1px 3px $medium-light-grey;
+  box-shadow: 0 4px 3px -2px $medium-light-grey;
   height: 34px;
   padding: 4px 10px;
   width: 100%;
   line-height: 24px;
+  position: sticky;
```
azangru, line 22: What effect will this rule have?

src/ensembl/src/content/app/entity-viewer/gene-view/_constants.scss
jyothishnt: This is used in Geneview overview image grid too.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
jyothishnt: Waiting for confirmation from Andrea

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
jyothishnt: It is good to use the constants and keep the grid layout same as GeneOverview image (as per XD). If you look at the design you can see this column being aligned to the GeneOverview Image.

I left the padding there deliberately as it is being removed in a different PR (Replacing Default/Left text PR)

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
jyothishnt: Oh cool. Thanks. Will change it.

src/ensembl/src/shared/components/panel/Panel.scss
jyothishnt: We have box shadow on the subtab panel which on scroll is visible as in the picture. Now we have shadow to the bottom.
Screenshot 2020-09-04 at 08 48 11
Screenshot 2020-09-04 at 08 48 06

src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
```diff
@@ -1,9 +1,9 @@
 @import 'src/styles/common';
-@import '../default-transcripts-list/constants';
+@import '../../constants';
```
azangru, line 2: This one needs changing as well. Otherwise, error:

image

jyothishnt: Yes, this was missed. Added now

src/ensembl/src/shared/components/panel/Panel.scss
azangru: Ok, I see now what the position: sticky is doing for the panel. It looks cool, but there are two things to consider:

- First, `Panel` is a shared component. I am almost sure its header should not always have `position: sticky`. I think, if you really want to use the `position: sticky`, you should customise the header through Panel's [`classNames` props](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/panel/Panel.tsx#L29)
- The scrolling solution with `posititon: sticky` is very neat. It wouldn't have occurred to me. I would have probably fixed the height of the panel and used `overflow-y: auto` on it. The result would be less pretty (there would be an extra scrollbar inside the panel), but possibly more robust. I wonder whether the solution with `position:sticky` will present problems when we will try to develop the code for scrolling to the selected protein when user clicks on the protein id on a transcript.

azangru (approved):

src/ensembl/src/shared/components/panel/Panel.scss
```diff
@@ -14,6 +14,7 @@
   border-top-left-radius: 5px;
   border-top-right-radius: 5px;
   background-color: $light-grey;
+  box-shadow: 0px 1px 3px $medium-light-grey;
```
azangru: Yeah, I thought you might want it back 😄

PR #344, opened 2020-09-07
Add 404 screen

azangru:
## Type
- Bug fix
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-749

## Description
- Show a 404 page (no design available) if the route does not match any of the specified routes
- Move the `ErrorBoundary` component from inside the `Switch` of the router (it was breaking the router, preventing the matcher to reach the last route that handles the not found case)

## Deployment URL
http://404-screen.review.ensembl.org/

## Views affected
Unmatched routes will show the 404 page.

### DIFF
--- a/src/ensembl/src/content/app/App.tsx
+++ b/src/ensembl/src/content/app/App.tsx
@@ -20,12 +20,12 @@ import { connect } from 'react-redux';
 
 import { changeCurrentApp } from 'src/header/headerActions';
 
-import Header from 'src//header/Header';
+import Header from 'src/header/Header';
 
 const HomePage = lazy(() => import('../home/Home'));
 const GlobalSearch = lazy(() => import('./global-search/GlobalSearch'));
-const SpeciesSelector = lazy(() =>
-  import('./species-selector/SpeciesSelector')
+const SpeciesSelector = lazy(
+  () => import('./species-selector/SpeciesSelector')
 );
 const SpeciesPage = lazy(() => import('./species/SpeciesPage'));
 const CustomDownload = lazy(() => import('./custom-download/CustomDownload'));
--- a/src/ensembl/src/content/app/App.tsx
+++ b/src/ensembl/src/content/app/App.tsx
@@ -15,13 +15,12 @@
  */
 
 import React, { useEffect, lazy, Suspense } from 'react';
-import { Route, Switch, useLocation } from 'react-router-dom';
+import { Route, Switch, Redirect, useLocation } from 'react-router-dom';
 import { connect } from 'react-redux';
 
 import { changeCurrentApp } from 'src/header/headerActions';
 
-import ErrorBoundary from 'src/shared/components/error-boundary/ErrorBoundary';
-import { NewTechError } from 'src/shared/components/error-screen';
+import Header from 'src//header/Header';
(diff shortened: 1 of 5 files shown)


imransl (approved):
I am happy with this PR so will approve it. Just one question about the `is404` property. Are we okay with using numbers in variables? Would like to know what you guys think @msmanoj @azangru @ens-ma7 @jyothishnt.

azangru:
@imransl 

- generally speaking, I am completely fine with variables containing numbers (considering, of course, javascript's rule that variables can't start with a digit)
- if we migrate over to Next, we won't have ReactRouter, and thus won't even have the is404 variable :-)

msmanoj:
@azangru are we sure that we don't want to give the user an option to go back to homepage from the 404 page?

Almost all the examples here seems to have a link back to homepage:
https://www.google.com/search?q=404+page+example&source=lnms&tbm=isch&sa=X&ved=2ahUKEwibocjE1O_rAhU_VRUIHW_aCz4Q_AUoAXoECAwQAw&biw=1440&bih=821#imgrc=ydbyWU_mahAfeM

src/ensembl/src/content/app/App.tsx
```diff
@@ -15,13 +15,12 @@
...
 import React, { useEffect, lazy, Suspense } from 'react';
-import { Route, Switch, useLocation } from 'react-router-dom';
+import { Route, Switch, Redirect, useLocation } from 'react-router-dom';
 import { connect } from 'react-redux';
 
 import { changeCurrentApp } from 'src/header/headerActions';
 
-import ErrorBoundary from 'src/shared/components/error-boundary/ErrorBoundary';
-import { NewTechError } from 'src/shared/components/error-screen';
+import Header from 'src//header/Header';
```
msmanoj, line 23: `src//` -> `src/` ?

src/ensembl/src/shared/components/error-screen/NotFoundError.tsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  const secondLineStyles = {
+    fontSize: '32px'
+  };
+
+  return (
+    <div style={containerStyles}>
+      <p style={firstLineStyles}>404</p>
+      <p style={secondLineStyles}>page not found</p>
```
msmanoj, line 40: `errorCodeStyles` & `errorMessageStyles` ?

azangru:
> are we sure that we don't want to give the user an option to go back to homepage from the 404 page?

I don't know :smiley: This 404 page has not been designed; its only purpose at the moment is to signal to us that we hit a non-existing route (e.g. the one with the `/app` in it). In the future I am sure Andrea will design a proper page. I am also almost sure the properly designed page won't show the number 404 — Andrea doesn't find http codes useful.

src/ensembl/src/shared/components/error-screen/NotFoundError.tsx
azangru: Let's leave styles / CSS class names until the page is properly designed :smiley:

msmanoj (approved):

PR #347, opened 2020-09-15
Add use / don't use / remove functionality to species page

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-754

## Description
- Add to Species page elements to enable, disable, or remove the currently active species
- When the `Remove` element is clicked, the species is removed and the user is redirected to to Species Selector

## Deployment URL
http://use-dont-use-remove.review.ensembl.org/species/homo_sapiens_GCA_000001405_28

## Views affected
Species page

### DIFF
--- a/src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.test.tsx
+++ b/src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.test.tsx
@@ -167,7 +167,7 @@ describe('SpeciesSelectionControls', () => {
     );
   });
 
-  it('correctly togglew removal dialog', () => {
+  it('correctly toggles removal dialog', () => {
     const wrapper = wrapInRedux();
     const removeLabel = wrapper
       .find('span')
--- a/src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.test.tsx
+++ b/src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.test.tsx
@@ -30,7 +30,9 @@ import {
 
 import { createSelectedSpecies } from 'tests/fixtures/selected-species';
 
-import SpeciesSelectionControls from './SpeciesSelectionControls';
+import SpeciesSelectionControls, {
+  speciesRemovalConfirmationMessage
+} from './SpeciesSelectionControls';
 import SlideToggle from 'src/shared/components/slide-toggle/SlideToggle';
 
 jest.mock('connected-react-router', () => ({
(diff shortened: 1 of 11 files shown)


src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.tsx
```diff
@@ -0,0 +1,147 @@
...
+    </div>
+  );
+};
+
+const speciesUseToggleHelpMessage = `When 'Use' is selected, this species will appear in the species list in all apps.
+'Don't use' will disable this species in other apps, but will not remove it from your list in Species selector.`;
+
+const SpeciesUseToggle = (props: SpeciesUseToggle) => {
+  const doNotUseLabelProps: Partial<LabelProps> = {};
+  const useLabelProps: Partial<LabelProps> = {};
```
msmanoj, line 106: Can we not make both the fields optional in `LabelProps` instead of using `Partial` here ?

msmanoj:
Would it be better to wrap the text so that the button is always visible? 
Screenshot 2020-09-22 at 14 47 40

It could go hidden on smaller screens and might confuse the users.

msmanoj:
The species name placeholder and the use / don't use toggle section seems to slightly jump down as we open the remove confirmation.

azangru:
> Would it be better to wrap the text so that the button is always visible?

@msmanoj I think you'll need to ask this question again when we merge https://github.com/Ensembl/ensembl-client/pull/351 in. On the screenshot that you are showing, Sidebar is overlaying part of the main section of the standard layout, which it no longer will once that other PR is merged.

We'll still have a rather ugly situation. i.e.:
image

But may want to show this to Andrea and fix it in a subsequent PR.

Meanwhile, I'll make sure "Don't use" doesn't get wrapped over several lines.

azangru:
> The species name placeholder and the use / don't use toggle section seems to slightly jump down as we open the remove confirmation.
@msmanoj yes, I've noticed that too. It should go away once we start showing the species icon; but will probably still be an issue because of potential wrapping of the text. This is something we'll need to ask Andrea for guidance.

src/ensembl/src/content/app/species/components/species-selection-controls/SpeciesSelectionControls.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+      .filterWhere((wrapper) => wrapper.text() === 'Use')
+      .first();
+    useLabel.simulate('click');
+
+    expect(toggleSpeciesUseAndSave).toHaveBeenCalledWith(
+      selectedSpecies.genome_id
+    );
+  });
+
+  it('correctly togglew removal dialog', () => {
```
msmanoj, line 170: `togglew` -> `toggle`

msmanoj, line 170: `togglew` - > `toggles`

azangru:
@msmanoj I've added a more realistic species icon placeholder to prevent the vertical shift when you click on the "Remove" link (at least as long as the warning together with the Remove button fit in a single line).

I also talked to Andrea, and she will add to the prototype what should happen to the warning on smaller screens.

msmanoj (approved):

PR #348, opened 2020-09-16
Question icon

azangru:
## Type
- Refactoring

## Description
- Updated QuestionButton
  - use the new svg icon (it contains a circle around the question mark; and so the width of the circle will scale consistently with the width of the question mark)
  - updated its size: next to regular text (font-size 12-13px), the svg itself will be will typically be 12px high; and also added 2px of padding to the surrounding div so that the cursor does not need to be placed exactly on top of the icon to trigger the tooltip
  - refactored to use the `useHover` hook
  - updated the appearance of the question button in EntityViewer (10px distance between the word Selected and the button; line height 1)
- Updated Launchbar
  - new icon for help launchbar button
  - new icon for custom download launchbar button
- Refactored BrowserTrackConfig bar
  -  removed `dl` and `dd`
  - removed default 40px width from the `SlideToggle` element, which was used to control the width of empty space around the toggle (bad idea):

image

## Deployment URL
http://question-icon.review.ensembl.org/

## Views affected
- Genome Browser
- Entity Viewer


src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -30,43 +32,40 @@ export enum QuestionButtonOption {
...
-  styleOption?: QuestionButtonOption;
+  styleOption: QuestionButtonOption;
   className?: { [key in QuestionButtonOption]?: string };
 };
 
 const QuestionButton = (props: Props) => {
-  const [isHovering, setIsHovering] = useState(false);
-  const elementRef = useRef<HTMLDivElement>(null);
+  const [hoverRef, isHovered] = useHover<HTMLDivElement>();
+  const [shouldShowTooltip, setShouldShowTooltip] = useState(isHovered);
```
msmanoj, line 41: Would it not be possible to just use `isHovered` instead of having `shouldShowTooltip`?

azangru: @msmanoj what I am trying to achieve here is:
- hide the tooltip when user removes mouse from the button
- hide the tooltip when user starts scrolling

`isHovered` doesn't help with the second point

Currently, the tooltip itself is aware of scrolling, and signals to the parent component that it wants to close when user starts scrolling. So I am combining the signals from the hook about whether user is mousing over the question button with the signal from the tooltip about whether user is scrolling. I could, of course, add a scroll listener to the QuestionButton itself, but I thought it better to use the one in the tooltip.

msmanoj: @azangru do we need to hide the tooltip as soon as the user starts to scroll?

I just tried to scroll 1-2px but the tooltip goes away when I'm still hovered. Is this expected?

azangru: I'd say yes. You usually don't scroll when you want to read a tooltip.

msmanoj (approved):

PR #264, opened 2020-03-02
Make sidebar slide over EV content view when open

imransl:
## Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement to existing feature

## Related JIRA Issue(s)
[ENSWBSITES-517](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-517)

## Description
This PR adds the capability of the standard app layout sidebar to slide over the main content. This is necessary for EV, whereas the GB behaviour will remain the same (sidebar does not slide over).

## Views affected
Entity Viewer

PR #351, opened 2020-09-20
Update StandardAppLayout

azangru:
## Type
- Refactoring

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-767

### Before the change
Part of the main content section of the standard app layout used to covered by the open sidebar like so (change introduced in https://github.com/Ensembl/ensembl-client/pull/264).

image

### After the change
We are now backtracking on this decision, because with the current layout, the vertical scrollbar is not visible to the user when the sidebar is open. The updated layout should behave like this regarding the scrollbars:

image

## Additional changes
Fixed extra width bug that was caused by moving transcript-wrapping `div`s via `transform: translate`. The divs, because of their `display: block` property, maintained their original width, and were pushing GeneView border to the right. The effect can be seen on the screenshot below:

image

## Deployment URL
http://layout.review.ensembl.org/

## Views affected
- Genome Browser
- Entity Viewer
- Species Page

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.scss
+++ b/src/ensembl/src/content/app/browser/Browser.scss
@@ -2,7 +2,6 @@
 
 .browserInnerWrapper {
   display: grid;
-  height: 100%;
   grid-template-rows: 80px 1fr;
 }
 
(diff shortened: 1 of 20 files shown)


msmanoj:
Hi @azangru - can you please resolve the conflicts when you get a chance?

azangru:
@msmanoj conflicts resolved

msmanoj:
The cursor seems to remain as pointer when we hover over `DefaultTranscriptListItemInfo`. Could you check if that should be the case?

azangru:
@msmanoj the pointer behaviour is something that already exists in dev (screenshot from internal-2020 below, if that's what you have in mind):

image

It's likely that Andrea will want us to change this, but that definitely looks like a separate PR.

azangru:
@msmanoj I've fixed the pointer behaviour by adding another div to wrap the transcript :-) I agree that the middle column should not have the cursor: pointer style associated with the whole column.

msmanoj (approved):

PR #352, opened 2020-09-24
Species homepage stats

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-758

## Description
- Created the shared component `ExpandableSection`
- Using a local sample data file which is populated using the following sheet:
https://docs.google.com/spreadsheets/d/1NYANlVmB4tzw4cIluOI4kUWtXRd1P9jU-CNbYOCqSzA/edit#gid=0

## Deployment URL
http://species-homepage-stats.review.ensembl.org/species/homo_sapiens_GCA_000001405_28

## Views affected
Species homepage

### Other effects


src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -6,5 +8,124 @@
 }
 
 .speciesLabelBlock {
-  padding: 0 75px 0 60px;
+  padding: 0 75px 0 45px;
+}
+
+// TODO: Remove Me
+.statsWrapper {
```
msmanoj, line 15: TODO: This needs to be updated once the PR https://github.com/Ensembl/ensembl-client/pull/351 is merged

src/ensembl/src/shared/components/expandable-section/ExpandableSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+      {isExpanded && (
+        <div className={expandedContentClassNames}>{props.expandedContent}</div>
+      )}
+
+      {!isExpanded && (
+        <div className={collapsedContentClassNames}>
+          {props.collapsedContent}
+        </div>
+      )}
```
imransl, line 83: How about using the ternary operator for this?

src/ensembl/src/content/app/species/SpeciesPage.tsx
```diff
@@ -32,30 +31,21 @@ import SpeciesAppBar from './components/species-app-bar/SpeciesAppBar';
...
-  const currentSpecies = useSelector((state: RootState) =>
-    getCommittedSpeciesById(state, genomeId)
-  );
+
   const sidebarStatus = useSelector(isSidebarOpen);
   const dispatch = useDispatch();
 
   useEffect(() => {
     dispatch(setActiveGenomeId(genomeId));
+    dispatch(fetchGenomeData(genomeId));
```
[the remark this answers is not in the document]
azangru: image

[link](https://react-redux.js.org/api/batch)

Apparently, it only makes sense from within something like thunks, not from inside React components.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -1,3 +1,5 @@
+@import 'src/styles/common';
```
azangru, line 1: Lots of 3 (or 4?) spaces used as indents in this file. Don't we normally use two spaces?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -6,5 +8,120 @@
...
+  .title{
+      font-size: 12px ;
+      grid-column: title;
+      color: $medium-dark-grey;
+  }
+
+  .stats{
+      grid-column: stats;
+  }
+}
```
azangru, line 127: Also, our editors will automatically add an empty line in the end of a file. What happened to your VSCode? 😄

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { fetchStatsForActiveGenome } from 'src/content/app/species/state/general/speciesGeneralSlice';
+
+import { RootState } from 'src/store';
+import { GenomeStats } from '../../state/general/speciesGeneralSlice';
+import ViewInApp, { urlObj } from 'src/shared/components/view-in-app/ViewInApp';
+import {
+  StatsSection,
+  sectionGroupsMap
+} from '../../state/general/speciesGeneralHelper';
+import { ExampleFocusObject } from 'src/shared/state/genome/genomeTypes';
```
azangru, line 41: order of imports

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+          ref={anchorRef}
+          onClick={() => setShowPointerBox(!showPointerBox)}
+        >
+          {exampleLinkText}
+          {showPointerBox && (
+            <PointerBox
+              anchor={anchorRef.current as HTMLDivElement}
+              onOutsideClick={() => setShowPointerBox(false)}
+              position={Position.BOTTOM_RIGHT}
+              container={containerRef.current}
```
azangru, line 80: The point of the container prop is to make the pointer box try to avoid extending beyond the borders of a given outer element:

image

It defaults to `window` and is useless here.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+    <div className={styles.exampleLink}>
+      <span ref={containerRef}>
+        <span
+          className={styles.exampleLinkText}
+          ref={anchorRef}
+          onClick={() => setShowPointerBox(!showPointerBox)}
+        >
+          {exampleLinkText}
+          {showPointerBox && (
+            <PointerBox
```
azangru, line 76: This popup is recurring theme with Andrea; there was time when she had it on the home page. It looks very much like a separate component.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+              anchor={anchorRef.current as HTMLDivElement}
```
azangru, line 77: The type above says that it's `HTMLSpanElement`. But probably better just to add `anchorRef.current` to the condition above (`showPointerBox && anchorRef.current && ...`).

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+}) => {
+  const { exampleLinks, exampleLinkText } = props;
+
+  const [showPointerBox, setShowPointerBox] = useState(false);
+  const anchorRef = useRef<HTMLSpanElement>(null);
+  const containerRef = useRef<HTMLSpanElement>(null);
+
+  if (!exampleLinks) {
+    return null;
+  }
```
azangru, line 64: ❓ 
why would ExampleLinks be a valid component without the example links?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+  activeGenomeId: string | null;
+  genomeStats: GenomeStats | undefined;
+  exampleFocusObjects: ExampleFocusObject[];
+  fetchStatsForActiveGenome: () => void;
+};
+
+const ExampleLinks = (props: {
+  exampleLinks?: Partial<urlObj>;
+  exampleLinkText?: string;
+}) => {
```
azangru, line 55: Several considerations:
- The component appears as a single link, although it might show the popup with buttons (compare our other ExampleLinks components, where we are showing multiple links). Perhaps better call it `ExampleFeatureLink` or `ViewExample`? Is it possible that this is a reusable component? It doesn't look particularly specific to `SpeciesMainViewStats`.
- The `example` part of the `exampleLinks` prop is redundant. `links` should be sufficient, as it is called in the `ViewInApp` component.
- The `exampleLinkText` can then simply become `children`
- When there is only one app in which the example can be viewed (e.g. example region can only be viewed in GenomeBrowser), I am sure there is no point in showing the popup. We should just redirect the user there.

src/ensembl/src/content/app/species/state/general/speciesGeneralSlice.ts
```diff
@@ -13,15 +13,71 @@
...
+    return;
+  }
+
+  const exampleFocusObjects = getGenomeExampleFocusObjects(
+    state,
+    activeGenomeId
+  );
+
+  const speciesStats = [
+    ...Object.values(SpeciesStatsSections).map((section) =>
```
azangru, line 66: why the spread?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,419 @@
...
+import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
+
+import { SpeciesStatsProps } from 'src/content/app/species/components/species-stats/SpeciesStats';
+import { ExampleFocusObject } from 'src/shared/state/genome/genomeTypes';
+
+import { sampleData } from '../../sample-data';
+import { buildFocusIdForUrl } from 'src/shared/state/ens-object/ensObjectHelpers';
+import { urlObj } from 'src/shared/components/view-in-app/ViewInApp';
+
+export enum SpeciesStatsSections {
```
azangru, line 27: 1) Enums are usually named in singular, because they represent not a set of things, but one thing belonging to a set. See [example in typescript docs](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums).

2) Aren't we using enums a bit too much? What would this code look like if we didn't use enums?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,419 @@
...
+    exampleLinkText?: string;
+  };
+};
+
+// Maps the groups to it's respective section
+export const sectionGroupsMap: SpeciesStatsSectionGroups = {
+  [SpeciesStatsSections.CODING_STATS]: {
+    title: Stats.CODING_GENES,
+    hasExampleLink: true,
+    exampleLinkText: 'Example gene',
```
azangru, line 109: `hasExampleLink` and `exampleLinkText` seem always to coincide. `hasExampleLink` looks like a redundant property; let's get rid of it

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+          </div>
+        );
+      })}
+    </div>
+  );
+};
+
+const SpeciesMainViewStats = (props: Props) => {
+  useEffect(() => {
+    if (!props.genomeStats && props.exampleFocusObjects?.length) {
```
azangru, line 174: Ouch! This is a very confusing dependency. Fetching genome stats should not depend on whether or not we have fetched example focus objects.

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
```diff
@@ -3,7 +3,8 @@
 $secondary-text-color: $dark-grey;
 
 .wrapper {
-  width: 250px;
+  min-width: 250px;
```
azangru, line 6: why min-width rather than width?

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
```diff
@@ -3,7 +3,8 @@
...
+  margin-bottom:30px;
```
azangru, line 7: space after colon

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+                return (
+                  <div key={group_index}>
+                    {groupStats.map((stat, stat_index: number) => {
+                      return <SpeciesStats key={stat_index} {...stat} />;
+                    })}
+                  </div>
+                );
+              })}
+            </div>
+            {group_index === 0 && (
```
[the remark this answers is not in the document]
azangru: I'd argue the way it's written now (`group_index === 0`) is ever so slightly more readable. With `!group_index` in this context you would have to mentally translate it back: "why is the group_index falsey? oh, that's because it's zero". The way it's written now saves this mental step.

Probably both ways of writing it are equally fine.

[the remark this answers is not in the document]
imransl: If so then shouldn't we be using the same method in a single file as line 140 uses the other method

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+};
+
+const getExpandedContent = (statsSection: StatsSection) => {
+  const { groups, exampleLinks, section } = statsSection;
+  const { exampleLinkText } = sectionGroupsMap[section];
+  return (
+    <div className={styles.expandedContent}>
+      {groups.map((group, group_index: number) => {
+        const { title, stats } = group;
+        const primaryStatsClassName = classNames(styles.primaryStats, {
```
azangru, line 138: Am I reading this correctly and the calculation of this class name does not depend on anything happening inside the loop? If so, then it should be calculated outside the `.map`

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -6,5 +8,120 @@
...
+      font-size: 11px;
+  }
+}
+
+
+.primaryStats{
+  padding: 13px 20px 0;
+  display: grid;
+  grid-template-columns: [title] 120px [stats] auto [example-link] 150px;
+  align-items: flex-start;
```
azangru, line 90: the value of align-items for `display: grid` does not use the prefix `flex-`

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -6,5 +8,120 @@
...
+      padding-top:2px;
+  }
+}
+
+
+.secondaryStats{
+  padding: 13px 20px 0;
+  display: grid;
+  grid-template-columns: [title] 120px [stats] auto;
+  align-items: flex-start;
```
azangru, line 116: The value of align-items for display: grid does not use the prefix flex-

Also, there is a lot of repetition between the `primaryStats` and `secondaryStats` classes. Won't a single class with some modifiers be better?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -6,5 +8,120 @@
...
+
+  .secondary{
+      grid-column: secondary;
+
+      .secondaryValue{
+          font-size: 22px;
+      }
+
+      .secondaryUnit{
+          color: #6f8190;
```
azangru, line 73: this color is defined in our design system as a variable

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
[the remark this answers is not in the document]
azangru: 🤷‍♂️

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+          [styles.primaryStatsNoLink]: exampleLinks
+            ? !Object.keys(exampleLinks).length
+            : true
+        });
+        const primaryOrSecondaryClassName =
+          group_index === 0 ? primaryStatsClassName : styles.secondaryStats;
```
azangru, line 144: I tried to figure this out from the code and from the way the page looks, and couldn't. What's the difference between "primary stats" and "secondary stats"? Is it just the first stats group versus all others? Then shouldn't it be something like `statsGroup` vs `firstStatsGroup`? Or, for more convenience with SCSS, expressing it in the BEM style: `statsGroup` vs `statsGroup_first`?

src/ensembl/src/content/app/species/state/general/speciesGeneralSelectors.ts
```diff
@@ -18,3 +18,13 @@ import { RootState } from 'src/store';
 
 export const getActiveGenomeId = (state: RootState) =>
   state.speciesPage.general.activeGenomeId;
+
+export const getActiveGenomeStats = (state: RootState) => {
+  const activeGeonmeId = getActiveGenomeId(state);
```
azangru, line 23: Geonme – 🙅‍♂️

src/ensembl/src/shared/components/expandable-section/ExpandableSection.scss
```diff
@@ -0,0 +1,39 @@
...
+    top: 12px;
+    position: absolute;
+    left: 35px;
+    transform: rotate(90deg);
+    transition: transform linear .2s;
+    &Down {
+      transform: rotate(-90deg);
+      
+    }
+  }
```
azangru, line 39: Also suddenly 4-space indents and no empty line in the end of the file 😞

src/ensembl/stories/shared-components/expandable-section/ExpandableSection.stories.scss
```diff
@@ -0,0 +1,87 @@
...
+    .title{
+        font-size: 12px ;
+        grid-column: title;
+        color: $medium-dark-grey;
+    }
+
+    .stats{
+        grid-column: stats;
+    }
+}
```
azangru, line 87: again, 4-space indents and no trailing empty line 😞

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,605 @@
...
+      'Assembly date': '2019-02-28',
+      'Contig N50': '56413054',
+      'Total genome length': '3272116950',
+      'Total coding sequence length': '34317174',
+      'Total gap length': '161368351',
+      'Spanned gaps': '661',
+      Chromosomes: '25',
+      'Toplevel sequences': '640',
+      'Component sequences': '36734',
+      '% GC': '38.87'
```
azangru, line 43: The structure presented here is very UI-specific, where the key in the json object is the same string as the label in the UI. I strongly feel that the proper long-term approach is to design a data structure that is likely to come from the backend, then figure out how to transform this data structure into something consumable by the frontend, and finally pass the proposed data structure to the backend for implementation. I am almost positive the the backend will not want to format the keys of the JSON this way.

A conversation with the backend team might help decide on the best shape of the json. @ens-st3, @jgtate what is your advise?

src/ensembl/src/content/app/species/state/general/speciesGeneralSlice.ts
```diff
@@ -13,15 +13,71 @@
...
+import { getActiveGenomeId } from './speciesGeneralSelectors';
+
+import { getGenomeExampleFocusObjects } from 'src/shared/state/genome/genomeSelectors';
+
+import { RootState } from 'src/store';
+import {
+  getStatsForSection,
+  StatsSection,
+  SpeciesStatsSections
+} from 'src/content/app/species/state/general/speciesGeneralHelper';
```
azangru, line 32: order of imports

src/ensembl/src/shared/components/expandable-section/ExpandableSection.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  beforeEach(() => {
+    jest.resetAllMocks();
+    wrapper = renderExpandableSection();
+  });
+
+  it('renders without error', () => {
+    expect(() => wrapper).not.toThrow();
+  });
+
+  it('hides expandedContent when isExpanded is false', () => {
```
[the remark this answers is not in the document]
azangru: `it('shows only collapsed content when collapsed'...)`

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,419 @@
...
+ * limitations under the License.
+ */
+
+import * as urlFor from 'src/shared/helpers/urlHelper';
+import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
+
+import { SpeciesStatsProps } from 'src/content/app/species/components/species-stats/SpeciesStats';
+import { ExampleFocusObject } from 'src/shared/state/genome/genomeTypes';
+
+import { sampleData } from '../../sample-data';
```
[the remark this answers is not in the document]
msmanoj: I thought relative paths up to two levels are okay?

src/ensembl/src/shared/components/expandable-section/ExpandableSection.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    expect(wrapper.find('.expandedContent')).toHaveLength(0);
+    expect(wrapper.find('.collapsedContent')).toHaveLength(1);
+  });
+
+  it('shows expandedContent when isExpanded is true', () => {
+    expect(wrapper.find('.expandedContent')).toHaveLength(1);
+    expect(wrapper.find('.collapsedContent')).toHaveLength(0);
+  });
+
+  it('calls onToggle with expanded status when the chevron wrapper is clicked', () => {
```
azangru, line 68: `it('calls onToggle prop when collapsing or expanding'...)`

and then the test should check that onToggle was called with `false` if the section was expanded and with `true` if it was collapsed.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: Yes, that's right.

Any idea how the following can be prevented? Should it be closed as soon as the user starts to scroll?
Kapture 2020-10-06 at 14 21 59

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: > Aren't we using enums a bit too much? What would this code look like if we didn't use enums?
Any suggestions for alternatives?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: That's right. Removed `props.exampleFocusObjects?.length` from here and moved it to the if statement below.

msmanoj: Example links are getting generated while the stats are getting processed. This is why I'm trying to make sure the exampleObjects are available before fetching the genomeStats.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
azangru: > Should it be closed as soon as the user starts to scroll?

That's what the tooltips do. I'd say yes.

One other option is to try the `renderInsideAnchor` prop; then the pointer box will be placed into the "Example gene" span rather than at the top level of the window. This might cause other layout problems — or it may work out fine :-)

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
azangru: Yep.

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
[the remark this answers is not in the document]
imransl: I can't remember if we had a consensus on this. If we did, then let's just leave it as it is.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: > Is it possible that this is a reusable component? It doesn't look particularly specific to SpeciesMainViewStats.

How about creating something like `ViewInAppPopup`?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+    </div>
+  );
+};
+
+const getExpandedContent = (statsSection: StatsSection) => {
+  const { groups, exampleLinks, section } = statsSection;
+  const { exampleLinkText } = sectionGroupsMap[section];
+  return (
+    <div className={styles.expandedContent}>
+      {groups.map((group, group_index: number) => {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+            ? !Object.keys(exampleLinks).length
+            : true
+        });
+        const primaryOrSecondaryClassName =
+          group_index === 0 ? primaryStatsClassName : styles.secondaryStats;
+        return (
+          <div key={group_index} className={primaryOrSecondaryClassName}>
+            <span className={styles.title}>{title}</span>
+            <div className={styles.stats}>
+              {stats.map((groupStats, group_index: number) => {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,217 @@
...
+                return (
+                  <div key={group_index}>
+                    {groupStats.map((stat, stat_index: number) => {
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,419 @@
...
+          focus: focusId
+        })
+      : undefined;
+
+    exampleLinks.entityViewer = focusId
+      ? urlFor.entityViewer({
+          genomeId: genome_id,
+          entityId: focusId
+        })
+      : undefined;
```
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/shared/components/expandable-section/ExpandableSection.test.tsx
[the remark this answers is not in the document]
msmanoj: That's done

src/ensembl/src/shared/components/expandable-section/ExpandableSection.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralSlice.ts
msmanoj: That's done

src/ensembl/stories/shared-components/expandable-section/ExpandableSection.stories.scss
msmanoj: Fixed it

src/ensembl/src/shared/components/expandable-section/ExpandableSection.scss
msmanoj: Fixed it

src/ensembl/src/content/app/species/state/general/speciesGeneralSelectors.ts
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: Updated to use `first-of-type` and removed the line above.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: Updated it

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: >The value of align-items for display: grid does not use the prefix flex-

Removed it

>Also, there is a lot of repetition between the primaryStats and secondaryStats classes. Won't a single class with some modifiers be better?

That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: Removed `align-items` as it's not required

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
msmanoj: Update it

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
msmanoj: Missed to clean up

src/ensembl/src/content/app/species/state/general/speciesGeneralSlice.ts
msmanoj: Removed it

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: Removed the check and made the prop mandatory

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: I always get confused with this :(

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: I've added line break but I'm not sure what's causing it. I'll keep an eye on it.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: That's done

src/ensembl/src/shared/components/expandable-section/ExpandableSection.tsx
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
msmanoj: That's done

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
azangru: Sounds good.

src/ensembl/src/shared/components/view-in-app-popup/ViewInAppPopup.scss
```diff
@@ -0,0 +1,17 @@
...
+    color: white;
+  
+    p {
+      max-width: 200px;
+    }
+  }
+  
+  .pointerBoxPointer {
+    fill: $black;
+  }
```
azangru, line 17: Indents and terminal empty line again 😞

src/ensembl/stories/shared-components/expandable-section/ExpandableSection.stories.scss
```diff
@@ -0,0 +1,85 @@
...
+    color: $blue;
+    grid-column: example-link;
+  }
+}
+
+.primaryStats {
+  padding: 13px 20px;
+  display: grid;
+  grid-template-columns: [title] 120px [stats] auto [example-link] 200px;
+  align-items: flex-start;
```
azangru, line 53: here and below — no flex

src/ensembl/stories/shared-components/expandable-section/ExpandableSection.stories.scss
```diff
@@ -0,0 +1,85 @@
...
+
+  .primary {
+    grid-column: primary;
+
+    .primaryValue {
+      font-size: 22px;
+    }
+
+    .primaryUnit {
+      color: #6f8190;
```
azangru, line 22: here and below — use scss variable

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
-  assembly_stats: {
-    chromosome_count: 23,
-    genome_size: 3096649726,
-    total_sequence_length: 4537931177
+import { SpeciesStatsSection } from 'src/content/app/species/state/general/speciesGeneralHelper';
+
+type RawSpeciesStats = {
+  [genomeId: string]: {
+    [key in SpeciesStatsSection]: {
+      [key: string]: string;
```
azangru, line 22: This is something I would definitely urge we change. The type of field value can be at least one of the following: string, number, null

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+    };
+  };
+};
+
+export const sampleData: RawSpeciesStats = {
+  homo_sapiens_GCA_000001405_28: {
+    assembly_stats: {
+      scientific_name: 'Homo sapiens',
+      sex: 'NA',
```
azangru, line 31: All the NAs should become nulls

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+      breed_cultivar_strain: 'NA',
+      taxonomy_id: '9606',
+      assembly_name: 'GRCh38.p13',
+      assembly_accession: 'GCA_000001405.28',
+      assembly_date: '2019-02-28',
+      contig_n50: '56413054',
```
azangru, line 37: All the numbers should be of type number.

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+      average_exons_per_transcript: '2.36',
+      total_introns: '25165',
+      average_intron_length: '4336.85'
+    }
+  },
+  homo_sapiens_GCA_000001405_14: {
+    assembly_stats: {
+      scientific_name: 'Homo sapiens',
+      sex: 'NA',
+      breed_cultivar_strain: 'NA',
```
azangru, line 106: This field can probably just be called "strain". Whichever is the preferred term for this taxonomic group at Ensembl.

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+      gc_percentage: '38.87'
+    },
+    coding_stats: {
+      scientific_name: 'Homo sapiens',
+      coding_genes: '20420',
+      average_genomic_span: '66885.16',
+      average_sequence_length: '3443.15',
+      average_CDS_length: '1139.02',
+      shortest_gene: '8',
+      longest_gene: '2473539',
```
azangru, line 54: shortest_gene_**length**? longest_gene_**length**?

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+      toplevel_sequences: '640',
+      component_sequences: '36734',
+      gc_percentage: '38.87'
+    },
+    coding_stats: {
+      scientific_name: 'Homo sapiens',
+      coding_genes: '20420',
+      average_genomic_span: '66885.16',
+      average_sequence_length: '3443.15',
+      average_CDS_length: '1139.02',
```
azangru, line 52: CDS -> cds (lower-case)

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: That's done

src/ensembl/src/content/app/species/sample-data.ts
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+  let primaryValue = props.primaryValue;
+
+  if (!primaryValue) {
+    return;
+  }
+
+  const formattingOptions = statsFormattingOptions[props.primaryKey];
+
+  if (!isNaN(Number(primaryValue))) {
+    primaryValue = getCommaSeparatedNumber(Number(primaryValue));
```
azangru, line 312: Why don't we expect numbers to be of type `number`? Why are they string/number?

_(And, consequently, why isn't the if-condition `if(typeof primaryValue === 'number)`?)_

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -0,0 +1,173 @@
...
+import {
+  StatsSection,
+  sectionGroupsMap
+} from '../../state/general/speciesGeneralHelper';
+
+import styles from './SpeciesMainView.scss';
+
+type Props = {
+  activeGenomeId: string | null;
+  genomeStats: GenomeStats | undefined;
```
azangru, line 44: Why would it ever be undefined? Does it make sense for a component responsible for displaying stats to receive an undefined instead of stats?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+type StatsGroup = {
+  title: string;
+  stats: [IndividualStat[]];
+};
+
+export type StatsSection = {
+  section: SpeciesStatsSection;
+  exampleLinks?: Partial<urlObj>;
+  primaryStats?: IndividualStat;
+  secondaryStats?: IndividualStat;
```
azangru, line 291: What do primary stats and secondary stats mean for a stats section?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+    }
+  } else if (section === SpeciesStatsSection.ASSEMBLY_STATS) {
+    const regionExample = exampleFocusObjects.find(
+      (object) => object.type === 'region'
+    );
+
+    const focusId = regionExample?.id
+      ? buildFocusIdForUrl({
+          type: 'region',
+          objectId: regionExample?.id
```
azangru, line 383: at this point, `regionExample` definitely exists (same above with `geneExample`).

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+  [Stats.TOTAL_GENOME_LENGTH]: {
+    primaryUnit: 'bp',
+    label: 'Total genome length'
+  },
+  [Stats.TOTAL_CODING_SEQUENCE_LENGTH]: {
+    primaryUnit: 'bp',
+    label: 'Total coding sequence length'
+  },
+  [Stats.AVERAGE_GC_CONTENT]: {
+    primaryValuePostfix: '%',
```
azangru, line 220: what's the difference between `primaryUnit` and `primaryValuePostfix`?

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
```diff
@@ -81,12 +83,17 @@ const SpeciesStats = (props: SpeciesStatsProps) => {
...
+        <span className={styles.preLabel}>{props.preLabel}</span>
+      )}
 
       <div className={styles.label}>{props.label}</div>
 
       <div>
-        <span className={styles.primaryValue}>{props.primaryValue}</span>
+        <span className={styles.primaryValue}>
+          {props.primaryValue}
+          {props.primaryValuePostfix || null}
```
azangru, line 95: no need for `|| null`

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+};
+
+const buildIndividualStat = (
+  props: BuildStatProps
+): IndividualStat | undefined => {
+  let primaryValue = props.primaryValue;
+
+  if (!primaryValue) {
+    return;
+  }
```
azangru, line 307: Erm, can primaryValue be `0`?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,462 @@
...
+
+  if (!isNaN(Number(primaryValue))) {
+    primaryValue = getCommaSeparatedNumber(Number(primaryValue));
+  }
+
+  return {
+    label: formattingOptions?.label || props.primaryKey,
+    primaryValue,
+    primaryUnit: formattingOptions?.primaryUnit,
+    primaryValuePostfix: formattingOptions?.primaryValuePostfix
```
azangru, line 319: If part of the formatting (`getCommaSeparatedNumber`) is happening in this function, why not add primaryValuePostfix to the comma-separated number here, without passing it to the component?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: I've removed the need to `primaryValuePostfix` as it's simpler to append the postfix value in the helper function itself.

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: They are the two stats that gets displayed in the header when the section is collapsed. Would `firstHeaderStat` & `secondHeaderStat` work?

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: Removed it as this block is no longer required.

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.tsx
msmanoj: Removed the prop `primaryValuePostfix` as mentioned in https://github.com/Ensembl/ensembl-client/pull/352#discussion_r503607775

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: Updated it

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: That's all updated now

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,615 @@
...
+export const sampleData: RawSpeciesStats = {
+  homo_sapiens_GCA_000001405_28: {
+    assembly_stats: {
+      scientific_name: 'Homo sapiens',
+      sex: null,
+      strain: null,
+      taxonomy_id: 9606,
+      assembly_name: 'GRCh38.p13',
+      assembly_accession: 'GCA_000001405.28',
+      assembly_date: '2019-02-28',
```
azangru, line 36: Do you want to logically separate the fields that describe the assembly (scientific_name, sex, taxonomy_id, assembly_name, assembly_accession, assembly_date) into a dedicated section (`assembly_description`?, `assembly_summary`?)? It feels like a perceptibly different kind of data.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -60,6 +60,7 @@
   grid-column: 3/4;
   position: relative;
   top: 2px;
+  cursor: pointer;
```
azangru, line 63: I'm surprised. How home you have something for entity viewer in this PR? 😄

msmanoj: I once merged the layout update branch before it got merged into dev. That could be why this got left out. I think I need to remove this since it is shown as a difference.

src/ensembl/src/content/app/species/sample-data.ts
msmanoj: That sounds like a good idea. `assembly_summary` feels like a better name.

msmanoj: I tried to make this change but currently these sections each represents an individual `ExpandableComponent`. Creating a separate section called `assembly_summary` which will not have it's own `ExpandableComponent` feels like it will cause more confusion. Would it be worth getting back to this later on?

azangru: > Would it be worth getting back to this later on?

I think it would be nice to have this separation by the time we give our json back to Fergal so that he can update it programmatically. Were you thinking of getting back to it earlier than that (e.g. during your work on the sidebar)?

msmanoj: I've tried to do it in this PR itself. Please check if it looks okay.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
azangru: What if we call them something like `summaryStats` or `briefStats` and rename `groups` to `fullStats`?

Then `primaryStats` and `secondaryStats` will become a single property, which is of type `IndividualStat[]`.

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -0,0 +1,477 @@
...
+  primaryValue: string | number;
+  primaryKey: Stats;
+};
+
+const buildHeaderStat = (
+  props: BuildHeaderStatProps
+): IndividualStat | undefined => {
+  let primaryValue = props.primaryValue;
+
+  const { primaryValuePostfix, headerUnit } = statsFormattingOptions[
```
azangru, line 323: if `StatsFormattingOptions` type really has a field `{primaryValuePostfix?: string;}`, then you can use the default value for primaryValuePostfix when destructuring (`const { primaryValuePostfix = '' , ... } = ...`). Then you wouldn't need to apply a ternary operator to check that primaryValuePostfix exists below.

src/ensembl/src/content/app/species/sample-data.ts
```diff
@@ -14,30 +14,637 @@
...
+      total_coding_sequence_length: 34317174,
+      total_gap_length: 161368351,
+      spanned_gaps: 661,
+      chromosomes: 25,
+      toplevel_sequences: 640,
+      component_sequences: 36734,
+      gc_percentage: 38.87
+    },
+    coding_stats: {
+      scientific_name: 'Homo sapiens',
```
azangru, line 56: This has been moved to assembly summary. Same for other sections and other species below.

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: That's done

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
msmanoj: That's done

src/ensembl/src/content/app/species/sample-data.ts
msmanoj: Remove it

azangru:
As a reminder from code review:

image

- stuff that's currently in assembly_summary should not appear in the assembly stats section
- need to re-check groupings of data in sections with the XD (we noticed that the coding genes section does not contain all stats that are in the XD)

src/ensembl/src/content/app/species/sample-data.ts
azangru: I'm still seeing it in multiple places in the file (https://github.com/Ensembl/ensembl-client/blob/ac354582bcc65b250a2e298b70d189246e16669f/src/ensembl/src/content/app/species/sample-data.ts).

msmanoj: Should be gone now?

msmanoj:
> As a reminder from code review:
> 
> image
> 
> * stuff that's currently in assembly_summary should not appear in the assembly stats section
> * need to re-check groupings of data in sections with the XD (we noticed that the coding genes section does not contain all stats that are in the XD)

That's all done

azangru:
@msmanoj something isn't right:

- pseudogenes among coding genes:
image

- pseudogenes in Analysis:
image

- pseudogenes in Transcripts:
image

- pseudogenes everywhere...
image

msmanoj:
> @msmanoj something isn't right:
> 
> * pseudogenes among coding genes:
>   image
> * pseudogenes in Analysis:
>   image
> * pseudogenes in Transcripts:
>   image
> * pseudogenes everywhere...
>   image

That should now be fixed. It was because of keys getting overridden in the `statsFormattingOptions`

azangru:
Coding genes, analysis

Design:
image

Deployment:
image

msmanoj:
@azangru - the missing `Average CDS length` is now back. I think `Average coding exons per transcript` is the duplicate of the one that's displayed in the `Transcripts` group. I will check this with Andrea and get back.

I've increase the PrimaryValue font size to 25px and it seems to be matching the XD.

azangru:
Notice that font size in the collapsed and expanded section are the same in the design:

image

But different in the deployment:

image

> I've increased the PrimaryValue font size to 25px and it seems to be matching the XD.

Probably should ask Andrea about font sizes to make sure?

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
```diff
@@ -13,7 +13,8 @@
 
 .statsWrapper {
   padding: 5px 20px 40px 120px;
-  max-width: 1360px;
+  max-width: 1381px;
+  font-size: 13px;
```
azangru, line 17: This is the default font size (see _settings.scss). I tried to disable this rule in the deployment, and didn't see any change. It's probably redundant.

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
```diff
@@ -7,38 +7,31 @@ $secondary-text-color: $dark-grey;
   margin-bottom: 30px;
   display: inline-block;
   vertical-align: top;
+  font-size: 13px;
```
azangru, line 10: same here: default font size, inherited from the `html` element, doesn't change anything if disabled, so likely redundant

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainView.scss
msmanoj: Removed it

src/ensembl/src/content/app/species/components/species-stats/SpeciesStats.scss
msmanoj: Removed it

azangru:
Non-coding genes, analysis, human:
image

should have `bp` units

image
genes -> gene, for both longest and shortest

image
no `bp`

image
`bp` for both

image
no `bp` for all

image
Ah! we are getting wrapped labels! Check w/Andrea what to do

image
"Top level" is written as two words in XD, although I am sure it should be spelled with a hyphen (top-level). If no-one cares about the hyphen, spell in two words.

Only checked for human. Hopefully, it will be the same for other species.

msmanoj:
@azangru  - I have added the missing units and removed where they are not needed. Regarding the wrapping of labels, Andrea mentioned not to worry about them now as she is planning to review them all and also get some stats re-arranged.
We can get those changes done in a subsequent PR once the design is updated.

azangru:
I pushed a commit for the labels you missed in your last one.

Let's merge this thing in 😃

azangru (approved):

PR #353, opened 2020-09-24
Add an ADR for server-side rendering

azangru:
## Type
- documentation

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-763

## Description
An Architecture Decision Record capturing the state of our conversation about server-side rendering. It seems that we are generally in agreement that server-side rendering is a desirable goal for the new Ensembl site, but are still unsure which solution will suit us the best.

### DIFF
--- a/src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
+++ b/src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
@@ -8,20 +8,20 @@ Under review.
 - **Client-Side Rendering** means building the DOM tree of a web page directly in the browser, by executing a script that the browser receives from the server.
 
 ## Context
-As of the time of writing this document, the new Ensembl site is rendered entirely client-side, which means that no matter which page is requested, the browser always receives the same minimal html payload with a link to the bootstrap javascript file. That file, in its turn, will trigger the download of the rest of the required javascript files, which will then fetch the necessary data, and render the page.
+As of the time of writing this document, the new Ensembl site is rendered entirely client-side, which means that no matter which page is requested, the response is always a minimal html payload with a link to the bootstrap javascript file. That file, in its turn, will trigger the download of the rest of the required javascript files, which will then fetch the necessary data, and finally show the contents of the page on the screen.
 
 ## Motivation
 ### Shortcomings of client-side rendering
 #### Loss of control over HTTP status codes
-With client-side rendering, the browser always gets the same html file, in a response with an HTTP status code `200: OK`. Thus, there is no way to signal, from the server, the intent to redirect the browser to a different page (300-range codes) or to message that the requested page does not exist (400-range errors).
+With client-side rendering, the browser always gets the same html payload, in a response with an HTTP status code `200: OK`. Thus, there is no way to signal, from the server, the intent to redirect the browser to a different page (300-range codes) or to message that the requested page does not exist (400-range errors).
 
-_**Note:** Use of improper http status codes can negatively impact search engine ranking. Google’s recommended solution for avoiding what it calls "soft 404 errors" (404-error pages with status code 200) is either to use javascript to send the browser to a different url with an actual 404-error page, or to add a `<meta name="robots" content="noindex">` tag in order to prevent search engine's web crawler from indexing the page._
+_**Note:** Use of improper http status codes can negatively impact search engine ranking by creating what Google calls "soft 404 errors" (404-error pages with status code 200). Google’s advice for client-side-rendered sites in this case is either to do a client-side redirect to a different url containing an actual 404-error page, or to add a `<meta name="robots" content="noindex">` tag to the page in order to prevent search engine's web crawler from indexing it._
 
 #### Negative impact on search engine discoverability and ranking
 While Google is very open about the capabilities of the modern version of its web crawler (Googlebot), which can execute client-side javascript, we do not know to what extent other web crawlers can do the same.
 
-##### Experiment (conducted in September 2020)
-HGNC's site `genenames.org` is an Angular-based site, with its metadata tags filled in dynamically in the browser. The html layout returned from the server has the following metatags (note that they are empty and contain Angular bindings to be activated after Angular is loaded in):
+##### Example
+As of September 2020, HGNC's site `genenames.org` is an Angular-based site, with its metadata tags filled in dynamically in the browser. The html layout returned from the server has the following metatags (note that they are empty and contain Angular bindings for the data that is loaded on the client side):
 
 ```html
 <title ng-bind="vm.headTitle"></title>
@@ -35,9 +35,9 @@ A sample query `brca2 site:https://www.genenames.org/` in different search engin
 This experiment confirms that Google is capable of indexing client-side-rendered web pages, but that other search engines struggle. With server-side-rendered pages, all search engines should be perfectly capable of indexing them.
 
 #### Lack of contextual information for 3rd-party bots
-When someone shares a url to a web page on social media or in a messaging app such as Slack, this service will often have a web crawler follow the shared url and scrape the data from the page, such as its title, its description, and possibly an image, to be used for a preview. These crawlers do not execute javascript; so with client-side rendering, they will not be able to retrieve the title and the description of the shared page.
+When someone shares a url to a web page on social media or in a messaging app such as Slack, this service will often send a web crawler to the shared page to scrape its data (such as its title, description, and possibly an image) for a preview. These aren't sophisticated crawlers, and they do not execute javascript; so with client-side rendering, they will not be able to retrieve the title and the description of the shared page.
 
-### Additional benefits of server-side rendering
+### Additional advantages of server-side rendering
 The points below are expected to offer only a marginal benefit in the context of the new Ensembl website, and are mentioned only for the sake of completeness.
 
 #### Higher page reliability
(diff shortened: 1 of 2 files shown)


jgtate:
Is it possible to estimate the amount of work involved with implementing a custom rendering server versus a Next.js-based implementation of server-side rendering? How much does the opinionated nature of Next.js tip the balance towards a custom renderer?

Intuitively, it feels like implementing our own renderer should be more difficult and more time-consuming than using a third-party tool. It will also add presumably a significant volume of code that will require on-going maintenance. However, if we could find, six months after implementing Next.js, that we can't do something essential, maybe that extra workload is justified.

azangru:
> Is it possible to estimate the amount of work involved with implementing a custom rendering server versus a Next.js-based implementation of server-side rendering? How much does the opinionated nature of Next.js tip the balance towards a custom renderer?

For each of those, there is its own set of known unknowns and unknown unknowns.

For Next.js, the unknowns are clustered around the question, what happens when you want your rendering server to do something else as well. E.g. proxy a subset of requests to other servers, add custom headers to response, run custom logger at every request. There's a section on [using Next.js with a custom server](https://nextjs.org/docs/advanced-features/custom-server) in the docs, but I have never before tried that. In any case, I expect the experiments around Next to only take several days.

For a custom implementation, the unknowns are around the best ways to code split ([React's stock solution doesn't work server-side at the moment, but a blessed alternative exists](https://reactjs.org/docs/code-splitting.html#reactlazy)), to fetch data server-side, and achieving good developer experience with proper code updates without having to rebuild the project. It should probably take around a week to get a clearer picture.

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Higher page reliability
+With client-side rendering, if anything happens to the javascript bundle, the page will not render. In contrast, with server-side rendering, the browser will display the contents of the page in the way it was rendered on the server.
+
+_**However**, we recognize that in the case of the Ensembl site, there is little value in being able to view the page, but not interact with it._
+
+#### Perception of faster page load (not guaranteed)
+While the time to the first byte with server-side rendering will inevitably be slower than with client-sidee rendering, it is, in principle, possible that the time it takes to build a page on the server and deliver it to the browser is less than what it would take a client-rendered page to fetch the script, then fetch the data, and then finally display the page content.
+
+_**However**, we do recognize that in practice, these gains, if at all attainable, will be negligible. We expect that Ensembl users are bioinformaticians or clinicians accessing Ensembl from office computers, which are sufficiently powerful and are on sufficiently fast networks that performance gains will not be meaningful._
```
andrewyatz, line 51: LMIC users might not be in as such a good place with their machines or internet

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
```diff
@@ -0,0 +1,149 @@
...
+### Additional advantages of server-side rendering
+The points below are expected to offer only a marginal benefit in the context of the new Ensembl website, and are mentioned only for the sake of completeness.
+
+#### Higher page reliability
+With client-side rendering, if anything happens to the javascript bundle, the page will not render. In contrast, with server-side rendering, the browser will display the contents of the page in the way it was rendered on the server.
+
+_**However**, we recognize that in the case of the Ensembl site, there is little value in being able to view the page, but not interact with it._
+
+#### Perception of faster page load (not guaranteed)
+While the time to the first byte with server-side rendering will inevitably be slower than with client-sidee rendering, it is, in principle, possible that the time it takes to build a page on the server and deliver it to the browser is less than what it would take a client-rendered page to fetch the script, then fetch the data, and then finally display the page content.
```
andrewyatz, line 49: `client-sidee`

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: Absolutely, but the way we are building ensembl-2020 is effectively disregarding LMICs.

- we are not, while building the site, routinely checking its performance on lower-end hardware
- we are not checking its performance on slow connections
- we know that for some applications, e.g. genome browser, we are expecting the client to have a pretty fast connection (lots of data to download)

If LMIC are really a concern, something needs to change about our development process. But I believe our target audience, even in such countries, are on reasonably powerful desktops/laptops and on a decent internet connection rather than on underpowered phones on patchy 3G connection. For the latter group, which Google likes to describe as a likely target for modern web developers, our site will perform poorly.

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Motivation
+### Shortcomings of client-side rendering
+#### Loss of control over HTTP status codes
+With client-side rendering, the browser always gets the same html payload, in a response with an HTTP status code `200: OK`. Thus, there is no way to signal, from the server, the intent to redirect the browser to a different page (300-range codes) or to message that the requested page does not exist (400-range errors).
+
+_**Note:** Use of improper http status codes can negatively impact search engine ranking by creating what Google calls "soft 404 errors" (404-error pages with status code 200). Google’s advice for client-side-rendered sites in this case is either to do a client-side redirect to a different url containing an actual 404-error page, or to add a `<meta name="robots" content="noindex">` tag to the page in order to prevent search engine's web crawler from indexing it._
+
+#### Negative impact on search engine discoverability and ranking
```
ens-st3, line 20: Search engine visibility is considered strategically important

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
```diff
@@ -0,0 +1,149 @@
+# 16. Server-Side Rendering
```
ens-st3, line 1: We're now up to 17

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Status
+Under review.
```
ens-st3, line 4: Published

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: The reason I put search visibility second is because:

- Google is capable of indexing purely client-side-rendered sites
- Even if we server-side-render, the only thing we should be able to really guarantee is that the html in the response will contain the appropriate meta tags in the `head` of the html document (title, description, maybe ld-json if there exists a schema on schema.org fitting for our use case). For pages that are mostly text, e.g. help&docs, we can easily make a case that they should be rendered server-side; but entity viewer, which is strategically the most important, is tricky.

But I'll move this section up, sure.

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: ✔️

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: Moved this section above the previous.

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: Updated.

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
azangru: ✔️

ens-st3 (changes requested):
Can you add the Consequences section ?

src/ensembl/doc/architecture/decisions/0016-server-side-rendering.md
andrewyatz: I'm not sure it is disregarding LMICs per-say, but it is discounting those who have the issues you outlined and they may disproportionally affect LMIC researchers. Point 1 is a hard one to resolve so I'm not going to address this.

Points 2 and 3 though. Dodgy network connections can happen anywhere. I have them time to time in my own home. It is a fact of life not the exception. Data locality will help and drives the point for CDN like behaviour. Also whilst yes we will be shipping a lot of data to the client, I've not seen the volumes to be excessive to the point I'm worried.

azangru: > Dodgy network connections can happen anywhere. I have them time to time in my own home. It is a fact of life not the exception.

Yes, I agree.

Server-side rendering brings with it a couple of UI-related concerns that I can think of:
- Time between when the contents of the page appear in the browser (defined by when the HTML document reaches the browser) and when they become interactive (extra time for a round trip for javascript). On slow networks, this may become noticeable and is an unpleasant experience [see the demo](https://youtu.be/0A-2BhEZiM4?t=337)
- Some of our data that's important for rendering is stored on the client side and is unavailable on the server (e.g. the list of selected species that we show as species lozenges in the top area of the site).

It is likely then that we will only be able to introduce some minimal aspects of server-side rendering, such as populating the `<head>` of the html document with relevant data and maybe showing a rough skeleton of the page, which we will then fill with data when javascript is loaded and relevant data has been requested. It should be better than looking at a white screen anyway. We need to bring Andrea on board with this additional design complexity.

cc @ens-st3

jgtate (approved):
I'm still nervous about the problems we might eventually encounter with the various solutions but I support the proposal to start with building prototypes to narrow down the uncertainties and inform the final decision.

ens-st3 (approved):

PR #354, opened 2020-09-27
Fix checkbox styles

azangru:
## Type
- Bug fix

## Description
Missed in a previous PR (https://github.com/Ensembl/ensembl-client/pull/303), where we updated Checkbox styles. The checkbox wrapper currently has two columns with the widths defined as follows:

```css
.checkboxHolder {
  display: grid;
  grid-template-columns: 27px 100%;
}
```

What this means is that the right column will have the same width as the whole container of the checkbox wrapper, as can be easily seen in the BrowserTrackConfig component:

image

image

Notice how the right column of the wrapper's grid continues outside the content box of its parent component.

Here's the behaviour after the fix:

image

image

Notice how the grid stays within the content box of the parent element.

### Deployment URL
http://fix-checkbox.review.ensembl.org

### DIFF
--- a/src/ensembl/src/shared/components/checkbox/Checkbox.scss
+++ b/src/ensembl/src/shared/components/checkbox/Checkbox.scss
@@ -2,7 +2,8 @@
 
 .checkboxHolder {
   display: grid;
-  grid-template-columns: 27px 100%;
+  width: 100%;
+  grid-template-columns: 27px 1fr;
 }
 
 .defaultCheckbox {


ens-ma7:
Hi @azangru do you have a deploy URL by any chance, i just want to see how the checkboxes in the filter are affected.
Thanks

azangru:
@ens-ma7 http://fix-checkbox.review.ensembl.org/

ens-ma7 (approved):

PR #355, opened 2020-09-27
Do not mutate redux state for Custom Download

azangru:
## Type
- Bug fix

### 1. Redux state mutations
If you run Ensembl client in development mode, and switch to Custom Download, you will see an error screen and multiple errors in the console:

image

The cause of the error is that the code uses the lodash's `set` function, which mutates the source object, both in CustomDownload-related redux actions, and in the ContentBuilder component.

This PR is a quick fix to these errors. It changes the regular lodash's `set` to the `set` function from `lodash/fp`, which creates a copy of the source object instead of mutating it. We shall revisit this later when we return to CustomDownload.

### 2. Redux wrapper re-rendering during the render of the `ContentBuilder`:

If you navigate to the initial CustomDownload screen and press the Gene/Transcripts button:

image

you will see an error/warning in the console about how `react-redux`'s component (`ConnectFunction`) tried to re-render while the `ContentBuilder` component was rendering:

image

This PR wraps the update function in a setTimeout to move it to the end of the event loop queue.

### DIFF
--- a/src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
+++ b/src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
@@ -15,7 +15,7 @@
  */
 
 import React from 'react';
-import set from 'lodash/set';
+import set from 'lodash/fp/set';
 import get from 'lodash/get';
 import classNames from 'classnames';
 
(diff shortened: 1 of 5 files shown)


src/ensembl/src/content/app/custom-download/components/content-builder/ContentBuilder.tsx
```diff
@@ -166,18 +159,25 @@ const ContentBuilder = (props: ContentBuilderProps) => {
...
+          [...currentPath, option.id],
+          true,
+          newSelectedData
+        );
         shouldUpdateSelectedData = true;
       }
     });
     if (shouldUpdateSelectedData) {
-      props.onChange(newSelectedData);
+      // Rhe setTimeout below is a temporary hack.
```
msmanoj, line 177: Rhe -> The

msmanoj (approved):

PR #356, opened 2020-10-01
Use thoas as data source for EntityViewer

azangru:
## Type
- Refacforting

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-623
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-775

## Description
Moving from the toy graphql server over to Thoas as data source for EntityViewer.

### Most notable changes
Entity types got updated.

Briefly, the relationships between the entities can now be diagrammed like this:
- `gene → transcript → spliced_exon → exon`, where transcript has relative location with coordinates relative to the gene, and spliced_exons have relative location with coordinates relative to the transcript
- `gene → transcript → product_generating_context → ( cds, cdna, product )`, where a single transcript has a one-to-many relationship with product-generating contexts.

For the full description of the schema, see the docs section in the graphiql interface of [Thoas deployment](http://web-thoas.review.ensembl.org) or schema docs in the [Thoas repository](https://github.com/Ensembl/ensembl-thoas/tree/master/common/schemas).

### Fixed bugs:
- Reverse-strand genes are now displaying their transcripts properly. Example: human BRAF gene (ENSG00000157764).

### Deleted
- Storybook stories for `UnsplicedTranscripts` and `ProteinDomainImage` components. These are one-off components that don't get repeated in other views; and while Storybook helped us develop them, we now have them in the main code. There's no need to keep these stories anymore.
- Rest fetchers and adaptors that we used in EntityViewer:
  - geneData (containing the fetchGene function)
  - rest-gene-adaptor
  - rest-transcript-adaptor

## Deployment URL
http://use-thoas.review.ensembl.org/

## Views affected
EntityViewer

PR #358, opened 2020-10-05
gene stable_id will have version, so dont concatenate

jyothishnt:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-725

## Description
Stable IDs will have version added to it by default in Thoas. So remove previously concatenated version.

## Views affected
Entity viewer


azangru:
This change is already present in PR #356 😄 
https://github.com/Ensembl/ensembl-client/pull/356/files#diff-223addf712ecdd3e3c886e998f8a70acR100

Is it ok if I close this PR?

jyothishnt:
Ah ok fine. I will close it. Thanks.

PR #359, opened 2020-10-05
Add selected transcript to protein view

jyothishnt:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-769

## Description
Add selected transcript info in the left column in protein list view 
We may need to work on the alignment of each row. But I would like to know your comments.

## Deployment URL
http://proteinview-selected-transcript.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000139618?view=protein

## Views affected
Entity viewer - protein list view


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -48,6 +48,16 @@ export type DefaultTranscriptListItemProps = {
...
 
+export const defaultTranscriptLabelMap = {
+  selected: {
+    label: 'Selected',
+    helpText:
+      'The selected transcript is a default single transcript per protein coding gene that is representative of biology, well-supported, expressed and highly conserved'
+  }
+};
+
+export const canonicalType = 'selected'; // TODO Change this to transcript.mane/plus etc when available
```
azangru, line 59: To reviewers: see if you'd prefer this to be a separate component.

To Jyo: see if you can figure out why transcript images are at a different height from the text.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -61,24 +65,14 @@ export const DefaultTranscriptListItem = (
...
     <div className={styles.defaultTranscriptListItem}>
       <div className={transcriptsListStyles.row}>
         {props.isDefault && (
           <div className={styles.defaultTranscriptLabel}>
-            <span>{defaultTranscriptLabelMap[canonicalType]?.label}</span>
+            <span>{DefaultTranscriptLabelMap[CanonicalType]?.label}</span>
             <QuestionButton
-              helpText={defaultTranscriptLabelMap[canonicalType]?.helpText}
+              helpText={DefaultTranscriptLabelMap[CanonicalType]?.helpText}
             />
```
azangru, line 76: I am confused. The `DefaultTranscript.tsx` file is not a React component (as it currently stands, it doesn't even need to be a tsx file; it could be a ts if there's no React code there). Wasn't the React code above also supposed to go to that component?

Also, it would be probably misleading to call that component `DefaultTranscript`. What is it that this label is about? Is it `TranscriptConfidenceLabel` or something?

jyothishnt: yes there is no react code so I should be using .ts. BTW I didnt understand your comment on moving the above react code.

Naming - atm the config contains label and helptext for different transcript types. It shows the confidence/quality of that transcript, so what would be the best name? open to suggestions

azangru: > I didnt understand your comment on moving the above react code.

I thought this bit of code:

```
          <div className={styles.defaultTranscriptLabel}>
            <span>{DefaultTranscriptLabelMap[CanonicalType]?.label}</span>
            <QuestionButton
              helpText={DefaultTranscriptLabelMap[CanonicalType]?.helpText}
            />
          </div>
```

was going to be extracted as a component. It would receive transcript as a prop, and would have the auxiliary code to decide which label to show for such transcript and which text to pass to the QuestionButton.

jyothishnt: Ok got it.

src/ensembl/src/content/app/entity-viewer/shared/components/default-transcript-label/DefaultTranscriptLabel.tsx
```diff
@@ -0,0 +1,45 @@
...
+const transcriptLabel = {
+  selected: {
+    label: 'Selected',
+    helpText:
+      'The selected transcript is a default single transcript per protein coding gene that is representative of biology, well-supported, expressed and highly conserved'
+  }
+};
+
+export const DefaultTranscriptLabel = (props: DefaultTranscriptLabelProps) => {
+  const CanonicalType = props.transcript && 'selected'; // TODO Change this to props.transcript.mane/plus etc when available
```
jyothishnt, line 37: eslint was complaining about props not being used. So just added this props.transcript here. Will have to change accordingly when we get the actual data from Thoas.

azangru: Ok, this looks good; I just have several minor nitpicking remarks:

- We don't usually use pascal case (as in `CanonicalType`) for variable names
- The word `canonical` refers to only a single transcript. I.e. one and only one transcript among all gene transcripts will be designated as canonical. Therefore, the name "canonical type" is somewhat misleading. I think what you are after is `transcriptSupportLevel`

jyothishnt: transcriptSupportLevel or transcriptSupportLabel?

azangru: I was thinking on level.

On the other hand, googling now, I see that transcript support level is what gave the acronym TSL (TSL-1, TSL-2, etc). Ensembl seems to be struggling with proposing a generic term for all of that (see https://www.ensembl.org/info/genome/genebuild/transcript_quality_tags.html). The url contains a good suggestion — `transcriptQuality`. In the [docs](https://github.com/Ensembl/ensembl-e2020-datafiles/blob/master/docs/genes_and_transcripts.md) to the script that builds data files for genome browser we are calling it "transcript designation". I think Dan used the word "prestige"; not sure whether he is still using it.

jyothishnt: transcriptQuality sounds simple and easy to understand, but @ens-st3 do you have any suggestions?

azangru (approved):
`transcriptLabelMap[transcriptQuality]?.label` will return `undefined` if there is no `transcriptLabelMap[transcriptQuality]`. I would argue that in this case, `TranscriptQualityLabel` should not even render (should return null). I've pushed a commit with this change not to delay the PR further.

PR #366, opened 2020-10-15
New shared mixin css

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-774

## Description
Create a new shared mixin for lozenges, roundbuttons, etc

## Deployment URL
We dont have one yet as it is still in storybook and not used in any components.

## Views affected
SS, EV and GB


src/ensembl/src/styles/_mixins.scss
```diff
@@ -0,0 +1,16 @@
+$selectedSpeciesBorderRadius: 20px;
+$selectedSpeciesPadding: 6px 20px;
+$deleteButtonPadding: 8px 20px 8px 28px;
+
+@mixin selected-species-common {
```
azangru, line 5: I was thinking more in terms of a mixin that's the same for RoundButton, SelectedSpecies, and whatever it is we'll call this component:

image

If that is the case, the mixin shouldn't be called `selected-species-common`, but rather by the shape that it describes (`lozenge`?).

We probably need to confirm with Andrea that SelectedSpecies and RoundButton have the same shape. Perhaps their height is different, in which case the mixin can accept a `height` parameter.

jyothishnt: yes agreed. I will change that name to lozenge

jyothishnt:
All passed in gitlab. So good to review.

azangru:
Aren't you going to update the RoundButton to use the mixin?

src/ensembl/src/styles/_mixins.scss
```diff
@@ -0,0 +1,20 @@
...
+  display: inline-block;
+  padding: map-get($lozengeDefaults, "padding");
+  border-radius: $borderRadius;
+  line-height: $lineHeight;
+  user-select: none;
+  margin-bottom: 4px;
+
+  &:not(:last-child) {
+    margin-right: 12px;
+  }
```
azangru, line 19: That's not the responsibility of a lozenge

jyothishnt: I agree, it should be removed

src/ensembl/src/styles/_mixins.scss
```diff
@@ -0,0 +1,16 @@
...
+);
+
+@mixin lozenge($lineHeight: map-get($lozengeDefaults, "line-height"), $borderRadius: map-get($lozengeDefaults, "border-radius")) {
+  display: inline-block;
+  padding: map-get($lozengeDefaults, "padding");
+  border-radius: $borderRadius;
+  line-height: $lineHeight;
+  user-select: none;
+  margin-bottom: 4px;
+}
```
azangru, line 16: Have you tried optional arguments and keyword arguments (https://sass-lang.com/documentation/at-rules/mixin)

E.g. why would we even need a map here? Isn't this simpler?

```
lozenge($lineHeight: 1, $borderRadius: 20px, $padding: 6px 20px) {}
```

Also, from the consumer perspective (`RoundButton.scss`), wouldn't named arguments be better:

```
@include lozenge($borderRadius: 15px);
```

hopefully this should help us avoid passing the same arguments as the default ones.

jyothishnt: yes true. Thanks.

azangru (approved):

PR #371, opened 2020-10-20
Update popular species button behaviour

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-790

## Description
Clicking on a popular species button for a committed species will open the species homepage instead of removing the species.

## Deployment URL
http://favourite-species.review.ensembl.org

## Views affected
Species selector -> Popular species

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.test.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.test.tsx
@@ -123,7 +123,7 @@ describe('<PopularSpeciesButton />', () => {
       );
     });
 
-    it('opens species homepage when it is clicked', () => {
+    it('opens species page when it is clicked', () => {
       const wrapper = mount(
         <PopularSpeciesButton {...commonProps} isCommitted={true} />
       ).find('.popularSpeciesButton');
@@ -131,7 +131,7 @@ describe('<PopularSpeciesButton />', () => {
       wrapper.simulate('click');
 
       expect(push).toHaveBeenCalledWith(
-        urlFor.speciesHomepage({
+        urlFor.speciesPage({
           genomeId: commonProps.species.genome_id
         })
       );
--- a/src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.test.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.test.tsx
@@ -17,6 +17,9 @@
 import React from 'react';
 import { mount, render } from 'enzyme';
 import set from 'lodash/fp/set';
+import { push } from 'connected-react-router';
+
+import * as urlFor from 'src/shared/helpers/urlHelper';
 
 import { PopularSpeciesButton } from './PopularSpeciesButton';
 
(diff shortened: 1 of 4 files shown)


src/ensembl/src/shared/helpers/urlHelper.ts
```diff
@@ -31,6 +31,16 @@ type EntityViewerUrlParams = {
   view?: string | null;
 };
 
+type SpeciesHomepage = {
+  genomeId: string;
+};
+
+export const speciesHomepage = (params: SpeciesHomepage) => {
+  const speciesHomepageRootPath = '/species';
```
imransl, line 39: When looking at these root paths being inside a function as a local variable, I wonder if it is better if we need to put them at a higher level in the same file.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
```diff
@@ -18,34 +18,32 @@ import React from 'react';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 import find from 'lodash/find';
+import { push } from 'connected-react-router';
 
+import Tooltip from 'src/shared/components/tooltip/Tooltip';
+import InlineSVG from 'src/shared/components/inline-svg/InlineSvg';
```
azangru, line 24: Why did you move these two imports? We usually try to group imports by their type; and now these two, which are components, are sitting together with helpres and hooks.

azangru:
We have a bit of a naming confusion. I know that Andrea likes to talk about "species home page". In our code, it is currently referred to simply as "species page". Is there any reason to add the word "home" to it? Are there other species pages we are expecting? If not, I would advise against using the phrase 'species home page' in our code.

@ens-st3 might have an opinion.

src/ensembl/src/shared/helpers/urlHelper.ts
azangru: That would help if there were multiple functions interested in the same path segments, but there aren't.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
msmanoj: I've put it back. I'm still getting confused with the order. Can you please remind me the correct order once again?

msmanoj:
> We have a bit of a naming confusion. I know that Andrea likes to talk about "species home page". In our code, it is currently referred to simply as "species page". Is there any reason to add the word "home" to it? Are there other species pages, as opposed to 'home', that we are expecting? If not, I would advise against using the phrase 'species home page' in our code.
> 
> @ens-st3 might have an opinion.

Fair enough. I have renamed it to `speciesPage` for now.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.tsx
azangru: I'm usually doing it like this:

1) libraries from node modules
2) non-components (redux actions or selectors, helpers, hooks), tend to group them by type
3) react components
4) typescript types (if not imported together with imports in one of the previous groups)
5) styles and, on occasion other static assets, such as images (not the svgs that we import as React components though; those go into the react components group).

What are your conventions?

msmanoj: Thanks. I think I didn't separate non-component and react component and kept mixing them up. It's all clear now.

src/ensembl/src/content/app/species-selector/components/popular-species-button/PopularSpeciesButton.test.tsx
```diff
@@ -118,13 +123,19 @@ describe('<PopularSpeciesButton />', () => {
       );
     });
 
-    it('deletes committed species when clicked', () => {
+    it('opens species homepage when it is clicked', () => {
```
azangru, line 126: homepage -> page?

msmanoj: That's done

azangru (approved):

PR #387, opened 2020-11-05
open external links in new tab

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-811

## Views affected
Where ever we have external links

### DIFF
--- a/src/ensembl/src/shared/components/external-link/ExternalLink.tsx
+++ b/src/ensembl/src/shared/components/external-link/ExternalLink.tsx
@@ -38,7 +38,7 @@ const ExternalLink = (props: ExternalLinkProps) => {
   return (
     <span className={styles.container}>
       <LinkIcon className={iconClass} />
-      <a className={linkClass} href={props.to}>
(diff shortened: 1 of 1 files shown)


src/ensembl/src/shared/components/external-link/ExternalLink.tsx
```diff
@@ -38,7 +38,7 @@ const ExternalLink = (props: ExternalLinkProps) => {
   return (
     <span className={styles.container}>
       <LinkIcon className={iconClass} />
-      <a className={linkClass} href={props.to}>
+      <a className={linkClass} href={props.to} target="_blank" rel="noreferrer">
```
azangru, line 41: Noreferrer is great! Perhaps `nofollow noreferrer` (to instruct Google crawler not to follow this link and thus not to go to the other site) would be even better?

jyothishnt: yup I can add that.

azangru (approved):

PR #394, opened 2020-11-12
Show only text for External Ref without URL

imransl:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-812

## Description
The ExternalLink component always shows the text passed to it as a link, even when there is no URL for it (`to` prop). This PR fixes that by showing the text as a link only if there is a URL for it.

## Views affected
Entity viewer -> sidebar panel

### Other effects
- [x] I have created/modified any required tests

### DIFF
--- a/src/ensembl/src/shared/components/external-link/ExternalLink.scss
+++ b/src/ensembl/src/shared/components/external-link/ExternalLink.scss
@@ -13,7 +13,7 @@
   position: relative;
 }
 
-.link {
+.text {
   display: inline-block;
   grid-column: link;
   position: relative;
--- a/src/ensembl/src/shared/components/external-link/ExternalLink.scss
+++ b/src/ensembl/src/shared/components/external-link/ExternalLink.scss
@@ -13,7 +13,7 @@
   position: relative;
 }
 
-.text {
+.link {
   display: inline-block;
   grid-column: link;
   position: relative;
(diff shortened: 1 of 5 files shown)


azangru:
But why?

If an external reference does not have a link, then we shouldn't use the ExternalLink component for it. The ExternalLink component should only render links, not non-links.

imransl:
> But why?
> 
> If an external reference does not have a link, then we shouldn't use the ExternalLink component for it. The ExternalLink component should only render links, not non-links.

I was wondering what to do about this. Whether to create a new component called ExternalText or just modify the ExternalLink to make it decide how to display the link. I went ahead with the latter. So do you think we should create that new component then?

azangru:
Can't this be handled by the `ExternalReference` component? So that it calls the `ExternalLink` component if the reference has a link, or renders the text for the reference if there is no link.

imransl:
@azangru We still need ExternalLink's class and icon that are inside it. We can duplicate all that inside ExternalReference, but don't think it'll be a bit messy.

azangru:
> We still need ExternalLink's class and icon that are inside it

Do we? Do we have an example of what an external reference without a link should look like?

imransl:
@azangru I don't think we have. But we don't need to show the icon for an external reference text as Andrea confirmed. So I'll just add it in the `ExternalReference` component as you suggested.

azangru:
Turns out we have a reference design:



As Andrea said, `font-weight: $light` and a left margin of 7px-10px.

imransl:
@azangru I've added the styling in the latest commit: 439279c

azangru (approved):

msmanoj (approved):

PR #395, opened 2020-11-13
Image width constant

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-800

## Description

Use a constant for image width importable both in the script and the style files

## Deployment URL
http://image-width-constant.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_14/gene:ENSG00000139618

## Views affected
Entity viewer
 - transcript, gene overview, protein images

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
@@ -46,7 +46,7 @@ export type DefaultTranscriptListItemProps = {
   toggleTranscriptInfo: (id: string) => void;
 };
 
-// NOTE: the width of the middle column is the same as the width of GeneOverviewImage, i.e. $image_width in global settings
+// NOTE: the width of the middle column is the same as the width of GeneOverviewImage, i.e. $gene_image_width in EV constants
 
 export const DefaultTranscriptListItem = (
   props: DefaultTranscriptListItemProps
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
@@ -46,8 +46,6 @@ export type DefaultTranscriptListItemProps = {
   toggleTranscriptInfo: (id: string) => void;
 };
 
-// NOTE: the width of the middle column is the same as the width of GeneOverviewImage, i.e. $gene_image_width in EV constants
-
 export const DefaultTranscriptListItem = (
   props: DefaultTranscriptListItemProps
 ) => {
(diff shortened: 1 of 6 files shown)


jyothishnt:
Issue with deployed review app due to Thoas. https://genomes-ebi.slack.com/archives/CC2RKR525/p1605266385181400?thread_ts=1605265485.179600&cid=CC2RKR525

src/ensembl/src/styles/_settings.scss
```diff
@@ -16,6 +16,9 @@ $global-radius: 0;
 $global-padding-left: 100px;
 $global-button-cursor: auto;
 
+// Image configurations
+$image_width: 695;
```
azangru, line 20: Oh, that's far too global! I think we need to put it in some variables file inside the entity viewer folder. [This one](https://github.com/Ensembl/ensembl-client/tree/dev/src/ensembl/src/content/app/entity-viewer/gene-view/styles) looks appealing.

The name image_width is also not particularly informative. `gene_image_width`?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.tsx
```diff
@@ -46,7 +46,7 @@ export type DefaultTranscriptListItemProps = {
   toggleTranscriptInfo: (id: string) => void;
 };
 
-// NOTE: the width of the middle column is the same as the width of GeneOverviewImage, i.e. 695px
+// NOTE: the width of the middle column is the same as the width of GeneOverviewImage, i.e. $gene_image_width in EV constants
```
azangru, line 49: We can probably safely delete this comment; it doesn't seem to be communicating anything useful.

src/ensembl/src/content/app/entity-viewer/gene-view/styles/_constants.scss
```diff
@@ -1,5 +1,13 @@
+$gene_image_width: 695;
```
azangru, line 1: To avoid multiplying by one pixel in other scss files (which looks odd there) we should probably confine all the ugliness just to this file, by doing something like:

```scss
$gene_image_width_number: 695;
$gene_image_width: #{695}px;

...

:export {
  gene_image_width: $gene_image_width_number;
}
```

then, I hope, you'll be able to remove `* 1px` from the scss files that are using the `$gene_image_width` variable.

jyothishnt: Agree about having it in one place. The reason I changed it to this is because I read about sting interpolation in scss is a bad idea and it has a powerful unit arithmetic that you can use instead  https://sass-lang.com/documentation/interpolation

azangru: You can do the multiplication if you prefer, but why is the interpolation a bad idea? I haven't heard about this.

src/ensembl/src/content/app/entity-viewer/gene-view/styles/_constants.scss
```diff
@@ -1,5 +1,14 @@
...
 $left_column: 160px;
 $middle_column_left_gap: 18px;
-$middle_column: 695px;
+$middle_column: $gene_image_width;
 $middle_column_right_gap: 27px;
 $right_column: 180px;
+
+/* stylelint-disable */
+:export {
+  gene_image_width_value: $gene_image_width_value;
```
azangru, line 12: wouldn't you prefer to export it under the name gene_image_width (without `value`)?

jyothishnt: OK

azangru (approved):

PR #398, opened 2020-11-17
Ensure species is enabled

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-867

## Description
- Fixed the issue mentioned in the ticket by ensuring that the species is enabled when it is accessed using the direct url

## Deployment URL
http://missing-species-tab.review.ensembl.org

## Views affected
- Species tabs across apps

### DIFF
--- a/src/ensembl/src/shared/state/genome/genomeActions.ts
+++ b/src/ensembl/src/shared/state/genome/genomeActions.ts
@@ -32,10 +32,7 @@ import {
   getGenomeTrackCategories,
   getGenomeKaryotype
 } from 'src/shared/state/genome/genomeSelectors';
-import {
-  ensureSpeciesIsCommitted,
-  ensureSpeciesIsEnabled
-} from 'src/content/app/species-selector/state/speciesSelectorActions';
+import { ensureSpeciesIsCommitted } from 'src/content/app/species-selector/state/speciesSelectorActions';
 import { getBrowserActiveGenomeId } from 'src/content/app/browser/browserSelectors';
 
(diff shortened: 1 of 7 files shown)


src/ensembl/src/shared/state/genome/genomeActions.ts
```diff
@@ -51,6 +54,7 @@ export const fetchGenomeData = (
   ]);
 
   dispatch(ensureSpeciesIsCommitted(genomeId));
+  dispatch(ensureSpeciesIsEnabled(genomeId));
```
azangru, line 57: I feel a bit uncomfortable about this. Logically, making sure that a species is enabled should be coupled with setting the active genome id, not with fetching data for genome.

msmanoj: That's true. Do you know if there is a better option than calling this function separately from all the apps?

azangru: Some tests are failing now.

Also, why did you add it in the react component for genome browser, but in a thunk action for entity viewer?

msmanoj: Fixed the test.

>Also, why did you add it in the react component for genome browser, but in a thunk action for entity viewer?

I just thought of adding it next to where `fetchGenomeData` is called.

azangru (approved):

PR #401, opened 2020-11-24
Single source xrefs

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-866

## Description
- Multiple xrefs from the same group are now displayed within an accordion
- If the xref source `description` and `accession_id` matched, the label will be hidden
- If the xref source `description` and the xref group name matched, the label will be hidden
- Fixed a minor UI issue with the gene details div where the name was getting wrapped when there is enough space.
Screenshot 2020-11-24 at 11 50 05

## Deployment URL
https://single-source-xrefs.review.ensembl.org

## Views affected
Entity viewer -> Sidebar -> External references

### Other effects


azangru:
Sorry if this is unrelated. I was checking how the styles work, and found in the overview panel that our external reference component seems to have child grid-associated styles on it (`grid-column` on the screenshot below) even though there is no corresponding grid parent. This looks wrong to me:

image

azangru:
Would you be able to change this section in the top of the external references panel so that it's the same as for the overview panel (see XDs), or do you feel it's better done in a separate PR?

image

msmanoj:
> Would you be able to change the top section of the external references panel so that it's the same as for the overview panel (see XDs and the screenshot fragment below), or do you feel it's better done in a separate PR?
> 
> image

Do you mean by using two spans instead of grid?

msmanoj:
> Sorry if this is unrelated. I was checking how the styles work, and found in the overview panel that our external reference component seems to have child grid-associated styles on it (`grid-column` on the screenshot below) even though there is no corresponding grid parent. This looks wrong to me:

- Removed it

azangru:
> Do you mean by using two spans instead of grid?

I mean to use gene symbol instead of the xref that we seem to be using instead.

msmanoj:
> I mean to use gene symbol rather than the xref that we seem to be using instead.

That's done

msmanoj:
> I mean to use gene symbol rather than the xref that we seem to be using instead.

That's done

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
```diff
@@ -45,6 +45,7 @@ const QUERY = gql`
     gene(byId: { stable_id: $stable_id, genome_id: $genome_id }) {
       name
```
azangru, line 46: is this redundant then?

msmanoj: Ah yes! I thought I removed it 🤦

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.scss
```diff
@@ -14,7 +14,7 @@
 
 .geneDetails {
   display: grid;
-  grid-template-columns: [gene-symbol] min-content 10px [stable-id] auto;
+  grid-template-columns: [gene-symbol] minmax(min-content, max-content) 10px [stable-id] auto;
```
azangru, line 17: Sorry, I keep getting confused by `minmax`. Do you know how `minmax(min-content, max-content)` will behave here, and is it any different from `auto`?

msmanoj: I tried `auto` but I was getting something like this:
Screenshot 2020-11-26 at 15 18 50


`minmax(min-content, max-content)`:

Screenshot 2020-11-26 at 15 19 01

azangru: Ah! I see. Just tried, and it would have worked with `[gene-symbol] auto [stable-id] 1fr;`

I don't have particularly strong feelings about the minmax — see for yourself whether you prefer that or the variant with auto, but `10px` though — that should be grid-column-gap, shouldn't it?

azangru: By the way, do you remember why we are using a grid for this element? Why not just an inline-block next to an  inline? Or not a flex with flex-wrap?

msmanoj: >Why not just an inline-block next to an inline? 

Not sure! The overview section uses two spans. I guess I might have tried to replicate what was on the XD where both the values are always on the same level.

azangru (approved):
If we are keeping `.geneDetails` as a grid, let's at least make it into a proper 2-column grid rather than 3-column, with the space between the columns coming from grid-columns-gap.

Otherwise – ✅

msmanoj:
> If we are keeping `.geneDetails` as a grid, let's at least make it into a proper 2-column grid rather than 3-column, with the space between the columns coming from grid-columns-gap.
> 
> Otherwise – ✅

Was just about to push that change. 
I noticed the following when I switch between overview and xrefs panels:
Gene details

Is is okay if I update the overview section to match with xrefs in this PR?

azangru:
> Is is okay if I update the overview section to match with xrefs in this PR?

Sure. Wouldn't you rather have just two spans in both panels? I'm still confused about the purpose of the grid.

msmanoj:
> > Is is okay if I update the overview section to match with xrefs in this PR?
> 
> Sure. Wouldn't you rather have just two spans in both panels? I'm still confused about the purpose of the grid.

Cool, will use two spans. I till it'll look better this way when it comes to longer values.

msmanoj:
> Sure. Wouldn't you rather have just two spans in both panels? I'm still confused about the purpose of the grid.

That's done

azangru (approved):
Looks good 👍

PR #405, opened 2020-11-27
Add label to the blue track representing the full protein in EntityViewer

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-786

## Description
Adding a label to the previously unlabelled protein track in EntityViewer (protein view):

image

## Deployment URL
http://protein-track-label.review.ensembl.org/

## Views affected
EntityViewer (protein view)

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-image/ProteinImage.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-image/ProteinImage.tsx
@@ -42,12 +42,12 @@ const ProteinImage = (props: ProteinImageProps) => {
     .range([0, props.width])
     .clamp(true);
 
-  const midStyles = transcriptsListStyles.middle;
-  const rightStyles = classNames(transcriptsListStyles.right, styles.label);
+  const trackContainerStyles = transcriptsListStyles.middle;
+  const labelStyles = classNames(transcriptsListStyles.right, styles.label);
 
   return (
     <div className={transcriptsListStyles.row}>
-      <div className={midStyles}>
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/protein-image/ProteinImage.tsx
```diff
@@ -42,7 +42,8 @@ const ProteinImage = (props: ProteinImageProps) => {
     .range([0, props.width])
     .clamp(true);
 
-  const midStyles = classNames(transcriptsListStyles.middle, styles.middle);
+  const midStyles = transcriptsListStyles.middle;
+  const rightStyles = classNames(transcriptsListStyles.right, styles.label);
```
msmanoj, line 46: Would it be better to rename these to something like `trackContainerStyles` and `labelStyles`?

azangru: I don't feel strongly one way or the other. Renamed them if you prefer these names.

msmanoj:
Just wondering what is the purpose of this label here?  It feels more like we should be displaying some value here.

azangru:
Andrea just wanted to give a name to what this track represents 😃 

For what it's worth, I also think that "amino acid length" or "protein length" is not a proper label, because this track represents the protein itself rather than its length. But it's up to Andrea, Andy or @ens-st3 to decide on the exact wording of this label if they want to change it 🤷‍♂️

PR #406, opened 2020-11-27
Fix for transcript order to match default order same as in transcript image

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-826

## Description
Matching the transcript order in the EV Sidebar in the external references to be the same as the default order in the transcript image. 

Note that the order doesnt change even if the transcript image order is changed by sorting. It will always be the default order

## Deployment URL
http://ev-xref-order-fix.review.ensembl.org

## Views affected
Entity viewer

- [ ] I have checked any requirements for Google Analytics
- [ ] I have created any required tests
- [ ] The PR has as its final commit a WASM/JS update commit if any Rust files were updated.

PR #408, opened 2020-11-30
Generalize parameter types in helper and sorter functions for transcripts

azangru:
## Related JIRA Issue(s)
Part of https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-879

## Description
We might have different components requesting different sets of fields for transcripts (compare the `GeneView` component and the `GeneExternalReferences` component, see https://github.com/Ensembl/ensembl-client/pull/406), but re-using the same helper functions or sorting functions. At the moment, helper functions and sorter functions expect their parameter to be of type `Transcript` as requested in the `GeneView` component, but that doesn't work for the `GeneExternalReferences` component. This PR is an attempt to describe more carefully the type requirements of these functions.

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
+++ b/src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
@@ -18,12 +18,56 @@ import {
   Slice,
   SliceWithLocationOnly
 } from 'src/content/app/entity-viewer/types/slice';
-import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
-import { Gene } from 'src/content/app/entity-viewer/types/gene';
+import { PhasedExon } from 'src/content/app/entity-viewer/types/exon';
 import {
   Product,
   ProductType
 } from 'src/content/app/entity-viewer/types/product';
+
+export type IsProteinCodingTranscriptParam = {
+  product_generating_contexts: Array<{
+    product_type: ProductType;
+  }>;
+};
+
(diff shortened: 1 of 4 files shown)


src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -18,13 +18,57 @@ import {
...
 } from 'src/content/app/entity-viewer/types/slice';
-import { Transcript } from 'src/content/app/entity-viewer/types/transcript';
-import { Gene } from 'src/content/app/entity-viewer/types/gene';
+import { PhasedExon } from 'src/content/app/entity-viewer/types/exon';
 import {
   Product,
   ProductType
 } from 'src/content/app/entity-viewer/types/product';
 
+export type IsProteinCodingTranscriptParam = {
```
msmanoj, line 27: Would it be better to define these types immediately before the respective functions so that it is more clear why it is named in such a way?

imransl (commented):
Looks good to me. I'd let @msmanoj approve this once his comment is addressed.

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
azangru: Don't really have a strong opinion. Moved the types around if that helps. They still look ugly as sin 😄

msmanoj (approved):

PR #411, opened 2020-12-03
Fix messaging to genome browser during first-time navigation to genome brower page

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-895

### Via UI (longer)
- Clear Storage
- Select Human
- View Human (species page)
- Click on view example gene in browser

### Directly through browser url (quicker)
- clear storage
- navigate directly to https://internal-2020.ensembl.org/genome-browser/homo_sapiens_GCA_000001405_28?focus=gene:ENSG00000139618

Problem in both scenarios above: empty genome browser canvas.

## Cause of bug
Direct navigation to genome browser page when active genome id has not yet been set brings us to [this conditional clause](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts#L109) in useBrowserRouting:

```
 dispatch(changeFocusObject(newFocusId as string));
```

However, `changeFocusObject` [will do nothing](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browserActions.ts#L351-L356) if active genome id has not yet been set. Therefore, we need to set active genome id before calling `changeFocusObject`.

## Deployment URL
http://fix-browser-routing.review.ensembl.org/

## Views affected
Genome Browser

## Possible complications
There is a comment in the code with an explicit warning that `changeFocusObject` needs to be called before `setDataFromUrlAndSave`. I cannot remember why this was the case. Need to dig in the code to find out what it is we might have broken.

### DIFF
--- a/src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
+++ b/src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
@@ -40,6 +40,7 @@ import {
 import {
   changeBrowserLocation,
   changeFocusObject,
+  setActiveGenomeId,
   setDataFromUrlAndSave
 } from '../browserActions';
 
@@ -101,11 +102,14 @@ const useBrowserRouting = () => {
       const newFocus = buildFocusIdForUrl(parseEnsObjectId(activeEnsObjectId));
       dispatch(replace(urlFor.browser({ genomeId, focus: newFocus })));
     } else if (focus && !chrLocation) {
+      if (!activeGenomeId) {
+        dispatch(setActiveGenomeId(genomeId));
+      }
       /*
        changeFocusObject needs to be called before setDataFromUrlAndSave
-       in order to prevent creating a previouslyViewedObject entry
-       for the focus object that is viewed first.
-       */
+       because it will also try to bookmark the Ensembl object that is stored in redux state
+       before it gets changed by setDataFromUrlAndSave
+      */
       dispatch(changeFocusObject(newFocusId as string));
     } else if (chrLocation) {
       const isSameLocationAsInRedux =
--- a/src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
+++ b/src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
@@ -98,13 +98,19 @@ const useBrowserRouting = () => {
       chrLocation
     };
 
+    if (!activeGenomeId) {
+      /*
+        Means that this is the first time user visits genome browser.
+        We need to make sure that active genome id is set properly in redux
+        for the actions below (e.g. changeFocusObject) to work
+      */
+      dispatch(setActiveGenomeId(genomeId));
+    }
+
     if (!focus && activeEnsObjectId) {
       const newFocus = buildFocusIdForUrl(parseEnsObjectId(activeEnsObjectId));
       dispatch(replace(urlFor.browser({ genomeId, focus: newFocus })));
     } else if (focus && !chrLocation) {
-      if (!activeGenomeId) {
-        dispatch(setActiveGenomeId(genomeId));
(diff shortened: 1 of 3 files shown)


msmanoj:
It looks like the bookmarks are broken as the entries are not getting created.

imransl:
> It looks like the bookmarks are broken as the entries are not getting created.

Strangely it's not broken for me. Am I checking it in the wrong way?

msmanoj:
> > It looks like the bookmarks are broken as the entries are not getting created.
> 
> @msmanoj Strangely it's not broken for me. Am I checking it in the wrong way?

The bookmarks modal will open fine but the problem is with the previously viewed links. Is it getting populated for you when you switch between genes?

azangru:
As a future reminder: the comment with a warning was added in https://github.com/Ensembl/ensembl-client/pull/145; and I can confirm that automatic bookmarking is broken 😢

imransl:
> The bookmarks modal will open fine but the problem is with the previously viewed links. Is it getting populated for you when you switch between genes?

Ah didn't notice that. Just checked and can confirm it is broken for me too.

azangru:
Ok, the reason bookmarks got broken, if I understand it properly, was twofold:

- If `setDataFromUrlAndSave` gets called first, we change activeEnsObjectId to a new object, and thus are losing the information about the previous one
- Besides, when we switch to a new ens object, we need to request information about if from the api; however, while the request is still ongoing, we already were trying to execute the synchronous action to add this object to previously viewed. The function attempted to retrieve this object from the redux store, realised that the object was not there, and [quit early](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts#L142-L144).

So, `changeFocusObject` had to be called **before** `setDataFromUrlAndSave` in order to retrieve data about active ens object from redux store before it got overwritten by the new ens object caused by the url change. But, at the same time, `changeFocusObject` had to be called **after** `setDataFromUrlAndSave`, because otherwise `changeFocusObject` may not have access to active genome id in redux store during the first visit to genome browser. Clearly a bad situation to be in.

I've refactored the `updatePreviouslyViewedObjectsAndSave` function so that it receives the ens object data directly from the component rather than try to retrieve it from redux store. It seems to be working correctly now; and I think it also reads more clearly. What do you say?

azangru:
Sorry, it turns out that moving `setDataFromUrlAndSave` above all those if-statements breaks various other things as well; specifically switching between different species. Very sad!

So I had to acknowledge my defeat, and just make sure that when a focus object gets changed, active genome id is always set in redux. Very ashamed of this. It shows how brittle and complicated our genome browser code is that tries to reconcile url in the address bar with the internal state of the application. I think it should work correctly now.

src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
```diff
@@ -101,11 +102,14 @@ const useBrowserRouting = () => {
       const newFocus = buildFocusIdForUrl(parseEnsObjectId(activeEnsObjectId));
       dispatch(replace(urlFor.browser({ genomeId, focus: newFocus })));
     } else if (focus && !chrLocation) {
+      if (!activeGenomeId) {
+        dispatch(setActiveGenomeId(genomeId));
+      }
```
msmanoj, line 107: Any chance this can be moved next to the if statement above?
https://github.com/Ensembl/ensembl-client/pull/411/files#diff-8f87f1f84e9255c291cc1bd6a5c52e16efb0474cfab2bb10053f99267b6fa994R94

msmanoj (approved):

PR #413, opened 2020-12-07
Preserve active species and entity in entity viewer

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-902

## Description
- Entity viewer activeGenomeId and activeObjectId are now preserved in the local storage so the if we close the tab and go back, it will take us to the view were on previously

## Deployment URL
http://ev-storage.review.ensembl.org

## Views affected
Entity viewer

### Other effects


src/ensembl/src/content/app/entity-viewer/services/entity-viewer-storage-service.ts
```diff
@@ -0,0 +1,56 @@
...
+import { EntityViewerGeneralState } from 'src/content/app/entity-viewer/state/general/entityViewerGeneralState';
+import storageService, {
+  StorageServiceInterface,
+  StorageType
+} from 'src/services/storage-service';
+
+export enum StorageKeys {
+  GENERAL_STATE = 'entityViewer.generalState',
+  SIDEBAR_STATE = 'entityViewer.sidebarState',
+  GENE_VIEW_STATE = 'entityViewer.geneViewState'
```
azangru, line 26: `SIDEBAR_STATE` and `GENE_VIEW_STATE` don't seem to be used anywhere?

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -94,9 +97,19 @@ export const setDataFromUrl: ActionCreator<ThunkAction<
...
+    }
+    dispatch(loadSidebar());
+    const storedState = entityViewerStorageService.getGeneralState();
+    entityViewerStorageService.updateGeneralState(
+      set(
+        'activeGenomeId',
+        genomeIdFromUrl,
+        set(`activeEnsObjectIds.${genomeIdFromUrl}`, entityId, storedState)
+      )
+    );
```
azangru, line 112: Like in the previous PR about sidebar state, why can't we update just the fragment we are interested in? Why do we need to get the whole state here?

src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts
```diff
@@ -93,6 +94,28 @@ export const updateEntityState = createAction(
...
+  const genomeId = getEntityViewerActiveGenomeId(state);
+  const entityId = getEntityViewerActiveEnsObjectId(state);
+
+  if (!genomeId || !entityId) {
+    return;
+  }
+
+  dispatch(setSidebarInitialState(genomeId));
+  dispatch(fetchSidebarPayload());
+};
```
azangru, line 117: I'm confused about what's happening to the sidebar in `loadSidebar` and `setSidebarInitialState`. What's the purpose of calling `setSidebarInitialState` in the `loadSidebar` here?

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
msmanoj: I am not quite clear about how to update just the fragment. Is it possible to share a snippet?

azangru: With lodash:

```js
const obj = { foo: { foo: { foo: 'foo' }  } };
merge(obj, {foo: { foo: { bar: 'bar' } }}); // adding a new property

// => obj is now  { foo: { foo: { foo: 'foo', bar: 'bar' }  } }

merge(obj, {foo: { foo: { foo: 'updated_foo' } }}); // updating existing property

// => obj is now  { foo: { foo: { foo: 'updated_foo', bar: 'bar' }  } }
```


Equally:

With stored state defined as `{ activeEnsObjectIds: { human: 'brca2' } }`:

```js
entityViewerStorageService.updateGeneralState({ activeEnsObjectIds: { human: 'fry' } })
```

should update the state to `{ activeEnsObjectIds: { human: 'fry' } }`

src/ensembl/src/content/app/entity-viewer/services/entity-viewer-storage-service.ts
msmanoj: Removed them

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
msmanoj: Thanks! I've updated it. I hope I remember it next time.

src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts
msmanoj: I'm trying to ensure that the initial state for the `genomeId` is created before fetching the sidebar data. I guess the name `setSidebarInitialState` is misleading? 
I have renamed it to `setSidebarInitialStateForGenome`.

azangru: > I'm trying to ensure that the initial state for the genomeId is created

Wouldn't it be better done in the reducer in response to active genome id change? It looks like a reducer responsibility.

We already have this by the way. Don't know if it's sufficient:
https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarReducer.ts#L37-L43

msmanoj: It is sufficient but I found it a bit confusing to use `generalActions` among the other actions here. 
The other problem I had was that `setActiveGenomeId` was not getting dispatched when we restore the state from the localStorage.

azangru: > The other problem I had was that setActiveGenomeId was not getting dispatched when we restore the state from the localStorage.

I don't think we need the `setActiveGenomeId` at this point. Wouldn't we just add the stored parts of the state to sidebar's initial state here: 

https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarState.ts#L99

azangru: > It is sufficient but I found it a bit confusing to use generalActions among the other actions here.

I understand how it can be a bit surprising to see actions from one slice of the state influence the other slice, but I like this solution much more than creating an action for the same purpose that another action is already serving.

msmanoj: > I don't think we need the `setActiveGenomeId` at this point. Wouldn't we just add the stored parts of the state to sidebar's initial state here:

If we don't dispatch `setActiveGenomeId`, the initial state for the genome will not be available when we fetch the sidebar data for the genome. This is why I thought it would be to have a separate action called `loadSidebar` to take care of setting the initial state as well as fetching the payload.

azangru: `loadSidebar` is ok (although, after Imran's [PR](https://github.com/Ensembl/ensembl-client/pull/422), we might not need it any longer); but I would have expected the reducer to be able to figure out for itself what the initial state is when the app is bootstrapped.

msmanoj: I guess is better to wait for Imran's PR to get merged first?

azangru: Imran's PR was merged in. Meet the conflicts 😄

msmanoj: Resolved the conflicts 👍

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -94,8 +95,18 @@ export const setDataFromUrl: ActionCreator<ThunkAction<
       })
     : null;
 
-  if (entityId && entityId !== activeEntityId) {
-    dispatch(updateEnsObject(entityId));
+  if (entityId && genomeIdFromUrl) {
+    if (entityId !== activeEntityId) {
+      dispatch(updateEnsObject(entityId));
```
azangru, line 100: I completely forgot that we still are using the notion of "ensObjects" in entity viewer 😞 

Wouldn't you agree that we should change all references to ensObjects in entity viewer to "entities" (unless they are too tightly entangled with ensObjects from genome browser)?

I also noticed that the `updateEnsObject` action fetches the ens object if it's missing from the state. I suspect we don't need this for the Entity Viewer state anymore.

Does this sound like something for a separate PR to you?

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -94,8 +95,18 @@ export const setDataFromUrl: ActionCreator<ThunkAction<
...
+    }
+    dispatch(setSidebarInitialStateForGenome(genomeIdFromUrl));
```
azangru, line 102: Still think that there shouldn't be a dedicated action for this, but rather the sidebar reducer should do this when it receives new entity id. We may address this in the same PR as the cleaning up of ensObjects.

azangru (approved):
Looks good, but see my last two comments about ensObjects. I suppose we can tackle them in a separate PR.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
msmanoj: >Wouldn't you agree that we should change all references to ensObjects in entity viewer to "entities" (unless they are too tightly entangled with ensObjects from genome browser)?
I also noticed that the updateEnsObject action fetches the ens object if it's missing from the state. I suspect we don't need this for the Entity Viewer state anymore.

Yes, that makes sense. `ensObjects` is not used elsewhere in entity viewer except this action so I guess it's not tightly entangled.

msmanoj:
> Looks good, but see my last two comments about ensObjects. I suppose we can tackle them in a separate PR.

Thanks. I have created a separate ticket to address those two comments.
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-928

PR #418, opened 2020-12-14
Genome browser brings back removed species

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-900

## Description
- Fixed the issue where removed species was getting added to the committed items list again

## Deployment URL
http://browser-removed-species.review.ensembl.org

## Views affected
- Genome browser

### Other effects


azangru:
I am very confused about this PR — primarily because I do not understand how the PR addresses the root cause of the bug.

Here's the simplified reproduction scenario for this bug (no need to open genome browser in a new tab as described in the ticket):

0) Clear site data in the browser (to start from clean slate)
1) In SpeciesSelector, select Human and Wheat
2) Navigate to genome browser (you'll be focused on Human)
3) Return to SpeciesSelector
4) Delete Human
5) Click on GenomeBrowser icon in the launchbar

Expected: only Wheat should remain.

Observed: both Wheat and Human.

The problem, I believe, lies entirely in our redux and local storage. When I delete human, redux store should no longer contain data about human, but it does:



Likewise, everything about human should be cleared from our permanent storage, but it isn't:

image

This, I believe, is what's causing the bugs we are observing. Let's try to clean up the deleted genome from redux and from the permanent storage first, and then see if we need to do anything about `fetchGenomeData` (I don't think so).

src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
```diff
@@ -123,10 +122,17 @@ const useBrowserRouting = () => {
...
-
     dispatch(setDataFromUrlAndSave(payload));
   }, [genomeId, focus, location]);
 
+  useEffect(() => {
+    const { genomeId } = params;
+    if (!genomeId) {
+      return;
+    }
+    dispatch(fetchGenomeData(genomeId));
```
imransl, line 133: Can we simplify this to:

```
if (params.genomeId) {
  dispatch(fetchGenomeData(genomeId));
}
```

msmanoj:
> I am very confused about this PR — primarily because I do not understand how the PR addresses the root cause of the bug.
> 
> Here's the simplified reproduction scenario for this bug (no need to open genome browser in a new tab as described in the ticket):
> 
> 1. Clear site data in the browser (to start from clean slate)
> 2. In SpeciesSelector, select Human and Wheat
> 3. Navigate to genome browser (you'll be focused on Human)
> 4. Return to SpeciesSelector
> 5. Delete Human
> 6. Click on GenomeBrowser icon in the launchbar
> 
> Expected: only Wheat should remain.
> 
> Observed: both Wheat and Human.
> 
> The problem, I believe, lies entirely in our redux and local storage. When I delete human, redux store should no longer contain data about human, but it does:
> 
> 
> 
> Likewise, everything about human should be cleared from our permanent storage, but it isn't:
> 
> image
> 
> This, I believe, is what's causing the bugs we are observing. Let's try to clean up the deleted genome from redux and from the permanent storage first, and then see if we need to do anything about `fetchGenomeData` (I don't think so).
> 
> P.S.: we'll need to do similar cleanup for entity viewer as well; not necessarily in this PR.

I did think about this approach of removing the entries from redux and local storage but I was not sure if it is good to alter the states across apps. Now it makes more sense to do it the way you have mentioned as we will leave some unwanted data stored in the permanent storage otherwise. 

I'm thinking about having a separate action within the browser to handle deleted species and clean things up. Please shout if it can be done in a different way.

azangru:
> I'm thinking about having a separate action within the browser to handle deleted species and clean things up.

Called from [this action in species selector](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts#L328-L354), I imagine?

msmanoj:
> > I'm thinking about having a separate action within the browser to handle deleted species and clean things up.
> 
> Called from [this action in species selector](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts#L328-L354), I imagine?

Yes, each app to its own `deleteSpeciesAndSave` action which will be called from the line above.

azangru:
Sounds good.

azangru:
There's an [unset](https://lodash.com/docs/4.17.15#unset) lodash function that may be useful for the removeAt method if we decide to add it.

src/ensembl/src/content/app/browser/hooks/useBrowserRouting.ts
msmanoj: That's done

src/ensembl/src/services/storage-service.ts
```diff
@@ -44,10 +50,20 @@ const defaultOptions: options = {
 };
 
 // We need to overwrite the arrays instead of merging them so that it is easier to remove entries
-const overwriteArray = (currentValue: JSONValue, newValue: JSONValue) => {
+const overwriteArray = (
```
azangru, line 53: The name of the function no longer properly reflects what it does. Perhaps rename the function to something like mergeStrategy?

azangru:
Could you revert changes in https://github.com/Ensembl/ensembl-client/pull/418/commits/bab07e8aa0a7670eb62703ba409227c0bab7268b? Unless there is any reason why we need them.

msmanoj:
> Could you revert changes in [bab07e8](https://github.com/Ensembl/ensembl-client/commit/bab07e8aa0a7670eb62703ba409227c0bab7268b)? Unless there is any reason why we need them.

Wouldn't it be better call `fetchGenomeData` from within the the hook instead of the component??

azangru:
Hm. I'm of two minds about that. On the one hand, this change isn't responsible for fixing the bug at all, so it both doesn't need to be in this PR and adds confusion to it. On the other hand, one might argue that fetching a species is the responsibility of the browser router hook rather than of the component itself. On the yet other hand, the routing hook is already complex and doesn't have any tests, while in the component this part of the behaviour was at least tested...

I probably don't strong enough conviction to argue one way or the other. If you think this is where the fetchGenomeData belongs, let it stay.

msmanoj:
> Hm. I'm of two minds about that. On the one hand, this change isn't responsible for fixing the bug at all, so it both doesn't need to be in this PR and adds confusion to it. On the other hand, one might argue that fetching a species is the responsibility of the browser router hook rather than of the component itself. On the yet other hand, the routing hook is already complex and doesn't have any tests, while in the component this part of the behaviour was at least tested...
> 
> I probably don't strong enough conviction to argue one way or the other. If you think this is where the fetchGenomeData belongs, let it stay.

I am happy to leave it as it is.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -395,3 +395,36 @@ export const toggleRegionEditorActive = createAction(
...
   'browser/toggle-region-field-active'
 )<boolean>();
+
+export const clearActiveGenomeId = createAction(
+  'browser/clear-active-genome-id'
+)();
+
+export const clearActiveEnsObjectIdsForGenome = createAction(
+  'browser/clear-active-ens-object-ids-for-genome'
+)<string>();
```
azangru, line 405: Can't we do all that in one single action — something like `deleteGenome` or `removeGenome`? This action would be checked by several reducers, which will need to:
- set active genome id to null if this genome is an active genome;
- remove this genome from `browserEntity.activeEnsObjectIds`
- remove it from the `browserLocation` part of the genome browser state
- remove it from the trackPanel state

We don't need to spawn multiple actions for this purpose.

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -395,3 +395,36 @@ export const toggleRegionEditorActive = createAction(
...
+    const updatedActiveEnsObjectIds = {
+      ...currentActiveEnsObjectIds
+    };
+    delete updatedActiveEnsObjectIds[genomeIdToRemove];
+    dispatch(updateBrowserActiveEnsObjectIds(updatedActiveEnsObjectIds));
+
+    browserStorageService.updateActiveEnsObjectIds({
+      ...currentActiveEnsObjectIds,
+      [genomeIdToRemove]: undefined
+    });
```
azangru, line 428: We also need to clean all this: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browser-storage-service.ts#L69-L89

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -395,3 +395,36 @@ export const toggleRegionEditorActive = createAction(
...
+
+export const clearActiveGenomeId = createAction(
+  'browser/clear-active-genome-id'
+)();
+
+export const clearActiveEnsObjectIdsForGenome = createAction(
+  'browser/clear-active-ens-object-ids-for-genome'
+)<string>();
+
+export const deleteSpeciesAndSaveBrowser = (
```
azangru, line 407: Maybe without `Browser` in the end? You aren't calling `clearActiveGenomeId` action above `clearActiveGenomeIdBrowser` :-)

I understand that we'll have similarly named actions for other apps (EntityViewer, CustomDownload, etc.). I don't think it's going to be a problem.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: That's done

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: That's done

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: That's done

msmanoj:
Hi @azangru - please hold on from reviewing this PR for now. I will update once I get some issues sorted.

msmanoj:
> Hi @azangru - please hold on from reviewing this PR for now. I will update once I get some issues sorted.

All sorted now.

src/ensembl/src/services/storage-service.ts
msmanoj: Changed it to `mergeStrategy`

src/ensembl/src/content/app/browser/browserReducer.ts
```diff
@@ -80,7 +81,11 @@ export function browserEntity(
       const genomeIdToRemove = action.payload;
       const activeGenomeId = state.activeGenomeId;
       if (activeGenomeId === genomeIdToRemove) {
-        return { ...state, activeGenomeId: null };
+        return {
+          ...state,
+          activeGenomeId: null,
+          activeEnsObjectIds: omit(state.activeEnsObjectIds, activeGenomeId)
```
azangru, line 87: I would advise against using `omit`. It sure looks nice, but lodash is planning to remove it from its next major release (whenever _that_ happens). [Link to the roadmap](https://github.com/lodash/lodash/wiki/Roadmap#v500-2021) and [link to a discussion on github](https://github.com/lodash/lodash/issues/2930#issuecomment-272298477).

You can do pickBy; or you can do something like:

```
const updatedEnsObjectIds = { ...state.activeEnsObjectIds }
delete updatedEnsObjectIds[activeGenomeId]

/* other stuff */
return {
          ...state,
          activeGenomeId: null,
          activeEnsObjectIds: updatedEnsObjectIds
}
```

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -20,6 +20,7 @@ import { replace } from 'connected-react-router';
 import { ThunkAction } from 'redux-thunk';
 import isEqual from 'lodash/isEqual';
 import get from 'lodash/get';
+import omit from 'lodash/omit';
```
azangru, line 23: the comment I left for the browserReducer file applies to this `omit` as well

src/ensembl/src/content/app/browser/browserActions.ts
```diff
@@ -399,3 +401,43 @@ export const toggleRegionEditorActive = createAction(
...
+    });
+    browserStorageService.updateChrLocation({
+      [genomeIdToRemove]: undefined
+    });
+    browserStorageService.updateTrackPanels({
+      [genomeIdToRemove]: undefined
+    });
+    browserStorageService.saveTrackStates({
+      [genomeIdToRemove]: undefined
+    });
```
azangru, line 437: Just wondering whether we wouldn't prefer to have a single `browserStorageService` method (e.g. `browserStorageService.deleteGenome` that would call all the above four methods.

src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import { ActionType, getType } from 'typesafe-actions';
-
+import omit from 'lodash/omit';
```
azangru, line 18: All said about the `omit` above applies here.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -25,6 +25,7 @@ import apiService from 'src/services/api-service';
 import speciesSelectorStorageService from 'src/content/app/species-selector/services/species-selector-storage-service';
 import analyticsTracking from 'src/services/analytics-service';
 import buildAnalyticsObject from 'src/analyticsHelper';
+import { deleteSpeciesAndSave as deleteSpeciesAndSaveBrowser } from 'src/content/app/browser/browserActions';
```
azangru, line 28: `deleteSpeciesAndSaveBrowser` -> `deleteSpeciesInGenomeBrowser`? We will also presumably have `deleteSpeciesInEntityViewer` etc. later on.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: I thought the same but I felt it is better to keep the storage service simpler.

src/ensembl/src/content/app/browser/browserReducer.ts
msmanoj: Created a custom `omit` function as discussed.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: That's done 👍

src/ensembl/src/shared/helpers/utils.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      [key]: {
+        [key1]: {
+          [key2]: {
+            bar: 1
+          }
+        }
+      }
+    };
+
+    expect(omit(obj, [key, key1, key2, 'foo'])).toEqual(updateObj);
```
azangru, line 48: Mmm, the signature of your `omit` is similar to lodash's, but the behaviour is different.

Normally, `omit` (such as in lodash or in ramda, or in the typescript `Omit` utility type) will take an object and an array of keys, and return an object without the passed keys:

```js
// from lodash docs:
var object = { 'a': 1, 'b': '2', 'c': 3 };
 
_.omit(object, ['a', 'c']);
// => { 'b': '2' }
``` 

What you wrote is an immutable unset rather than an omit 😄

src/ensembl/src/content/app/browser/browserActions.ts
azangru: It won't complicate the service, but will make the code of the consumer of the service simpler. Deleting a species is practically what in the database world would amount to a transaction. We don't want to expose the complexity of removing a species to the consumer; rather, it's much more preferable to hide the complexity inside the service.

src/ensembl/src/shared/helpers/utils.test.ts
msmanoj: Oh yea! my bad 😄  I've updated it now.

src/ensembl/src/content/app/browser/browserActions.ts
msmanoj: That's done

src/ensembl/src/shared/helpers/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import unset from 'lodash/unset';
+
+export const omit = (
+  originalObject: Record<string, any>,
+  keysToOmit: Array<string>
```
azangru, line 21: This is losing all type information. It should be typed using generics.

Here's an example of the type definition for lodash's `omit`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/common/object.d.ts#L1892-L1903

msmanoj: Would this work? [Typescript playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBArpAprAvDAFAQwFw02ATwBoYAjXfAgShhQD4YBvGAJ2ThbAG4YBfLgFACYImANCRYIALYBLKLgA8AFRiIAHlERgAJhBgAlRKBY7F0FrLABzEggDWYEAHcwdEgGk1m7Xpj3EAhAAMxhlOnQQUgArYwUwkgCCCFwAQRYWTAJFLwAfGA8AbQBdOhp6GAB5OSgVTwY0SMtrK0wAG0qYuMTAiHKGRmFREQloGGA28EQdTtjgVCYYADoVkGbWjq75vkEh4ZFgtYxR2AAHTCgACxgQ-16aQf2n+CQodAmpma2oEnOrqkEzz4QiBbCgHDA40mYGmszigl4AiAA)

src/ensembl/src/content/app/browser/browserReducer.ts
```diff
@@ -76,6 +76,15 @@ export function browserEntity(
         ...state,
         trackStates: merge({}, state.trackStates, action.payload)
       };
+    case getType(browserActions.deleteGenome): {
+      const genomeIdToRemove = action.payload;
+      const activeGenomeId = state.activeGenomeId;
+      if (activeGenomeId === genomeIdToRemove) {
+        const newState = { ...state, activeGenomeId: null };
+        delete newState.activeEnsObjectIds[activeGenomeId];
```
azangru, line 84: I suppose this is what's producing this error 😄 :

image

`{ ...state }` is a shallow copy of the state. Everything that's nested one and more levels deep are the same objects as in the original state. By changing newState.activeEnsObjectIds[activeGenomeId] you are actually mutating the state.

There are several ways out of this predicament. Ideally, we should refactor our browser redux state to use the `createSlice` function from react-toolkit; then we can just pretend that the state is mutable. But in the meantime, you can use either the `pickBy` for the activeEnsObjectIds part of the state or lodash's `cloneDeep` to get a true clone of the state. `pickBy` should be better for performance.

msmanoj: Thanks. I have replaced it all with `pickBy`.

azangru (approved):

PR #419, opened 2020-12-14
Check if Uniprot ID exists on Interpro

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-891

## Description
- This PR is to adda check if the Uniprot ID is present on Interpro before displaying in the protein item info.

## Deployment URL
http://uniprot-validation.review.ensembl.org

## Views affected
Entity viewer -> Protein -> ProteinItemInfo

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
+++ b/src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
@@ -87,3 +87,20 @@ export const fetchProteinSummaryStats = async (
 
   return restProteinSummaryAdaptor(proteinStatsData[xrefId]);
 };
+
+type InterproResponse = {
+  entries: unknown;
+};
+
+export const checkInterproUniprotId = async (
+  uniprotId: string,
+  signal?: AbortSignal
+): Promise<boolean> => {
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
```diff
@@ -87,3 +87,20 @@ export const fetchProteinSummaryStats = async (
 
   return restProteinSummaryAdaptor(proteinStatsData[xrefId]);
 };
+
+type InterproResponse = {
+  entries: unknown;
+};
+
+export const checkInterproUniprotId = async (
```
imransl, line 95: I am wondering if `isInterproUniprotIdValid` is a better name. Not that I don't like this function's name.

msmanoj: I think it will cause some confusion as we are only checking if the data is available for the given ID.

azangru (approved):
The `ProteinsListItemInfo`, with its three xhr requests, all updating component's state, is getting depressingly messy. On the other hand, we'll be getting uniprot id from Thoas later on, so that's one xhr request fewer. Let's merge it in and see how it develops from here.

PR #420, opened 2020-12-14
Fix cannot set property expandedTranscriptIds of undefined

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-906

## Description
Steps to reproduce:
- go to entity viewer for brca2
- select gene function tab
- go back to genome browser
- open entity viewer for a different gene using the zmenu
- select gene function tab

## Deployment URL
http://cannot-set-property-expanded.review.ensembl.org

## Views affected
- Entity viewer -> Protein view

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
+++ b/src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
@@ -107,7 +107,11 @@ const proteinsSlice = createSlice({
     ) {
       const { activeGenomeId, activeObjectId, expandedIds } = action.payload;
       if (!state[activeGenomeId]) {
-        state[activeGenomeId] = { [activeObjectId]: defaultStatePerGene };
+        state[activeGenomeId] = {
+          [activeObjectId]: { ...defaultStatePerGene }
+        };
+      } else if (!state[activeGenomeId][activeObjectId]) {
+        state[activeGenomeId][activeObjectId] = { ...defaultStatePerGene };
       }
       state[activeGenomeId][activeObjectId].expandedTranscriptIds = expandedIds;
     }


src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
```diff
@@ -108,8 +108,15 @@ const proteinsSlice = createSlice({
       const { activeGenomeId, activeObjectId, expandedIds } = action.payload;
       if (!state[activeGenomeId]) {
         state[activeGenomeId] = { [activeObjectId]: defaultStatePerGene };
+      } else if (!state[activeGenomeId]) {
```
azangru, line 111:
```
      if (!state[activeGenomeId]) {
        ...
      } else if (!state[activeGenomeId]) {
```
❓

msmanoj: Updated it! 🤦

src/ensembl/src/content/app/entity-viewer/state/gene-view/proteins/geneViewProteinsSlice.ts
```diff
@@ -107,7 +107,11 @@ const proteinsSlice = createSlice({
     ) {
       const { activeGenomeId, activeObjectId, expandedIds } = action.payload;
       if (!state[activeGenomeId]) {
-        state[activeGenomeId] = { [activeObjectId]: defaultStatePerGene };
+        state[activeGenomeId] = {
+          [activeObjectId]: { ...defaultStatePerGene }
+        };
+      } else if (!state[activeGenomeId][activeObjectId]) {
+        state[activeGenomeId][activeObjectId] = { ...defaultStatePerGene };
```
azangru, line 114: I don't think we need the spread, either here or on line 111.

msmanoj: I get the following error when I don't use the spread:

`TypeError: Cannot assign to read only property 'expandedTranscriptIds' of object '#<Object>'
    at updateExpandedProteins (geneViewProteinsSlice.ts?773c:116)`

Am I missing something?

azangru: Ah, immer! And we are not returning the state, but modifying it further on line 116. In this case, both you and immer are right 😄

azangru (approved):

PR #422, opened 2020-12-15
Retrieve gene data for gene overview from thoas

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-824](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-824)

## Description
Use Thoas instead of Ensembl REST to retrieve data for Gene overview in entity viewer sidepanel.

**IMPORTANT**: This is still a WIP, but I would like someone to have a look at the code to see if I'm heading in the right direction :)

## Views affected
Entity viewer -> Sidepanel -> Overview

### Other effects


imransl:
This is the thoas query I used.

```curl 'http://internal-2020.ensembl.org/thoas' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://internal-2020.ensembl.org' --data-binary '{"query":"# Write your query or mutation here\n{\n  gene(\n    byId: {\n      genome_id: \"homo_sapiens_GCA_000001405_28\"\n      stable_id: \"ENSG00000139618\"\n    }\n  ) {\n    stable_id\n    symbol\n    name\n    alternative_symbols\n    so_term\n    external_references {\n      name\n      description\n      source {\n        name\n        id\n      }\n      accession_id\n      url\n    }\n    transcripts {\n      symbol\n      external_references {\n        url\n        description\n        accession_id\n        name\n        source {\n          id\n          name\n        }\n      }\n      stable_id\n    }\n  }\n}\n"}' --compressed```

File: src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx [hunk 1]
```diff
@@ -32,38 +34,82 @@ import JSONValue from 'src/shared/types/JSON';
...
+      external_references {
+        accession_id
+        description
+        name
+        source {
+          name
+          id
+        }
+        url
+      }
```
azangru, line 54: What do you want to use the external references for?

src/ensembl/src/content/app/entity-viewer/types/gene.ts
```diff
@@ -37,5 +39,15 @@ export type Gene = {
...
+  attributes: string[];
+  function?: {
+    description: string;
+    source: Source;
+  };
+  filters?: {
+    [key: string]: JSONValue;
+  };
+  metadata?: Metadata;
+  external_references?: ExternalReference[];
```
azangru, line 52: Several things.

One is that we need to up our types game. On the one hand, we want to describe the types of entities that we can request from Thoas. On the other hand, since every field coming from Thoas is optional, we'll probably need multiple types — one type listing all the fields as potentially undefined, and also narrowed-down types that remove the optionality from the fields that have actually been requested. This is going to be painful and is definitely for another PR.

Another is that among the fields that you added here, the gene will only have metadata and external_references. Alternative symbols, function, etc. will be fields inside the metadata field. We are still working on the shape of metadata items — see https://github.com/Ensembl/ensembl-cdm-docs/pull/18. I don't expect it to be finished before mid-January; and then the data will have to be added to Thoas. So let's not attempt to add these fields to our code prematurely — they won't be ready for at least a couple of months.

The `filters` field does not exist anywhere on the Gene type at all.

I suppose the only fields that we can now show dynamically for the overview panel are the ones I marked with the green checkmark here:



@ens-st3 are you expecting anything more from this task?

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
imransl: I removed it

src/ensembl/src/content/app/entity-viewer/types/gene.ts
imransl: I have removed the UI sections and the thoas data properties in `GeneOverview.ts`

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
```diff
@@ -15,119 +15,78 @@
...
+        <span>{gene.stable_id}</span>
       </div>
 
       <div className={styles.sectionHead}>Gene name</div>
 
       <div className={styles.geneName}>
         <ExternalReference
-          label={gene.metadata?.name?.description}
-          linkText={gene.metadata?.name?.source?.id || ''}
+          label={gene.name as string}
```
azangru, line 75: But gene.name can really be null, typescript isn't lying.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
```diff
@@ -15,119 +15,78 @@
...
+
+const GeneOverview = () => {
+  const params: EntityViewerParams = useParams();
+  const { entityId, genomeId } = params;
+  const stableId = entityId ? parseEnsObjectIdFromUrl(entityId).objectId : null;
+
+  const { data, loading } = useQuery<{ gene: Gene }>(QUERY, {
+    variables: {
+      stable_id: stableId,
+      genome_id: genomeId
```
azangru, line 49: Does this work for you? It's broken for me; the names of the variables should be the same as the names of the variables prefixed with $ in the query, only the $ has to be removed:

```
variables: {
      geneId: stableId,
      genomeId
}
```

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
```diff
@@ -15,119 +15,78 @@
...
       <div className={styles.geneDetails}>
         <span className={styles.geneSymbol}>{gene.symbol}</span>
-        <span>{gene.id}</span>
+        <span>{gene.stable_id}</span>
       </div>
 
       <div className={styles.sectionHead}>Gene name</div>
 
       <div className={styles.geneName}>
         <ExternalReference
```
azangru, line 74: Gene name should not be an external reference yet. We are not retrieving any source for the gene name from Thoas.

azangru:
@msmanoj should we delete https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/sampleData.ts and this action https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts#L96 in this PR, or should we do it in a separate PR so that it doesn't conflict with https://github.com/Ensembl/ensembl-client/pull/413?

msmanoj:
> @msmanoj should we delete https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/sampleData.ts and this action https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts#L96 in this PR, or should we do it in a separate PR so that it doesn't conflict with #413?

I think it's okay to clean it up in this PR itself.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
imransl: I had used them incorrectly. It's fixed now ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
imransl: I've removed `ExternalReference` to resolve a comment below.

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
imransl: `ExternalReference` is replaced with `gene.name` to display the gene name.

imransl:
> > @msmanoj should we delete https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/sampleData.ts and this action https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarActions.ts#L96 in this PR, or should we do it in a separate PR so that it doesn't conflict with #413?
> 
> I think it's okay to clean it up in this PR itself.

@azangru @msmanoj Those are now deleted

azangru:
Pushed a couple of more commits:

- rather than use the `Gene` type itself, which contains lots of fields not requested in GeneOverview component, I've narrowed the type down to only the requested fields
- Andrea wanted to keep the Synonyms section even if there are no synonyms (with a message that there are no synonyms). I updated the code accordingly, and did the same for gene name.
- removed the remaining unused parts of redux state responsible for EntityViewer sidebar data

imransl:
Can we merge this

imransl:
@azangru @msmanoj Sorry about that. Accidentally pressed the 'Close with comment' button. BTW can we merge this PR?

azangru:
Pushed a minor update to implement Steve's suggestion not to show the Gene name section if there isn't any. Tested that on C.elegans gene WBGene00049335, which does not have gene name.

azangru:
Pushed another commit with updated tests for GeneOverview.

Instead of using a special library to mock Apollo, and instead of mocking react-router, I just used `jest.mock` to mock their hooks. I think this makes the test simpler, and its usefulness does not suffer.

PR #424, opened 2020-12-17
Add link to Europe PMC to Entity Viewer sidebar

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-863

## Description
Adding link to Europe PMC for genes.

### Note
The query generated for some organisms isn't ideal. For some organisms (such as wheat) this is because the sample data in the json file is incorrect ("TraesCS3D02G273600.1" is not gene symbol); but also, we are assuming the common_name and the scientific_name of a species are always different, which currently they aren't. Don't know whether we should check on the client if the scientific_name and the common_name values are identical or distinct, or whether we should just wait until our apis figure this out.

## Deployment URL
http://publications-link.review.ensembl.org/

## Views affected
Entity viewer -> Sidebar -> Overview

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
@@ -30,6 +30,10 @@ jest.mock('@apollo/client', () => ({
   useQuery: jest.fn()
 }));
 
+jest.mock('../publications/GenePublications', () => () => (
+  <div className="genePublications" />
+));
+
 const genomeId = 'genome_id';
 const geneId = 'unversioned_gene_id';
 const geneName = 'gene_name';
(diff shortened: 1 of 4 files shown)


msmanoj:
>Don't know whether we should check on the client if the scientific_name and the common_name values are identical or distinct, or whether we should just wait until our apis figure this out.

I don't have a strong opinion either but I feel it would be nice to temporarily implement some checks to prevent this:
Screenshot 2020-12-17 at 16 28 43

azangru:
> I don't have a strong opinion either but I feel it would be nice to temporarily implement some checks to prevent this:

The more I think about it, the more I am convinced this is the task for the backend. The common name returned from tthe backend should not be the same as scientitfic name. I'll ask Kamal if he can update genome search server. Meanwhile, can we ignore this for this PR?

ens-st3 (approved):
If there are data changes for wheat needed then that's for later. Good idea to deal with duplicated names now

azangru:
> If there are data changes for wheat needed then that's for later. Good idea to deal with duplicated names now

I can add a check on the client-side, no problem; but all temporary fixes have an annoying tendency of becoming permanent. If a change to [this line in genome-search api](https://github.com/Ensembl/ensembl-2020-genome-search/blob/e8934f95a00ea28b9857a0ffc2ce45aafc1418ac/resources/genome.py#L34) is sufficient to nullify common_name when it's the same as scientific name, then just changing it there and redeploying the genome search api should fix the issue.

azangru:
Using the updated genome search api in http://publications-link.review.ensembl.org, the `("Saccharomyces cerevisiae" OR "Saccharomyces cerevisiae")` problem is gone.

PR #426, opened 2020-12-17
Add sort filter indicator to EV transcripts list

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-913](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-913)

## Description
This PR adds a red dot indicator next to the `Filter & sort` text in the EV gene view. The indicator shows up when you have chosen a sort option that isn't `Default` and selected any option/s in the filter.

## Views affected
Entity viewer -> gene view -> filter and sort

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -72,6 +72,18 @@
   cursor: pointer;
 }
 
+.filterActivityIndicator {
+  &::after {
+    border-radius: 100%;
+    background: $red;
+    display: inline-block;
+    height: 5px;
+    width: 5px;
+    position: relative;
+    bottom: 10px;
+  }
+}
+
 .chevron {
   margin-left: 10px;
   margin-bottom: -3px;
@@ -81,14 +93,4 @@
 
 .hidden {
   display: none;
-}
-
-.activeIndicator {
-  border-radius: 100%;
-  background: $red;
-  display: inline-block;
-  height: 5px;
-  width: 5px;
-  position: relative;
-  bottom: 10px;
 }
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -82,3 +82,13 @@
 .hidden {
   display: none;
 }
+
+.filterSortIndicator {
+  border-radius: 100%;
+  background: $red;
+  display: inline-block;
+  height: 5px;
+  width: 5px;
+  position: relative;
+  bottom: 10px;
+}
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -72,14 +72,19 @@
   cursor: pointer;
 }
 
-.filterActivityIndicator {
-  border-radius: 100%;
-  background: $red;
-  display: inline-block;
-  height: 5px;
-  width: 5px;
+.labelWithActivityIndicator {
   position: relative;
-  bottom: 10px;
+
+  &::after {
+    content: '';
+    position: absolute;
+    border-radius: 100%;
+    height: 5px;
+    width: 5px;
+    background: $red;
+    right: -6px;
+    top: -2px;
+  }
 }
 
 .chevron {
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -83,7 +83,7 @@
   display: none;
 }
 
-.filterSortIndicator {
+.activeIndicator {
   border-radius: 100%;
   background: $red;
   display: inline-block;
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
@@ -73,15 +73,13 @@
 }
 
 .filterActivityIndicator {
-  &::after {
-    border-radius: 100%;
-    background: $red;
-    display: inline-block;
-    height: 5px;
-    width: 5px;
-    position: relative;
-    bottom: 10px;
-  }
+  border-radius: 100%;
+  background: $red;
+  display: inline-block;
+  height: 5px;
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -45,29 +45,45 @@ import styles from './DefaultTranscriptsList.scss';
...
+      dispatch(toggleTranscriptInfo(sortedTranscripts[0].stable_id));
     }
   }, []);
 
+  const shouldShowIndicator = () => {
+    if (sortingRule !== SortingRule.DEFAULT) {
+      return true;
+    }
+
+    if (Object.values(filters).some((filterValue) => filterValue)) {
```
azangru, line 80: `.some` returns a boolean. Which means that starting from this line, the body of the function can be rewritten to:

`return Object.values(filters).some((filterValue) => filterValue)`

or `return Object.values(filters).some(Boolean)`

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -84,7 +100,10 @@ const DefaultTranscriptslist = (props: Props) => {
         <div className={styles.row}>
           {gene.transcripts.length > 5 && !isFilterOpen && (
             <div className={styles.filterLabel} onClick={toggleFilter}>
-              Filter & sort
+              Filter & sort{' '}
```
azangru, line 103: the empty string shouldn't be necessary. The indicator's horizontal position can be adjusted entirely via CSS.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
```diff
@@ -82,3 +82,13 @@
 .hidden {
   display: none;
 }
+
+.filterSortIndicator {
```
azangru, line 86: I wonder if there may be a better name for this. `activeIndicator`, `activityIndicator`, `onIndicator`?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: This was done by prettier when formatting the code. I'll remove it nevertheless.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.scss
imransl: I replaced it with `activeIndicator`

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✅

azangru: I meant

```
  const shouldShowIndicator = () => {
    if (sortingRule !== SortingRule.DEFAULT) {
      return true;
    }

    return Object.values(filters).some(Boolean);
  };
```

or, even terser:

```
  const shouldShowIndicator = () =>
    sortingRule !== SortingRule.DEFAULT
      || Object.values(filters).some(Boolean);
```

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -85,6 +101,9 @@ const DefaultTranscriptslist = (props: Props) => {
           {gene.transcripts.length > 5 && !isFilterOpen && (
             <div className={styles.filterLabel} onClick={toggleFilter}>
               Filter & sort
+              {shouldShowIndicator() && (
+                <span className={styles.activeIndicator}></span>
```
azangru, line 105: self-closing span?

imransl: I am curious as to why self-closing span is better in this case as HTML/CSS purists wouldn't like it 😉  And also wondering if we need to make Prettier do it for us automatically.

azangru: A self-closing tag just communicates more explicitly that this tag isn't intended to contain anything. Which arguably makes it preferable for React purists 😄 Definitely a style thing. I probably shouldn't have left the comment to that line. I guess a more interesting question is whether to represent the indicator as a stand-alone empty span or as a styled pseudoelement.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
imransl: ✅

azangru: You'd need `content: ''` for the `after` to work 🙂 

But I guess I was wrong, and using a span really is better here. Either that, or wrap the label text into a span of its own. Because otherwise we are getting this:

image

Sorry for wasting your time 🙁

imransl: Ah that's fine. It didn't take long to switch it back to using `span` 😄

imransl: ✅

azangru:
So sorry again, but I've been playing with your PR, and noticed that with the `position: relative`, the label got shifted to the side when the indicator was added:

image

I've updated the indicator once again, to make it into a pseudoelement with `position: absolute`, anchored to the label. The label now firmly remains in the same place when the indicator is added.

azangru (approved):

imransl:
@azangru Thanks for fixing it 😃

PR #407, opened 2020-11-29
Add ADR for moving to React testing library, and migrate several test files from Enzyme

azangru:
## Description
An ADR containing rationale for moving from `enzyme` to `react-testing-library`, and some test files migrated over to react-testing-library to serve as examples.

PR #427, opened 2020-12-20
Fix Species Selector search field

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-917

### Manifestation of the bug
_present on internal deployment_

1) Search for a species in species selector; click on a match from the dropdown. The text in the input does not get replaced with the name of the selected species

or

2) Click on a popular species button. The content of the search field does not get updated with the name of that species. 

### Cause of the bug
The bug was introduced in https://github.com/Ensembl/ensembl-client/pull/407, where the `Input` component was updated from being controlled entirely by props to storing the changes in its local state. The reason for the update was because the new tests were failing. Turns out, this breaks the Species Selector, where the Input needs to show whatever is coming with the props.

### Solution
In this PR, the Input is changed back into a controlled React component, and the tests are updated to work properly.

## Deployment URL
http://fix-species-selector.review.ensembl.org/species-selector

## Views affected
Species Selector

### DIFF
--- a/src/ensembl/src/shared/components/input/Input.test.tsx
+++ b/src/ensembl/src/shared/components/input/Input.test.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { fireEvent, render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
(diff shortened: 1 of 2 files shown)


imransl (approved):

imransl:
Just noticed the label `in progress` is still tagged. Is there anything more that needs to go into this PR?

azangru:
> Just noticed the label in progress is still tagged. Is there anything more that needs to go into this PR?

Nope, sorry, forgot to remove it.

PR #428, opened 2020-12-21
Update tests to use react-testing-library

azangru:
## Type
- Refactoring
 
## Description
Migrate genome browser-related tests from enzyme to react-testing-library

## Notes
- this behaviour of BrowserRegionField [described in an existing test](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx#L178-L180) looks strange; not sure if it's correct. I've replicated it in the updated tests; but we probably want to confirm/fix the behaviour in later PRs.

### DIFF
--- a/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
@@ -67,7 +67,7 @@ describe('<BrowserRegionEditor />', () => {
       expect(container.querySelector('.select')).toBeTruthy();
     });
 
-    it('contains two Input elements', () => {
+    it('contains two input elements', () => {
       const { container } = render(<BrowserRegionEditor {...defaultProps} />);
       expect(container.querySelectorAll('input').length).toBe(2);
     });
--- a/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
+++ b/src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
@@ -15,31 +15,36 @@
  */
 
 import React from 'react';
-import { mount } from 'enzyme';
+import { screen, render, fireEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import faker from 'faker';
 
 import {
   BrowserRegionEditor,
   BrowserRegionEditorProps
 } from './BrowserRegionEditor';
-import Input from 'src/shared/components/input/Input';
-import Select from 'src/shared/components/select/Select';
-import Tooltip from 'src/shared/components/tooltip/Tooltip';
-import Overlay from 'src/shared/components/overlay/Overlay';
 
 import { createGenomeKaryotype } from 'tests/fixtures/genomes';
-import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
+import {
+  getCommaSeparatedNumber,
+  getNumberWithoutCommas
+} from 'src/shared/helpers/formatters/numberFormatter';
 import {
   createChrLocationValues,
   createRegionValidationMessages
 } from 'tests/fixtures/browser';
 import * as browserHelper from '../browserHelper';
 
-describe('<BrowserRegionEditor />', () => {
-  afterEach(() => {
-    jest.resetAllMocks();
-  });
+jest.mock('src/shared/components/select/Select', () => () => (
+  <div className="select" />
+));
+jest.mock('src/shared/components/overlay/Overlay', () => () => (
+  <div className="overlay" />
+));
 
+jest.mock('../browserHelper');
+
+describe('<BrowserRegionEditor />', () => {
   const initialChrLocation = createChrLocationValues().tupleValue;
   const defaultProps: BrowserRegionEditorProps = {
     activeGenomeId: faker.lorem.words(),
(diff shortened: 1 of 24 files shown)


src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -52,64 +57,66 @@ describe('<BrowserRegionEditor />', () => {
...
+    it('contains two input elements', () => {
+      const { container } = render(<BrowserRegionEditor {...defaultProps} />);
+      expect(container.querySelectorAll('input').length).toBe(2);
     });
 
-    test('contains submit and close buttons', () => {
-      expect(wrapper.find('button[type="submit"]')).toHaveLength(1);
+    it('contains submit and close buttons', () => {
+      const { container } = render(<BrowserRegionEditor {...defaultProps} />);
+      expect(container.querySelector('button[type="submit"]')).toBeTruthy();
```
msmanoj, line 77: We are not checking for the close button as mentioned in the description.

azangru: Yeah, the BrowserRegionEditor and BrowserRegionField components are pretty broken now, despite the passing tests. Don't know if I want to change anything about it in this PR, or fix both these two components and their tests in a separate PR.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -68,85 +64,97 @@ describe('<TrackPanelBookmarks />', () => {
...
   });
 
-  it('renders correct number of previously viewed links', () => {
-    const previouslyViewedLinksWrapper = wrapper.find(PreviouslyViewedLinks);
-
-    expect(previouslyViewedLinksWrapper.find('.link').length).toBe(
-      numberOfPreviouslyViewedObjects
+  it('closes the bookmarks modal when a bookmark link is clicked', () => {
+    render(<TrackPanelBookmarks {...props} />);
+    const previouslyViewedLinksContainer = screen.getByText('Previously viewed')
```
msmanoj, line 87: Is this approach better than adding an ID to the container element? 
This test is more likely to break if we introduce a different element after the header.

src/ensembl/src/content/app/browser/track-panel/track-panel-tabs/TrackPanelTabs.test.tsx
```diff
@@ -39,46 +40,59 @@ describe('<TrackPanelTabs />', () => {
...
-
-      beforeEach(() => {
-        wrapper = mount(
-          <TrackPanelTabs
-            {...defaultProps}
-            isTrackPanelOpened={false}
-            isDrawerOpened={true}
-          />
-        );
+      // wrapper.find('.trackPanelTab').first().simulate('click');
```
msmanoj, line 56: Can this be removed?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
azangru: This is a very good point. I've been thinking about starting using data-test-ids for such cases, which the testing-library community recommends ([link](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid)). I didn't want to add CSS selectors, such as ids or classes, specifically for testing purposes, but there is a babel plugin that can remove unnecessary html attributes, such as these test ids, from the production build, so that they do not get shipped to the end user (feels a bit dirty if they do). I've been a bit apprehensive of adding the extra babel plugin that removes these ids from the final build (or maybe just lazy), but now I guess is a good time to set this up.

azangru: ✔️

src/ensembl/src/content/app/browser/track-panel/track-panel-tabs/TrackPanelTabs.test.tsx
azangru: ✔️

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
msmanoj: A separate PR would be better I think.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
msmanoj: Looks good 👍 
Should we use snake casing for the test IDs or it doesn't matter? It feels a little odd to use text with spaces in an ID field.

msmanoj (approved):

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -124,11 +131,11 @@ describe('<BrowserRegionEditor />', () => {
 
       const locationStartInput = getCommaSeparatedNumber(faker.random.number());
       const locationEndInput = getCommaSeparatedNumber(faker.random.number());
-      const startError = faker.lorem.words();
-      const endError = faker.lorem.words();
+      const startError = 'start error message';
+      const endError = 'end error message';
```
imransl, line 135: Just curious to know why we aren't using random words anymore for strings in tests?

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
```diff
@@ -118,79 +109,75 @@ describe('<BrowserRegionField />', () => {
...
+        expect(defaultProps.changeFocusObject).toHaveBeenCalledWith(regionId);
       });
 
-      test('changes the browser location in same region if regionInput has only start and end (no stick/chromosome)', () => {
-        const start = faker.random.number();
-        const end = faker.random.number();
-        const regionInput = `${start}-${end}`;
+      it('preserves the same chromosome if input contains only new start and end coordinates', () => {
+        const oldChrLocation = ['13', 1, 1000] as ChrLocation;
+        const newChrLocation = ['13', 500, 1000];
```
imransl, line 163: Can't we use the same old and new location arrays used in the above test case? If we can then maybe we don't need to define them separately in individual test cases?

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
azangru: @imransl thanks for noticing 😄 

Random values turn out to be pretty difficult to debug when a test breaks. When I was migrating the tests and was checking that the rendered output was correct or that callbacks were called with correct arguments, seeing random Latin words in the error messages of failed tests (and some tests did fail for me at first) wasn't really helpful for discovering where the problem lies. In this component, for example, it's far easier to understand which wrong error message has been shown if they are explicitly written down and not generated randomly. 

I know that some people prefer not to use random values at all, choosing 100% reproducibility of a test over the fact that the test satisfies all imaginable inputs. I don't have sufficient confidence to advise whether we should stop using random values. At this point, I think, it's more of a gut feeling.

src/ensembl/src/content/app/browser/browser-region-field/BrowserRegionField.test.tsx
azangru: These two tests check different things. The previous one checks what happens when the chromosome changes but the location remains the same. This one checks what happens when the location changes but the chromosome stays the same. Don't think it's worth changing.

src/ensembl/src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
imransl: @azangru Thanks for the explanation. If it helps us debug test more easily, then this is definitely better 😄

imransl (approved):

PR #430, opened 2021-01-08
Show only Swissprot description in gene function

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-909](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-909)

## Description
We need to show only protein descriptions from Swissprot as the other sources are not reliable. Also when there are more than one description for a protein list item, we need to show the first one. If there are no protein descriptions from Swissprot, then we show nothing for it.

## Views affected
Entity viewer -> Gene function -> Protein list item

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
@@ -86,8 +86,8 @@ const ProteinsListItem = (props: Props) => {
   }, [proteinIdToFocus]);
 
   const getProteinDescription = () => {
-    const swissprotReference = product.external_references.find((reference) =>
-      reference.source.id.includes(SWISSPROT_SOURCE)
+    const swissprotReference = product.external_references.find(
+      (reference) => reference.source.id === SWISSPROT_SOURCE
     );
 
     return swissprotReference?.description;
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx
```diff
@@ -15,90 +15,90 @@
...
+        dispatch(toggleExpandedProtein(product.stable_id));
       }
     }
   }, [proteinIdToFocus]);
 
   const getProteinDescription = () => {
-    const uniprotReference = product.external_references.find((reference) =>
-      reference.source.id.includes(UNIPROT_SOURCE)
+    const swissprotReference = product.external_references.find((reference) =>
+      reference.source.id.includes(SWISSPROT_SOURCE)
```
azangru, line 90: why `.includes` and not `===`?

image

imransl: It's because the `UNIPROT_SOURCE` constant had only `Uniprot` before. But now we can switch to using `===` as you've suggested.

imransl: ✅

azangru (approved):

PR #431, opened 2021-01-12
Fix hidden lozenges

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-933

## Description
- Fixed the species lozenges overflow issue on the species page

## Deployment URL
http://hidden-lozenges.review.ensembl.org

## Views affected
http://hidden-lozenges.review.ensembl.org/homo_sapiens_GCA_000001405_28

### Other effects

### DIFF
--- a/src/ensembl/src/root/Root.scss
+++ b/src/ensembl/src/root/Root.scss
@@ -1,6 +1,6 @@
 .root {
   display: grid;
-  grid-template-rows: auto 1fr;
+  grid-template-rows: auto minmax(min-content, max-content) 1fr;
   position: relative;
   height: 100%;
 }
--- a/src/ensembl/src/shared/components/app-bar/AppBar.scss
+++ b/src/ensembl/src/shared/components/app-bar/AppBar.scss
@@ -10,7 +10,6 @@
   align-items: center;
   font-size: 12px;
   padding: 8px 20px;
-  height: max-content;
   min-height: 80px;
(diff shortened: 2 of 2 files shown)


azangru:
Doesn't work on Firefox:

image

Although I am fascinated that the height of the container works correctly in Safari, even without the fix.

Here's an idea, because the current CSS grid styles on the root element are wrong. Works on Firefox as well:

image

msmanoj:
> Doesn't work on Firefox:
> 
> image
> 
> Although I am fascinated that the height of the container works correctly in Safari, even without the fix.
> 
> Here's an idea, because the current CSS grid styles on the root element are wrong. Fix them as below, and it will work in Chrome, Safari and FF:
> 
> image

Thanks @azangru. That's done

imransl (approved):
Works for me

azangru (approved):
Strangely, the last commit didn't get deployed to the review deployment (you can see there's no green checkmark next to it):

image

but I checked the branch locally, and it works fine in Chrome, Safari and FF.

🚀

msmanoj:
> Strangely, the last commit didn't get deployed to the review deployment (you can see there's no green checkmark next to it):

Not quite sure what happened there. I guess it's something to keep an eye on.

PR #435, opened 2021-01-15
Track descriptions

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-822

## Description
- Created a `sampleData.ts` file that has the track descriptions stored in JSON format
- Created a `TrackDetails` view under drawer views which is used to display the track details for the selected track
- Updated the tests on `TrackPanelListItems.test.tsx`

## Deployment URL
http://track-descriptions.review.ensembl.org

## Views affected
Browser -> Track Panel -> Drawer

### Other effects


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -65,6 +66,22 @@ import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
...
+        keyFields: ['stable_id'],
+        fields: {
+          slice: {
+            merge: false
+          }
+        }
+      }
+    }
+  })
+});
```
azangru, line 83: After our conversation with @imransl last code review, we decided to move the code for the instantiation of Apollo client into its own file, so that it can be initialized only once and can be reused outside of React components. Imran is doing it [here](https://github.com/Ensembl/ensembl-client/pull/437/files#diff-3cae4751bd482c8e57ada5e39c83ae417e731ed82235b1b2b59190b187ca30fbR1). This is something to be aware of and later reuse in genome browser.

src/ensembl/src/content/app/browser/drawer/Drawer.scss
```diff
@@ -6,49 +6,49 @@
...
-  .secondaryDetail {
-    font-weight: $light;
-    margin-left: 10px;
-  }
-}
+.drawerView {	
+  overflow-y: auto;	
+  margin-top: 10px;	
+
+  h2 {
```
azangru, line 13: Wow, this (and below, and above) is a lot of trailing white space!

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.scss
```diff
@@ -0,0 +1,51 @@
+@import 'src/styles/common';
+
+.trackName{
+    font-weight: $bold;
+    display: inline-block;
+}
```
azangru, line 6: here and below — 2 spaces rather than 4 for indent.

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.scss
```diff
@@ -0,0 +1,51 @@
...
+  font-weight: $bold;
+}
+.stableId{
+  font-weight: $bold;
+}
+.featureSymbol + .stableId{
+  margin-left: 12px;
+  color: $dark-grey;
+  font-weight: $light;
+}
```
azangru, line 51: missing empty last line

src/ensembl/src/content/app/browser/drawer/sampleData.ts
```diff
@@ -0,0 +1,518 @@
...
+import { Source } from 'src/content/app/entity-viewer/types/source';
+
+export type TrackDetails = {
+  track_id: string;
+  track_name: string;
+  strand: 'forward' | 'reverse' | null;
+  shared_description?: string;
+  specific_description?: string;
+  source?: Source;
+};
```
azangru, line 27: Wonder what determines the colour of the squares. I suppose it's just protein coding vs non-protein-coding. Wonder where that information should go.

As well as where should the information responsible for the name of tracks group (e.g. "Genes & transcripts") go.

image

We probably need some kind of collective brainstorming session to discuss what tracks are and what fundamental properties they have. `shared_desciption` and `specific_desription` look suspicious (I understand that you want to separate them into their own paragraphs).

src/ensembl/src/content/app/browser/drawer/Drawer.scss
msmanoj: Removed the spaces

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.scss
msmanoj: Changed it all to 2 spaces

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.scss
msmanoj: added it 👍

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
```diff
@@ -18,7 +18,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import get from 'lodash/get';
 
-import { toggleDrawer, changeDrawerView } from '../../drawer/drawerActions';
+import { toggleDrawer } from '../../drawer/drawerActions';
```
azangru, line 21: `toggleDrawer` isn't used in `TrackPanelList` either 😄

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
```diff
@@ -26,92 +30,177 @@ import {
...
+jest.mock('react-redux', () => {
+  const originalModule = jest.requireActual('react-redux');
+  return {
+    ...originalModule,
+    useDispatch: () => jest.fn(),
+    useSelector: (selector: any) => {
+      return selector(mockState);
+    }
+  };
+});
```
azangru, line 105: We shouldn't need to mock react-redux. In fact, I have a vague suspicion that if you remove this mock, the tests will continue working just fine.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
```diff
@@ -26,92 +30,177 @@ import {
...
+jest.mock('src/content/app/browser/browserActions', () => {
+  const originalModule = jest.requireActual(
+    'src/content/app/browser/browserActions'
+  );
+
+  return {
+    ...originalModule,
+    updateTrackStatesAndSave: jest.fn()
+  };
+});
```
azangru, line 59: 1) Is there any point in keeping the actual implementations with `jest.requireActual`?
2) If you do not pass the constructor function as a second argument to `jest.mock`, jest should automatically mock all exports from the module (with jest.fn(), I believe). Will it not be sufficient?

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
```diff
@@ -26,92 +30,177 @@ import {
...
+jest.mock('src/content/app/browser/track-panel/trackPanelActions', () => {
+  const originalModule = jest.requireActual(
+    'src/content/app/browser/track-panel/trackPanelActions'
+  );
+
+  return {
+    ...originalModule,
     updateCollapsedTrackIds: jest.fn()
   };
+});
```
azangru, line 70: same as for the mock of the previous module

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
```diff
@@ -26,92 +30,177 @@ import {
...
+const defaultProps: TrackPanelListItemProps = {
+  categoryName: faker.lorem.words(),
+  trackStatus: Status.SELECTED,
+  defaultTrackStatus: Status.SELECTED,
+  track: createMainTrackInfo()
+};
+
+const wrapInRedux = (
+  props?: Partial<TrackPanelListItemProps>,
+  state: typeof mockState = mockState
```
azangru, line 118: 1) I don't think you are passing custom props to this wrapper in any of the tests
2) If we want to keep the option of passing the props; perhaps consider a single object with the `state` and the `props` fields as a parameter of `wrapInRedux`. Passing `undefined` as a first parameter to `wrapInRedux` is awkward.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
```diff
@@ -114,8 +120,41 @@ export const closeDrawer: ActionCreator<ThunkAction<
...
       })
     );
   });
 };
+
+export const setTrackDetails = createAction('browser/set-track-details')<{
+  [genomeId: string]: { [trackId: string]: TrackDetails };
+}>();
+
+export const fetchTrackDetails: ActionCreator<ThunkAction<
```
azangru, line 133: 🙅‍♂️ `ActionCreator`. `ActionCreator` [bad](https://github.com/reduxjs/redux/blob/9d3273846aa8906d38890c410b62fb09a4992018/src/types/actions.ts#L52-L54)

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.tsx
```diff
@@ -0,0 +1,78 @@
...
+import styles from './TrackDetails.scss';
+
+const TrackDetails = () => {
+  const trackDetails = useSelector(getActiveTrackDetails);
+
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (!trackDetails) {
+      dispatch(fetchTrackDetails());
```
azangru, line 33: What's the benefit, at this point, to getting track details via redux rather than directly from the component? These details don't seem to be reused in other components; the code that "fetches" them is restricted to the drawer; so why go through redux at all?

_(I understand that in the future, when we start fetching track data for real from a track api, track data will likely be reused by different components, e.g. TrackPanel and Drawer. But if that happens, fetching track data will not be the drawer's concern anyway, so the fetching code probably won't even be in the drawer state)_

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
msmanoj: I tried removing it but it doesn't seem to work.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.tsx
msmanoj: Removed it 👍

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
msmanoj: Thanks! That makes is a lot cleaner.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
msmanoj: That's done

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
msmanoj: I've removed the param `props` for now.

src/ensembl/src/content/app/browser/drawer/drawerActions.ts
msmanoj: That's done.

src/ensembl/src/content/app/browser/drawer/drawer-views/track-details/TrackDetails.tsx
msmanoj: That's a fair point. I have removed it from redux. As you mentioned, we can look into storing it somewhere generic when it needs to be reused.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
azangru: I updated the tests and pushed a new commit.

In your tests, you were trying to tests that action creators were called by mocking the action creators with functions that return nothing. Redux didn't like that — it wants to receive actions that are objects containing at least the type field. I changed the tests to use the real action creators, and to check which actions were generated by getting them from the mock store. I suppose, it would also be possible to make mock action creators return some mock action object, and that would also work.

msmanoj: Great! Thanks. I spent quite a long time trying to find a workaround for that error that was getting thrown when the `type` field was missing and I stopped looking further once I got it to work. the changes you have made seems to make more sense now.

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: I have removed the graphql related changes from this PR as discussed.

azangru (approved):
Looks good 👍

PR #437, opened 2021-01-15
Use Refget and thoas for instant download

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-929](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-929)

## Description
Instant download needs to use Thoas and Refget rather than REST API to download the sequences in FASTA format. This PR makes that change. However, the sequence checksums are not available for genomic sequences yet. Therefore, the REST API is used for it.

## Deployment URL
http://feature-instant-download-thoas.review.ensembl.org/

## Views affected
Entity viewer -> transcripts -> transcript list item -> download
Entity viewer -> gene function -> protein list item -> download

## Possible complications
The sequence checksums via Thoas are available only through http://thoas-checksum.review.ensembl.org/thoas at the moment. So this code will not work unless the sequence checksums become available in the staging Thoas server.


azangru:
One problem and one question.

The problem is genome browser. From what I am seeing in the PR, this code should break instant download in the genome browser. That's bad.

The question is whether it's better to query for checksums from components, like you are doing now; or to query for them from the fetcher functions (fetchForTranscript and fetchForProtein), in which case every fetch function will become a sequence of two fetches: one for the checksum and the other for the sequence itself. The latter approach looks more flexible (you are currently repeating the code that's fetching the checksums in InstantDownloadTranscript and InstantDownloadProtein components), but will likely require sending plain graphql query strings via POST requests to thoas using the browser's `fetch` function rather than using apollo's `useQuery` React hook.

azangru:
Update regarding the Apollo client.

1) Apollo client that we initialise [here](https://github.com/Ensembl/ensembl-client/blob/ebe7c8ac5712a2a55dc22895638b67f7237b2e0a/src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx#L66-L80) can be exported, imported in a fetchers file, and used directly as `client.query(...)`
2) Which suggests that we probably want to move the code for initialisation of ApolloClient in its own file, much like we [initialise redux store in its own file](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/store.ts). The client can then be imported both in the React component that uses ApolloProvider to pass it down the component tree, and independently by the functions that need it. I hope it will work.

See [SO discussion](https://stackoverflow.com/a/40663036/3925302) for reference.

imransl:
@azangru I've made the changes accordingly

imransl:
@azangru The tests are now fixed. I managed to fix it by mocking the instant download components in Zmenu.test.tsx and TranscriptsListItemInfo.test.tsx, so think it should be better than what we anticipated (mocking apollo client objects).

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -28,10 +29,19 @@ import ViewInApp from 'src/shared/components/view-in-app/ViewInApp';
...
+  gql: jest.fn(),
+  useQuery: jest.fn()
+}));
+
 jest.mock('src/shared/components/view-in-app/ViewInApp', () => () => (
   <div>ViewInApp</div>
 ));
 
+jest.mock('src/shared/components/instant-download', () => ({
+  InstantDownloadTranscript: () => <div>ViewInApp</div>
```
azangru, line 42: The mock text should say something different.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -62,6 +72,11 @@ describe('<TranscriptsListItemInfo /', () => {
 
   beforeEach(() => {
     wrapper = renderComponent();
+
+    (useQuery as any).mockImplementation(() => ({
+      data: null,
+      loading: true
+    }));
```
azangru, line 79: Would it work if the whole `.mockImplementation` thing for useQuery is moved up into the `jest.mock('@apollo/client'` factory function? `useQuery: jest.fn().mockImplementation ...`?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
```diff
@@ -151,7 +152,7 @@ export const TranscriptsListItemInfo = (
             </span>
           )}
         </div>
-        {props.expandDownload && renderInstantDownload(props)}
+        {props.expandDownload && renderInstantDownload(genomeId, props)}
```
azangru, line 155: would `renderInstantDownload({...props, genomeId})` make a better signature? 1 argument rather than 2.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
```diff
@@ -20,42 +20,61 @@ import {
...
   };
-  const typeParam = sequenceTypeToTypeParam[sequenceType];
+  const contextType = sequenceTypeToContextType[
+    sequenceType
+  ] as keyof ProductGeneratingContextFragment;
+  const sequenceChecksum =
+    productGeneratingContext[contextType]?.sequence_checksum;
 
-  return `https://rest.ensembl.org/sequence/id/${id}?content-type=text/x-fasta&type=${typeParam}`;
+  return `http://refget.review.ensembl.org/refget/sequence/${sequenceChecksum}?accept=text/plain`;
```
azangru, line 79: 1) For the time being, we should request the sequence in the fasta format (`accept=text/x-fasta` rather than accept=text/plain). This won't work when we start getting genomic sequences from ENA; but by that time, we'll think of something.
2) Remember to update the host to point to an actual refget server available from outside Ensembl rather than to a feature deployment of refget proxy.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
```diff
@@ -29,51 +32,88 @@ type Options = {
...
+  );
+
   if (geneOptions.genomicSequence) {
-    urls.push(buildFetchUrl(geneId, 'genomicSequence'));
+    urls.push(buildFetchUrl({ geneId }, 'genomicSequence'));
   }
+
   const sequencePromises = urls.map((url) =>
-    fetch(url).then((response) => response.text())
+    fetch(url as string).then((response) => response.text())
```
azangru, line 67: why the `as string`?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
```diff
@@ -29,51 +32,88 @@ type Options = {
...
+  genomeId: string;
   geneId: string;
+  transcriptId: string;
   options: Options;
 };
 
+type ProductGeneratingContextFragment = Pick<
+  ProductGeneratingContext,
+  'product' | 'cdna' | 'cds'
+>;
```
azangru, line 44: don't you have it defined in `fetchSequenceChecksums.ts`?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -0,0 +1,63 @@
...
+type GeneFragment = {
+  transcript: {
+    product_generating_contexts: ProductGeneratingContextFragment[];
+  };
+};
+
+export type ProductGeneratingContextFragment = Pick<
+  ProductGeneratingContext,
+  'cdna' | 'cds' | 'product'
+>;
```
azangru, line 32: Umm, this doesn't look like the correct type. CDNA, CDS and product requested from Thoas contain just a single field.

src/ensembl/src/shared/components/instant-download/instant-download-protein/InstantDownloadProtein.tsx
```diff
@@ -43,14 +44,16 @@ const proteinOptionLabels: Record<keyof ProteinOptions, string> = {
...
   const [isProteinSeqSelected, setProteinSeqSelected] = useState(false);
   const [isCdsSeqSelected, setCdsSeqSelected] = useState(false);
+  const { genomeId, transcriptId } = props;
 
   const onProteinCheckboxChange = () =>
     setProteinSeqSelected(!isProteinSeqSelected);
   const onCdsCheckboxChange = () => setCdsSeqSelected(!isCdsSeqSelected);
 
-  const onSubmit = () => {
+  const onSubmit = async () => {
```
azangru, line 53: 1) why?
2) without the `await`, this functions is just immediately returning a Promise of undefined anyway

src/ensembl/src/shared/components/instant-download/instant-download-transcript/InstantDownloadTranscript.tsx
```diff
@@ -110,6 +113,20 @@ const InstantDownloadTranscript = (props: Props) => {
     setTranscriptOptions(filterTranscriptOptions(so_term));
   }, [so_term]);
 
+  const onSubmit = async () => {
```
azangru, line 116: Same as in InstantDownloadProtein — there is no point in making a function async if it isn't awaiting for anything. And there is no point in awaiting if no subsequent logic is depending on the awaited result 🤷‍♂️

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.tsx
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
imransl: #1 is done. Will look into the other two when the PR is approved.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
imransl: ✅

imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
imransl: Created a new type for this

imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-protein/InstantDownloadProtein.tsx
imransl: This was a remnant of the previous attempt to get the download to work. It's removed now.

imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-transcript/InstantDownloadTranscript.tsx
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
imransl: This was because of the if statements in the `bulidFetchUrl` function having conditional returns (two `if` statements with no `else`) that resulted in a return type of `string | undefined`. I've made the changes in the function and also have removed the type assertion here.

imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
azangru: Points 2 and 3 have transformed into just using `/refget`, no hostname necessary.

Webpack dev server config will have to be updated here: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/webpack/environments/webpack.config.dev.js#L48 to include the `/refget` endpoint and point to the staging server.

(plus a reminder to change the `accept` query parameter in the other fetcher file to text/x-fasta)

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -28,10 +28,23 @@ import ViewInApp from 'src/shared/components/view-in-app/ViewInApp';
 import { createGene } from 'tests/fixtures/entity-viewer/gene';
 import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
 
+jest.mock('@apollo/client', () => ({
+  gql: jest.fn(),
+  useQuery: () =>
+    jest.fn().mockImplementation(() => ({
```
msmanoj, line 34: This could just be:
```suggestion
    jest.fn(() => ({
```

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
```diff
@@ -29,51 +34,82 @@ type Options = {
...
-  return `https://rest.ensembl.org/sequence/id/${id}?content-type=text/x-fasta&type=${typeParam}`;
+  if (sequenceType === 'genomicSequence') {
+    return `https://rest.ensembl.org/sequence/id/${data.geneId}?content-type=text/x-fasta&type=${sequenceTypeToContextType.genomicSequence}`;
+  } else {
+    const contextType = sequenceTypeToContextType[
+      sequenceType
+    ] as keyof ProductGeneratingContextFragment;
+    const sequenceChecksum =
+      data.productGeneratingContext &&
+      data.productGeneratingContext[contextType]?.sequence_checksum;
```
msmanoj, line 110: Is it safe to assume that the checksum will always be defined?
Screenshot 2021-01-28 at 09 01 51

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+import { gql } from '@apollo/client';
+
+import { client } from 'src/gql-client';
+
+type GeneFragment = {
+  transcript: {
+    product_generating_contexts: ProductGeneratingContextFragment[];
```
msmanoj, line 23: Should we consider placing the definition of `ProductGeneratingContextFragment` above ?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
msmanoj: I wonder if we can be more specific about what the fragment contains. It don't help much as any set of data inside `ProductGeneratingContext` can be a fragment.

imransl: @msmanoj I've changed it to this: https://github.com/Ensembl/ensembl-client/blob/624548ee763c32519a7050381cda51742c6af291/src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts#L27

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
```diff
@@ -29,51 +34,82 @@ type Options = {
...
 };
 
-const buildFetchUrl = (id: string, sequenceType: TranscriptOption) => {
-  const sequenceTypeToTypeParam: Record<TranscriptOption, string> = {
+const buildFetchUrl = (
+  data: {
+    geneId: string;
+    productGeneratingContext?: ProductGeneratingContextFragment;
+  },
+  sequenceType: TranscriptOption
```
azangru, line 93: - If we are transforming the first parameter into an object; I think we should include the second parameter in this object as well, rather than have it as an independent string
- The name `productGeneratingContext` is awkward. Let's change the name to `checksums`, and the object representing it to:

```
{
  cdna: string;
  cds: string;
  product: string;
}
```

notice that for non-coding transcripts we won't be able to get the cDNA at the moment, because Thoas doesn't currently return product-generating contexts for them:

image

Something to be discussed w/Steve and Andrea.

azangru: P.S.: I am also not sure it's a good idea to call a protein a product.

At the thinking about CDM currently stands, Thoas will return either proteins or mature RNAs as products. The user in the UI is requesting a protein; so I suppose we should be explicit in our code that such-and-such code path is for protein.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
msmanoj: @imransl Sorry, I meant if the name `ProductGeneratingContextFragment` can be changed to be something more specific.

azangru: I think it shouldn't have the `ProductGeneratingContext` in its name at all. What we are dealing with here is just checksums.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
azangru: > Is it safe to assume that the checksum will always be defined?

We have coding and non-coding transcripts; those will have different sets of checksums.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
imransl: Change it to `SequenceChecksums`  ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
imransl: > > Is it safe to assume that the checksum will always be defined?
> 
> We have coding and non-coding transcripts; those will have different sets of checksums.

I assume we don't need to change anything for this then?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
imransl: Will this do for now: 9a42f55?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
```diff
@@ -28,10 +28,23 @@ import ViewInApp from 'src/shared/components/view-in-app/ViewInApp';
 import { createGene } from 'tests/fixtures/entity-viewer/gene';
 import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
 
+jest.mock('@apollo/client', () => ({
+  gql: jest.fn(),
+  useQuery: () =>
+    jest.fn(() => ({
+      data: null,
+      loading: true
+    }))
```
azangru, line 37: useQuery in this case would actually be the actual jest.fn rather that a function that returns jest.fn. Funnily enough, tests didn't care.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+`;
+
+type Variables = {
+  genomeId: string;
+  transcriptId: string;
+};
+
+export const fetchSequenceChecksums = (variables: Variables) =>
```
azangru, line 62: This is actually a special case of fetching sequence checksums. It works for a transcript. It won't work for a gene. I suppose we want to rename it to `fetchTranscriptSequenceChecksums` (and the `const query` to `const transcriptChecksumsQuery`).

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -0,0 +1,68 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { gql } from '@apollo/client';
+
+import { client } from 'src/gql-client';
+
+export type SequenceChecksums = {
```
azangru, line 21: `TranscriptSequenceChecksums`, I suppose?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
```diff
@@ -20,42 +20,62 @@ import {
...
-  const typeParam = sequenceTypeToTypeParam[sequenceType];
+  const contextType = sequenceTypeToContextType[
+    sequenceType
+  ] as keyof SequenceChecksums;
+  const sequenceChecksum =
+    productGeneratingContext[contextType]?.sequence_checksum;
 
-  return `https://rest.ensembl.org/sequence/id/${id}?content-type=text/x-fasta&type=${typeParam}`;
+  // TODO: Change this before merging the PR
+  return `http://refget.review.ensembl.org/refget/sequence/${sequenceChecksum}?accept=text/x-fasta`;
```
azangru, line 80: `http://refget.review.ensembl.org/refget/` -> `/refget/`

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
```diff
@@ -29,51 +34,78 @@ type Options = {
...
+    return `https://rest.ensembl.org/sequence/id/${data.geneId}?content-type=text/x-fasta&type=${sequenceTypeToContextType.genomicSequence}`;
+  } else {
+    const contextType = sequenceTypeToContextType[
+      sequenceType
+    ] as keyof SequenceChecksums;
+    const sequenceChecksum =
+      data.checksums && data.checksums[contextType]?.sequence_checksum;
+
+    // TODO: Change this before merging the PR
+    return `http://refget.review.ensembl.org/refget/sequence/${sequenceChecksum}?accept=text/x-fasta`;
```
azangru, line 109: `http://refget.review.ensembl.org/refget` -> `/refget`

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForProtein.ts
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForTranscript.ts
imransl: ✅

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/transcripts-list-item-info/TranscriptsListItemInfo.test.tsx
imransl: ✅

azangru (approved):
Looks good. Don't know if you want to move the `gql-client.ts` file some place more convenient.

imransl:
> Looks good. Don't know if you want to move the `gql-client.ts` file some place more convenient.

I can't figure out where it will fit. I'll leave it as it is for now, and move it in a later PR when there is a better place to put it.

PR #439, opened 2021-01-20
Use CloseButton component consistently

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-864

## Description
This PR unifies the way we are using the CloseButton component across the site. Up until now, we have been using both the CloseButton and, independently, the close svg icon.

The following components have been changed (the images below show the original appearance of the components):

- `BrowserNavBarMain`

image

- `ToolboxExpandableContent` (part of ZMenu in GenomeBrowser)

- `CustomDownloadInfoCard`

- `PreviewDownload`

- `TranscriptsListItemInfo`

- `TranscriptsFilter`

image

- `StandardAppLayout` (specifically, the drawer)

image

- `Panel` (we aren't really using the Panel component with the close button anywhere in the app itself; so I used Storybook for reference)

## Deployment URL
http://consistent-close-button.review.ensembl.org/


src/ensembl/src/shared/components/layout/StandardAppLayout.test.tsx
```diff
@@ -65,54 +66,57 @@ const minimalProps = {
 };
 
 describe('StandardAppLayout', () => {
-  afterEach(() => {
+  beforeEach(() => {
```
imransl, line 69: Think most of our test cases use `afterEach` instead of `beforeEach`. Shouldn't we stick with just one of them?

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -15,14 +15,15 @@
...
+import userEvent from '@testing-library/user-event';
 import faker from 'faker';
 
 import Panel, { PanelProps } from './Panel';
 
-const panelContent = <p>{faker.lorem.words()}</p>;
+const panelContent = faker.lorem.words();
 
-const panelHeader = <p>{faker.lorem.words()}</p>;
+const panelHeader = faker.lorem.words();
```
imransl, line 26: Didn't we want to start using strings to make testing easier instead of random words? 😉

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
```diff
@@ -162,7 +162,5 @@ $drawerWindowWidth: 45px;
   position: absolute;
   right: calc(#{$drawerWindowWidth} + 10px);
   top: 10px;
-  width: 18px;
-  height: 18px;
   cursor: pointer;
```
msmanoj, line 165: I think this can be removed as well.

src/ensembl/src/shared/components/layout/StandardAppLayout.test.tsx
azangru: Yeah. The reason I am switching to beforeEach is because I think react-testing-library might do its cleanups a bit too late to be caught in afterEach. I remember one of the tests failing in a genome browser component when I was doing the cleanup in afterEach rather than in the beforeEach.

src/ensembl/src/shared/components/layout/StandardAppLayout.scss
azangru: good spot!

src/ensembl/src/shared/components/panel/Panel.test.tsx
azangru: 🤷‍♂️ 

✔️

src/ensembl/src/shared/components/panel/Panel.test.tsx
```diff
@@ -40,71 +41,68 @@ const defaultProps: PanelProps = {
...
-    expect(
-      wrapper.find('.body').first().hasClass(panelClassNames.body)
-    ).toBeTruthy();
-    expect(
-      wrapper.find('.header').first().hasClass(panelClassNames.header)
-    ).toBeTruthy();
+    const { container } = renderPanel();
+    const panel = container.querySelector('.panel') as HTMLElement;
+    const panelBody = container.querySelector('.body') as HTMLElement;
+    const panelHeader = container.querySelector('.header') as HTMLElement;
```
msmanoj, line 93: Since `panelHeader` is already defined on line 26, should we use a different variable name here?

azangru: 🤷‍♂️ 

I've updated the names of variables on lines 24-26 if that helps

msmanoj (approved):

src/ensembl/src/shared/components/layout/StandardAppLayout.test.tsx
imransl: That's interesting. So we'll gradually need to change the test cases to use `beforeEach` then 🤔

PR #444, opened 2021-01-22
Reorder protein list when transcripts are ordered

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-747](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-747)

## Importance
n/a

## Description
Reordering the transcripts in the protein list Under Gene Function in EV when we apply a sorting on the transcripts in the transcripts image. The order should be reflected across. 
Also moving the selected label (same as we did in https://github.com/Ensembl/ensembl-client/pull/441)

## Deployment URL
http://protein-order.review.ensembl.org/

## Views affected
Protein view in Entity Viewer (Under Gene Functions Tab)

### Other effects
none

## Possible complications
none

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { useEffect } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { useLocation } from 'react-router';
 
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/ProteinsList.tsx
```diff
@@ -42,8 +47,17 @@ const ProteinsList = (props: ProteinsListProps) => {
   const dispatch = useDispatch();
   const { search } = useLocation();
   const proteinIdToFocus = new URLSearchParams(search).get('protein_id');
+  const defaultTranscriptId = useMemo(() => {
+    const defaultTranscripts = defaultSort(props.gene.transcripts);
+    return defaultTranscripts[0].stable_id;
```
msmanoj, line 52: Just curious to know why we are recalculating this within each view without storing it in redux?

ens-ma7: From my understanding first of all this should be a temporary fix and when we get the data from thoas we wouldn't need this. Also we thought this will only be needed in the transcript image but i just found out that we needed it in the protein list as well. But still its only 2 place for now. And from what i heard in the planning meeting this morning, twice repeating is still ok :) 
adding @azangru since he might have a better explanation.

azangru: I agree that the solution is temporary anyway. We might move the code for determining which transcript is default up to the GeneView component, not to repeat it twice. Moving it all the way to redux may be a bit of a hassle; not sure if it's worth it.

ens-ma7: Thanks @azangru . since it is temporary is it ok to leave it as it is for now and when we need to implement the actual work when the data is available we can look into moving up the parent component. how does that sound @msmanoj ?

azangru: I think we are ok with repeating. Not the prettiest code, but we know it's changing and are not yet sure how.

msmanoj: Cool, let's leave it as it is for now. Thanks 👍

msmanoj (approved):

PR #445, opened 2021-01-25
Update Gene & Transcript drawer view to retrieve data from Thoas

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-934

## Description
- Updated the Gene and Transcript drawer view to use Thoas
- Moved the types defined within entity-viewer to the shared folder

## Deployment URL
http://browser-thoas.review.ensembl.org

## Views affected
Browser -> Drawer

### Other effects


src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -65,6 +66,22 @@ import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
...
+        keyFields: ['stable_id'],
+        fields: {
+          slice: {
+            merge: false
+          }
+        }
+      }
+    }
+  })
+});
```
azangru, line 83: After Imran's PR got accepted, this should no longer be necessary. See [EntityViewer component](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx).

I am also wondering whether it might not be a good idea to separate, inside the Browser component, the Browser from, say, ApolloWrappedBrowser, as two different exports. The default export would be Apollo-wrapped, but the export for unit-testing won't have any notion of Apollo. Of course that would mean mocking its children components that will complain of the absence of the Apollo context.

Kinda what our App.tsx [used to do](https://github.com/Ensembl/ensembl-client/blob/c3f5a8520ccbd255364ee894a687491dafa75d7b/src/ensembl/src/content/app/App.tsx#L92-L111) a long time ago.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
```diff
@@ -0,0 +1,41 @@
+@import 'src/styles/common';
+
+.standardLabelValue {
+    display: grid;
```
azangru, line 4: 2 spaces for indent are mixed with 4 spaces in this file. We use 2.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
```diff
@@ -0,0 +1,41 @@
...
+    grid-template-columns: [label] 120px 10px [value] auto;
```
azangru, line 5: In this case, where we only have two columns, better define the space between them using grid-column-gap. Makes the intent clearer.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
+    GeneFromGraphql,
+    | 'stable_id'
+    | 'symbol'
+    | 'name'
+    | 'alternative_symbols'
+    | 'so_term'
+    | 'transcripts'
+    | 'slice'
+  >
+>;
```
azangru, line 72: As a side note: this type isn't really reflecting what you are getting from Thoas (e.g. you are definitely not getting the whole `transcripts`). Hopefully, this will be mitigated with the approach that's being developed in https://github.com/Ensembl/ensembl-client/pull/452.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
+  const entityViewerUrl = urlFor.entityViewer({
+    genomeId: ensObjectGene.genome_id,
+    entityId: focusId
+  });
 
   return (
-    <div className={styles.drawerView}>
-      <div className={styles.container}>
+    <div className={styles.container}>
+      <div className={styles.standardLabelValue}>
```
azangru, line 108: What does standardLabelValue mean?

image

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
+  });
+  const entityViewerUrl = urlFor.entityViewer({
+    genomeId: ensObjectGene.genome_id,
+    entityId: focusId
+  });
 
   return (
-    <div className={styles.drawerView}>
-      <div className={styles.container}>
+    <div className={styles.container}>
```
azangru, line 107: there is no such class in the scss

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
+      <div className={styles.standardLabelValue}>
         <div className={styles.label}>Gene</div>
-        <div className={styles.details}>
-          <span className={styles.mainDetail}>{ensObject.label}</span>
+        <div className={styles.value}>
+          <div className={styles.featureDetails}>
+            {gene.symbol && (
+              <span className={styles.featureSymbol}>{gene.symbol}</span>
+            )}
+            {gene.symbol !== stableId && (
```
azangru, line 115: does it ever?

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
+              <span className={styles.stableId}>{stableId}</span>
+            )}
+            {gene.so_term && <div>{gene.so_term.toLowerCase()}</div>}
+            {gene.slice.strand.code && (
```
azangru, line 119: why are we checking this? the api swears that there will always be a slice.strand.code. Don't we trust the api?

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
```diff
@@ -16,30 +16,138 @@
...
-          <span className={styles.mainDetail}>{ensObject.label}</span>
+        <div className={styles.value}>
+          <div className={styles.featureDetails}>
+            {gene.symbol && (
+              <span className={styles.featureSymbol}>{gene.symbol}</span>
+            )}
+            {gene.symbol !== stableId && (
+              <span className={styles.stableId}>{stableId}</span>
+            )}
+            {gene.so_term && <div>{gene.so_term.toLowerCase()}</div>}
```
azangru, line 118: thoas promises that there will always be an so_term on the gene. If there isn't, thoas will return an error.

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -14,70 +14,320 @@
...
+  const [shouldShowDownload, showDownload] = useState(false);
 
-  const transcriptTrack = getTranscriptTrack();
+  const { data, loading } = useQuery<{
+    gene: Gene;
+    transcript: Transcript;
+  }>(GENE_AND_TRANSCRIPT_QUERY, {
+    variables: {
+      geneId: ensObjectGene.stable_id,
+      transcriptId: transcriptTrack?.stable_id,
```
azangru, line 153: This isn't working for wheat, because the data from `/api/object/track_list` isn't consistent with Thoas. Ideally, this part of the track panel should be getting its data from Thoas as well:

image

Since we seem to be in a hurry, we might want to postpone that for a separate PR.

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
```diff
@@ -0,0 +1,82 @@
+@import 'src/styles/common';
```
azangru, line 1: Comaring current implementation:

image

With the design in the XD:
image

- first line in the XD (Transcript) seems to be in its own group. Don't know why the distance between that line and the next group of lines (transcript name, transcript length, exons) is different than the distance between this second group and the third group. Maybe confirm w/Andrea
- missing bp after transcript length
- transcript length not formatted
- I have a suspicion that the light-grey colour in the XD may be achieved through the use of a thinner font rather than of the same font with a different colour. Confirm w/Andrea

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -14,70 +14,320 @@
...
+        <div className={styles.value}>{transcript.slice.location.length}</div>
+      </div>
+
+      <div className={styles.standardLabelValue}>
+        <div className={styles.value}>
+          {transcript.spliced_exons.length} exons
+        </div>
+      </div>
+
+      <div className={`${styles.standardLabelValue} ${styles.spaceAbove}`}>
```
azangru, line 235: I wonder if this might not be a better substitution for the spaceAbove class:

```
<section>
  <div className="styles.row">
    <div className="styles.label">Blah</div>
    <div className="styles.value">Blah</div>
  </div>
</section>
```

the sections can then be spaced in the way we like

image

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -14,70 +14,320 @@
...
+          {shouldShowDownload && (
+            <div className={styles.downloadWrapper}>
+              <InstantDownloadTranscript
+                genomeId={ensObjectGene.genome_id}
+                transcript={{
+                  id: transcript.unversioned_stable_id,
+                  so_term: transcript.so_term
+                }}
+                gene={{ id: gene.unversioned_stable_id }}
+                theme={'dark'}
```
azangru, line 299: Whoa! Isn't it the opposite? Isn't this the light theme, like in your browser or your IDE that would be the light theme?

image

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
ens-st3: Agreed

ens-st3: Can you create a ticket, for BE initially

azangru: @ens-st3 what will the backend need to do? Don't we already get everything we need from Thoas?

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
msmanoj: I have renamed `standardLabelValue` to `row`. Do we need the enclosing `section` tags? We could just apply additional stylings to `row` to add the spaces.

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
msmanoj: I am not sure. InstantDownload component only controls the colour of the text. When we ask it to give `dark` theme, it uses dark text.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
msmanoj: Removed this check.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
msmanoj: Removed the check

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
msmanoj: I guess not. I've removed it.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
msmanoj: Removed it 👍

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.tsx
msmanoj: It's just a name I've given to the container that has the label field on the left and the value on the right with 10px gap in between. I have renamed it to `row` for now.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
msmanoj: Updated it to use `grid-column-gap`. 
Regarding the second point, I think we can look into better ways to structure it when we have more contents to add in.

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
msmanoj: All, except first and the last are done. I will check the spacing with Andrea during the code review.

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: I have made the changes to import the client and removed the lines above. Is it necessary to export Browser without any notation of Apollo right now?

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
msmanoj: Updated it

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
ens-st3: Was wondering what we could do to get it working for wheat, both short term and longer term. Feels like a JIRA discussion ?

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
msmanoj: That's all done now.

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
msmanoj: Swapped the theme as discussed

src/ensembl/src/content/app/browser/Browser.tsx
azangru: What's the purpose of still mocking apollo-client in the test file?

image

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.featureSymbol {
+  font-weight: $bold;
+}
+.stableId{
+  font-weight: $bold;
+}
+.featureSymbol + .stableId{
+  margin-left: 12px;
+  color: $dark-grey;
```
azangru, line 46: I believe it shouldn't be gray since it's light? Compare how it looks:



with the way similar elements look in the transcript drawer:

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -14,70 +14,323 @@
...
+          {shouldShowDownload && (
+            <div className={styles.downloadWrapper}>
+              <InstantDownloadTranscript
+                genomeId={ensObjectGene.genome_id}
+                transcript={{
+                  id: transcript.unversioned_stable_id,
+                  so_term: transcript.so_term
+                }}
+                gene={{ id: gene.unversioned_stable_id }}
+                theme={'light'}
```
azangru, line 302: you don't need the interpolation if the property is a string (`theme="light"`)

src/ensembl/src/content/app/browser/Browser.tsx
msmanoj: I get the following error when I don't:
Screenshot 2021-02-10 at 13 48 47

But you are right that we shouldn't be mocking it here.

azangru: That is so strange! Mocking of `'src/gql-client'` should have gotten rid of this problem. But I am seeing the same error locally as well...

azangru: this works for me: `jest.mock('src/gql-client');` -> `jest.mock('src/gql-client', () => ({ client: jest.fn() }));`.

Don't understand why jest isn't doing this automatically, but there we are.

src/ensembl/src/content/app/browser/drawer/drawer-views/gene-summary/GeneSummary.scss
msmanoj: Yes, I missed to update it here. That's done

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
msmanoj: That's done

azangru (approved):

PR #446, opened 2021-02-01
Fix focus entity on GB being stored in web browser history

imransl:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-919](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-919)

## Description
The focus entity in genome browser was been stored in the web browser history as the `push` function was used rather than `replace`. This PR has changed the code to use the latter.

Also, some redundant code has been removed from `TrackPanelBookmarks.tsx` and also hooks used instead of the react-redux connected wrapper. This required changes to be made to its test file, which is currently broken.

**TODO**: Fix the broken tests

### Other effects


imransl:
@azangru @msmanoj Would like to know what you think of using hooks for this component (`TrackPanelBookmarks`) instead of making it a connected component. The tests needed to be changed a bit and aren't working properly yet. I'll leave it as it is for now due to taking too much time just to fix them. Will wait till you comment on this PR :)

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -25,12 +26,25 @@ import { TrackPanelBookmarks } from './TrackPanelBookmarks';
...
-    </div>
-  )
+jest.mock('react-redux', () => ({
+  useSelector: jest.fn().mockImplementation((selector) => selector()),
+  useDispatch: jest.fn()
+}));
+
+jest.mock('src/shared/state/ens-object/ensObjectSelectors', () => ({
+  getExampleEnsObjects: jest.fn()
+}));
```
msmanoj, line 36: Here and below, please refer to the point number 2 mentioned by Andrey here:
https://github.com/Ensembl/ensembl-client/pull/435#discussion_r567514153

msmanoj:
> @azangru @msmanoj Would like to know what you think of using hooks for this component (`TrackPanelBookmarks`) instead of making it a connected component. The tests needed to be changed a bit and aren't working properly yet. I'll leave it as it is for now due to taking too much time just to fix them. Will wait till you comment on this PR :)


@imran - I recently updated [TrackPanelListItem.tsx](https://github.com/Ensembl/ensembl-client/pull/435/files#diff-119433d3abe689659604d56627491b2696c26b71d43512eff38845aa83ae4a7cL35)  to use `hooks` instead of `connect` and I felt that is has made the code look a bit cleaner.

azangru:
Using redux from inside components through hooks, rather than through the `connect` function is the current recommendation of the maintainers of `react-redux`. One of the strongest arguments for this is that when we use the `mapDispatchToProps` function/object, we are not guaranteeing that the types of the action creator functions will be correct. You can see this for yourself: in most places where we use mapDispatchToProps, the action creator functions are actually thunks, i.e. they should return another function (this is how we type them in actions files); yet, we declare in our Props types that these functions return `void`.

So I agree with using the react-redux hook; although it makes testing of these components more involved.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -52,26 +64,31 @@ describe('<TrackPanelBookmarks />', () => {
...
   beforeEach(() => {
     jest.resetAllMocks();
+
+    (useSelector as any).mockReturnValue(exampleEnsObjects);
+    (useSelector as any).mockReturnValue(previouslyViewedObjects);
   });
 
-  it('renders previously viewed links', () => {
-    const { container } = render(<TrackPanelBookmarks {...props} />);
+  it.only('renders previously viewed links', () => {
```
azangru, line 82: 🙅‍♂️ `only`!

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -25,12 +26,25 @@ import { TrackPanelBookmarks } from './TrackPanelBookmarks';
...
-jest.mock('react-router-dom', () => ({
-  Link: (props: any) => (
-    <div {...props} className={'link'}>
-      {props.children}
-    </div>
-  )
+jest.mock('react-redux', () => ({
+  useSelector: jest.fn().mockImplementation((selector) => selector()),
+  useDispatch: jest.fn()
+}));
```
azangru, line 32: I would much rather we mocked the relevant part of redux state and used the mock-redux-store to inject it instead of mocking redux and redux hooks.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -52,26 +64,31 @@ describe('<TrackPanelBookmarks />', () => {
...
+    const { container } = render(<TrackPanelBookmarks />);
     const links = [...container.querySelectorAll('.link')] as HTMLElement[];
-    const linkTexts = props.previouslyViewedObjects.map(({ label }) => label);
+    const linkTexts = previouslyViewedObjects.map(({ label }) => label);
+
+    linkTexts.forEach((text) =>
+      links.find((link) => {
+        return link.innerHTML === text;
+      })
+    );
```
azangru, line 91: what's going on in this forEach?

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -96,41 +113,32 @@ describe('<TrackPanelBookmarks />', () => {
...
-    );
+    newPreviouslyViewedObjects.push(createPreviouslyViewedLink());
+    rerender(<TrackPanelBookmarks />);
 
     expect(previouslyViewedSection.querySelector('button')).toBeTruthy();
   });
 
   it('calls changeDrawerViewAndOpen when the ellipsis is clicked', () => {
-    const previouslyViewedObjects = times(21, () =>
+    const newPreviouslyViewedObjects = times(21, () =>
```
azangru, line 135: I would much rather reuse the same variable 😞 I find `newPreviouslyViewedObjects` more confusing than reusing the variable from the outer scope. @msmanoj might think differently.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imransl: ✅

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imransl: ✅

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imransl: ✅

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imransl: Removed it ✅

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
imransl: I've used mockState so this is now redundant ✅

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx
```diff
@@ -15,63 +15,157 @@
...
+          genome_id: genomeId
+        }
+      }
+    }
+  },
+  speciesPage: {
+    general: {
+      activeGenomeId: null
+    }
+  }
```
azangru, line 144: what's the speciesPage part of the mock state?

azangru:
I pushed a PR to update some tests:

- used a spy to detect that an action creator for closing the drawer is called
- added data-test-ids to separate example links from previously viewed links
- the test that you had [here](https://github.com/Ensembl/ensembl-client/blob/0e595e56087e5b4a4ed55d532a9a7848c52f1e95/src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx#L185-L200) was checking the result of a click on the first link. Which is an example link. This made the [last test](https://github.com/Ensembl/ensembl-client/blob/0e595e56087e5b4a4ed55d532a9a7848c52f1e95/src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx#L270-L282) redundant, so I deleted it
- removed species page from the mock state (it wasn't used)

imransl:
> I pushed a PR to update some tests:
> 
> * used a spy to detect that an action creator for closing the drawer is called
> * added data-test-ids to separate example links from previously viewed links
> * the test that you had [here](https://github.com/Ensembl/ensembl-client/blob/0e595e56087e5b4a4ed55d532a9a7848c52f1e95/src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx#L185-L200) was checking the result of a click on the first link. Which is an example link. This made the [last test](https://github.com/Ensembl/ensembl-client/blob/0e595e56087e5b4a4ed55d532a9a7848c52f1e95/src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.test.tsx#L270-L282) redundant, so I deleted it
> * removed species page from the mock state (it wasn't used)

@azangru Thanks for fixing the test case issues and improving them

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
```diff
@@ -15,58 +15,47 @@
...
+import { useSelector, useDispatch } from 'react-redux';
 import upperFirst from 'lodash/upperFirst';
+import { Link } from 'react-router-dom';
 
 import analyticsTracking from 'src/services/analytics-service';
 import * as urlFor from 'src/shared/helpers/urlHelper';
 import { buildFocusIdForUrl } from 'src/shared/state/ens-object/ensObjectHelpers';
-import { getBrowserActiveGenomeId } from 'src/content/app/browser/browserSelectors';
+
+import { getActiveGenomeId } from 'src/content/app/species/state/general/speciesGeneralSelectors';
```
azangru, line 26: Nearly missed this one 😞 Discovered it while manually testing locally.
This is the wrong selector. It gets the active genome id on the species page, while we are interested in the active genome id on the browser page. Used on the browser page, it was returning `null`, which broke the app.

I've fixed the selector and updated the tests to check the generated .

azangru:
I'm not sure what this PR aims to fix. The browser history stack still seems to grow when I switch from one focus object to another one using the zmenu. Also, the list of example objects on the genome browser interstitial needs to be updated the same way the links were updated in the bookmarks modal.

src/ensembl/src/content/app/browser/track-panel/track-panel-modal/modal-views/TrackPanelBookmarks.tsx
imransl: Thanks for fixing it @azangru

imransl:
> I'm not sure what exactly this PR aims to fix, but the browser history stack still seems to grow when I switch from one focus object to another one using the zmenu, and so pressing the browser back button will switch me to the previous focus object. Also, the list of example objects on the genome browser interstitial needs to be updated the same way the links were updated in the bookmarks modal.

It was supposed to make the user navigate straight from genome browser to species selector after clicking on the browser back button. However, I have not been clear in the description. Having said that, the issues that you've been mentioned probably can be fixed in this PR itself :)

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
```diff
@@ -62,25 +70,32 @@ describe('<ViewInApp />', () => {
...
-      expect(push).toHaveBeenCalledWith(link);
+      userEvent.click(
+        appButtonContainer.querySelector('button') as HTMLButtonElement
+      );
+
+      expect(replaceFn).toHaveBeenCalledWith(link);
     });
+
+    replaceFn.mockRestore();
   });
```
imransl, line 100: @azangru @msmanoj This test case unfortunately fails with an error: `Actions must be plain objects. Use custom middleware for async actions.` Appreciate any suggestions on how it can be fixed :)

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
```diff
@@ -45,7 +45,7 @@ export type urlObj = Record<AppName, string>;
 
 export type ViewInAppProps = {
   links: Partial<urlObj>;
-  push: Push;
+  shouldReplaceState?: boolean;
```
azangru, line 48: This looks far too indiscriminate. Different links may have different behaviour. I think this should go into the urlObj; and the type of the urlObj should be updated to:

```
{
  [appName: AppName]: {
    link: string;
    replaceState: boolean
  }
}
```

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
azangru: You've mocked the push and the replace functions from connected react router with just `jest.fn()`. Which means that these functions will now return `undefined`. Naturally, redux doesn't like it. It wants action creators to return actions that it can handle (normally — objects with the `type` field).

imransl: @azangru Thanks I'll try it that way

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
imransl: That's done now. Did some small changes to `urlObj` too.

imransl: ✅

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
imransl: I've got the test to work even though typescript doesn't like the signature of the function I've used to mock `replace`

azangru: Pushed a commit to slightly update the test. I cheated, and used the `any` type 😄  But we don't really care whether the mock `replace` function actually returns the realistic replace action; nothing in our test depends on that. As long as it returns an object, we should be fine 😄 

I also updated the `toHaveBeenCalledWith` to `toHaveBeenLastCalledWith`.

src/ensembl/src/shared/components/view-in-app/ViewInApp.tsx
azangru: >  Did some small changes to urlObj too.

Hmm, interesting. I was expecting the urlObj to always have the values of `{ link: string, replaceState?: boolean}`, but your suggestion may be more convenient. We'll discuss this during the code review.

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
imransl: @azangru Thanks. It looks much better now with less code 😄

azangru: Now that we have a test for `replace`, I couldn't resist adding a test for `push` (and simplified these tests slightly, removing the uncertainty in the number of rendered buttons — that is already checked in the previous tests).

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -569,33 +569,42 @@ const getExampleLinks = (props: {
...
+          })
+        },
+        entityViewer: {
+          url: urlFor.entityViewer({
+            genomeId: genome_id,
+            entityId: focusId
+          })
+        }
+      };
+    } else if (section === SpeciesStatsSection.ASSEMBLY) {
```
azangru, line 586: There's something wrong with the nesting here — this condition ended up inside the `SpeciesStatsSection.CODING_STATS` block.



I noticed it only accidentally when running the type checking script 😞 

image

imransl: Thanks for spotting that. I'll look into it now.

src/ensembl/src/shared/components/view-in-app/ViewInApp.test.tsx
imransl: Awesome. It looks much better now 😃

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
imransl: That's fixed now ✅

azangru (approved):
Looks good.

There's one more thing related to web browser history on the genome browser page. I think Andrea wanted to add switching between species to web browser history (it currently isn't). If that's the case, it can be easily done in a separate PR.

PR #447, opened 2021-02-03
Improve sort and filter indicator for transcripts list

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-944](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-944)

## Description
Sort and filter indicator for transcripts list in EV appears only after it is closed. This is not the ideal behaviour, so this PR makes it appear immediately when a change is made to the sort/filter.

## Deployment URL
http://transcript-filter-indicator.review.ensembl.org/

## Views affected
Entity viewer -> Transcripts list -> Sort and filter

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -104,7 +104,7 @@ const DefaultTranscriptslist = (props: Props) => {
       <div className={styles.header}>
         {isFilterOpen && (
           <TranscriptsFilter
-            filterLabel={filterLabel}
+            label={filterLabel}
             toggleFilter={toggleFilter}
             transcripts={sortedTranscripts}
           />
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -93,6 +93,7 @@ const DefaultTranscriptslist = (props: Props) => {
       <div className={styles.header}>
         {isFilterOpen && (
           <TranscriptsFilter
+            shouldShowFilterIndicator={shouldShowFilterIndicator}
             toggleFilter={toggleFilter}
             transcripts={sortedTranscripts}
           />
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -16,6 +16,7 @@
 
 import React, { useState, useEffect, useMemo } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
+import classNames from 'classnames';
 
 import { getFeatureCoordinates } from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
 import {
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -93,6 +93,7 @@ const DefaultTranscriptslist = (props: Props) => {
       <div className={styles.header}>
         {isFilterOpen && (
           <TranscriptsFilter
+            shouldShowFilterIndicator={shouldShowFilterIndicator}
```
azangru, line 96: 😲 why? TranscriptsFilter already contains all the information about the current sorting rule and the selected filters.

imransl: Just wanted to avoid code duplication

azangru: We may have something like the following:

In the `DefaultTranscriptsList` component:

```tsx
import classNames from 'classnames'

// other code

const filterSortBoxLabel = (
  <span
    className={classNames({[transcriptListStyles.labelWithActivityIndicator]: shouldShowFilterIndicator })
  >
    Filter & sort
  </span>
);

// then where you use this label in DefaultTranscriptsList:

{ gene.transcripts.length > 5 && !isFilterOpen && filterSortBoxLabel }

// and for TranscriptsFilter component

<TranscriptsFilter label={filterSortBoxLabel} ... />

// in which case the props of TranscriptsFilter component should include
// { label: ReactNode }
```

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -93,22 +104,15 @@ const DefaultTranscriptslist = (props: Props) => {
       <div className={styles.header}>
         {isFilterOpen && (
           <TranscriptsFilter
+            filterLabel={filterLabel}
```
azangru, line 107: prop name `filterLabel` -> `label`?

imransl: It's named as `filterLabel` everywhere else so wanted to be consistent with the naming. Do you still think we should change this to `label`?

azangru: > everywhere else so wanted to be consistent with the naming

I don't understand what you mean by consistency here. TranscriptsFilter component expects a list of properties from the outside world. Inside the TranscriptsFilter component, it doesn't make much sense in prefixing the word "label' with the word "filter", because there seems to be only one label that this component receives from the outside with the props. In the world that lies outside the TranscriptsFilter component, i.e. in the `DefaultTranscriptsList` component, the name `filterLabel` makes a bit more sense (although I still don't like that we refer to the filter and sort box as just filter). There is no requirement for the name of a property of a chid component to match the name of the variable outside that component. My suggestion was to change `filterLabel={filterLabel}` to `label={filterLabel}`. Now that I'm re-reading my last comment, I realise that I didn't make it clear.

imransl: > My suggestion was to change filterLabel={filterLabel} to label={filterLabel}.

That sounds better :) I was thinking you were telling to change it everywhere.

imransl: It is changed now ✅

azangru (approved):

PR #450, opened 2021-02-07
Download sequences in plain text and transform them to FASTA in web worker

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-982
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-983

## Description
With the switch to refget came the following problems:
- refget servers do not guarantee the implementation of the fasta format
- refget servers do not have access (or have very limited access) to the information about sequences other then their checksums; therefore, even if a refget server implements the fasta format, it cannot properly fill in the sequence description line and just uses sequence checksum for description (not at all helpful to end users)

This PR implements the following:
- fetching of unmodified, plain-text sequences and formatting them into the fasta format: a sequence gets split into lines 60 characters long (same to the fasta implementation of ensembl rest or ensembl refget servers) and is prepended with an informative description line
- since this sequence formatting job has no relation to the rendering done by the main thread, it is delegated to a web worker, which should make the least impact on the responsiveness of the UI

Since the native worker communication is event-based, which is inconvenient to handle in the middle of a function, the PR introduces the `comlink` library for convenient communication with the worker. This tiny library by Surma from Google abstracts away the event-based API and exposes a much handier promised-based API.

## Deployment URL
TODO (if gitlab starts working again)

### DIFF
--- a/src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
+++ b/src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
@@ -18,21 +18,35 @@ import { gql } from '@apollo/client';
 
 import { client } from 'src/gql-client';
 
-export type TranscriptChecksums = {
-  cdna: {
-    sequence_checksum: string;
+export type TranscriptSequenceData = {
+  cdna?: {
+    checksum: string;
+    label: string;
   };
-  cds: {
-    sequence_checksum: string;
+  cds?: {
+    checksum: string;
+    label: string;
   };
-  product: {
-    sequence_checksum: string;
+  protein?: {
+    checksum: string;
+    label: string;
   };
 };
 
-type GeneFragment = {
+type TranscriptQueryResult = {
   transcript: {
-    product_generating_contexts: TranscriptChecksums[];
+    product_generating_contexts: Array<{
+      cdna: {
+        sequence_checksum: string;
+      };
+      cds: {
+        sequence_checksum: string;
+      };
+      product: {
+        stable_id: string;
+        sequence_checksum: string;
+      };
+    }>;
   };
 };
 
@@ -47,6 +61,7 @@ const transcriptChecksumsQuery = gql`
           sequence_checksum
         }
         product {
+          stable_id
           sequence_checksum
         }
       }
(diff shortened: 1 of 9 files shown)


src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -59,10 +74,37 @@ type Variables = {
   transcriptId: string;
 };
 
-export const fetchTranscriptChecksums = (variables: Variables) =>
-  client
-    .query<GeneFragment>({
+export const fetchTranscriptSequenceData = (
```
msmanoj, line 77: I am not sure about a better name but this sounds like we are fetching the sequence within this function itself.

azangru: We are fetching information about the sequence that will later be used to fetch the sequence itself. Only it's no longer just checksums.

Would fetchTranscriptSequenceMetadata be better? 😄

msmanoj: Yes, `fetchTranscriptSequenceMetadata` sounds better.

azangru: Updated data to metadata and resolved conflicts.

msmanoj:
I think we need some indication that we have clicked on the download button and something is going on in the background.
I ended up clicking the button multiple times as I wasn't sure if I clicked it right and Chrome blocked automatic downloads.

I wonder if we should temporarily disable the button until the file is ready?

azangru:
> I think we need some indication that we have clicked on the download button and something is going on in the background.

I think so too, but that's for a different PR. Andrea had an idea [here](https://xd.adobe.com/view/a7fa7bcf-7dd5-4f9a-6190-6da6f2794465-d1cb/screen/a279bdbb-a041-456e-b02a-1ee629d57336/?fullscreen)

msmanoj (approved):

azangru:
test the button

imransl (approved):
This PR looks good to merge. I'll need to keep a note to myself to merge this PR into the instant download gene branch.

PR #451, opened 2021-02-09
Remove commas before validating region

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-938

## Description
- Removing the commas from user input before validating the region to fix the error

## Deployment URL
http://fix-navigation-bar.review.ensembl.org

## Views affected
Browser - > Region editor

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browserHelper.ts
+++ b/src/ensembl/src/content/app/browser/browserHelper.ts
@@ -178,12 +178,8 @@ export const validateRegion = async (params: {
     try {
       const url = `/api/genome/region/validate?genome_id=${genomeId}&region=${regionInput}`;
       const response: RegionValidationResponse = await apiService.fetch(url);
-      const regionId = buildEnsObjectId({
-        genomeId,
-        type: 'region',
-        objectId: regionInput
-      });
-      response.region_id = regionId;
+
+      response.region_id = `${genomeId}${response.region_id}`;
 
       processValidationMessages(
         getRegionValidationMessages(response),


imransl (approved):

azangru:
Sorry, I am confused now. I am trying to reproduce the original problem, and cannot anymore. Do you know if there really was a bug?

I remember not being able to change a chromosome, but now that I tried to do that again, I realise that I was forgetting to click on the green "Submit" button next to the region editor. I also remember we were quite happy to submit the original user's input, commas and all, for server-side validation. So do you know if the changes proposed in this PR are fixing anything that was broken?

msmanoj:
I think it happens only when we try to change the chromosome using the region editor.
That is when it calls the following endpoint and gets 500 in return:
http://internal-2020.ensembl.org/browser/locale/homo_sapiens_GCA_000001405_28:region:3:32364704-32,364,755

azangru:
The region id returned from the region validation endpoint ([example](https://2020.ensembl.org/api/genome/region/validate?genome_id=homo_sapiens_GCA_000001405_28&region=13:31,595,292-34,150,280)) is currently inconsistent with the region id of example objects returned from the genome info endpoint ([example](http://2020.ensembl.org/api/genome/info?genome_id=homo_sapiens_GCA_000001405_28)). I suppose it happened around [this PR](https://github.com/Ensembl/ensembl-client/pull/282), when we changed the way example object ids are reported by the api. We must have forgotten about the region validate endpoint. I asked Kamal to take a look.

msmanoj:
@azangru  - thanks for speaking to Kamal about this. He has now updated the endpoint and deployed it on to production.

Please refer to the following discussion on Slack for more info:
https://genomes-ebi.slack.com/archives/CC2RKR525/p1613608663137700

Closing this PR as it is no longer required.

PR #452, opened 2021-02-09
Refactor the types of entities that we get from Thoas

azangru:
## Type
- refactoring

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-879

## Description
Type definitions of our entities, in essence, describe the full set of properties that can be retrieved from our graphql api. However, we've been using these types, without any modification, when declaring the types of props of our React components.

This gave rise to several kinds of problems:

- Components, and occasionally helper functions, did not declare their true data requirements, i.e. the bare minimum of data that is sufficient for them to function properly, but rather reused the graphql response types that contained superfluous data.
- We lied to our type checker by pretending that a component had access to many more fields on an entity than what it actually requested or received from its parent (e.g. all the fields of the Gene type and of all of its nested types).
- There was no correspondence between what data a component actually requested using a graphql query, and the types that we declared that this component received.

This PR addresses the first and the second of these problems. Now components declare in their prop types only the bare minimum of data that is necessary for them to render. Due to typescript's structural typing, if incoming props in fact contain more data, this will be safely ignored.

There seem to be two alternative approaches to describing the types of entities that can be fetched from a graphql service. One approach recognises the fact that any field in any given query may be either requested or omitted and thus marks every field in the graphql response type as optional. This, however, makes such types very hard to work with. The other approach, on the contrary, marks every field in the response type as mandatory, but then this response type is used for deriving more narrow types that actually correspond to what is being fetched. This is the approach taken in this PR.

Notice that the third problem – inferring the type of the graphql response given the actual graphql query string in a component – remains unsolved by this PR.


src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -40,31 +40,27 @@ import ExternalReference from 'src/shared/components/external-reference/External
...
-    | 'stable_id'
-    | 'unversioned_stable_id'
-    | 'symbol'
-    | 'so_term'
-    | 'external_references'
-    | 'slice'
-    | 'spliced_exons'
-    | 'product_generating_contexts'
-  >
+// TODO: narrow down the types for spliced exons and product+generating_contexts
```
msmanoj, line 48: Is this something for a separate PR?

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -167,7 +163,7 @@ const TranscriptSummary = () => {
   const { gene, transcript } = data;
   const defaultProductGeneratingContext = transcript.product_generating_contexts.find(
     (entry) => entry.default
-  ) as ProductGeneratingContext;
+  ) as Transcript['product_generating_contexts'][number];
```
msmanoj, line 166: Would it be better if we define `Transcript['product_generating_contexts'][number]` separately and refer it here?
Like `type ProductGeneratingContext = Transcript['product_generating_contexts'][number]`

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -36,19 +36,29 @@ import * as urlFor from 'src/shared/helpers/urlHelper';
...
-import GeneFunction from 'src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction';
+import GeneFunction, {
+  Props as GeneFunctionProps
+} from 'src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction';
 import GeneRelationships from 'src/content/app/entity-viewer/gene-view/components/gene-relationships/GeneRelationships';
 import ViewInApp from 'src/shared/components/view-in-app/ViewInApp';
 import { CircleLoader } from 'src/shared/components/loader/Loader';
 
-import { Gene } from 'src/shared/types/thoas/gene';
+// import { Gene } from 'src/shared/types/thoas/gene';
```
msmanoj, line 53: Can this be removed?

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -36,19 +36,29 @@ import * as urlFor from 'src/shared/helpers/urlHelper';
 import { buildFocusIdForUrl } from 'src/shared/state/ens-object/ensObjectHelpers';
 import { parseFocusIdFromUrl } from 'src/shared/state/ens-object/ensObjectHelpers';
 
-import GeneOverviewImage from './components/gene-overview-image/GeneOverviewImage';
-import DefaultTranscriptslist from './components/default-transcripts-list/DefaultTranscriptsList';
+import GeneOverviewImage, {
+  GeneOverviewImageProps
+} from './components/gene-overview-image/GeneOverviewImage';
+import DefaultTranscriptslist, {
+  Props as DefaultTranscriptslistProps
```
msmanoj, line 43: `DefaultTranscriptslistProps` -> `DefaultTranscriptsListProps`?

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
azangru: Definitely! 😅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
azangru: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
azangru: ✅

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
azangru: ✅

src/ensembl/src/content/app/entity-viewer/shared/helpers/transcripts-sorter.ts
```diff
@@ -27,11 +28,13 @@ import {
 
 import { SortingRule } from 'src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice';
 
-import { SliceWithLocationOnly } from 'src/shared/types/thoas/slice';
+import { Slice } from 'src/shared/types/thoas/slice';
+
+type SliceWithOnlyLength = Pick2<Slice, 'location', 'length'>;
```
msmanoj, line 33: Would it be worth defining subset types like these within the `slice.ts` file?

azangru: Not sure. It depends on how commonly such types are reused or how much they make sense on their own. I didn't feel a strong urge to define such subtypes in the thoas-related types files.

msmanoj:
This PR looks good apart from the conflicts 👍

azangru:
Conflicts resolved 😄

msmanoj (approved):

PR #455, opened 2021-02-11
Handle empty uniprot xref in Entity Viewer protein view

azangru:
## Type
- Bug fix

## Description
At the moment, if Thoas does not return any links to Uniprot/Swissprot, we see a never-ending spinner. Example:  https://2020.ensembl.org/entity-viewer/triticum_aestivum_GCA_900519105_1/gene:TraesCS3D02G273600?view=protein

I think this change might solve the problem. I tried it locally, and it seems to work. Although the code in our ProteinListItemInfo component has become quite complex, so we might want to consider a refactor in the future.

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
@@ -88,10 +88,10 @@ const ProteinsListItemInfo = (props: Props) => {
   const proteinId =
     transcript.product_generating_contexts[0].product.unversioned_stable_id;
 
-  const { product } =
+  const { product: productWithProteinDomains } =
     transcriptWithProteinDomains?.product_generating_contexts[0] || {};
 
-  const uniprotXref = product?.external_references.find(
+  const uniprotXref = transcript.product_generating_contexts[0].product.external_references.find(
     (xref) => xref.source.id === SWISSPROT_SOURCE
   );
 
@@ -121,6 +121,7 @@ const ProteinsListItemInfo = (props: Props) => {
   useEffect(() => {
     const abortController = new AbortController();
     if (summaryStatsLoadingState === LoadingState.LOADING && !uniprotXref) {
+      // if uniprotXref is absent, we cannot fetch relevant data from PDBe; so pretend that we've successfully completed the request
       setSummaryStatsLoadingState(LoadingState.SUCCESS);
       return;
     }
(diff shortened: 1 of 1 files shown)


imransl (approved):

imransl (approved):

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -124,7 +120,7 @@ const ProteinsListItemInfo = (props: Props) => {
 
   useEffect(() => {
     const abortController = new AbortController();
-    if (xrefLoadingState === LoadingState.SUCCESS && !uniprotXref) {
+    if (summaryStatsLoadingState === LoadingState.LOADING && !uniprotXref) {
```
msmanoj, line 123: It looks like the SummaryStats will never get fetched as this condition will always be true during the initial render. This will in turn set the `summaryStatsLoadingState` to SUCCESS, which would mean that the second `if` statement condition will never return true.

azangru: > as this condition will always be true during the initial render

Why would it always be true during the initial render? `uniprotXref` comes from the transcript -> product_generating_context -> product -> external_references, which is already fetched and passed down to this component.

Try running this branch on your local machine. Does the problem appear there?

msmanoj: >uniprotXref comes from the transcript -> product_generating_context -> product -> external_references, which is already fetched and passed down to this component.

No, it's not. 
It is only getting set here:
https://github.com/Ensembl/ensembl-client/pull/455/files#diff-5f1e049df55be7bf2c2bccf34cb18736d9f52f263aa58d02c44cce3bb417ab59R106

msmanoj: > Try running this branch on your local machine. Does the problem appear there?

The problem mentioned in this PR doesn't seem to be occurring but I think the stats doesn't get displayed anymore.

azangru: I've updated the PR:

- get uniprotXref directly from the transcript passed as a property; do not wait until the transcript is enriched with protein domains
- improve the coordination between the top and the bottom part of the component while loading (PDBe is responding faster than our rest today; so I was getting protein stats appearing before the domains have even been downloaded)

msmanoj (approved):

msmanoj:
I've checked it locally and it looks good.

PR #456, opened 2021-02-12
Do not show arrow if summary and expanded stats have same content

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-932

## Description
If the content of the collapsed section is the same as the content of the expanded section, there is no point in expanding the stats section, and therefore we shouldn't be showing the up/down arrow in its right corner.

## Deployment URL
http://remove-stats-arrow.review.ensembl.org/

## Views affected
Species stats page


src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -120,6 +121,7 @@ const getExpandedContent = (props: ContentProps) => {
               [styles.statsGroupWithExampleLink]:
                 !group_index && !row_index && exampleLinkText
             });
+            groupsContent.push(...groupStats);
```
msmanoj, line 124: I think this could just be a simple counter like `totalGroupsContent` instead of being an array.

src/ensembl/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
```diff
@@ -145,7 +147,10 @@ const getExpandedContent = (props: ContentProps) => {
     })
     .filter(Boolean);
 
-  return expandedContent.length ? (
+  const hasExpandedContent =
+    summaryStats?.length !== groupsContent.length && expandedContent.length;
```
imransl, line 151: Is `expandedContent.length` used here to check if it is 0 or more?

azangru:
@jyo @msmanoj what should be the algorithm, in pseudocode, to decide whether a section has expanded content? I was reading through the `getExpandedContent` function, but got confused.

Is it just as easy as: "if there is more than one group or if the first group has more members than the summary stats, then it means we have expanded content.

If so, wouldn't it be a good idea to extract this in a dedicated function:

```ts
const hasExpandedContent = () => {
  const { groups, summaryStats } = statsSection;

  const [firstGroup] = groups;

  return groups.length > 1 || summaryStats.length === firstGroup.length
}
```

jyothishnt:
> @jyo @msmanoj what should be the algorithm, in pseudocode, to decide whether a section has expanded content? I was reading through the `getExpandedContent` function, but got confused.
> 
> Is it just as easy as: "if there is more than one group or if the first group has more members than the summary stats, then it means we have expanded content"?
> 
> If so, wouldn't it be a good idea to extract this in a dedicated function:
> 
> ```ts
> const hasExpandedContent = () => {
>   const { groups, summaryStats } = props.statsSection;
> 
>   const [firstGroup] = groups;
> 
>   return groups.length > 1 || summaryStats.length === firstGroup.length
> }
> ```

What if there are 2 groups eaach with one entry and we have both of those entries shown in the summary?

azangru:
> What if there are 2 groups eaach with one entry and we have both of those entries shown in the summary?

Will we ever? Can the summary contain combined content from several groups?

msmanoj:
>Will we ever? Can the summary contain combined content from several groups?

I would think it can.

I wonder if we can apply this filter within the `speciesGeneralHelper` file without having to do a lot of processing inside the UI component. I have created a draft PR which will actually compare all the summary and expanded stats and hide the hides the expanded stats if they both match: https://github.com/Ensembl/ensembl-client/pull/462/files.

jyothishnt:
> > Will we ever? Can the summary contain combined content from several groups?
> 
> I would think it can.
> 
> I wonder if we can apply this filter within the `speciesGeneralHelper` file without having to do a lot of processing inside the UI component. I have created a draft PR which will actually compare all the summary and expanded stats and hide the hides the expanded stats if they both match: https://github.com/Ensembl/ensembl-client/pull/462/files.

Yes thats a very good idea. Not giving back the expanded result would get rid off all the calculations in the UI :-) thanks @msmanoj  If you could get this merged I can change the code accordingly in this component

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -660,38 +661,53 @@ export const getStatsForSection = (props: {
       })
     : undefined;
 
+  const isExpandedContentSame = isEqual(
+    Object.keys(filteredData).sort(),
+    summaryStatsKeys?.sort()
+  );
```
msmanoj, line 667: Just to confirm, why do we need to sort and check for equality?

jyothishnt: that should also work. This will do a strict check (just in case if the content are different)

jyothishnt: decided to go with .length() check

msmanoj (approved):

src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -660,38 +660,51 @@ export const getStatsForSection = (props: {
...
+
+  const processedGroups = groups.map((group) => {
+    const groupStats = groupsStatsMap[group];
+
+    const processedStats = groupStats
+      .map((subGroupStats) => {
+        const processedSubGroupStats: IndividualStat[] = [];
+
+        subGroupStats.forEach((stat) => {
+          if (filteredData[stat] || filteredData[stat] === 0) {
```
azangru, line 682: I'm confused. Haven't we [guaranteed, by this point](https://github.com/Ensembl/ensembl-client/blob/fdfee6f80fe1bf5abe23f4a93d37edb999ff94b8/src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts#L625-L633), that we are interested in everything that's in `filteredData`, so we do not need this extra if check?

jyothishnt: I also thought the same but not really. This is looping through subGroupStats which is not filtered.

azangru (approved):

PR #440, opened 2021-01-20
EV bookmarks

jyothishnt:
## Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-461

## Description
Adding new bookmarks feature showing example links and 20 previously viewed links.

## Deployment URL
http://ev-bookmarks.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000139618?view=protein

## Views affected
EntityViewer

- [ ] I have checked any requirements for Google Analytics
- [ ] I have created any required tests
- [ ] The PR has as its final commit a WASM/JS update commit if any Rust files were updated.

PR #459, opened 2021-02-15
Species specific sidebar state

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-896

## Description
- Updated the species page sidebar open/close state to be species specific

## Views affected
Species page sidebar

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
+++ b/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
@@ -23,7 +23,9 @@ export const isSidebarOpen = (state: RootState) => {
   if (!activeGenomeId) {
     return false;
   }
-  return !!state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen;
+  return (
+    state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? false
+  );
 };
 
 export const getActiveGenomeSidebarPayload = (state: RootState) => {
--- a/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
+++ b/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
@@ -24,7 +24,7 @@ export const isSidebarOpen = (state: RootState) => {
     return false;
   }
   return (
-    state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? false
+    state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? true
   );
 };
 
--- a/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
+++ b/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
@@ -23,9 +23,7 @@ export const isSidebarOpen = (state: RootState) => {
   if (!activeGenomeId) {
     return false;
   }
-  return (
-    state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? true
-  );
+  return state.speciesPage.sidebar[activeGenomeId]?.isSidebarOpen ?? true;
 };
 
 export const getActiveGenomeSidebarPayload = (state: RootState) => {
@@ -35,5 +33,5 @@ export const getActiveGenomeSidebarPayload = (state: RootState) => {
     return null;
   }
 
-  return state.speciesPage.sidebar.species[activeGenomeId]?.payload || null;
+  return state.speciesPage.sidebar[activeGenomeId]?.payload || null;
 };
--- a/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
+++ b/src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
@@ -17,8 +17,14 @@
 import { getActiveGenomeId } from 'src/content/app/species/state/general/speciesGeneralSelectors';
 import { RootState } from 'src/store';
 
-export const isSidebarOpen = (state: RootState) =>
-  state.speciesPage.sidebar.isOpen;
+export const isSidebarOpen = (state: RootState) => {
+  const activeGenomeId = getActiveGenomeId(state);
+
+  if (!activeGenomeId) {
+    return false;
+  }
+  return !!state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen;
+};
 
 export const getActiveGenomeSidebarPayload = (state: RootState) => {
   const activeGenomeId = getActiveGenomeId(state);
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
```diff
@@ -17,8 +17,14 @@
...
 
-export const isSidebarOpen = (state: RootState) =>
-  state.speciesPage.sidebar.isOpen;
+export const isSidebarOpen = (state: RootState) => {
+  const activeGenomeId = getActiveGenomeId(state);
+
+  if (!activeGenomeId) {
+    return false;
+  }
+  return !!state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen;
```
azangru, line 26: It's strange to see the boolean (`isSidebarOpen`) converted to a boolean with the `!!`.
Don't you mean return `state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? false`?

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSlice.ts
```diff
@@ -110,7 +116,10 @@ const speciesPageSidebarSlice = createSlice({
       }>
     ) {
       if (!state.species[action.payload.genomeId]) {
-        state.species[action.payload.genomeId] = { payload: null };
+        state.species[action.payload.genomeId] = {
+          payload: null,
+          isSidebarOpen: true
+        };
```
azangru, line 122: There are several strange things going on; I am not sure I am reading them right.

1) `setSidebarPayloadForGenomeId` handles situations when there isn't `state.species[action.payload.genomeId]`, but `toggleSidebar` doesn't. Is this a good thing?
2) The selector's default `isSidebarOpen` value for when there is no active genome id is `false`, but the reducer sets the default value as `true`. Is this inconsistent?
3) More importantly, do we know whether the code as currently written will even work? I see `state.species[action.payload.genomeId]` is first assigned an object value, and then this object is modified. I believe immer shouldn't even allow this to happen — from what I recall, it freezes the modified part of the subtree.

Consider the following alternative:

```ts
import get from 'lodash/get';
import merge from 'lodash/merge';


type StateForGenome = {
  payload: SpeciesSidebarPayload | null;
  isSidebarOpen: boolean;  
};

const initialStateForGenome: StateForGenome = {
  payload: null,
  isSidebarOpen: true
};

const updateStateForGenome = (state: SpeciesPageSidebarState, genomeId: string, fragment: Partial<StateForGenome>) => {
  const stateForGenome = get(state, `species.${genomeId}`, initialStateForGenome);
  const updatedStateForGenome = merge({}, stateForGenome, fragment);
  state.species[genomeId] = stateForGenome; // here we will be relying on the immer magic
};


// then, in the action creator inside the createSlice

    toggleSidebar(
      state,
      action: PayloadAction<{
        genomeId: string;
      }>
    ) {
      const isSidebarOpen = state.species[action.payload.genomeId].isSidebarOpen ?? true;
      updateStateForGenome(state, action.genomeId, { isSidebarOpen });

    }

// etc.
```

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
msmanoj: That's done

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSlice.ts
msmanoj: >setSidebarPayloadForGenomeId handles situations when there isn't state.species[action.payload.genomeId], but toggleSidebar doesn't. Is this a good thing?

I thought the sidebar will not get displayed without the payload but I slightly had a strange feeling about this change. thanks for pointing it out.

>The selector's default isSidebarOpen value for when there is no active genome id is false, but the reducer sets the default value as true. Is this inconsistent? 

Good spot. No, I think the selector should return true by default.

>More importantly, do we know whether the code as currently written will even work? I see state.species[action.payload.genomeId] is first assigned an object value, and then this object is modified. I believe immer shouldn't even allow this to happen — from what I recall, it freezes the modified part of the subtree.

Yes, you are right. I did test it and it was working so I am not sure how. I have now updated it with the suggested alternative. Thanks

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
```diff
@@ -17,8 +17,16 @@
...
-  state.speciesPage.sidebar.isOpen;
+export const isSidebarOpen = (state: RootState) => {
+  const activeGenomeId = getActiveGenomeId(state);
+
+  if (!activeGenomeId) {
+    return false;
+  }
+  return (
+    state.speciesPage.sidebar.species[activeGenomeId]?.isSidebarOpen ?? true
+  );
```
azangru, line 28: image

shouldn't the default return values be the same?

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSlice.ts
```diff
@@ -58,17 +60,23 @@ export type SpeciesSidebarPayload = {
...
-    [genomeId: string]: {
-      payload: SpeciesSidebarPayload | null;
-    };
+    [genomeId: string]: StateForGenome;
   };
 };
 
 const initialState: SpeciesPageSidebarState = {
-  isOpen: true,
   species: {}
```
azangru, line 80: As a side note, why did we decide to have the `species` key in the `SpeciesPageSidebarState`? Are we expecting any other keys at the same level?

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSelectors.ts
msmanoj: I thought this would mean that we keep the sidebar closed when there is no `activeGenomeId` and open it by default when we have `activeGenomeId`.

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSlice.ts
msmanoj: Good question. I am not sure if we will be having any other keys at the moment. I wonder if we will be having more sections within the sidebar like we have on other apps like bookmarks.

azangru: In EntityViewer (see PR https://github.com/Ensembl/ensembl-client/pull/440, which we pair-programmed with Jyo), we decided to put bookmarks in their own slice of the state. Sidebar is more for the UI purposes, whereas bookmarks is a more abstract concept, not strictly related to any UI element.

I'm inclined to deleting this species field, but I am leaving the decision to you. We may wait and see if anything else pops up in the sidebar.

azangru (approved):

src/ensembl/src/content/app/species/state/sidebar/speciesSidebarSlice.ts
msmanoj: I have now deleted the key as discussed during the code review.

PR #460, opened 2021-02-22
Fix region validation error

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-938

## Description
- Using the region_id returned form the validation endpoint to instead of the one from user input

## Views affected
Browser - > navigatione ditor

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/browser/browserHelper.ts
+++ b/src/ensembl/src/content/app/browser/browserHelper.ts
@@ -181,7 +181,7 @@ export const validateRegion = async (params: {
       const regionId = buildEnsObjectId({
         genomeId,
         type: 'region',
-        objectId: regionInput
+        objectId: response.region_id || ''
       });
       response.region_id = regionId;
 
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/browser/browserHelper.ts
```diff
@@ -181,7 +181,7 @@ export const validateRegion = async (params: {
       const regionId = buildEnsObjectId({
         genomeId,
         type: 'region',
-        objectId: regionInput
+        objectId: response.region_id || ''
```
azangru, line 184: the `''` would surely create an invalid ens object id.

I understand that this is because of the way we defined the response type. We should either convince typescript that `response.region_id` at this point will always be `string` (`objectId: response.region_id as string`), or clarify what the possible response shapes are and update our types.

msmanoj: I've updated it to `objectId: response.region_id as string` for now.

azangru: 👌

azangru (approved):

PR #462, opened 2021-02-23
Filter stats within the helper

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-932

## Description
- The expandedStats are now filtered out before getting sent to the component if it matches with the summary stats

## Views affected
Species page stats

### DIFF
--- a/src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
+++ b/src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
@@ -651,38 +651,51 @@ export const getStatsForSection = (props: {
       })
     : undefined;
 
+  const processedGroups = groups.map((group) => {
+    const groupStats = groupsStatsMap[group];
+    const expandedStatsKeys: Stats[] = [];
+    const processedStats = groupStats
+      .map((subGroupStats) => {
+        const processedSubGroupStats: IndividualStat[] = [];
+
+        subGroupStats.forEach((stat) => {
+          if (filteredData[stat]) {
+            expandedStatsKeys.push(stat);
+            const individualStat = buildIndividualStat({
+              primaryKey: stat,
+              primaryValue: filteredData[stat],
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -651,38 +651,51 @@ export const getStatsForSection = (props: {
...
+      .filter(Boolean);
+
+    const summaryExpandedStatsDifference = expandedStatsKeys?.filter(
+      (key) => !summaryStatsKeys?.includes(key)
+    );
+
+    if (!summaryExpandedStatsDifference.length || !processedStats.length) {
+      return {
+        title: groupTitles[group]
+      };
```
azangru, line 686: Could you remind me of the purpose of these objects, with title, but no stats?

msmanoj: It will still display the section with the title and the summary stats are handled separately. There will be no expandedStats to display which will hide the expanded section.

azangru (approved):
I am slightly uneasy about the function that produces the `processedGroups` variable. It's large, it tries to do several things — although very closely related things — at once, it uses ambiguous words (e.g. a `group` seems to be a group name rather than an actual group of objects) and it may be rather hard to understand. It might benefit from splitting into several smaller functions. Or it might not 🤷‍♂️ After staring at it for half an hour, I convinced myself that I am probably OK with it.

@imransl @jyothishnt you are very welcome to offer your opinions about this function.

jyothishnt:
I think this looks ok except that it was a little hard for me to understand due to the complexity of the groups. Please go ahead and merge it @msmanoj and I will continue with my other PR.

PR #465, opened 2021-02-24
Add transcript filter functionality to entity viewer

ens-ma7:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-888

## Importance
N/A

## Description
This is for the filtering functionality in Entity viewer.  The component was already implemented together with the test script. This PR is for adding the functionality to filter the transcripts. When you click on one of the filters, it will only show the transcripts with that filter. 

For now we are only doing the filter by so_term category, there might be more in the future and we can make the code more generic then as we know more about design and requirements.

## Deployment URL
http://entityviewer-filter.review.ensembl.org/

## Views affected
Entity Viewer (Transcript Image)

### Other effects
None

## Possible complications
None

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-filter.test.ts
+++ b/src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-filter.test.ts
@@ -1,0 +1,136 @@
+import { filterTranscriptsBySOTerm } from '../transcripts-filter';
+
+import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
+
+/* Creating filters with different filter set to true */
+const proteinCodingfilters = {
+  protein_coding: true,
+  retained_intron: false,
+  processed_transcript: false,
+  nonsense_mediated_decay: false
+};
+
+const retainedIntronfilters = {
+  protein_coding: false,
+  retained_intron: true,
+  processed_transcript: false,
+  nonsense_mediated_decay: false
+};
+
+const processedTranscriptfilters = {
+  protein_coding: false,
+  retained_intron: false,
+  processed_transcript: true,
+  nonsense_mediated_decay: false
+};
+
+const nonsenseMediatedDecayfilters = {
+  protein_coding: false,
+  retained_intron: false,
+  processed_transcript: false,
+  nonsense_mediated_decay: true
+};
+
+/* Creating dummy transcripts with different different so_term  to test so_term filtering */
+const createProteinCodingTranscript = () => {
+  const transcript = createTranscript();
+  transcript.so_term = 'protein_coding';
+  return transcript;
+};
+
+const createProcessedTranscript = () => {
+  const transcript = createTranscript();
+  transcript.so_term = 'processed_transcript';
+  return transcript;
+};
+
+const createRetainedIntronTranscript = () => {
+  const transcript = createTranscript();
+  transcript.so_term = 'retained_intron';
+  return transcript;
+};
+
+const createNonsenseMediatedDecayTranscript = () => {
+  const transcript = createTranscript();
+  transcript.so_term = 'nonsense_mediated_decay';
+  return transcript;
+};
+
+const ProteinCodingTranscript = createProteinCodingTranscript();
+const ProcessedTranscript = createProcessedTranscript();
+const RetainedIntronTranscript = createRetainedIntronTranscript();
+const NonsenseMediatedDecayTranscript = createNonsenseMediatedDecayTranscript();
+
+describe('filter transcripts by so_term', () => {
+  it('filters transcripts by protein_coding correctly', () => {
+    const Transcripts = [
+      ProteinCodingTranscript,
+      ProcessedTranscript,
+      RetainedIntronTranscript,
+      NonsenseMediatedDecayTranscript
+    ];
+
+    const expectedTranscripts = [ProteinCodingTranscript];
+
+    const filteredTranscripts = filterTranscriptsBySOTerm(
+      Transcripts,
+      proteinCodingfilters
+    );
+
+    expect(filteredTranscripts).toEqual(expectedTranscripts);
+  });
+
+  it('filters transcripts by retained_intron correctly', () => {
+    const Transcripts = [
+      ProteinCodingTranscript,
+      ProcessedTranscript,
+      RetainedIntronTranscript,
+      NonsenseMediatedDecayTranscript
+    ];
+
+    const expectedTranscripts = [RetainedIntronTranscript];
+
+    const filteredTranscripts = filterTranscriptsBySOTerm(
+      Transcripts,
+      retainedIntronfilters
+    );
+
+    expect(filteredTranscripts).toEqual(expectedTranscripts);
+  });
+
+  it('filters transcripts by processed_transcript correctly', () => {
+    const Transcripts = [
+      ProteinCodingTranscript,
+      ProcessedTranscript,
+      RetainedIntronTranscript,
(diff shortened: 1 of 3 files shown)


azangru:
Umm, if our goal is to create a set of filters that can eventually cover various properties of a transcript, we will need to be smarter than our original implementation of the filters state:

image

The way the filters state in redux looks now:

https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts#L55

will work for filtering by a single attribute, such as the so_term, but is unlikely to scale to the point when we start filtering by different attributes (biotype and appris and tsl). I wonder whether it might be better to take a pause and figure out a way for filtering by several attributes now rather, because we already know that we will need that, than to refactor/rewrite this code later.

ens-ma7:
@azangru i have explained the reason for doing just this category in the description. I think if we have something for Andrea to have a look at, she will have a better feeling of how things are working. I remember she or steve said we might just have SO Term for MAP. Probably the code will need refactoring at that time. For now if we have something for Andrea/Steve to play with they will have a better understanding what else to add.

azangru:
I'm not saying we should postpone the filters until TSLs and APRIS become available. I'm saying we should probably build the filters with the anticipation of multiple filter categories, such as TSLs and APRIS, which we expect will be coming. As the code is written now, it doesn't seem to offer a way for adding new filter categories without a rewrite. That's a technical debt.

ens-st3:
> I'm not saying we should postpone the filters until TSLs and APRIS become available. I'm saying we should probably build the filters with the anticipation of multiple filter categories, such as TSLs and APRIS, which we expect will be coming. As the code is written now, it doesn't seem to offer a way for adding new filter categories without a rewrite. That's a technical debt.

There are distinct JIRAs for adding content as it becomes available eg 1017. What level of technical debt we are willing to accept now is a decision we need to make.  

Re biotypes specifically, we will use them when we have them. Are they essential for MAP ? Prob not.

ens-ma7:
Sorry i dont agree it will be a technical debt as we dont know yet for sure how this will pan out. Not sure if you missed something but we dont know yet what the other filters will be or if there will be other categories and how will they work with the sorting. We might have to rewrite the whole thing again if lets say the sorting change. I think doing more on it than required will be the opposite of technical debt maybe technical  overpayment :)  
What are the changes are you suggesting to make, i thought it was generic enough already. Maybe something for code review, will be good to get Andrea as well.

ens-ma7:
Probably best lets see what Andrea said in the code review and then if we could merge this one and after that i will create another ticket to make the code able to handle more filters. This will also help me understand the code (its easier to learn the code when you see exactly whats happening)

azangru:
ENSWBSITES-1017 being in the backlog should probably point us in the direction of building the filters as if we were building them for ENSWBSITES-1017 — writing code in a way that will handle multiple filter criteria and using just a single so_term filter criterion which we'll then replace with biotype when it becomes available.

azangru:
> Not sure if you missed something but we dont know yet what the other filters will be.

I'm confused then. How confident are we in ENSWBSITES-1017 then?

> What are the changes are you suggesting to make

There are several indications that filters want to be grouped.

One indication comes from the UI, where three groups of filters are separated into three different columns:

image

The other indication comes from the filter function, which knows which property of a transcript to check (line 13 on screenshot below):

image

Currently, the filters in redux aren't being grouped at all.

ens-ma7:
i agree with the grouping with you @azangru , let me pass this by andrea in the code review or earlier if i can (probably best in code review) and then we will understand better the grouping and what to do with the code. Question to ask andrea, will there be a grouping and how will it be?

ens-st3:
The associated EA-695 ticket describes what data we are expecting. The bottom 3 are not filters

src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-filter.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  transcript.so_term = 'nonsense_mediated_decay';
+  return transcript;
+};
+
+const ProteinCodingTranscript = createProteinCodingTranscript();
+const ProcessedTranscript = createProcessedTranscript();
+const RetainedIntronTranscript = createRetainedIntronTranscript();
+const NonsenseMediatedDecayTranscript = createNonsenseMediatedDecayTranscript();
+
+describe('filter transcripts by so_term', () => {
```
azangru, line 64: **Minor point:** the string passed to the `describe` function is usually the name of the function or a class that is being tested. In this case `describe('filterTranscriptsBySOTerm', () => {...})`. It even reads like an English sentence: `describe('a thing') ... it<i.e. that thing you mentioned in the describe block>('does so-and-so')`.

**Major point:** The tests in this file are very repetitive. We don't really need to test 4 different specific so_term strings.

If we want to test filtering transcripts by so_term, then we only need a single test. In pseudocode, the test would look something like this:

- Generate a bunch of transcripts (e.g. five transcripts)
- Choose a couple of transcripts randomly out of that bunch and add some unique string to their so_term field. It doesn't have to be a real so_term, because our filter function doesn't care about whether so_terms are real.
- Apply your filter function to this bunch: `filterTranscriptsBySOTerm(transcripts, { my_unique_string: true })`
- Check that the output from the filter function contains only transcripts that have the so_term field set to that unique string.

azangru:
For future reference: the majority opinion during the code review was not to bother about the support for more than one kind of filter now. Which means that we are postponing a more scalable implementation of transcript filters until later, and the moment we add another filter we'll need to rework, from ground up, the way we are setting up the filters, selecting them from the state, and choosing the appropriate filtering function to apply to transcripts.

ens-ma7:
@azangru thanks for the PR feedback, i have pushed the changes. Note that i have valid so_term in the test cases just so that it make sense when reading it but i do agree that we can put any string instead.

ens-ma7:
> **For future reference:** the majority opinion during the code review was not to bother about the support for more than one kind of filter now. Which means that we are postponing a more scalable implementation of transcript filters until later, and the moment we add another filter we'll need to rework, from ground up, the way we are setting up the filters, selecting them from the state, and choosing the appropriate filtering function to apply to transcripts.

Just to add because there is too many uncertainties regarding how the filters will pan out (we don't know where the filters will be applied, what groupings we will have,....), its better we keep making the code scalable as we know more about the filters instead of having to rewrite/change them later. I will like to carry on the work on making it scalable, another opportunity to understand react, redux and testing more.

src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-filter.test.ts
```diff
@@ -0,0 +1,62 @@
+import { filterTranscriptsBySOTerm } from '../transcripts-filter';
+
+import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
+
+/* Creating filters with different filter set to true/false */
+const proteinCodingfilters = {
```
azangru, line 6: proteinCodingfilters -> proteinCodingFilters

azangru (approved):
Probably want to capitalise the word Filters in `proteinCodingfilters`; but otherwise ok

ens-ma7:
QA: Andrea has approved the expected behaviour (see comment in ENSWBSITES-888). There is another task that came out of the QA which I have created a ticket for (applying the filtering on protein view - ENSWBSITES-1038).

PR #466, opened 2021-02-24
ENSWBSITES-937: show "None" when there is no gene name and synonyms i…

ens-ma7:
…n Entity Viwer RH panel

## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-937

## Importance
N/A

## Description
Showing "None" in the entity viewer Right Hand Panel when there is no gene name or no synonyms and also updating corresponding test.

## Deployment URL
http://no-genename.review.ensembl.org
Example Gene with no Gene Name and no Synonyms:
http://no-genename.review.ensembl.orgentity-viewer/triticum_aestivum_GCA_900519105_1/gene:TraesCS3D02G273500?view=transcripts

## Views affected
Entity viewer Right Hand Panel

### Other effects
None

## Possible complications
None

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
@@ -138,7 +138,7 @@ describe('<GeneOverview />', () => {
       expect(geneDetailsElement?.textContent).toBe(stableId);
     });
 
-    it('does not render gene name section if gene name is not available', () => {
+    it('shows "None" message for gene without gene name', () => {
       const geneData = { ...completeGeneData, name: null };
       (useQuery as any).mockImplementation(() => ({
         data: { gene: geneData },
@@ -148,10 +148,10 @@ describe('<GeneOverview />', () => {
       const { container } = render(<GeneOverview />);
 
       const geneNameElement = container.querySelector('.geneName');
-      expect(geneNameElement).toBeFalsy();
+      expect(geneNameElement?.textContent).toBe('None');
     });
 
-    it('shows "no synonyms" message for gene without synonyms', () => {
+    it('shows "None" message for gene without synonyms', () => {
       const geneData = { ...completeGeneData, alternative_symbols: [] };
       (useQuery as any).mockImplementation(() => ({
         data: { gene: geneData },
@@ -161,7 +161,7 @@ describe('<GeneOverview />', () => {
       const { container } = render(<GeneOverview />);
 
       const synonymsElement = container.querySelector('.synonyms');
-      expect(synonymsElement?.textContent).toBe('No synonyms');
+      expect(synonymsElement?.textContent).toBe('None');
     });
   });
 });
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
@@ -138,7 +138,7 @@ describe('<GeneOverview />', () => {
       expect(geneDetailsElement?.textContent).toBe(stableId);
     });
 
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.tsx
```diff
@@ -79,18 +79,16 @@ const GeneOverview = () => {
...
 
-      {!!gene.name && (
-        <>
-          <div className={styles.sectionHead}>Gene name</div>
-          <div className={styles.geneName}>{gene.name}</div>
-        </>
-      )}
+      <div className={styles.sectionHead}>Gene name</div>
+      <div className={styles.geneName}>
+        {gene.name && gene.name.length ? gene.name : 'None'}
```
imransl, line 84: Can we change this `gene.name && gene.name.length` to just `gene.name?.length`?

azangru: I'm wondering whether we even need to check the length of gene.name. The backend should send either a string or a null as a value of this field. It shouldn't ever send an empty string.

ens-ma7: @imransl thanks for suggestion, it makes sense, i have made the changes.
@azangru whether the backend send us null or empty we will still need to make a check to display "None" unless we get the backend to send us the string "None" when it is empty. How long will it take them to do this? Also i feel like these kind of check is better done on our side just so that we dont have to ask the backend to make the change everytime. It will be easier and quicker for us to make the change :)

azangru: > whether the backend send us null or empty we will still need to make a check to display "None" unless we get the backend to send us the string "None" when it is empty.

I don't understand...

We expect the backend to send us either a fully formed string if a gene has a name, or `null` if it doesn't. We do not expect them to ever send us an empty string.

Even having said that, an expression `{gene.name || 'None'}` should be just as effective as `{gene.name && gene.name.length ? gene.name : 'None'}`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/overview/GeneOverview.test.tsx
```diff
@@ -138,7 +138,7 @@ describe('<GeneOverview />', () => {
       expect(geneDetailsElement?.textContent).toBe(stableId);
     });
 
-    it('does not render gene name section if gene name is not available', () => {
+    it('shows "None" message for gene without gene name', () => {
```
azangru, line 141: Not insisting upon this change, but wondering whether the description of the test might benefit from a little bit less detail?

E.g.: `it('shows that the gene does not have a name')` and `it('shows that the gene does not have synonyms')`?

ens-ma7: In my head the test was to check if it is showing the message "None" and hence the description. Because before it was showing No synonyms instead but we had to change this. I wanted to put the emphasis on "None" since it looks like that the right word. Is that a good enough reason to not change it :) if not i am happy to change it.

azangru: The word "None", as the phrase "no synonyms" before it (mea culpa), is a very specific implementation detail, which we are already checking in the body of the test. It's generally advisable to write test descriptions at somewhat higher level than the actual implementation details that are being tested.

ens-ma7: Makes sense @azangru i have pushed the changes.

azangru (approved):

PR #469, opened 2021-02-25
display multiple uniprot entries if available

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-910
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-892

## Description

Update the display of xrefs as below

1. when we have a SWISSPROT we show just that
2. when we don't have a SWISSPROT we show one or more TrEMBL

## Deployment URL
http://protein-xrefs.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000204252?view=protein
http://protein-xrefs.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000108622?view=protein

Multiple Uniprot/Trembl  http://protein-xrefs.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000112561?view=protein
Open transcript ENST00000424495.2

## Views affected
Gene View


jyothishnt:
Found a problem with ENST00000532392.2 in http://protein-xrefs.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000164733?view=protein
If there are < 3 ids, it lists them horizontally, not as a column. Will fix it soon.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -21,9 +18,9 @@ $content-width: $gene_image_width;
 }
 
 .proteinSummaryTop {
-  align-items: center;
   display: grid;
-  grid-template-columns: [references] 300px [download] 395px;
+  grid-template-columns: [references] 350px [download] 345px;
+  align-items: flex-start;
```
azangru, line 23: Flex-start is supposed to only apply to flexboxes ([docs](https://drafts.csswg.org/css-align/#positional-values)).

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +57,42 @@ $content-width: $gene_image_width;
   justify-content: center;
   margin-top: 40px;
 }
+.errorMessage{
```
azangru, line 60: I see this class is added in this PR, but can't find where it is used. Is it?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +57,42 @@ $content-width: $gene_image_width;
...
+  min-width: 50px;
+
+  .chevron {
+    margin-left: 10px;
+    margin-bottom: -1px;
+    height: 12px;
+    width: 12px;
+  
+  }
+  .close {
```
azangru, line 80: Although we don't follow the BEM methodology in its entirety, some of its CSS conventions, such as identifying an element and a modifier, are useful, because they help with readability. In the original BEM syntax it would look like so: `.element_modifier`, or so: `element--modifier`; but we often write it like so: `elementModifier`.

In this case, the element is the chevron, and this additional class is modifying the behaviour of the chevron. We would probably name it something like `.chevronOpened` or `.chevronRotated`.

azangru:
Agreed during code review that the toggle should not get pushed away to the right from the first link when the list of the references is expanded.

I suppose the "+38" should also disappear when you expand a list of references.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
jyothishnt: used chevronUp/Down

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -21,22 +18,22 @@ $content-width: $gene_image_width;
...
   display: grid;
-  grid-template-columns: [references] 300px [download] 395px;
+  grid-template-columns: [references] 350px [download] 345px;
 }
 
 .downloadWrapper {
   display: flex;
   justify-content: flex-end;
   grid-column: download;
+  align-items: flex-start;
```
azangru, line 29: It seems that the reason you are adding this flex-start is because the height of the `.downloadWrapper` becomes too large:

image

and the button gets distorted without this rule:



I would argue that what we really should try to do is to prevent the `.downloadWrapper` from getting this high in the first place. The reason it is too high is because it is trying to occupy the full height of the grid cell. If you add `align-items: start` to the `.proteinSummaryTop` class, the `.downloadWrapper` element will keep its natural height:

image

(The same result can be achieved by adding `align-self: start` to `.downloadWrapper`; but `align-items: start` on `.proteinSummaryTop`) makes more sense.

azangru:
As a reminder from the code review meeting, it was suggested that the "+ X" label be greyed out when the rest of the links are shown.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +57,41 @@ $content-width: $gene_image_width;
...
+
+.chevron {
+  margin-left: 6px;
+  margin-bottom: -1px;
+  height: 12px;
+  width: 12px;
+}
+
+.chevronUp {
+  transform: rotate(0deg);
```
azangru, line 81: Isn't this redundant? It would help if you animated the chevron's transition between the up and down state (as is done e.g. in expandable sections with species stats on the species page; but if you aren't going to animate the transition, what's the point of this rule?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +57,41 @@ $content-width: $gene_image_width;
...
+
+.xrefsWrapper {
+  padding-bottom: 10px;
+  display: flex;
+  flex-direction: column;
+}
+
+.xrefGroupWrapper {
+  display: flex;
+  flex-direction: column;
```
azangru, line 96: I don't think these two CSS rules are doing anything. Getting stacked on top of one another is the default behaviour of elements with display: block, which is the display property of this wrapper's children.

By the way, why is there an xrefGroupWrapper inside another xrefGroupWrapper?

image

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +57,41 @@ $content-width: $gene_image_width;
...
+}
+
+.chevronDown {
+  transform: rotate(-180deg);
+}
+
+.xrefsWrapper {
+  padding-bottom: 10px;
+  display: flex;
+  flex-direction: column;
```
azangru, line 91: `display: flex` and `flex-direction: column` don't seem to be doing anything for this element.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +274,108 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+                accessionId={displayXref.accession_id}
+                name={displayXref.name}
+              />
+              <div className={styles.xrefGroupChevron} onClick={toggleXrefGroup}>
+                + {xrefs.length - 1}
+                <ChevronDown className={chevronClass} />
+              </div>
+            </div>
+          ) : (
+            <div className={styles.xrefGroupWrapper}>
```
azangru, line 333: Looks like it's here that one group wrapper gets wrapped in another group wrapper.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +274,108 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+
+  const chevronClass = classNames(
+    styles.chevron,
+    isXrefGroupOpen? styles.chevronDown : styles.chevronUp
+  );
+
+  if (xrefs.length > 3) {
+    const displayXref = xrefs[0];
+
+    if (!displayXref) {
```
azangru, line 313: How can this happen?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +274,108 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+                />
+                <div className={styles.xrefGroupChevron} onClick={toggleXrefGroup}>
+                  + {xrefs.length - 1}
+                  <ChevronDown className={chevronClass} />
+                </div>
+              </div>
+              { xrefs.map((xref, i) => {
+                if (i === 0) { // Ignore first element as it is displayed above with chevron.
+                  return;
+                }
```
azangru, line 349: Instead of mapping over all xrefs and ignoring the first one, create a slice that excludes the first xref and map over it (`xrefs.slice(1).map(...)`)

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
jyothishnt: This was just to make the css naming good when used inside the component

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -223,15 +240,10 @@ const ProteinsListItemInfo = (props: Props) => {
               </div>
             )}
           </div>
-          {proteinSummaryStats &&
-            uniprotXref &&
+          { proteinSummaryStats &&
```
imransl, line 243: A very minor issue but noticed there is this unnecessary space between the opening curly brace and the variable

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +274,108 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+
+  const [isXrefGroupOpen, setXrefGroupOpen] = useState(false);
+
+  const toggleXrefGroup = () => {
+    setXrefGroupOpen(!isXrefGroupOpen);
+  };
+
+  const chevronClass = classNames(
+    styles.chevron,
+    isXrefGroupOpen? styles.chevronDown : styles.chevronUp
```
imransl, line 307: A bit similar to the whitespace comment above. This time there is no space between `isXrefGroupOpen` and `?`. Isn't Prettier working in your VS Code?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
jyothishnt: thanks for pointing that. It was just an artefact :-)

azangru (approved):
I haven't checked it locally another time to see a list of multiple xrefs, but the code looks good.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +275,91 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+  const toggleXrefGroup = () => {
+    setXrefGroupOpen(!isXrefGroupOpen);
+  };
+
+  const chevronClasses = classNames(
+    styles.chevron,
+    {
+      [styles.chevronDown]: isXrefGroupOpen
+    }
+    // isXrefGroupOpen? styles.chevronDown : styles.chevronUp
```
msmanoj, line 311: Can this be removed?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -42,12 +44,14 @@ import { Product } from 'src/shared/types/thoas/product';
...
 import { ExternalReference as ExternalReferenceType } from 'src/shared/types/thoas/externalReference';
 
-import { SWISSPROT_SOURCE } from '../protein-list-constants';
+import { SWISSPROT_SOURCE } from 'src/content/app/entity-viewer/gene-view/components/proteins-list/protein-list-constants';
 
 import styles from './ProteinsListItemInfo.scss';
 import settings from 'src/content/app/entity-viewer/gene-view/styles/_constants.scss';
 
-type ProductWithoutDomains = Pick<
+import { ReactComponent as ChevronDown } from 'static/img/shared/chevron-down.svg';
```
msmanoj, line 52: Can this import be moved after Line 28?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -111,19 +115,21 @@ const ProteinsListItemInfo = (props: Props) => {
...
   const proteinId =
     transcript.product_generating_contexts[0].product.unversioned_stable_id;
 
   const { product: productWithProteinDomains } =
     transcriptWithProteinDomains?.product_generating_contexts[0] || {};
 
-  const uniprotXref = transcript.product_generating_contexts[0].product.external_references.find(
-    (xref) => xref.source.id === SWISSPROT_SOURCE
+  const proteinXrefs = getProteinXrefs(
+    transcript.product_generating_contexts[0].product.external_references
```
msmanoj, line 130: It might be better if we just pass `transcript` to the function `getProteinXrefs` and let it handle the rest.

jyothishnt: thanks @msmanoj so I export Transcript type from this component and use as the function props?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.scss
```diff
@@ -60,3 +55,38 @@ $content-width: $gene_image_width;
...
+  min-width: 50px;
+}
+
+.chevron {
+  margin-left: 6px;
+  margin-bottom: -1px;
+  height: 12px;
+  width: 12px;
+  transition: transform linear 0.2s;
+  &Down {
```
azangru, line 77: Don't you mean "up"?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
jyothishnt: changes made :-)

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -262,9 +272,87 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
...
+        <div className={styles.xrefWithChevron}>
+          <ProteinExternalReference
+            key={displayXref.accession_id}
+            source={source}
+            accessionId={displayXref.accession_id}
+            name={displayXref.name}
+          />
+          <div className={styles.xrefGroupChevron} onClick={toggleXrefGroup}>
+            <span
+              className={isXrefGroupOpen ? styles.xrefCountChevronOpen : ''}
```
azangru, line 320: Meh. We don't really need a class attribute without any value. If you default to `undefined` instead of the `''`, the span shouldn't have an empty `class` attribute.

src/ensembl/src/content/app/entity-viewer/shared/helpers/entity-helpers.ts
```diff
@@ -152,12 +158,32 @@ export const getLongestProteinLength = (gene: GetLongestProteinLengthParam) => {
...
+  let proteinXrefs = xrefs.filter(
+    (xref) => xref.source.id === SWISSPROT_SOURCE
+  );
+
+  if (!proteinXrefs.length) {
+    proteinXrefs = xrefs.filter((xref) => xref.source.id === SPTREMBL_SOURCE);
+  }
+
+  return proteinXrefs;
+};
```
azangru, line 189: I've played with the types some more. Here's the type I suggest to use for this function:

```ts
export const getProteinXrefs = <T extends Pick2<ExternalReference, 'source', 'id'>>(transcript: {
  product_generating_contexts: Array<{
    product: {
      external_references: T[];
    }
  }>;
}) => {
  const xrefs =
    transcript.product_generating_contexts[0].product.external_references;
  let proteinXrefs = xrefs.filter(
    (xref) => xref.source.id === SWISSPROT_SOURCE
  );

  if (!proteinXrefs.length) {
    proteinXrefs = xrefs.filter((xref) => xref.source.id === SPTREMBL_SOURCE);
  }

  return proteinXrefs;
};
```

**Reasoning:**
- You will notice that this function only cares about the `xref.source.id` field, whereas the `ProductWithoutDomains` has `external_references` that contain other fields that are irrelevant for this function.
- Using the generic type `T extends Pick2<ExternalReference, 'source', 'id'>` allows us to pass to this function external references that have any set of fields as long as one of those fields is `source` with an `id` — and to make typescript understand that whatever external_references shape was passed to this function will be exactly the same shape as the function is going to return.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+  name: 'UniProtKB/TrEMBL',
+  id: 'trembl1'
+};
+const tremblXrefs = times(4, () =>
+  createExternalReference({ source: tremblSource })
+);
+
+times(4, (index) => (tremblXrefs[index].accession_id = `trembl_${index}`));
+
+describe('<ProteinsListItemInfo /', () => {
```
azangru, line 49: The tests in this `describe` block do not refer to the `ProteinsListItemInfo` as a whole, but only to the `ProteinExternalReferenceGroup`. You can either change the describe message here to:

```
describe('<ProteinExternalReferenceGroup />', () => {
  ...
})
```

or add a nested `describe` block and put your tests inside this nested block:

```

describe('<ProteinsListItemInfo /', () => {

  describe('<ProteinExternalReferenceGroup />', () => {
    ...
  })

}
```

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    // ProteinExternalReferenceGroup expect xrefs from a single source
+    const { queryByText } = render(
+      <ProteinExternalReferenceGroup {...props} />
+    );
+    expect(queryByText(swissprotXref.source.name)).toBeTruthy();
+    expect(queryByText(swissprotXref.accession_id)).toBeTruthy();
+  });
+
+  describe('multiple xrefs', () => {
+    it('renders 2 xrefs', () => {
```
azangru, line 64: i suppose we'd better change this test to check for 3 xrefs to prove to ourselves that all three get rendered, and that there is no chevron.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <ProteinExternalReferenceGroup {...props} />
+      );
+      const chevron = container.querySelector('.chevron') as HTMLElement;
+      userEvent.click(chevron);
+      expect(getByText(props.xrefs[1].accession_id)).toBeTruthy();
+      expect(getAllByText(props.source)).toHaveLength(4);
+      expect(container.querySelectorAll('.externalLinkContainer')).toHaveLength(
+        4
+      );
+    });
```
azangru, line 106: It might be worth combining the above two tests into one (they are effectively two parts of the same test case, and their setup part is identical), in which the first half of the test will check the expectations before you click on the chevron, and the second half will check the expectations after you click on the chevron.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -282,15 +285,15 @@ const ProteinExternalReference = (props: ProteinExternalReferenceProps) => {
   );
 };
 
-type ProteinExternalReferenceGroupProps = {
+export type ProteinExternalReferenceGroupProps = {
```
azangru, line 288: no need to export it, I guess?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -316,7 +319,11 @@ const ProteinExternalReferenceGroup = (
             accessionId={displayXref.accession_id}
             name={displayXref.name}
           />
-          <div className={styles.xrefGroupChevron} onClick={toggleXrefGroup}>
+          <div
+            data-test-id="chevron"
```
azangru, line 323: we aren't using this test id anymore

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -150,14 +153,20 @@ const ProteinsListItemInfo = (props: Props) => {
...
+      !proteinStatsXref
+    ) {
+      // if proteinStatsXref is absent, we cannot fetch relevant data from PDBe; so pretend that we've successfully completed the request
       setSummaryStatsLoadingState(LoadingState.SUCCESS);
       return;
     }
 
-    if (summaryStatsLoadingState === LoadingState.LOADING && uniprotXref) {
-      fetchProteinSummaryStats(uniprotXref.accession_id, abortController.signal)
+    if (summaryStatsLoadingState === LoadingState.LOADING && proteinStatsXref) {
```
azangru, line 165: Missed it before. Why do you call it `proteinStatsXref`? Isn't it simply a protein xref?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -200,20 +209,27 @@ const ProteinsListItemInfo = (props: Props) => {
...
-                  name={uniprotXref.name}
-                />
-                <ProteinExternalReference
-                  source={ExternalSource.UNIPROT}
-                  accessionId={uniprotXref.accession_id}
-                  name={uniprotXref.name}
-                />
-              </div>
-            )}
+            {proteinXrefs?.length > 0 &&
```
azangru, line 212: `proteinXrefs` are returned by `getProteinXrefs`, right? Which always returns an array, right? If so, there's no need for the question mark before the `.length`

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -224,19 +240,17 @@ const ProteinsListItemInfo = (props: Props) => {
             )}
           </div>
           {proteinSummaryStats &&
-            uniprotXref &&
+            proteinXrefs &&
```
azangru, line 243: if `proteinXrefs` is returned by `getProteinXrefs`, it is always truthy. An empty array is still truthy.

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
jyothishnt: better to call it displaXref as this is the xref that is chosen to display

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -153,20 +153,14 @@ const ProteinsListItemInfo = (props: Props) => {
 
   useEffect(() => {
     const abortController = new AbortController();
-    if (
-      summaryStatsLoadingState === LoadingState.LOADING &&
-      !proteinStatsXref
-    ) {
-      // if proteinStatsXref is absent, we cannot fetch relevant data from PDBe; so pretend that we've successfully completed the request
+    if (summaryStatsLoadingState === LoadingState.LOADING && !displayXref) {
+      // if displayXref is absent, we cannot fetch relevant data from PDBe; so pretend that we've successfully completed the request
```
azangru, line 157: Something is wrong in this comment; and maybe also in the logic of this if-statement

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
```diff
@@ -126,7 +126,7 @@ const ProteinsListItemInfo = (props: Props) => {
     transcriptWithProteinDomains?.product_generating_contexts[0] || {};
 
   const proteinXrefs = getProteinXrefs(transcript);
-  const proteinStatsXref = proteinXrefs && proteinXrefs[0];
+  const displayXref = proteinXrefs && proteinXrefs[0];
```
azangru, line 129: aren't proteinXrefs always at least an empty array?

src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item-info/ProteinsListItemInfo.tsx
jyothishnt: discussed over slack and zoom during sprint review. Decision is to create a ticket to request for trembl ids and add "Similar proteins" to the protein stats.

azangru (approved):

PR #470, opened 2021-03-03
Species do not get removed from Entity Viewer

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-916

## Description
- Clearing the local storage entries for a particular genome within the Entity Viewer when the species is removed.

## Deployment URL
http://ev-removed-species.review.ensembl.org

## Views affected
Entity Viewer

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
+++ b/src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
@@ -15,10 +15,11 @@
  */
 
 import { createAction } from 'typesafe-actions';
-import { ActionCreator, Action } from 'redux';
+import { Action } from 'redux';
 import { batch } from 'react-redux';
 import { push, replace } from 'connected-react-router';
 import { ThunkAction } from 'redux-thunk';
+import pickBy from 'lodash/pickBy';
 
 import * as urlHelper from 'src/shared/helpers/urlHelper';
 import {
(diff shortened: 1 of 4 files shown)


src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -160,3 +156,26 @@ export const updateEntityId: ActionCreator<ThunkAction<
...
+    dispatch(deleteGenome(genomeIdToRemove));
+
+    const state = getState();
+
+    const updatedActiveEntityIds = pickBy(
+      getEntityViewerActiveEntityIds(state),
+      (value, key) => key !== genomeIdToRemove
+    );
+
+    dispatch(updateActiveEntityForGenome(updatedActiveEntityIds));
```
azangru, line 177: I know we've had similar conversations before, but I still think that the logic that's happening here on lines 170-177 is better suited for reducers. The reducers that are interested in `entityViewerGeneralActions.deleteGenome` action will react to it by removing what's related to the deleted genome id from the state slices that they are responsible for.

msmanoj: That's done. These lines were not required as the reducer already does a similar task.

azangru (approved):
Looks good.

PR #471, opened 2021-03-04
Remove AssemblySelector from Species Selector

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-990

## Description
This started as a bugfix (see reproduction steps in the Jira ticket), but after consultation with Andrea about the proper behaviour of the assembly dropdown it was decided to remove the dropdown altogether. Andrea says that it made sense in the context of switching between grch37 and grch38, but she doesn't feel this component serves the purpose anymore. She feels that if user should be allowed to pick another (past) assembly for a genome, this should be done in a different UI context.

**TO CONSIDER:** whether to remove the fetching of alternative assemblies in this PR.

## Deployment URL
http://no-assembly-selector.review.ensembl.org

## Views affected
Species selector


src/ensembl/src/content/app/species-selector/containers/species-search-panel/SpeciesSearchPanel.tsx
```diff
@@ -29,11 +36,32 @@ const SearchPanel = () => {
...
-      <div className={styles.assemblyWrapper}>
-        <AssemblySelector />
-      </div>
+      <SelectedAssembly />
     </section>
   );
 };
 
+const SelectedAssembly = () => {
+  const genomeId = useSelector(getCurrentSpeciesGenomeId);
```
imransl, line 45: How about renaming this to `activeGenomeId` instead?

msmanoj:
If the code that is responsible for fetching the alternative assemblies is not used anywhere, I guess it would be better to remove in this PR itself.

src/ensembl/src/content/app/species-selector/containers/species-search-panel/SpeciesSearchPanel.tsx
azangru: I made more extensive changes to remove the code related to the assembly selector from the codebase, and now this line looks different.

azangru:
@msmanoj Fair. I cleared the redux code from everything related to fetching and selecting an alternative assembly.

I realise that there still remains in the codebase some ancient code related to species strains that have never materialised; but that code, I think, can be removed in a different PR.

src/ensembl/src/content/app/species-selector/state/speciesSelectorState.ts
```diff
@@ -60,8 +57,7 @@ const initialState: SpeciesSelectorState = {
     search: LoadingState.NOT_REQUESTED
   },
   ui: {
-    isSelectingStrain: false,
-    isSelectingAssembly: false
+    isSelectingStrain: false
```
msmanoj, line 60: I think this can be removed as well?

azangru: There is still some more unused code related to strains in our codebase. I didn't want to dilute the purpose of this PR by removing it as well. I suggest we do it in a separate PR?

msmanoj: Sure, happy to leave that to a separate PR.

msmanoj (approved):

PR #473, opened 2021-03-05
Transcriptimage bugfix

ens-ma7:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-964

image

## Importance
N/A

## Description
This is a fix for a bug that showed up only in Chrome on Windows. The bug manifested in incorrect widths of transcripts that got drawn in the default transcripts list of the gene view in Entity Viewer (on the screenshot above, the top transcript, which should span the whole width of the diagram, is only a fraction of that width). The immediate cause of the bug was an incorrect `width` value that got set on the `svg` element. The true cause was that there was a confusion about variable names — we were passing the `length` variable to a function without defining it, which meant that javascript was using the global `length` attribute that's available on `window` element and that refers to the number of iframes in that window ([link to docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/length)) 🤦 . In our case the `length` variable was 0, and the width of the svg element got set to 1 rather than the expected 695. 

## Deployment URL
http://transcriptimage-bugfix.review.ensembl.org

## Views affected
Entity viewer transcript image

### Other effects
N/A

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
@@ -16,7 +16,7 @@
 
 import React from 'react';
 import classNames from 'classnames';
-import { scaleLinear, ScaleLinear, interpolateRound } from 'd3';
+import { scaleLinear, ScaleLinear } from 'd3';
 import { Pick3 } from 'ts-multipick';
 
 import { FullTranscript } from 'src/shared/types/thoas/transcript';
@@ -65,14 +65,10 @@ const UnsplicedTranscript = (props: UnsplicedTranscriptProps) => {
   } = slice;
   const scale = scaleLinear()
     .domain([1, transcriptLength])
-    .range([1, props.width])
-    .interpolate(interpolateRound)
+    .rangeRound([1, props.width])
     .clamp(true);
 
-  const transcriptClasses = classNames(
-    styles.transcript,
-    props.classNames?.transcript
-  );
+  const transcriptClasses = props.classNames?.transcript;
 
   const renderedTranscript = (
     <g className={transcriptClasses}>
@@ -120,10 +116,7 @@ const Backbone = (
     },
     scale
   } = props;
-  const backboneClasses = classNames(
-    styles.backbone,
-    props.classNames?.backbone
-  );
+  const backboneClasses = props.classNames?.backbone || undefined;
 
   if (!spliced_exons.length) {
     return (
@@ -201,12 +194,12 @@ const ExonBlock = (props: ExonBlockProps) => {
   const y = -3;
   const height = BLOCK_HEIGHT;
 
-  const exonClasses = classNames(styles.exon, props.className);
+  const exonClasses = props.className;
 
   if (cds && isNonCodingLeft) {
     const nonCodingPart = (
       <rect
-        className={classNames(exonClasses, styles.emptyBlock)}
+        className={classNames(exonClasses, styles.emptyBlock) || undefined}
         y={y}
         height={height}
         x={props.scale(exonStart)}
@@ -223,7 +216,7 @@ const ExonBlock = (props: ExonBlockProps) => {
       />
     );
     return (
-      <g key={exonStart}>
+      <g key={exonStart} data-test-id="exon">
         {nonCodingPart}
         {codingPart}
       </g>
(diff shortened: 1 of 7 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
```diff
@@ -199,12 +194,12 @@ const ExonBlock = (props: ExonBlockProps) => {
   const y = -3;
   const height = BLOCK_HEIGHT;
 
-  const exonClasses = classNames(styles.exon, props.className);
+  const exonClasses = props.className;
```
ens-ma7, line 197: sorry @azangru is this missing something here, should it not be:
props.classNames?.exon || undefined;
if i understand this right, the exon class is coming from the css but in your change you are omitting it

azangru: Line 197 is within a react component called ExonBlock. This component is receiving an optional className property that is just for the exon from UnsplicedTranscript [here](https://github.com/Ensembl/ensembl-client/blob/eb0423055e325364d257966f141dcb0bf0589a46/src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx#L77-L82)

src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
```diff
@@ -199,12 +194,12 @@ const ExonBlock = (props: ExonBlockProps) => {
...
 
   if (cds && isNonCodingLeft) {
     const nonCodingPart = (
       <rect
-        className={classNames(exonClasses, styles.emptyBlock)}
+        className={classNames(exonClasses, styles.emptyBlock) || undefined}
```
imransl, line 202: Think `classNames` will return `undefined` if the classes passed to it are not available. So don't think we need `||` to assign `undefined`.

src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
```diff
@@ -238,7 +233,7 @@ const ExonBlock = (props: ExonBlockProps) => {
     );
     const nonCodingPart = (
       <rect
-        className={classNames(exonClasses, styles.emptyBlock)}
+        className={classNames(exonClasses, styles.emptyBlock) || undefined}
```
imransl, line 236: Same as above unless there is a really good reason to do this

src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
azangru: Try it 😄 That's not what the [type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/classnames/types.d.ts#L11) says

imransl: @azangru Thanks, I checked it out and was surprised it sends back an empty string rather than `undefined`. Do we still need the `undefined` in there? Just curious about the usage of it here as we haven't used it in many other places.

azangru: If we pass an empty string to the `className` attribute of a jsx element, it will render an empty class attribute in the DOM:

image

If we pass `undefined` to the `className` attribute, the rendered DOM element will note have a class attribute at all:

image

I think the second option is much more preferable.

azangru: > Just curious about the usage of it here as we haven't used it in many other places.

I think we are fairly sure in most other places that the value that we are passing to `className` will not be an empty string. This component turned out to be a rare exception.

ens-ma7:
@imransl are you happy with the changes? The issue with className undefined as far as i understood is browser specific, firefox does show class="" like this whereas chrome just show class which i think both is wrong. Hence the fix by specifying the className to be undefined, it doesnt show the class attribute at all

src/ensembl/src/content/app/entity-viewer/gene-view/components/unspliced-transcript/UnsplicedTranscript.tsx
imransl: @azangru Thanks for the explanation. It makes a lot more sense now.

imransl:
@ens-ma7 Yes the changes look good. Was interesting know the usage of `undefined` with `classnames` thanks to this PR.

imransl (approved):

PR #476, opened 2021-03-09
Add QA checkbox group to PR template

msmanoj:
## Description
Added the following checkbox group to the PR template:
## QA
- [ ] Requested
- [ ] Approved
- [ ] Not needed

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -11,6 +11,11 @@
 - [ ] New feature
 - [ ] Improvement to existing feature
 
+## QA
+- [ ] Requested
+- [ ] Approved
+- [ ] Not needed
+
 ## Related JIRA Issue(s)
 <!--_Please provide the URL(s) for any JIRA issues related to this PR._-->
 


azangru:
If the QA happens (sometimes) _after_ the code review, and the person who merges the PR is the same person who submits it, then what benefit do these additional checkboxes offer to our communication? What extra information do they provide to the reviewers?

_(I was secretly hoping we could reduce the number of checkboxes in our templates, not increase them 😄 )_

ens-ma7:
We wanted a way to capture whether a ticket requires QA and hence it was suggested to add it to the PR template. I agree though that the QA hasn't got anything to do with the PR and should be the ticket probably (we have the kanban board showing it needs QA). But i think what we also want is to stop merging to dev in case QA is required. so somehow person reviewing this ticket will also need to know that it needs a QA.

azangru:
> so somehow person reviewing this ticket will also need to know that it needs a QA.

I'd have thought that it should not be the responsibility of the reviewer to know that a PR needs the QA. This should be the responsibility of the person who is merging the code, which in our case is the same person who has submitted the PR.

msmanoj:
Closing this as we have decided to indicate if QA is required or not within the JIRA ticket iteself.

PR #484, opened 2021-03-24
React testing refactor

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1045

## Importance
N/A

## Description
This is part of refactoring the test script from enzyme to react testing library.

## Deployment URL
N/A

## Views affected
N/A (Only test scripts have been updated)

### Other effects
None

### DIFF
--- a/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
+++ b/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
@@ -38,18 +38,18 @@ describe('BadgedButton', () => {
     jest.resetAllMocks();
   });
   it('renders the passed in button', () => {
-    const container = renderButton().container;
+    const { container } = renderButton();
     expect(container.querySelectorAll('.button')).toHaveLength(1);
   });
 
   it('assigns the "badgeDefault" class to the badge by default', () => {
-    const container = renderButton().container;
+    const { container } = renderButton();
     expect(container.querySelectorAll('.badgeDefault')).toHaveLength(1);
   });
 
   it('extends the badge class', () => {
     const fakeClassName = faker.lorem.word();
-    const container = renderButton({ className: fakeClassName }).container;
+    const { container } = renderButton({ className: fakeClassName });
     expect(
       container
         .querySelector('.badgeDefault')
@@ -58,14 +58,14 @@ describe('BadgedButton', () => {
   });
 
   it('trims the longer strings to three characters', () => {
-    const container = renderButton({ badgeContent: 'abcd' }).container;
+    const { container } = renderButton({ badgeContent: 'abcd' });
     expect(container.querySelector('.badgeDefault')?.textContent).toEqual(
       'abc'
     );
   });
 
   it('formats number greater than 99 to "99+"', () => {
-    const container = renderButton({ badgeContent: 100 }).container;
+    const { container } = renderButton({ badgeContent: 100 });
     expect(container.querySelector('.badgeDefault')?.textContent).toEqual(
       '99+'
     );
--- a/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
+++ b/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
@@ -15,7 +15,7 @@
  */
 
 import React from 'react';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 import faker from 'faker';
 import BadgedButton from './BadgedButton';
 import Button from '../button/Button';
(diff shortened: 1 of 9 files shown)


src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
```diff
@@ -26,66 +26,42 @@ const defaultProps = {
   badgeContent: faker.lorem.words()
 };
 
-const renderButton = (
-  ButtonComponent: React.FunctionComponent<any>,
-  props: any = defaultProps
-) => mount(<ButtonComponent {...props} />);
+const renderButton = (props: any = {}) =>
```
azangru, line 29: We should try to avoid `any`s when at all possible. Here, the best way to specify the type of the props is to export the `Props` type from the `BadgedButton` component [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/badged-button/BadgedButton.tsx#L21-L25), then import it in this file (`import { Props as BadgedButtonProps } from './BadgedButton`), and declare the props in this function to be `Partial<BadgedButtonProps>`.

src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
```diff
@@ -26,66 +26,42 @@ const defaultProps = {
...
   beforeEach(() => {
-    wrapper = renderButton(BadgedButton, {
-      ...defaultProps,
-      children: <Button onClick={onClick}>{faker.lorem.words()}</Button>
-    });
-  });
-
-  afterEach(() => {
     jest.resetAllMocks();
+    container = renderButton().container;
```
azangru, line 40: I would suggest running this line inside each `it`-block rather than in the `beforeEach`, for better flexibility, in the following form:

`const { container } = renderButton()`;

The reason for this is that the `render` function used in the `renderButton` returns an object with a number of useful fields besides `container`. For example, this object has a `rerender` method for re-rendering a component, and a `debug` method for pretty-printing the html representation of the component. You are losing access to those by plucking just the `container` from `renderButton`'s return value.

src/ensembl/src/shared/components/external-reference/ExternalReference.test.tsx
```diff
@@ -15,55 +15,67 @@
...
+    container = renderExternalReference({ label: undefined }).container;
+    expect(container.querySelectorAll('.label')).toHaveLength(0);
   });
 
   it('applies the passed in classNames', () => {
     expect(
-      wrapper.render().hasClass(defaultProps.classNames?.container)
+      screen
+        .getByTestId('external reference container')
+        .classList.contains(defaultProps.classNames?.container as string)
```
azangru, line 60: Is the `as string` required? It isn't specified in the next `expect` (`classList.contains(defaultProps.classNames?.label) `).

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -30,88 +30,115 @@ const defaultProps = {
...
+  }),
   readAsArrayBuffer: jest.fn(),
   readAsBinaryString: jest.fn(),
   readAsDataURL: jest.fn(),
   result: fileContents
 };
 
 describe('Upload', () => {
-  let wrapper: any;
+  let container: any;
```
azangru, line 55: Similar to the comment for the BadgedButton — better run the `const { container } = renderUpload(defaultProps);` in each test. Also, it looks like the `defaultProps` can be moved inside the `renderUpload` function in the same manner it's done in other tests, and the type of the props passed to the `renderUpload` function can be the `Partial` of Upload component's props rather than `any`.

src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
ens-ma7: if thats the case (we need the `rerender`  or `debug`) then we can just call the renderButton() like i did here https://github.com/Ensembl/ensembl-client/blob/762dbeaf9e218b30a6a23f21320f0d0cc01439d4/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx#L52 . does that not work?  I quite like the fact i didnt had to rewrite the renderButton() everytime and was finding it confusing that it was the same code being repeated and if there was a way to have a const that will make the code clearer. And the part where you have rerender it is clear you are passing some props to rerender.
I can see the benefit of `debug` but i have used container.innerHTML a lot for debugging. 
Let me know what you think? I just thought the code looks more understandable :)

azangru: > I quite like the fact i didnt had to rewrite the renderButton() everytime and was finding it confusing that it was the same code being repeated

Kent C. Dodds, the author of react-testing-library and a man of many opinions 😄, has written an [article](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) advising against putting the variables referencing the rendered components globally. I think I agree with him, at least in the part about the actual rendering of a component. In our tests, it's just a single line anyway. Easy to repeat, easy to modify, easy to read within the context of a single test.

ens-ma7: I am not going to upset Kent C. Dodds :)

src/ensembl/src/shared/components/external-reference/ExternalReference.test.tsx
ens-ma7: If i removed it, typescript complains with the following error:
`(property) container?: string | undefined
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.ts(2345)`
  
  i am also confused as to why some of the line it complains others it doesn't. It doesnt complain in the next `expect`

ens-ma7:
@azangru thanks for the review, I have implemented the suggested changes, please have a look and let me know. Thanks

azangru (approved):

src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
```diff
@@ -26,66 +26,48 @@ const defaultProps = {
...
-    });
-  });
-
-  afterEach(() => {
     jest.resetAllMocks();
   });
-
   it('renders the passed in button', () => {
-    expect(wrapper.find('button').length).toEqual(1);
+    const container = renderButton().container;
```
msmanoj, line 41: Just for reference, this could just be written as `const {container} = renderButton()`

src/ensembl/src/shared/components/image-button/ImageButton.test.tsx
```diff
@@ -110,73 +110,60 @@ describe('<ImageButton />', () => {
...
-      });
-      wrapper.update();
+      fireEvent.mouseEnter(container.querySelectorAll('.imageButton')[0]);
 
-      const tooltip = wrapper.find(Tooltip);
-      expect(tooltip.length).toBe(1);
-      expect(tooltip.text()).toBe(description);
+      const tooltip = container.querySelectorAll('.tooltip');
+      expect(tooltip).toHaveLength(1);
+      expect(tooltip[0].innerHTML).toBe(description);
```
msmanoj, line 148: I think `textContent` is safer than using `innerHTML` here.

src/ensembl/src/shared/components/upload/Upload.test.tsx
```diff
@@ -30,88 +30,116 @@ const defaultProps = {
...
+      const container = renderUpload().container;
+      expect(
+        container.querySelector('input')?.getAttribute('multiple')
+      ).toEqual('');
     });
 
     it('disables multiple file selection if allowMultiple is set to false', () => {
-      wrapper = renderUpload({ ...defaultProps, allowMultiple: false });
-      expect(wrapper.find('input').prop('multiple')).toBeFalsy();
+      const container = renderUpload({ ...defaultProps, allowMultiple: false })
```
msmanoj, line 75: `...defaultProps,` is not required here

src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
ens-ma7: I have made the changes and push it.

ens-ma7:
Thanks for the review @msmanoj. I have pushed the suggested the changes, can you please have a look and let me know if there any other issue. thanks

msmanoj (approved):

PR #486, opened 2021-03-29
Migrate remaining tests to react-testing-library, and remove Enzyme

azangru:
### Migrated tests for:
- Zmenu
- ZmenuContent
- BasePairsRuler
- DefaultTranscriptListItem
- TranscriptsListItemInfo
- ProteinDomainImage
- TranscriptsFilter
- UnsplicedTranscript
- PopularSpeciesButton
- SpeciesCommitButton
- SpeciesSearchField
- SpeciesSearchMatch
- SpeciesSelectorAppBar
- SpeciesRemove
- AutosuggestSearchField
- ErrorBoundary
- Button
- Checkbox
- FeatureSummaryStrip
- InlineSvg
- InstantDownloadTranscriptVisualisation
- PointerBox
- RoundButton
- SearchField
- Select
- SelectedSpecies
- Tabs
- Textarea
- ToolboxExpandableContent
- VisibilityIcon
- useApiService
- useOutsideClick
- useRefWithRerender

### Removed components
- LaunchbarContainer (serves no purpose)
- Account (not used)
- SpeciesTab (superseded by SelectedSpecies component; not used anywhere)

### Removed tests for
- Launchbar (these tests would have gotten outdated soon anyway; as far as I understand, Andrea won't want to disable launchbar buttons if no species have been selected)

### Also done as part of the PR
- Removed the `launchbarExpanded` and `accountExpanded` fields from redux and removed them from the places they were used in code
- Typescript marks the `keyCode` field on KeyboardEvent as deprecated. I have changed it to `event.key`. Browser support seems to be good these days.

### Finally
Removed dependency on Enzyme — 🎉 🎊 🍾 

### Remember to do after update of dependencies
When we update React to v17 (right after this PR is accepted and enzyme removed), investigate whether the `.persist()` function on events should be deleted. According to the [docs](https://reactjs.org/docs/legacy-event-pooling.html), it will no longer do anything.


src/ensembl/src/shared/components/tabs/Tabs.test.tsx
```diff
@@ -58,82 +59,97 @@ const defaultProps = {
   classNames: tabClassNames
 };
 
-const renderTabs = (props: TabsProps) => <Tabs {...props}></Tabs>;
+const renderTabs = (props: TabsProps) => render(<Tabs {...props} />);
```
msmanoj, line 62: I guess we don't need to `props` param here as we are only using `defaultProps` in all the test cases.

src/ensembl/src/shared/components/error-boundary/ErrorBoundary.test.tsx
```diff
@@ -15,62 +15,82 @@
...
+};
 
 const Fallback = ({ error }: { error: Error }) => (
-  <span>I am the fallback component that received error {error}</span>
+  <span className="fallback">
+    I am the fallback component that received error {error.message}
+  </span>
 );
 
 jest.spyOn(errorService, 'report');
```
msmanoj, line 38: Can we use `jest.spyOn(errorService, 'report').mockImplementation(noop);` here and remove the `beforeEach` block on L47?

src/ensembl/src/shared/components/autosuggest-search-field/AutosuggestSearchField.test.tsx
```diff
@@ -200,43 +206,46 @@ describe('<AutosuggestSearchField />', () => {
...
-            allowRawInputSubmission={true}
-            onSubmit={onSubmit}
-          />
-        );
-      });
+      const props = {
+        search,
+        onChange,
+        onSelect,
+        matchGroups: groupsOfMatches,
```
msmanoj, line 213: Would it be worth defining these as `commonProps`?

src/ensembl/src/shared/components/error-boundary/ErrorBoundary.test.tsx
azangru: I don't want errorService to execute its real report method during any of the tests (otherwise it will spam the console with the error message). What I can do is I can replace the `jest.spyOn(errorService, 'report').mockImplementation(noop);` with `errorService.report = jest.fn()` and put it in the `beforeAll` function, so that it runs once per the whole describe block.

src/ensembl/src/shared/components/tabs/Tabs.test.tsx
azangru: ✅

src/ensembl/src/shared/components/autosuggest-search-field/AutosuggestSearchField.test.tsx
azangru: ✅

src/ensembl/src/content/app/custom-download/containers/content/CustomDownloadContent.tsx
```diff
@@ -88,7 +84,6 @@ const mapDispatchToProps: DispatchProps = {
 
 const mapStateToProps = (state: RootState): StateProps => ({
   showSummary: getShowPreviewResult(state),
-  launchBarExpanded: getLaunchbarExpanded(state),
   showExampleData: getShowExampleData(state)
 });
```
imransl, line 89: Can we switch to using hooks for this component?

azangru: In due course, when we resume work on custom download. I imagine it won't be for a while.

msmanoj (approved):

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
```diff
@@ -31,20 +30,18 @@ import {
 } from 'tests/fixtures/entity-viewer/gene';
 
 jest.mock('../transcripts-list-item-info/TranscriptsListItemInfo', () => () => (
-  <div>TranscriptsListItemInfo</div>
+  <div className="transcriptsListItemInfo">TranscriptsListItemInfo</div>
```
imransl, line 33: Here and elsewhere, why are we using `className` instead of `data-test-id`?

src/ensembl/src/shared/components/select/Select.tsx
```diff
@@ -115,18 +114,13 @@ const Select = (props: SelectProps) => {
...
-    ) {
+    if (!focusRef.current) {
       return;
     }
 
-    if (event.keyCode === keyCodes.ENTER) {
+    if (event.key === 'Enter') {
       openPanel();
-    } else if (event.keyCode === keyCodes.ESC) {
+    } else if (event.key === 'Escape') {
```
imransl, line 123: Do we need the `else` here?

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
azangru: I don't have a good explanation other than we used classes even before we started using data-test-ids. In our source code I try never to use classes for anything other than styling purposes, so I would be opposed to adding a class to an element in a component for the sole purpose of identifying this element in a test, and would use the data-test-id attribute instead. On the other hand, in tests, we already commonly use classes of various elements in a component to identify them in a test. I think we do it out of convenience and economy — we don't wan't to litter our source code with the data-test-id attributes. Now when it comes to which kind of identifier we should use in a mock, data-test-id would be cleaner, but ultimately it shouldn't really matter.

In a word: no reason, data-test-id probably slightly better semantically. I can change the ones I can find in this PR, but it's unlikely to improve anything 😃

src/ensembl/src/shared/components/select/Select.tsx
azangru: Don't think so; we'll just ignore those key presses.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
imransl: Thanks for the great explanation. I was curious about why using both of them instead of one, especially in mocked components like this one.

azangru: I've replaced searches by className with searches by data-test-id for all mocked components I could find in this PR. Just for purity's sake 😃

imransl (approved):

PR #487, opened 2021-03-30
Fix the import of the web worker for gene sequence fetching

azangru:
## Type
- Bug fix

## Description
Forgot to update the import of the sequence downloading web worker into the gene sequence fetcher. Using the worker loader of webpack 5 now.

## Deployment URL
http://fix-web-worker.review.ensembl.org/

## Views affected
Gene sequence download


imransl (approved):
Tested it and works

PR #488, opened 2021-04-04
Update dependencies

azangru:
## Description
After https://github.com/Ensembl/ensembl-client/pull/486 gets accepted, we should be able to update React to version 17 (the update was blocked by our prior dependency on `enzyme`).

This PR updates all our dependencies, including notably:
- React to 17.0.2
- react-spring to 9.0.0
- typescript to 4.2.3
- husky to 6.0.0

**Breaking changes:**
- `react-spring` made breaking changes in `useTransition` when it updated to version 9.0.0. They have not yet updated their docs site, despite having released version 9.0.0 to the public; so the docs for the changes currently only exist [in their github repo](https://github.com/pmndrs/react-spring.io/blob/v9/content/v9/breaking-changes.mdx#the-usetransition-hook).
- `husky` got updated to v.6.0.0 and changed the way the configs are set up (for reference, see the [docs describing v4 to v6 migration](https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v6))
- `eslint-config-prettier` required an update of the config (see the [docs](https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21) for reference)

**Other notable changes:**
- `faker` received an update, as a result of which they moved some functions around. This doesn't seem to be documented anywhere other than in the [deprecation warnings](https://github.com/Marak/faker.js/blob/149289afa955d81bed410104d5f034eb1495fe9d/lib/random.js#L29-L35) in their source code.
- Hot module replacement seems to be broken in dev. The reason, curiously, is that we are using a browserlist configuration. See discussion in this issue: `https://github.com/webpack/webpack-dev-server/issues/2758`. Updating to webpack-dev-server beta looks like a reasonably good option. Despite it being a beta, it looks and feels mature.

**Temporarily added dependency**
Storybook has a dependency on an old version of the `dotenv-webpack` package, which doesn't support webpack 5. See the issue described in this link: `https://github.com/storybookjs/storybook/issues/14403`. To prevent this from being a blocker, the temporary workaround, as suggested in the linked issue, is to install the new version of `dotenv-webpack` as our own dev dependency, which will win over the storybook's outdated one.

We do not really need this package other than to help with the storybook build. When the issue is solved on storybook's side, we will be able to remove it.

relevant as an example of how messed up Storybook currently is: `https://github.com/storybookjs/storybook/issues/14497`

## Deployment URL
http://update-packages.review.ensembl.org

### Checks
- unit tests — ✔️ 
- typescript checker — ✔️ 
- development build — ✔️ 
- production build — ✔️ 
- storybook build — ✔️


msmanoj:
Not sure if it is just for me but I get some errors in the console when I try to run storybook:

Screenshot 2021-04-15 at 06 56 18

msmanoj:
Not sure if it is related but I get the following `Violation` message logged in the console when I click on `Example gene` in the Species home page:
>[Violation] Added non-passive event listener to a scroll-blocking <some> event. Consider marking event handler as 'passive' to make the page more responsive. 

This is also preventing the `View In` toolbox from being shown.

msmanoj:
Something is not right with the cogs here:
Kapture 2021-04-15 at 07 13 23

msmanoj:
Not sure if this is related but navigating away from entity viewer keeps throwing the following error:
>useRestoreScrollPosition.ts?fbd8:55 Uncaught TypeError: Cannot read property 'scrollTop' of null
    at eval (useRestoreScrollPosition.ts?fbd8:55)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:3945)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:3994)
    at invokeGuardedCallback (react-dom.development.js?61bb:4056)
    at flushPassiveEffectsImpl (react-dom.development.js?61bb:23543)
    at unstable_runWithPriority (scheduler.development.js?3069:468)
    at runWithPriority$1 (react-dom.development.js?61bb:11276)
    at flushPassiveEffects (react-dom.development.js?61bb:23447)
    at flushDiscreteUpdates (react-dom.development.js?61bb:22356)
    at flushDiscreteUpdatesIfNeeded (react-dom.development.js?61bb:3768)


This doesn't happen on both internal and 2020.

imransl:
I noticed a different issue when running the production build locally. 

image

imransl:
> Something is not right with the cogs here:
> Kapture 2021-04-15 at 07 13 23

Strangely

imransl:
Sorry for closing the PR by accident. I pressed the wrong button (`Close with comment` instead of `Comment`)

azangru:
> Not sure if this is related but navigating away from entity viewer keeps throwing the following error:
> `useRestoreScrollPosition.ts?fbd8:55 Uncaught TypeError: Cannot read property 'scrollTop' of null`

@msmanoj this one is actually quite interesting 👍 

In `useRestoreScrollPosition`, we are using the cleanup function of the `useEffect` hook to simulate the `.componentWillUnmount` lifecycle method. In that cleanup function, we need access to the DOM element whose scroll position we measure and record in the store.

Apparently, in React 16, the cleanup function was executed synchronously before the component got unmounted from the DOM; but now they have changed it to run asynchronously; so first the component gets unmounted, and then the useEffect cleanup function is executed. Which is why we lose the reference to the DOM element.

The docs suggest we use the `useLayoutEffect` hook for such rare cases when something needs to be done either before browser paint on mount, or before the component is removed from the DOM on unmount.

azangru:
>  However, the example links in species homepage don't work properly anymore. I wonder if it is related to this event listener issue in React 17

Yes, you are right, it was related. The problem was with the `useOutsideClick` hook — it was listening to clicks that occurred outside a component, such as a popup under the Example links element. I am not quite sure why, but when, after a click on the "Example link" element, the popup gets rendered and initialises the `useOutsideClick` hook, it still has time to catch the click event that happened on the "Example link" element by the time that click bubbles up all the way to the `document`. It's weird that this is the case, but it just shows that we now need to be more mindful of stopping event propagation. I fixed this in my latest commit.

azangru:
> This is also preventing the View In toolbox from being shown.

@msmanoj I haven't seen any `Violation` messages in my console; but the reason the ViewInApp popup wasn't showing is different; I discussed it in more detail in my previous comment.

azangru:
> Something is not right with the cogs here:

@msmanoj  I updated the behaviour of the cogs. Again, the problem was in the race condition between the code path that says that you clicked on a cog and the code path that says that you clicked outside the open browser track config bar. Currently, in production, if you click on another cog, the browser track config bar is closed, but the bar for the other cog doesn't get opened, and you need to click the second time to open it. I think the behaviour below is more in line with what the user would expect.

https://user-images.githubusercontent.com/6834224/115121734-280c0600-9fac-11eb-94ad-9d177abc90ce.mov

azangru:
> Not sure if it is just for me but I get some errors in the console when I try to run storybook:

@msmanoj yeah; that's because we still have a story for the InstantDownloadTranscript component; and this component imports the `fetchTranscript` module, which in turn imports a worker for downloading sequences. Communication with web workers is annoying; and so we are using the comlink library that hides the communication ugliness behind a neat promise-based api. But internally it's still based on postMessages; and apparently it encounters some problems when trying to post a message from within an iframe.

This error is already present in our [current deployment](https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/design-primitives-colours--colours-story) of the storybook:

image

We might want to consider removing the InstantDownloadTranscript component from the storybook; it's probably no longer useful there. If we do, this should make the error in the console go away.

ens-ma7:
when i did an npm ci, i get the following warning/error. Not sure if this is a problem or it came up before:

```
Binary found at /mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/node-sass/vendor/linux-x64-64/binding.node
Testing binary
Binary is fine
  ⚠ Command failed: /mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/cwebp-bin/vendor/cwebp -version
/mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/cwebp-bin/vendor/cwebp: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory
```

imransl:
> when i did an npm ci, i get the following warning/error. Not sure if this is a problem or it came up before:
> 
> ```
> Binary found at /mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/node-sass/vendor/linux-x64-64/binding.node
> Testing binary
> Binary is fine
>   ⚠ Command failed: /mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/cwebp-bin/vendor/cwebp -version
> /mnt/c/Users/ridwan/ensembl-client-wsl/ensembl-client/src/ensembl/node_modules/cwebp-bin/vendor/cwebp: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory
> ```

I think you'll need to install libGL locally in Ubuntu on Windows for this to work according to this Github issue: https://github.com/imagemin/imagemin-webp/issues/22

ens-ma7:
> I noticed a different issue when in both dev and production builds locally.
> 
> image
> 
> The genome browser navigation with the mouse is broken and also the zmenus don't work as well. Probably due to the panic errors.

I also get the same error, when i start moving the genome browser right or left, i see the same errors as imran and genome browser is then broken (cant zoom).
ZMenu is also broken.

imransl:
@azangru Can confirm that I still have the same issue as @ens-ma7 when trying to navigate or click on a gene in both dev and production builds. The navigation buttons however work fine.

imransl:
@azangru Screenshot of the console errors that I get 

image

imransl:
@azangru I checked locally again and didn't experience any other issue apart from the genome browser one in chrome.

imransl (approved):

msmanoj:
Should we allow multiple toolbox to be open at the same time?

Kapture 2021-04-26 at 16 32 17

azangru:
@msmanoj I believe the bug you pointed out should be fixed now. By the way, investigating its causes has led me to update the `useOutsideClick` hook, by listening to the document click event during the capture phase rather than the bubbling phase. This seems to have simplified the code of the hook drastically without breaking it (tested with the Storybook stories).

msmanoj (approved):

PR #491, opened 2021-04-19
ENSWBSITES-1037: show none when gene name is novel transcript

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1037

## Importance
N/A

## Description
Show 'none' when gene name is 'novel transcript' in both Entity Viewer Sidebar panel and in Genome browser track drawer. This is a temporary fix until the data is loaded properly. We will have to revert this change.

As a bonus we are cleaning the gene name where they contains source details for example: 
`Heat shock protein 101[Source:UniProtKB/TrEMBL;Acc:Q9SPH4]` becomes `Heat shock protein 101`

## Deployment URL
http://hide-novel-transcript.review.ensembl.org/

example gene for human: 
gene:ENSG00000270008

## Views affected
Entity viewer sidebar panel
Genome browser track bar

### Other effects
N/A

## Possible complications
None

### DIFF
--- a/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
+++ b/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
@@ -65,8 +65,8 @@
   font-weight: $light;
 }
 
-.leftSpacing {
-  margin-left: 12px;
+.geneSymbol {
+  margin-right: 12px;
 }
 
 
--- a/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
+++ b/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
@@ -65,6 +65,10 @@
   font-weight: $light;
 }
 
+.leftSpacing {
+  margin-left: 12px;
+}
+
 
 .downloadLink{
   color: $blue;
--- a/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
+++ b/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
@@ -20,6 +20,7 @@
    it('returns the correct gene display name', () => {
      expect(getGeneName('novel transcript')).toBe('None');
      expect(getGeneName('')).toBe('None');
-     expect(getGeneName('Heat shock protein 101[Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101');
+     expect(getGeneName(null)).toBe('None');
+     expect(getGeneName('Heat shock protein 101 [Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101 ');
    });
  });
--- a/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
+++ b/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
@@ -1,0 +1,25 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ import { getGeneName } from './geneFormatter';
+  
+ describe('getGeneName', () => {
+   it('returns the correct gene display name', () => {
+     expect(getGeneName('novel transcript')).toBe('None');
+     expect(getGeneName('')).toBe('None');
+     expect(getGeneName('Heat shock protein 101[Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101');
+   });
+ });
--- a/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
+++ b/src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
@@ -21,6 +21,6 @@
      expect(getGeneName('novel transcript')).toBe('None');
      expect(getGeneName('')).toBe('None');
      expect(getGeneName(null)).toBe('None');
-     expect(getGeneName('Heat shock protein 101 [Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101 ');
+     expect(getGeneName('Heat shock protein 101 [Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101');
    });
  });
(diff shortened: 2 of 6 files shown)


jyothishnt:
PR looks ok to me. But for some reason travelling to gene function tab breaks everything (even on 2020)
http://2020.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000270008?view=transcripts

jyothishnt (commented):
PR looks ok, though it breaks on clicking gene function (even in 2020)
http://2020.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000270008?view=transcripts

imransl:
I wonder if the doing the opposite in the condition will be more readable. Would like to know what others think.

```gene.name && gene.name !== 'novel transcript' ? gene.name :  'None'```

**Comment by azangru** [comment]
We are showing the gene name in the GenomeBrowser as well:

image

We may want to add a formatter function for gene name (somewhere alongside [these files](https://github.com/Ensembl/ensembl-client/tree/dev/src/ensembl/src/shared/helpers/formatters)). In which we may even remove this `[Source:HGNC Symbol;Acc:HGNC:6872]` ugliness with a clever regex. Hopefully in the future we won't need it.

**Comment by ens-ma7** [comment]
> We are showing the gene name in the GenomeBrowser as well:
> 
> image
> 
> We may want to add a formatter function for gene name (somewhere alongside [these files](https://github.com/Ensembl/ensembl-client/tree/dev/src/ensembl/src/shared/helpers/formatters)). In which we may even remove this `[Source:HGNC Symbol;Acc:HGNC:6872]` ugliness with a handy regex (e.g. something like ` (\[Source:.*\])`). Hopefully in the future we won't need it.

I did update the code for genome browser as well. This is a temporary fix which will need to be reverted once the database is updated. So not sure if we need a formatter function.

**Comment by ens-ma7** [comment]
> I wonder if the doing the opposite in the condition will be more readable. Would like to know what others think.
> 
> `gene.name && gene.name !== 'novel transcript' ? gene.name : 'None'`

Lets see what others say, i dont mind changing it but i find the other way clearer, i guess its because of the Jira request. :)

**Comment by azangru** [comment]
> I did update the code for genome browser as well, its already in the PR or have i missed something else.

Yes, sorry; thought both changes were for the genome browser.

> This is a temporary fix which will need to be reverted once the database is updated. So not sure if we need a formatter function.

🤷‍♂️  Since we are interfering with the return value from the backend for now anyway (which, yes, we shouldn't be doing), it could also afford us an opportunity to remove the ugliness in the square brackets, which may make Andrea happy. But I don't have particularly strong feelings about it.

@ens-st3 do you have an opinion?

**Comment by ens-st3** [comment]
I think I dislike it more than Andrea does

**Comment by ens-ma7** [comment]
I am going to add a formatter then. I will give it a go and update the PR.

File: src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ import { getGeneName } from './geneFormatter';
+  
+ describe('getGeneName', () => {
+   it('returns the correct gene display name', () => {
+     expect(getGeneName('novel transcript')).toBe('None');
+     expect(getGeneName('')).toBe('None');
```
azangru, line 22: check for null as well.

src/ensembl/src/shared/helpers/formatters/geneFormatter.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function getGeneName(geneName: string | null) {
+  if(geneName && geneName != 'novel transcript') {
```
azangru, line 18: `!=` -> `!==`

I believe we have an eslint setting that enforces using strict equality comparison (triple equals) rather than equality comparison after typecasting (double equals). Strange that it didn't complain here.

src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
```diff
@@ -0,0 +1,25 @@
...
+     expect(getGeneName('Heat shock protein 101[Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101');
```
azangru, line 23: gene names that we are getting from the api will have a space before the opening square bracket (e.g. `mitogen-activated protein kinase 10 [Source:HGNC Symbol;Acc:HGNC:6872]`). Your test string should probably have the space before the square bracket as well.

ens-ma7:
Thanks @azangru , made the suggested changes. Please have a look and let me know if there is anything else.

src/ensembl/src/shared/helpers/formatters/geneFormatter.test.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+ import { getGeneName } from './geneFormatter';
+  
+ describe('getGeneName', () => {
+   it('returns the correct gene display name', () => {
+     expect(getGeneName('novel transcript')).toBe('None');
+     expect(getGeneName('')).toBe('None');
+     expect(getGeneName(null)).toBe('None');
+     expect(getGeneName('Heat shock protein 101 [Source:UniProtKB/TrEMBL;Acc:Q9SPH4]')).toBe('Heat shock protein 101 ');
```
azangru, line 24: `toBe('Heat shock protein 101 ');`

Well, we should trim that trailing space now, shouldn't we? 😉 

`return geneName.replace(/\[Source:.*\]/,'').trim();`

ens-ma7: I thought you will miss this :)

azangru: Never!

ens-ma7: Fixed.

azangru:
image

Forgot this one (genome browser, transcript summary)

ens-ma7:
> image
> 
> Forgot this one (genome browser, transcript summary)

@azangru, its done.

azangru (approved):
Looks good.

If you would like to fix this little bug in the same PR, then go ahead, if not we'll have a new ticket for it:

image

ens-ma7:
thanks @azangru , i have also fixed the bug with the gene symbol hoping it is a simple one. can you please have a look and let me know if there is anything wrong. thanks

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.scss
```diff
@@ -65,6 +65,10 @@
   font-weight: $light;
 }
 
+.leftSpacing {
```
azangru, line 68: Unless we are repeating this class in multiple places in this view (which I don't think we are), better give it a more concrete name (`.geneSymbol`), add it to the `span` surrounding the gene symbol, and use a right margin instead of the left. So that when a gene does not have a symbol, the whole gene symbol span, with this class, won't even appear in the DOM.

ens-ma7: Done @azangru . Let me know if its all good now. Thanks

azangru (approved):
Looks good.

imransl (approved):

PR #495, opened 2021-04-29
Fix the unexpected expansion of species links on the home page

azangru:
## Type
- Bug fix

## Description
Steps to reproduce the bug (currently in production):
- Select at least one species
- Go to the home page
- Click somewhere on the home page

Bug: a species row will get expanded:

This is a result of my sloppy code that became apparent after we updated React to v.17. I am adding a quick fix here, just to avoid unnecessary confusion for home page visitors; but keep in mind that the home page is going to fundamentally change very soon, before the MAP.

## Deployment URL
http://fix-species-links-on-homepage.review.ensembl.org

## Views affected
Home page


imransl (approved):
Works for me now

azangru:
Got a thumbs up from Steve as well. Although the home page will be replaced within a month anyway 🙂

PR #499, opened 2021-05-10
Update dependencies and clean up babel config

azangru:
## Type
- Housekeeping/infrastructure

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1104

## Description
1. Routine update of dependencies. Of note:
  - rxjs got updated to v.7. There must have been an update to the types of rxjs's ajax, which switched the success type to `AjaxResponse<unknown>`. The [following change](https://github.com/Ensembl/ensembl-client/pull/499/files#diff-7170abaa200f2edccf35bce574b78835dc1f7fc83ace4769f80f28d93368e6f6R21-R23) was necessary to fix our types.
2. Replace `node-sass` (which is deprecated) with `sass` (which is pure js and doesn't require C bindings)
3. Remove babel transforms
- dynamic imports ([support table](https://caniuse.com/?search=dynamic-import))
- @babel/plugin-proposal-object-rest-sprea (inluded in `@babel/preset-env` that we are already using)
- @babel/plugin-proposal-class-properties (inluded in `@babel/preset-env` that we are using)
- @babel/plugin-proposal-optional-chaining (included in `@babel/preset-env` that we are using)
- @babel/plugin-proposal-nullish-coalescing-operator (included in `@babel/preset-env` that we are using)

The removal of the individual plugins resulted, given our list of supported browsers, in about 40kB savings:

before:
Screenshot 2021-05-10 at 18 22 46

after:
Screenshot 2021-05-10 at 18 27 22

## Deployment URL


imransl:
@azangru There are a few files that have been formatted by Prettier. Is this because we haven't formatted them before or due to a recent update?

imransl:
Everything looks good. Didn't notice any issues.

imransl (approved):

azangru:
> There are a few files that have been formatted by Prettier. Is this because we haven't formatted them before or due to a recent update?

I don't know, to be honest. These are fairly old files; and I think Prettier has updated its defaults at some point; but I don't believe it happened in this PR —  I've been seeing reformatted lines in new commits for a long time.

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
```diff
@@ -187,12 +177,9 @@ export const ensureSpeciesIsEnabled: ActionCreator<ThunkAction<
   dispatch(toggleSpeciesUseAndSave(genomeId));
 };
 
-export const fetchPopularSpecies: ActionCreator<ThunkAction<
-  void,
-  any,
-  null,
-  Action<string>
->> = () => async (dispatch) => {
+export const fetchPopularSpecies: ActionCreator<
```
msmanoj, line 180: Here and in other places, is it worth getting rid of `ActionCreator` in this PR itself ?

azangru: Done for this actions file.

Curiously, the redux team [has updated](https://github.com/reduxjs/redux/blame/master/src/types/actions.ts#L52) type definitions for the ActionCreator, which might have made it safer.

msmanoj (approved):

src/ensembl/src/content/app/species-selector/state/speciesSelectorActions.ts
msmanoj: Cool, I guess we don't have to worry about removing it going forward?

azangru: It's great if it no longer does harm. I guess we don't need to worry much about it, but we can justify the further gradual removal of this type by the fact that, for us, it doesn't seem to add anything of value.

PR #500, opened 2021-05-19
Replace download with new loading button 

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1058

## Description
Replace the current button with the new loading button on Gene and Transcript downloads. It should also reset the checkboxes once the download completes. 

## Deployment URL
http://replace-download-button.review.ensembl.org/

## Views affected
EV and GB

### Other effects
N/A

## Possible complications
Nothing that I am aware


src/ensembl/src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
```diff
@@ -110,7 +118,15 @@ const InstantDownloadGene = (props: Props) => {
       }
     };
 
-    fetchForGene(payload);
+    const fetch = await fetchForGene(payload);
```
imransl, line 121: Can we change this to something else like `genePromise` or `geneRequest` as naming it `fetch`, the web API`, could be a bit misleading.

azangru: 👍 yes to not naming our variables with names reserved for the web api.

But I do not understand what's happening in this function in general. What's the purpose of the new Promise on the next line?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForGene.ts
```diff
@@ -51,36 +51,41 @@ export const fetchForGene = async (payload: FetchPayload) => {
...
 
-  downloadAsFile(sequences, `${geneId}.fasta`, {
-    type: 'text/x-fasta'
-  });
+    worker.terminate();
+
+    downloadAsFile(sequences, `${geneId}.fasta`, {
+      type: 'text/x-fasta'
+    });
+    return true;
```
azangru, line 87: I am confused. What's the purpose of the returned boolean?

src/ensembl/src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
jyothishnt: Thansk @imransl  I will change it. 
Loading button onclick prop is a method that returns a promise?

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchForGene.ts
jyothishnt: It is used to check if the fetch was successful or not.

src/ensembl/src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
azangru: An async function returns a promise by definition 😄 No need to artificially create one.

Let's discuss this PR when you are back.

src/ensembl/src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
```diff
@@ -135,11 +147,12 @@ const InstantDownloadGene = (props: Props) => {
         theme={props.theme}
         onChange={onTranscriptOptionChange}
       />
-      <InstantDownloadButton
-        className={isButtonDisabled ? styles.downloadButtonDisabled : undefined}
-        isDisabled={isButtonDisabled}
-        onClick={onSubmit}
-      />
+      <div className={styles.downloadButton}>
```
azangru, line 150: Any reason why this class isn't passed directly to `InstantDownloadButton` as a prop?

jyothishnt: If I pass it to the component, it will only apply to that button. In this case I want to align the whole component to the right side. Is there any other work around?

azangru: Right. I see the problem. My fault: when I created the `LoadingButton` component, I did not add a way to style the wrapper div around the button [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/loading-button/ControlledLoadingButton.tsx#L47). We can ignore this for now, I guess, but I wonder if it's going to be a problem for us in the future.

azangru (approved):

src/ensembl/src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
jyothishnt: create ticket to fix this? @azangru

msmanoj (approved):
Looks good apart from the conflicts

PR #505, opened 2021-05-26
Add ESLint plugin for Jest

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1141](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1141)

## Description
Add ESLint rules to avoid jest tests using `.only` from getting committed. An example of it would be:
```
test.only('outputs string when input received as ChrLocation', () => {
      expect(getChrLocationStr(chrLocationValues.tupleValue)).toBe(
        chrLocationValues.stringValue
      );
});
```

I've also added rules to detect duplicate titles and erroneous usage of `expect`.

### DIFF
--- a/src/ensembl/.eslintrc.js
+++ b/src/ensembl/.eslintrc.js
@@ -3,9 +3,10 @@ module.exports = {
   extends: [
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
-    'plugin:prettier/recommended' // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
+    'plugin:prettier/recommended', // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
+    'plugin:jest/style'
   ],
-  plugins: ['react-hooks'],
+  plugins: ['react-hooks', 'eslint'],
   parserOptions: {
     ecmaVersion: 2019,
     sourceType: 'module',
@@ -20,7 +21,7 @@ module.exports = {
     '@typescript-eslint/interface-name-prefix': 0,
     '@typescript-eslint/prefer-interface': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
-    "@typescript-eslint/explicit-module-boundary-types": 0,
+    '@typescript-eslint/explicit-module-boundary-types': 0,
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/no-use-before-define': 0,
     '@typescript-eslint/no-unused-vars': ['warn', { args: 'after-used' }],
@@ -29,9 +30,12 @@ module.exports = {
     'react/no-unescaped-entities': 0,
     'react-hooks/rules-of-hooks': 2,
     'prettier/prettier': 0,
-    "no-unused-vars": "off",
+    'no-unused-vars': 'off',
     'no-unneeded-ternary': 'error',
-    'eqeqeq': 'error'
+    'eqeqeq': 'error',
+    'jest/no-focused-tests': 'error',
+    'jest/no-identical-title': 'error',
+    'jest/valid-expect': 'error'
   },
   settings: {
     react: {
--- a/src/ensembl/.eslintrc.js
+++ b/src/ensembl/.eslintrc.js
@@ -4,7 +4,6 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
     'plugin:prettier/recommended', // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
-    'plugin:jest/style'
   ],
   plugins: ['react-hooks', 'jest'],
   parserOptions: {
--- a/src/ensembl/.eslintrc.js
+++ b/src/ensembl/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     'plugin:prettier/recommended', // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
     'plugin:jest/style'
   ],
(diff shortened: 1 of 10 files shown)


src/ensembl/.eslintrc.js
```diff
@@ -29,9 +30,12 @@ module.exports = {
...
     'react-hooks/rules-of-hooks': 2,
     'prettier/prettier': 0,
-    "no-unused-vars": "off",
+    'no-unused-vars': 'off',
     'no-unneeded-ternary': 'error',
-    'eqeqeq': 'error'
+    'eqeqeq': 'error',
+    'jest/no-focused-tests': 'error',
+    'jest/no-identical-title': 'error',
+    'jest/valid-expect': 'error'
```
azangru, line 38: Yep, I made that error 👍 

image

src/ensembl/package.json
```diff
@@ -103,6 +103,7 @@
     "dotenv-webpack": "7.0.2",
     "eslint": "7.27.0",
     "eslint-config-prettier": "8.3.0",
+    "eslint-plugin-jest": "^24.3.6",
```
azangru, line 106: Let's do the exact version; no need for the `^`.

src/ensembl/.eslintrc.js
```diff
@@ -3,9 +3,10 @@ module.exports = {
   extends: [
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
-    'plugin:prettier/recommended' // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
+    'plugin:prettier/recommended', // Displays prettier errors as ESLint errors. This has to be the last in this array (not yet sure why).
+    'plugin:jest/style'
   ],
-  plugins: ['react-hooks'],
+  plugins: ['react-hooks', 'eslint'],
```
azangru, line 9: That's different from `plugins: ['jest']` as described in the [docs](https://github.com/jest-community/eslint-plugin-jest#usage). What does the eslint plugin do, and why did you prefer `'plugin:jest/style'` and `plugins: ['eslint']` to `plugins: ['jest']`?

imransl: `plugins: ['eslint`] is a mistake. I'll fix that. `plugin:jest/style` is to add more stylistic rules for jest. I thought of including that and see if it can spot more issues for us.

imransl:
I forgot to mention that running the linter spots a lot of issues in `jest`:

```
✖ 119 problems (104 errors, 15 warnings)
  93 errors and 15 warnings potentially fixable with the `--fix` option.
```

azangru:
@imransl could you post examples of what the linter is complaining of?

Also, if you disabled the `plugin:jest/style`, how many of those errors/warnings will go away? I see in the docs that it would complain of our `.toBe(null)` as opposed to `.toBeNull()`, which is probably meaningless.

imransl:
@azangru It was showing lost of those null errors that you mentioned and also the length check errors too:
```34:36  error  Use toHaveLength() instead  jest/prefer-to-have-length```

And once I disabled the `plugin:jest/style` as you suggested, it is showing only these errors related to `jest`:
```
/Users/imran/Projects/ensembl-client/src/ensembl/src/content/app/browser/track-panel/track-panel-tabs/TrackPanelTabs.test.tsx
  49:9  error  Expect must have a corresponding matcher call  jest/valid-expect

/Users/imran/Projects/ensembl-client/src/ensembl/src/content/app/browser/zmenu/ZmenuContent.test.tsx
  83:13  error  Expect must have a corresponding matcher call  jest/valid-expect
  86:13  error  Expect must have a corresponding matcher call  jest/valid-expect

/Users/imran/Projects/ensembl-client/src/ensembl/src/shared/components/badged-button/BadgedButton.test.tsx
  57:7  error  Matchers must be called to assert  jest/valid-expect

/Users/imran/Projects/ensembl-client/src/ensembl/src/shared/helpers/formatters/fastaFormatter.test.ts
  40:5  error  Expect must have a corresponding matcher call  jest/valid-expect

/Users/imran/Projects/ensembl-client/src/ensembl/src/shared/helpers/formatters/numberFormater.test.ts
  120:6  error  Test title is used multiple times in the same describe block  jest/no-identical-title
```

**Comment by azangru** [comment]
@imransl let's start with just these five errors. I don't know whether we, as a team, have a clear preference for `toBeNull` or for `toHaveLength`. I don't know why the jest plugin prefers those.

File: src/ensembl/package.json (hunk 2 above)
Comment (imransl) [reply src/ensembl/package.json]: ✅

**Comment by imransl** [comment]
@azangru Sure, I'll commit these changes then.

File: src/ensembl/.eslintrc.js (hunk 3 above)
Comment (imransl) [reply src/ensembl/.eslintrc.js]: ✅

**Comment by imransl** [comment]
@azangru I have fixed the `jest` lint errors. There are still some other errors related to `Function` e.g.:
```/Users/imran/Projects/ensembl-client/src/ensembl/src/content/app/browser/browser-messaging-service.test.ts
  23:42  error  Don't use `Function` as a type. The `Function` type accepts any function-like value.
It provides no type safety when calling the function, which can be a common source of bugs.
It also accepts things like class declarations, which will throw at runtime as they will not be called with `new`.
If you are expecting the function to accept certain arguments, you should explicitly define the function shape  @typescript-eslint/ban-types
```

Do they need fixing as well?

imransl:
@azangru Fixed the `Function` type related errors as well in the latest commit

azangru (approved):
Looks good.

The linting error in browser-messaging-service wasn't particularly important, because browser-messaging-service seems to be on its way out; but I suppose it won't hurt.

PR #506, opened 2021-06-02
Add a custom rendering server

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1069

## Description
This PR adds the infrastructure for server-side rendering to the Ensembl website, and contains the refactorings necessary to support it.

Noteworthy parts of this PR are:
- a Node-based rendering server
- refactored build process (added dedicated script files for development and production build)
- refactored routing, so that it can be reused both on the client and on the server side
- Page-type components in every app; currently they are the only components that run on the server and add the dynamic meta tags to the page
- updated redux code that it doesn't try to access browser-specific apis during the startup

## Immediate to-do list
- [x] fix tests
- [x] fix loading of data from browser storage into redux
- [x] fix api routing for requests that can be sent both client-side and server-side
- [x] clean up the code

## Subsequent to-dos (in separate PRs)
- [ ] rethink the storage services — they should not try to get initialized upon an import
- [ ] review the api service; specifically its use of the "apiHost" parameter from the config (this parameter probably doesn't make sense anymore)
- [ ] move all our api calls to use the base urls passed from the server to the config
- [ ] remove Connected React Router — we don't seem to be using it anyway (see [this section](https://github.com/ReactTraining/react-router/blob/42933fe141819e4662113ab2c320bf86be3490fb/packages/react-router/docs/api/hooks.md#uselocation) in React Router docs for how to use React Router's hooks for submitting analytics)
- [ ] dynamically detect free ports in the `start` script
- [ ] we don't really need to run `hasMounted` in every page component; in fact, it's clearly undesirable because it leads to a flash of white. `hasMounted` should probably be calculated once per app and then passed down through global context
- [ ] use a proper logger (winston?) on the Node side, rather than console.log
- [ ] fix the script for performing bundle analysis on the build
- [ ] dynamically generate the app manifest

## Miscellaneous notes
- After I added the `isServer` and `isClient` functions to the `environment.ts` helper file, and added a bit of client/server-dependent logic to the `config.ts`, I had to move the code for reading the environment variable from the `config.ts` file to the `environment.ts` file. Otherwise, a circular dependency between the config.ts and the environment.ts file occurred.

## Deployment URL
http://custom-ssr.review.ensembl.org/

## Views affected
All

## Possible complications
🙈


src/ensembl/Dockerfile
```diff
@@ -1,13 +1,29 @@
...
     npm ci && \
-    npm run test
+    npm run test && \
+    npm run build
+
+
+# PRODUCTION IMAGE
+FROM node:14.17.0-alpine AS runner
+
+WORKDIR /srv/ensembl-client/
```
msmanoj, line 20: Do we need to define this twice? (already defined on L9)

azangru: Yep, we are inside a different container.

src/ensembl/src/content/app/browser/browserState.ts
```diff
@@ -18,10 +18,16 @@ import browserStorageService from './browser-storage-service';
...
 
-const activeGenomeId = browserStorageService.getActiveGenomeId();
-const activeEnsObjectIds = browserStorageService.getActiveEnsObjectIds();
-const trackStates = browserStorageService.getTrackStates();
-const chrLocations = browserStorageService.getChrLocation();
+const isServer = typeof window === 'undefined';
+
+const activeGenomeId = !isServer
+  ? browserStorageService.getActiveGenomeId()
+  : null;
```
msmanoj, line 25: Here and below, wouldn't it be better to move the `isServer` logic inside the getters itself?

azangru: If you mean, move the` isServer` inside the storage services, then we will lose the type predictability of the getters — effectively, all their return values will become potentially `undefined`. However, storage services shouldn't even be run on the server — it doesn't make sense, at least as long as we keep all our data client-side, to do so. As far as a storage service is concerned, it is always run in the browser. So I am reluctant to move the server/client check into the storage services.

src/ensembl/package.json
```diff
@@ -118,13 +141,10 @@
     "jest": "26.6.3",
     "jest-fetch-mock": "3.0.3",
     "jest-transform-stub": "2.0.0",
-    "koa": "2.13.1",
-    "koa-connect": "2.1.0",
-    "koa-proxy": "1.0.0-alpha.3",
-    "koa-static": "5.0.0",
```
imransl, line 124: Since our prod test server was using `koa`, we'll need to rewrite `server.js` to get it to work again

imransl:
I am getting a bunch of warnings like below when running `storybook`:

```@babel/plugin-proposal-private-methods.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
	["@babel/plugin-proposal-private-property-in-object", { "loose": true }]```

File: src/ensembl/package.json (hunk 3 above)
Comment (azangru) [reply src/ensembl/package.json:124]: We won't need that server.js anymore. If I forgot to delete it, I will.

All the server-side code is now in the `src/server` folder

Comment (imransl) [reply src/ensembl/package.json:124]: Yes it is still there along with the `setup-ssl` script. We will need to modify `serve:prod` in `npm` too right?

**Comment by msmanoj** [comment]
I am not sure if this is something to look into in this PR.
When we try to display "Sorry, something seems to have gone wrong", if the issue is related to do with communicating with the node server itself, the error page will probably look like this without the images:
Screenshot 2021-06-15 at 17 53 52

File: src/ensembl/src/content/app/help/HelpPage.tsx [hunk 4]
```diff
@@ -0,0 +1,39 @@
...
+
+const LoadableHelp = loadable(() => import('./Help'));
+
+const HelpPage = () => {
+  const hasMounted = useHasMounted();
+
+  return (
+    <>
+      <Helmet>
+        <title>Help and documentations — Ensembl</title>
```
msmanoj, line 31: Should it be `documentation` instead of `documentations`?

src/ensembl/src/content/app/species/SpeciesPageContent.tsx
```diff
@@ -0,0 +1,101 @@
...
+import SpeciesMainView from 'src/content/app/species/components/species-main-view/SpeciesMainView';
+import CloseButton from 'src/shared/components/close-button/CloseButton';
+
+import styles from './SpeciesPage.scss';
+
+type SpeciesPageParams = {
+  genomeId: string;
+};
+
+const SpeciesPage = () => {
```
msmanoj, line 42: `SpeciesPage` - > `SpeciesPageContent`

src/ensembl/src/content/home/HomePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+const LoadableHome = loadable(() => import('./Home'));
+
+const HomePage = () => {
+  const hasMounted = useHasMounted();
+
+  return (
+    <>
+      <Helmet>
+        <title>Ensembl home page</title>
+        <meta name="description" content="This is Ensembl home page" />
```
msmanoj, line 32: Just wanted to check if the content should have `This is`?

src/ensembl/src/content/app/help/HelpPage.tsx
azangru: 👍 It should

src/ensembl/src/content/home/HomePage.tsx
azangru: Definitely not, and especially not for the home page 😄  This is just dummy text; we haven't been given real one.

azangru:
> I am getting a bunch of warnings like below when running storybook

@imransl thanks for checking the Storybook. As is frequently the case with Storybook, Storybook is tiring. It has its own default webpack config, in which it inlines babel-loader's options, and among those options, it lists the plugins that we recently removed: `plugin-proposal-class-properties`, `plugin-proposal-private-methods`, and `plugin-proposal-object-rest-spread`:

image

You will notice that these settings explicitly contain the `"loose": true` option. When babel-loader compares these inline settings with our babel config which doesn't even mention these plugins, it complains. I added a custom babelrc in the storybook directory that lists these two plugins. Hopefully, some time in the future, Storybook will remove these plugins form its own webpack config, and we will be able to delete the babelrc from the storybook folder.

azangru:
> I am not sure if this is something to look into in this PR.
> When we try to display "Sorry, something seems to have gone wrong", if the issue is something to do with communicating with the node server itself, the error page will probably look like this without the images:

Not sure what you mean, @msmanoj. In a production build, all our static files are saved into the `dist/static` folder, which will be served by Nginx. The images on the error page can be found among `dist/static/images/general-error-....jpg`.

src/ensembl/src/content/app/species/SpeciesPageContent.tsx
azangru: ✅

src/ensembl/src/content/app/help/HelpPage.tsx
azangru: ✅

src/ensembl/webpack/paths.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const getPaths = (env = 'development') => {
+  const isDev = ['dev', 'development'].includes(env);
+
+  const rootPath = projectRoot;
+  const nodeModulesPath = path.resolve(rootPath, 'node_modules');
+  const staticPath = path.resolve(rootPath, 'static');
+  const buildPath = path.resolve(rootPath, 'dist');
+  const serverEntryPath = path.resolve(rootPath, 'src/server/index.ts');
+
```
imransl, line 31: There are a couple of variables here that are not reused i.e. `nodeModulesPath` and `serverEntryPath`. Can we put these directly into the returning object?

src/ensembl/src/server/index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+import server from './server';
+
+const port = 8080;
+
+server.listen(port, () => {
+  // eslint-disable-next-line
+  console.log(`Example app listening at http://localhost:${port}`);
```
imransl, line 23: This is not an example app anymore right? :D

ens-ma7:
I haven't look at the code changes, just try the branch on my laptop. I did an npm ci and npm start and click through the site and everything seems to be working as expected (also tried the new 404). There were some errors about the static files (favicons and manifest.json) which we are aware of already and will need fixing.

ens-ma7:
Just when i already approved the PR that i spotted something thats broken. It seems like displaying youtube video isn't working:
http://ssr.review.ensembl.org/help/videos/how-to-find-and-add-species (works fine on internal) and i think there are some issues with static content on help pages, http://ssr.review.ensembl.org/help/articles/what-is-the-genome-browser: 
`
GET http://ssr.review.ensembl.org/api/docs/images/id-genome-browser.svg 404 (Not Found)
`

azangru:
> It seems like displaying youtube video isn't working

Should be fixed now.

Notice that the deployment url has now changed to http://custom-ssr.review.ensembl.org

src/ensembl/package.json
azangru: Removed the `server.js` and `setup-ssl.js` files, along with the npm scripts calling them in the `package.json`

src/ensembl/src/server/index.ts
azangru: Replaced with a more appropriate message.

src/ensembl/webpack/paths.ts
azangru: ✅

imransl:
> > It seems like displaying youtube video isn't working
> 
> Should be fixed now.
> 
> Notice that the deployment url has now changed to http://custom-ssr.review.ensembl.org

Just checked and this is working for me now

azangru:
Added an `ensemblRoot` alias for the root directory of the project (for the rare occasions not covered by the available aliases). Replaced the `../../../` in the imports of a file from a webpack directory. Seems to work fine.

imransl (approved):

ens-ma7:
The youtube video now works fine. Everything else is ok.

PR #508, opened 2021-06-07
ENSWBSITES-1129: Launchbar and header update

ens-ma7:
switch About ensembl and genome data & annotation in launchbar and header
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1129

## Importance
N/A

## Description
Based on xD provided in the jira ticket, moving 'About Ensembl...' into launchbar and moving 'Genome data...' into header

## Deployment URL
http://update-launchbar.review.ensembl.org

## Views affected
Its on the masthead so available on all views

### Other effects
N/A

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/header/Header.scss
+++ b/src/ensembl/src/header/Header.scss
@@ -61,25 +61,15 @@
   }
 }
 
-.aboutEnsembl {
+.topRight {
   display: flex;
-  align-items: baseline;
-  margin-top: 4px;
-  font-size: 12px;
-  color: $light-blue;
-}
+  align-items: center;
+  margin-left: 20px;
 
-.logotypeAbout {
-  height: 10px;
-  fill: $light-blue;
-  margin-left: 0.5rem;
-}
-
-// For disabling temporarily the about Ensembl link in production
-.aboutEnsemblDisabled {
-  color: $medium-dark-grey;
-
-  .logotypeAbout {
-    fill: $medium-dark-grey;
+  .topRightText {
+    font-size: 13px;
+//    font-weight: $light;
+    margin-right: 25px;
+    margin-top: 2px;
   }
 }
--- a/src/ensembl/src/header/Header.scss
+++ b/src/ensembl/src/header/Header.scss
@@ -68,7 +68,6 @@
 
   .topRightText {
     font-size: 13px;
-//    font-weight: $light;
     margin-right: 25px;
     margin-top: 2px;
   }
--- a/src/ensembl/src/header/Header.scss
+++ b/src/ensembl/src/header/Header.scss
@@ -61,14 +61,8 @@
   }
 }
 
-.topRight {
-  display: flex;
-  align-items: center;
-  margin-left: 20px;
-
-  .topRightText {
-    font-size: 13px;
-    margin-right: 25px;
-    margin-top: 2px;
-  }
+.topbarRight {
+  font-size: 13px;
+  margin-right: 25px;
 }
+
--- a/src/ensembl/src/header/launchbar/Launchbar.scss
+++ b/src/ensembl/src/header/launchbar/Launchbar.scss
@@ -81,7 +81,6 @@
   display: flex;
   align-items: center;
   padding-right: 46px;
-  font-size: 13px;
   color: $blue;
 }
 
--- a/src/ensembl/src/header/launchbar/Launchbar.scss
+++ b/src/ensembl/src/header/launchbar/Launchbar.scss
@@ -77,11 +77,6 @@
   }
 }
 
-.logotype {
-  height: 12px;
-  fill: white;
-}
-
 .aboutEnsembl {
   display: flex;
   align-items: center;
@@ -90,7 +85,7 @@
   color: $blue;
 }
 
-.logotypeAbout {
+.logotype {
   height: 10px;
   fill: $blue;
   margin-left: 0.5rem;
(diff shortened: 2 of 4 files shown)


src/ensembl/src/header/Header.scss
```diff
@@ -61,25 +61,15 @@
...
-.aboutEnsemblDisabled {
-  color: $medium-dark-grey;
+  align-items: center;
+  margin-left: 20px;
 
-  .logotypeAbout {
-    fill: $medium-dark-grey;
+  .topRightText {
+    font-size: 13px;
+//    font-weight: $light;
```
imransl, line 71: Don't think we need this comment

ens-ma7: That is very strange, i dont see on my checkout and i dont have any local changes. I remember removing it. Let me investigate why is it showing on github?

ens-ma7: Found it, i didnt save the file..doh :)

src/ensembl/src/header/Header.scss
```diff
@@ -61,25 +61,14 @@
   }
 }
 
-.aboutEnsembl {
+.topRight {
```
azangru, line 64: `.topbarRight`? To complement the existing topbarLeft?

image

src/ensembl/src/header/Header.scss
```diff
@@ -61,25 +61,14 @@
...
-  height: 10px;
-  fill: $light-blue;
-  margin-left: 0.5rem;
-}
-
-// For disabling temporarily the about Ensembl link in production
-.aboutEnsemblDisabled {
-  color: $medium-dark-grey;
+  align-items: center;
+  margin-left: 20px;
```
azangru, line 67: Why?

src/ensembl/src/header/Header.scss
```diff
@@ -61,25 +61,14 @@
...
 
-  .logotypeAbout {
-    fill: $medium-dark-grey;
+  .topRightText {
```
azangru, line 69: Confused. What's the purpose for this span?

image

Also, what's the purpose of `display: flex` on its immediate parent?

src/ensembl/src/header/launchbar/Launchbar.scss
```diff
@@ -77,14 +77,31 @@
   }
 }
 
-.about {
+.logotype {
+  height: 12px;
+  fill: white;
+}
```
azangru, line 83: Where is this class used?

src/ensembl/src/header/launchbar/Launchbar.scss
```diff
@@ -77,14 +77,31 @@
...
+.aboutEnsembl {
   display: flex;
   align-items: center;
-  margin-left: 20px;
+  padding-right: 46px;
+  font-size: 13px;
+  color: $blue;
+}
+
+.logotypeAbout {
```
azangru, line 93: Perhaps if the `logotype` class isn't used anywhere in the Launchbar component, this class can be renamed to just .logotype?

src/ensembl/src/header/Header.scss
ens-ma7: No idea why  : ) i think i just copied from where it was before and didnt question it :) In fact even the display flex can go. Added the changes.

src/ensembl/src/header/Header.scss
ens-ma7: sorry that was just copied from the previous launchbar.scss. Makes sense what you say, i will remove it unless the person who originally implemented it had a different expectation but looking at how the homepage is going to be now, seems like its not necessary the extra span.

ens-ma7:
@azangru thanks for the review, made the changes you suggested, please review and let me know. thanks

src/ensembl/src/header/launchbar/Launchbar.tsx
```diff
@@ -90,21 +92,33 @@ const Launchbar = () => {
...
   );
 };
 
+// Temporarily disable the link to About Ensembl in production
+const AboutEnsembl = () =>
+  isEnvironment([Environment.DEVELOPMENT, Environment.INTERNAL]) ? (
+    <Link to="/about" className={styles.aboutEnsembl}>
+      About the
+      <Logotype className={styles.logotype} />
+      team & its work
```
imransl, line 114: This bit is reused below too. Not necessary, but think it'll be nicer if we can carve this out into a separate node (like `AboutText`?) and use it in both the places.

ens-ma7: I think the "reuse" is only temporary since we dont have the help app/page on 2020. once this is available we will remove the code thats repeating completely. Does it then merit a separate node?

azangru: I agree with Ridwan — when we start showing the link to the About page in production, we'll just remove `isEnvironment` with the component inside it, and the repetition will go away by itself.

src/ensembl/src/header/launchbar/Launchbar.scss
```diff
@@ -77,14 +77,26 @@
   }
 }
 
-.about {
+.aboutEnsembl {
   display: flex;
   align-items: center;
-  margin-left: 20px;
+  padding-right: 46px;
+  font-size: 13px;
```
azangru, line 84: This is the default value of our body font; no need to specifically add it here.

azangru (approved):

imransl (approved):

PR #510, opened 2021-06-11
Add interstitial screen for Genome Browser and Entity Viewer 

jyothishnt:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1124
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1123

## Description
Currently, Genome Browser and Entity Viewer assumes that the user that visits it will have at least one species selected.

Following Andrea's design, which allows the user access to the these apps even without selecting a species first, we should update the interstitial to support this use case.

Initially I started to work on entity viewer, but then realised that genome browser as well uses the same shared component "Step". So it makes sense to get this both in one PR rather than wait for the PR to get merged to use the shared component.

## Deployment URL
http://ev-interstitial.review.ensembl.org/

## Views affected
Entity viewer
Genome Browser

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.test.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.test.tsx
@@ -87,29 +87,17 @@ describe('<Browser />', () => {
     );
 
   describe('rendering', () => {
-    test('render interstitial page when there is no species selected', () => {
+    it('renders an interstitial if no species is selected', () => {
       const { container } = mountBrowserComponent({ activeGenomeId: null });
       expect(container.querySelector('.browserInterstitial')).toBeTruthy();
     });
 
-    test('renders links to example objects only if there is no selected focus feature', () => {
-      const { container, rerender } = mountBrowserComponent();
-
-      expect(container.querySelectorAll('.exampleLinks')).toHaveLength(1);
-
-      rerender(
-        <Browser
-          {...defaultProps}
-          browserQueryParams={{
-            focus: faker.lorem.words()
-          }}
-        />
-      );
-
-      expect(container.querySelectorAll('.exampleLinks')).toHaveLength(0);
+    it('renders an interstitial if no feature has been selected', () => {
+      const { container } = mountBrowserComponent();
+      expect(container.querySelector('.browserInterstitial')).toBeTruthy();
     });
 
-    test('renders the genome browser and track panel only when there is a selected focus feature', () => {
+    it('renders the genome browser and track panel only when there is a selected focus feature', () => {
       const { container, rerender } = mountBrowserComponent();
 
       expect(container.querySelectorAll('.browserImage')).toHaveLength(0);
--- a/src/ensembl/src/content/app/browser/Browser.test.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.test.tsx
@@ -44,6 +44,9 @@ jest.mock('./track-panel/TrackPanel', () => () => (
 jest.mock('./browser-app-bar/BrowserAppBar', () => () => (
   <div className="browserAppBar">BrowserAppBar</div>
 ));
+jest.mock('./interstitial/BrowserInterstitial', () => () => (
+  <div className="browserInterstitial">BrowserInterstitial</div>
+));
 jest.mock('./track-panel/track-panel-bar/TrackPanelBar', () => () => (
   <div className="trackPanelBar">TrackPanelBar</div>
 ));
@@ -84,9 +87,9 @@ describe('<Browser />', () => {
     );
 
   describe('rendering', () => {
-    test('does not render when no activeGenomeId', () => {
+    test('render interstitial page when there is no species selected', () => {
       const { container } = mountBrowserComponent({ activeGenomeId: null });
-      expect(container.innerHTML).toBeFalsy();
+      expect(container.querySelector('.browserInterstitial')).toBeTruthy();
     });
 
     test('renders links to example objects only if there is no selected focus feature', () => {
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -66,6 +66,7 @@ import { ChrLocation } from './browserState';
 import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
 
 import styles from './Browser.scss';
+import BrowserInterstitial from './interstitial/BrowserInterstitial';
 
 export type BrowserProps = {
   activeGenomeId: string | null;
@@ -107,10 +108,6 @@ export const Browser = (props: BrowserProps) => {
   const shouldShowNavBar =
     props.browserActivated && props.browserNavOpenState && !isDrawerOpened;
 
-  if (!props.activeGenomeId) {
-    return null;
-  }
-
   const mainContent = (
     <>
       {shouldShowNavBar && <BrowserNavBar />}
@@ -122,7 +119,7 @@ export const Browser = (props: BrowserProps) => {
     <ApolloProvider client={client}>
       <div className={styles.browserInnerWrapper}>
         <BrowserAppBar onSpeciesSelect={changeGenomeId} />
-        {props.browserQueryParams.focus ? (
+        {props.activeGenomeId && props.browserQueryParams.focus ? (
           <StandardAppLayout
             mainContent={mainContent}
             sidebarContent={<TrackPanel />}
@@ -137,7 +134,7 @@ export const Browser = (props: BrowserProps) => {
             viewportWidth={props.viewportWidth}
           />
         ) : (
-          <ExampleObjectLinks {...props} />
+          <BrowserInterstitial />
         )}
       </div>
     </ApolloProvider>
--- a/src/ensembl/src/content/app/browser/Browser.tsx
+++ b/src/ensembl/src/content/app/browser/Browser.tsx
@@ -60,13 +60,13 @@ import Drawer from './drawer/Drawer';
 import { StandardAppLayout } from 'src/shared/components/layout';
 import ErrorBoundary from 'src/shared/components/error-boundary/ErrorBoundary';
 import { NewTechError } from 'src/shared/components/error-screen';
+import BrowserInterstitial from './interstitial/BrowserInterstitial';
 
 import { RootState } from 'src/store';
 import { ChrLocation } from './browserState';
 import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
 
 import styles from './Browser.scss';
-import BrowserInterstitial from './interstitial/BrowserInterstitial';
 
 export type BrowserProps = {
   activeGenomeId: string | null;
--- a/src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
+++ b/src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
@@ -1,0 +1,92 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { Link } from 'react-router-dom';
+import noop from 'lodash/noop';
+
+import { Step } from 'src/shared/components/step/Step';
+import { ImageButton } from 'src/shared/components/image-button/ImageButton';
+import { PrimaryButton } from 'src/shared/components/button/Button';
+import { Status } from 'src/shared/types/status';
+
+import { ReactComponent as SpeciesSelectorIcon } from 'static/img/launchbar/species-selector.svg';
+import { ReactComponent as BrowserIcon } from 'static/img/launchbar/browser.svg';
+import { ReactComponent as SearchIcon } from 'static/img/sidebar/search.svg';
+
+import styles from './BrowserInterstitialInstructions.scss';
+import classNames from 'classnames';
+
+const speciesSelectorButtonClasses = classNames(
+  styles.stepsWrapper,
+  styles.speciesSelectorButton
+);
+
+const BrowserInterstitialInstructions = () => {
+  return (
+    <section className={styles.instructionsPanel}>
+      <div className={styles.instructionsWrapper}>
+        <div className={styles.stepWrapper}>
+          <Step count={1} title="Find and add a species" />
+          <div className={styles.description}>
+            <ImageButton
+              className={styles.imageButtonIcon}
+              status={Status.DISABLED}
+              image={SpeciesSelectorIcon}
+            />
+            <div className={styles.iconLabel}>Species Selector</div>
+          </div>
+        </div>
+
+        <div className={styles.stepWrapper}>
+          <Step count={2} title="Return to this app" />
+          <div className={styles.description}>
+            <ImageButton
+              className={styles.imageButtonIcon}
+              status={Status.DISABLED}
+              image={BrowserIcon}
+            />
+            <div className={styles.iconLabel}>Genome Browser</div>
+          </div>
+        </div>
+
+        <div className={styles.stepWrapper}>
+          <Step
+            count={3}
+            title="Use Search or the example links to view a gene or region"
+          />
+          <div className={styles.searchDescription}>
+            <ImageButton
+              className={styles.searchButtonIcon}
+              status={Status.DISABLED}
+              image={SearchIcon}
+            />
+            <div className={styles.exampleText}>Example gene</div>
+            <div className={styles.exampleText}>Example region</div>
+          </div>
+        </div>
+
+        <div className={speciesSelectorButtonClasses}>
+          <Link to="/species-selector">
+            <PrimaryButton onClick={noop}>Go to Species Selector</PrimaryButton>
+          </Link>
+        </div>
+      </div>
+    </section>
+  );
+};
+
+export default BrowserInterstitialInstructions;
--- a/src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
+++ b/src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
@@ -15,8 +15,10 @@
  */
 
 import React from 'react';
-import { Link } from 'react-router-dom';
-import noop from 'lodash/noop';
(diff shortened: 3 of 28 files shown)


src/ensembl/src/content/app/App.tsx
```diff
@@ -61,6 +67,14 @@ const App = (props: AppProps) => {
...
           <Route path={`/global-search`} component={GlobalSearch} />
           <Route path={`/species-selector`} component={SpeciesSelector} />
+          <Route
+            path={`/entity-viewer-interstitial`}
+            component={EntityViewerInterstitial}
+          />
+          <Route
+            path={`/genome-browser-interstitial`}
+            component={BrowserInterstitial}
+          />
```
azangru, line 77: No, that's not right. The interstitials should live in the GenomeBrowser and EntityViewer apps. Same as how we are now showing the example links when the user visits these apps for the first time. They should not have their own urls.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -120,26 +117,30 @@ export const Browser = (props: BrowserProps) => {
...
-            isDrawerOpen={props.isDrawerOpened}
-            drawerContent={<Drawer />}
-            onDrawerClose={toggleDrawer}
-            viewportWidth={props.viewportWidth}
-          />
-        ) : (
-          <ExampleObjectLinks {...props} />
-        )}
-      </div>
+      {!props.activeGenomeId && !props.activeEnsObjectId ? (
```
imransl, line 120: Won't this be easier to read if the condition was like:
```
{props.activeGenomeId && props.activeEnsObjectId ? (
  <div className={styles.browserInnerWrapper}>
    <!-- Rest of the JSX code -->
  </div>
) : <BrowserInterstitial />
```

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { ReactComponent as SpeciesSelectorIcon } from 'static/img/launchbar/species-selector.svg';
+import { ReactComponent as BrowserIcon } from 'static/img/launchbar/browser.svg';
+import { ReactComponent as SearchIcon } from 'static/img/sidebar/search.svg';
+
+import styles from './BrowserInterstitialInstructions.scss';
+import { PrimaryButton } from 'src/shared/components/button/Button';
+import classNames from 'classnames';
+import { Link } from 'react-router-dom';
+
```
imransl, line 31: The imports aren't grouped properly

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Status } from 'src/shared/types/status';
+
+import { ReactComponent as SpeciesSelectorIcon } from 'static/img/launchbar/species-selector.svg';
+import { ReactComponent as BrowserIcon } from 'static/img/launchbar/browser.svg';
+import { ReactComponent as SearchIcon } from 'static/img/sidebar/search.svg';
+
+import styles from './BrowserInterstitialInstructions.scss';
+import { PrimaryButton } from 'src/shared/components/button/Button';
+import classNames from 'classnames';
+import { Link } from 'react-router-dom';
```
imransl, line 30: These imports aren't grouped properly

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+  styles.stepsWrapper,
+  styles.speciesSelectorButton
+);
+
+const BrowserInterstitialInstructions = () => {
+  return (
+    <section className={styles.instructionsPanel}>
+      <div className={styles.instructionsWrapper}>
+        <div className={styles.stepWrapper}>
+          <Step count={1} title={'Find and add a species'} />
```
imransl, line 42: The `title` attribute value can be wrapped in the usual way with single or double quotes

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+              className={styles.imageButtonIcon}
+              status={Status.DISABLED}
+              image={SpeciesSelectorIcon}
+            />
+            <div className={styles.iconLabel}>Species Selector</div>
+          </div>
+        </div>
+
+        <div className={styles.stepWrapper}>
+          <Step count={2} title={'Return to this app'} />
```
imransl, line 54: Same here too for the `title` value

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+              image={BrowserIcon}
+            />
+            <div className={styles.iconLabel}>Genome Browser</div>
+          </div>
+        </div>
+
+        <div className={styles.stepWrapper}>
+          <Step
+            count={3}
+            title={'Use Search or the example links to view a gene or region'}
```
imransl, line 68: And here as well

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+              status={Status.DISABLED}
+              image={SearchIcon}
+            />
+            <div className={styles.exampleText}>Example gene</div>
+            <div className={styles.exampleText}>Example region</div>
+          </div>
+        </div>
+
+        <div className={speciesSelectorButtonClasses}>
+          <Link to={'/species-selector'}>
```
imransl, line 82: Just like for the `title` attribute value you can wrap this with single/double quotes. Also, makes me wonder if we need any kind of consensus on when not to use them for attributes.

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <PrimaryButton onClick={() => null}>
```
imransl, line 83: We can use `noop` from lodash for this like in [`EntityViewerSidebarToolstrip.tsx`](https://github.com/Ensembl/ensembl-client/blob/393d0d14044b876968f15e1fc689925eeb751a37/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip.tsx#L72)

src/ensembl/src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -87,26 +88,34 @@ const EntityViewer = () => {
     <GeneViewSidebar />
   );
 
+  const showInterstitial = !genomeId && !entityId;
```
imransl, line 91: Do we need a variable for this condition when it is used only in one place?

src/ensembl/src/content/app/entity-viewer/interstitial/EntityViewerInterstitial.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import EntityViewerInterstitialInstructions from './entity-viewer-interstitial-instructions/EntityViewerInterstitialInstructions';
+import AppBar from 'src/shared/components/app-bar/AppBar';
+
+const EntityViewerInterstitial = () => {
+  return (
+    <div>
+      <AppBar
+        appName={'Entity Viewer'}
+        mainContent={'To start using this app...'}
```
imransl, line 27: Like I mentioned above we might need some consensus on what to use for attribute values that are static strings

src/ensembl/src/content/app/entity-viewer/interstitial/entity-viewer-interstitial-instructions/EntityViewerInterstitialInstructions.scss
```diff
@@ -0,0 +1,79 @@
...
+@media (max-width: $global-padding-left + $leftMargin + 4 * $columnWidth + 3 * $columnGap) {
+
+  .instructionsWrapper {
+    grid-template-columns: repeat(2, #{$columnWidth});
+  }
+
+  .speciesSelectorButton {
+    margin-left: 0;
+  }
+}
```
imransl, line 79: We need a new line. Wonder why Prettier didn't format the file properly 🤔

src/ensembl/src/content/app/entity-viewer/interstitial/entity-viewer-interstitial-instructions/EntityViewerInterstitialInstructions.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Status } from 'src/shared/types/status';
+
+import { ReactComponent as SpeciesSelectorIcon } from 'static/img/launchbar/species-selector.svg';
+import { ReactComponent as EntityViewerIcon } from 'static/img/launchbar/entity-viewer.svg';
+import { ReactComponent as SearchIcon } from 'static/img/sidebar/search.svg';
+
+import styles from './EntityViewerInterstitialInstructions.scss';
+import { PrimaryButton } from 'src/shared/components/button/Button';
+import classNames from 'classnames';
+import { Link } from 'react-router-dom';
```
imransl, line 30: Like above the imports here are not grouped properly

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -45,93 +45,90 @@ export const setActiveGenomeId = createAction(
...
-    if (!genomeInfo) {
-      dispatch(fetchGenomeData(activeGenomeId));
+export const setDataFromUrl =
+  (params: EntityViewerParams): ThunkAction<void, any, null, Action<string>> =>
+  (dispatch, getState: () => RootState) => {
+    const state = getState();
+    const { genomeId: genomeIdFromUrl } = params;
+
+    let activeGenomeId = getEntityViewerActiveGenomeId(state);
+    const activeEntityId = getEntityViewerActiveEntityId(state) || undefined;
```
imransl, line 55: Why can't this be `null` instead of `undefined`?

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -45,93 +45,90 @@ export const setActiveGenomeId = createAction(
...
+
+    const entityIdForUrl = activeEntityId
```
imransl, line 57: Same question as above: why `null` instead of `undefined`? Think we will need to take a look at the type `EntityViewerUrlParams` for this.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
```diff
@@ -45,93 +45,90 @@ export const setActiveGenomeId = createAction(
...
+      ? buildFocusIdForUrl(activeEntityId)
+      : undefined;
```
imransl, line 59: Same question as above: why `null` instead of `undefined`? Think we will need to take a look at the type `EntityViewerUrlParams` for this.

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -120,26 +117,30 @@ export const Browser = (props: BrowserProps) => {
...
-            drawerContent={<Drawer />}
-            onDrawerClose={toggleDrawer}
-            viewportWidth={props.viewportWidth}
-          />
-        ) : (
-          <ExampleObjectLinks {...props} />
-        )}
-      </div>
+      {!props.activeGenomeId ? (
+        <BrowserInterstitial />
```
azangru, line 121: Some general questions I wanted to address.

First, isn't it true that what we think about as an interstitial is shown in two situations:

1) When the user visiting the relevant "app" has not even selected a single species
2) When the user has selected a species, but has not yet selected a gene or region of interest

In the first case, we show a screen with instructions for how to select a species ([link to XD](https://xd.adobe.com/view/36dd8d77-32a6-48c5-a833-3b34d285bbe3-638e/screen/88b66037-ac26-4419-a590-21754a64201a?fullscreen)); in the second case, we show an in-app search screen ([link to XD](https://xd.adobe.com/view/36dd8d77-32a6-48c5-a833-3b34d285bbe3-638e/screen/c621bab7-7ed3-49f4-afe7-e4a3be06b58e/?fullscreen)). Both are the functions of an interstitial, aren't they?

The code that I am seeing in this PR narrowly specializes the interstitial to handle only case 1. For example, it calls the grey panel that contains the instructions an instructions panel, although the same panel will contain in-app search field in the future. Is this a good idea?

image

Second, shouldn't the grey panel in the interstitial be the same height as the grey panel in the species selector?

Third, wouldn't it be a good idea to temporarily move the example links block from the browser component into the new interstitial component, so that it becomes clearer that it handles the two cases? The example links will be removed from the interstitial when we add in-app search into it; but at least having them in the interstitial will remind us of the two scenarios that it needs to support.

src/ensembl/src/shared/components/step/Step.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import styles from './Step.scss';
+
+type StepProps = {
+  count: number;
+  title: string;
```
jyothishnt, line 23: change title to label

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
jyothishnt: The method itself is returning null so we can completely remove it.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralActions.ts
jyothishnt: I have changed the type of entityId in `EntityViewerUrlParams`.

src/ensembl/src/content/app/browser/Browser.tsx
jyothishnt: Yes I like the idea of moving Example links to the interstitial page B (with search). 
Except that the statement  "The example links will be removed from the interstitial when we add in-app search into it"  which is wrong as per our code review discussion with Andrea.

src/ensembl/src/shared/components/step/Step.tsx
jyothishnt: Create moreInfo react node and pass to Step component which then renders it

azangru: Let's do it in a separate PR; looks like a chunk of work that can be cleanly separated from this one.

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
```diff
@@ -0,0 +1,96 @@
...
+              className={styles.searchButtonIcon}
+              status={Status.DISABLED}
+              image={SearchIcon}
+            />
+            <div className={styles.exampleText}>Example gene</div>
+            <div className={styles.exampleText}>Example region</div>
+          </div>
+        </div>
+
+        <PrimaryButton
```
jyothishnt, line 85: Didnt we decide to use Wrap click buttons around a Link so that we can right-click and open in new tab?

jyothishnt:
In general @azangru, your code looks OK to me.

src/ensembl/src/content/app/browser/interstitial/browser-interstitial-instructions/BrowserInterstitialInstructions.tsx
azangru: Did we? I forgot... If we did, we should style the link so that it looks like a button; we shouldn't wrap a button with a link.

azangru: After another thought, I suggest we separate this task from this PR as well. It goes deeper than just replacing a button with a link here. If we want true links that look like buttons, then we should either create a dedicated component for that or update the existing PrimaryButton to handle this case. Is still feels wrong to me, because I would not expect people to interact with a button in the way they interact with a link (i.e. try to right-click on it and 'open' it in another window)

src/ensembl/src/content/app/browser/Browser.test.tsx
```diff
@@ -84,29 +87,17 @@ describe('<Browser />', () => {
...
-      );
-
-      expect(container.querySelectorAll('.exampleLinks')).toHaveLength(0);
+    it('renders an interstitial if no feature has been selected', () => {
+      const { container } = mountBrowserComponent();
+      expect(container.querySelector('.browserInterstitial')).toBeTruthy();
     });
 
-    test('renders the genome browser and track panel only when there is a selected focus feature', () => {
+    it('renders the genome browser and track panel only when there is a selected focus feature', () => {
```
imransl, line 100: Why was this changed from `test` to `it`?

src/ensembl/src/content/app/browser/Browser.tsx
```diff
@@ -66,6 +66,7 @@ import { ChrLocation } from './browserState';
 import { EnsObject } from 'src/shared/state/ens-object/ensObjectTypes';
 
 import styles from './Browser.scss';
+import BrowserInterstitial from './interstitial/BrowserInterstitial';
```
imransl, line 69: Shouldn't the styles imports come last?

src/ensembl/stories/shared-components/step/Step.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react';
+
+import { Step } from 'src/shared/components/step/Step';
+
+import styles from './Step.stories.scss';
+
+const steps = [
+  'Go to BurgerKing',
+  'Order a burger of your choice',
+  'Eat happily'
```
imransl, line 26: I'd prefer McDonalds 😄

src/ensembl/src/content/app/browser/Browser.test.tsx
azangru: For consistency with other test files. We sort of decided a while back to use the `it` function name rather than `test`.

src/ensembl/stories/shared-components/step/Step.stories.tsx
azangru: 🤷‍♂️

src/ensembl/src/content/app/browser/Browser.tsx
azangru: ✅

imransl (approved):

PR #514, opened 2021-06-21
Save the sidebar state for genomes in Genome browser and Entity viewer

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1088](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1088)

## Description
The sidebar (or track panel in Genome browser) behaviour needed some improvements by saving (or using?) the previous state in Entity viewer and Genome browser. When a user switches between species in either of these apps, the last state of the sidebar before switching has to be saved for the previously selected species. And when you switch back to that species, the saved state needs to be the default state.

This PR saves the following states in order to achieve this:

1. Sidebar state (open or close)
2. Sidebar modal state (open or close as well as view if open)
3. Drawer state in Genome browser (open or close as well as view if open)

## Deployment URL
https://trackpanel-modal-behaviour.review.ensembl.org/

## Views affected
Entity viewer -> sidebar
Genome browser -> track panel
Genome browser -> track panel -> drawer

### DIFF
--- a/src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
+++ b/src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
@@ -19,7 +19,8 @@ import pickBy from 'lodash/pickBy';
 
 import {
   getInitialTrackPanelState,
-  getTrackPanelStateForGenome,
+  getInitialTrackPanelStateForGenome,
+  getPersistentTrackPanelStateForGenome,
   TrackPanelState
 } from './trackPanelState';
 import * as browserActions from 'src/content/app/browser/browserActions';
@@ -34,9 +35,18 @@ export default function trackPanel(
   switch (action.type) {
     case getType(browserActions.setDataFromUrl):
       const { activeGenomeId } = action.payload;
+      const newTrackPanelStateForGenome = state[activeGenomeId]
+        ? {
+            ...state[activeGenomeId],
+            ...getPersistentTrackPanelStateForGenome(activeGenomeId)
+          }
+        : {
+            ...getInitialTrackPanelStateForGenome(activeGenomeId)
+          };
+
       return {
         ...state,
-        [activeGenomeId]: getTrackPanelStateForGenome(activeGenomeId)
+        [activeGenomeId]: newTrackPanelStateForGenome
       };
     case getType(trackPanelActions.updateTrackPanelForGenome):
       return {
--- a/src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
+++ b/src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
@@ -18,16 +18,15 @@ import { ActionType, getType } from 'typesafe-actions';
 import pickBy from 'lodash/pickBy';
 
 import {
-  getInitialTrackPanelState,
-  getInitialTrackPanelStateForGenome,
-  getPersistentTrackPanelStateForGenome,
+  getTrackPanelState,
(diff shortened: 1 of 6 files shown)


src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
```diff
@@ -34,9 +35,18 @@ export default function trackPanel(
...
     case getType(browserActions.setDataFromUrl):
       const { activeGenomeId } = action.payload;
+      const newTrackPanelStateForGenome = state[activeGenomeId]
+        ? {
+            ...state[activeGenomeId],
+            ...getPersistentTrackPanelStateForGenome(activeGenomeId)
+          }
+        : {
+            ...getInitialTrackPanelStateForGenome(activeGenomeId)
+          };
```
azangru, line 45: I'm confused. Why do we even need to read the data from the browser storage, when it should already be available in the redux state? Shouldn't we rather do the following:

```
if (!state[activeGenomeId]) {
  return { ...state, [activeGenomeId]: getTrackPanelStateForGenome(activeGenomeId)  }
} else {
  return state;
}
```

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -233,10 +229,6 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
       return;
     }
 
-    if (isSidebarOpen) {
-      dispatch(closeSidebarModal());
-    }
```
azangru, line 238: Couldn't find anywhere in the ticket or in the description of the PR the reason for this change. Don't we want to close the modal for EntityViewer when we switch from it to a different application anymore?

imransl: As discussed during daily scrum today we are keeping the modal open

src/ensembl/src/content/app/browser/track-panel/trackPanelReducer.ts
imransl: Changed the code to be like this in 19202e8

src/ensembl/src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSelectors.ts
```diff
@@ -54,5 +54,9 @@ export const isEntityViewerSidebarOpen = (state: RootState): boolean => {
...
 };
 
-export const getEntityViewerSidebarModalView = (state: RootState) =>
-  getEntityViewerGenomeState(state)?.sidebarModalView ?? null;
+export const getEntityViewerSidebarModalView = (state: RootState) => {
+  const activeGenomeId = getEntityViewerActiveGenomeId(state);
+  return activeGenomeId
+    ? getEntityViewerGenomeState(state)?.sidebarModalView
+    : null;
+};
```
azangru, line 62: What was the purpose for updating this function? Why is the new form preferable?

imransl: This shouldn't be preferable. I was checking if the getting the state with/without knowing the `genomeId` was causing any issues. So this is only a remnant of that and should be reverted.

imransl: ✅

azangru (approved):

PR #516, opened 2021-06-23
New Homepage layout

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1114](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1114)

## Importance
N/A

## Description
This is the new homepage layout (XD and dev spec attached to Jira ticket)

## Deployment URL
http://homepage.review.ensembl.org

## Views affected
None

### Other effects
N/A

## Possible complications
N/A


azangru:
One thing that doesn't look quite right is how the image behaves on viewports that are narrower than 980px. The body width is pinned at 980px, because we don't allow it to go lower; but the image in the bottom of the page remains positioned at the left of the screen, and the human is no longer in the centre of the page.

https://user-images.githubusercontent.com/6834224/123640919-90335f80-d819-11eb-9d70-074e254af177.mp4

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+  display: grid;
+  grid-template-rows: minmax(550px, auto) 85px;
+}
 
-  h2 {
-    color: $grey;
-    font-size: 14px;
-    font-weight: $normal;
-  }
+.container {
```
azangru, line 8: I see the `.container` class describes the area of the home page that is above the footer. Maybe call it `.main`?

image

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+  display: grid;
+  grid-template-rows: auto 1fr;
+  background-color: $blue;
```
azangru, line 11: why?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+  background-image: linear-gradient($black, $blue);
+  padding-top: 40px; 
+}
+
+.middleHomeWrapper {
```
azangru, line 16: Mmm, I am not sure it's a middle of anything. `.mainTop` perhaps?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
   position: relative;
 }
 
-.search {
-  background: $light-grey;
-  box-shadow: 0 3px 5px $medium-light-grey;
-  min-height: 235px;
-  padding: 15px 60px;
+.middleHome {
```
azangru, line 20: mainTopMiddle maybe?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
-  box-shadow: 0 3px 5px $medium-light-grey;
-  min-height: 235px;
-  padding: 15px 60px;
+.middleHome {    
+  display: grid;
+  grid-template-rows: 38px 110px 100px; 
+  justify-content: center;
+  height: 248px;
+
+  .logotypeContainer {
```
azangru, line 26: Why does this class exist? You can remove it entirely and add `justify-self: center` to the `.logotype`

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
 }
 
-.search {
-  background: $light-grey;
-  box-shadow: 0 3px 5px $medium-light-grey;
-  min-height: 235px;
-  padding: 15px 60px;
+.middleHome {    
+  display: grid;
+  grid-template-rows: 38px 110px 100px;
```
azangru, line 22: `38px` --> `auto`?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
-.search {
-  background: $light-grey;
-  box-shadow: 0 3px 5px $medium-light-grey;
-  min-height: 235px;
-  padding: 15px 60px;
+.middleHome {    
+  display: grid;
+  grid-template-rows: 38px 110px 100px; 
+  justify-content: center;
+  height: 248px;
```
azangru, line 24: why?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+
+  .logotypeContainer {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  .logotype {
+    width: 270px;
```
azangru, line 33: i suspect the width is unnecessary

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+    height: 36px;
+    fill: $white;
+  }
+
+  .text {
```
azangru, line 38: Should it be a heading, such as `h2` by any chance?

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
azangru, line 41: it doesn't need to be `display: flex`. `text-align: center` should be sufficient

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+    height: 36px;
+    fill: $white;
+  }
+
+  .text {
+    display: flex;
+    flex-direction: column;
+    align-items: center;      
+    font-size: 18px;
+    font-weight: 300;
```
azangru, line 43: We have a variable for that. It's called `$light`.

src/ensembl/src/content/home/Home.scss
```diff
@@ -1,44 +1,223 @@
...
+    margin-top: 15px;
+  }
+}
+
+.selectionWrapper {
+  display: grid;
+  grid-template-columns: 270px 270px 270px;
+} 
+
+.selectionContainer {
```
azangru, line 54: `.selection` is a strange name for the buttons row

src/ensembl/src/content/home/Home.tsx
```diff
@@ -15,25 +15,160 @@
...
+          <Logotype className={styles.logotype} />
+          <img src={ebiLogoUrl} className={styles.ebiLogo} />
+          <div className={styles.text}>
+            Wellcome Genome Campus, Hinxton, Cambridgeshire CB10 1SD, UK
+          </div>
+        </div>
+
+        <div className={styles.footerRightIcons}>
+          <a
+            href="//www.ensembl.info"
```
azangru, line 136: Here and below: there is no need to depend on the protocol over which our page is served when deciding which protocol to use for an external site. All these links should explicitly use `https://`

src/ensembl/src/content/home/Home.tsx
```diff
@@ -15,25 +15,160 @@
...
+          </a>
+          <a
+            href="//www.facebook.com/Ensembl.org"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <img src={facebookIconUrl} className={styles.mediaIcons} />
+          </a>
+          <a
+            href="//www.twitter.com/ensembl"
```
azangru, line 157: twitter account doesn't have a `www` in the url:

image

src/ensembl/src/content/home/Home.tsx
```diff
@@ -15,25 +15,160 @@
...
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <img src={twitterIconUrl} className={styles.mediaIcons} />
+          </a>
+          <a
+            href="//www.elixir-europe.org/platforms/data/core-data-resources"
```
azangru, line 164: Elixir doesn't have a `www` in the url:

image

azangru:
We need to remember that we broke the help and about pages by fixing the `.root` container. Will need to fix those.

By the way, shouldn't we keep hiding those link for the users:



or have we decided that it's ok to open these parts of the site to the world?

ens-ma7:
> We need to remember that we broke the help and about pages by fixing the `.root` container. Will need to fix those.
> 
> By the way, shouldn't we keep hiding those link for the users:
> 
> 
> 
> or have we decided that it's ok to open these parts of the site to the world?

These are hidden in production. The code is in place to disabled them in production.

src/ensembl/src/content/home/Home.scss
ens-ma7: Done

src/ensembl/src/content/home/Home.scss
ens-ma7: Removed. I was just playing with the background color before and forgot to remove it.

src/ensembl/src/content/home/Home.scss
ens-ma7: Thanks for the suggestion. It's not easy to come up with class name which makes sense specially when you are in the middle of a code crisis :)

src/ensembl/src/content/home/Home.scss
ens-ma7: Done

src/ensembl/src/content/home/Home.scss
ens-ma7: I tried this before but it didn't work. Maybe because its an svg and we cant center the svg itself but if you center the parent content then it works.

ens-ma7: sorry ignore above comment, turn out i need to remove the parent completely and it works with justify-self.

src/ensembl/src/content/home/Home.scss
ens-ma7: Done

src/ensembl/src/content/home/Home.scss
ens-ma7: No idea again :) I think it was when i was playing with the middle content that i might have forgotten to remove it after we got a proper fix. There is a pattern here, things that might have worked before and after the proper fix, they dont make any difference and then i forgot to remove them. But good spot, i have removed it.

src/ensembl/src/content/home/Home.scss
ens-ma7: True, i didnt even realise the width by default is closed to 270px ( i took this width from the template that Andrea provided).

src/ensembl/src/content/home/Home.scss
ens-ma7: Done

src/ensembl/src/content/home/Home.scss
ens-ma7: Done

src/ensembl/src/content/home/Home.scss
ens-ma7: I have changed to `h2` but i still had to style it. Do we have an `h2` with style already defined? Should we have one?

src/ensembl/src/content/home/Home.tsx
ens-ma7: Done

src/ensembl/src/content/home/Home.tsx
ens-ma7: I have removed the `www` but the link does work with `www`.

src/ensembl/src/content/home/Home.tsx
ens-ma7: Done. same as above

ens-ma7:
> One thing that doesn't look quite right is how the image behaves on viewports that are narrower than 980px. The body width is pinned at 980px, because we don't allow it to go lower; but the image in the bottom of the page remains positioned at the left of the screen, and the human is no longer in the centre of the page.
> 
>  homepage.mp4

Fixed.

ens-ma7:
Help page fixed as part of this PR.

src/ensembl/src/content/home/Home.scss
azangru: > Do we have an h2 with style already defined?

We don't

> Should we have one?

I don't know 😞

src/ensembl/src/content/home/Home.tsx
azangru: It does, but twitter redirects it to no www:

image

src/ensembl/src/content/home/Home.tsx
azangru: yep, Elixir also does a redirect:

image

ens-ma7:
@azangru all above changes (including help pages) and the one we discussed in the code review are now fixed and push. Please review and let me know. Thanks

azangru (approved):

azangru:
Approved and moved this to QA.

PR #518, opened 2021-06-30
add similar_proteins stat from pdb

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1156

## Description
The PDBeKB REST service provides information on 'Similar proteins'. We should add the results of this to 2020, for two reasons. 1) it's genuinely valuable for users and 2) showing the links to PDBeKB when the count for all other data is zero is confusing

## Deployment URL
http://similar-proteins.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000108622?view=protein

## Views affected
EV-Geneview

### Other effects
N/A

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
+++ b/src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
@@ -23,6 +23,7 @@ export type ProteinStatsInResponse = {
   ligands: number;
   interaction_partners: number;
   annotations: number;
+  similar_proteins: number;
 };
 
 export type ProteinAcessionInResponse = {
@@ -42,6 +43,7 @@ export type ProteinStats = {
   ligandsCount: number;
   interactionsCount: number;
   annotationsCount: number;
+  similarProteinsCount: number;
 };
 
 export const fetchProteinSummaryStats = async (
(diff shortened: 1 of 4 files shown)


imransl (approved):

src/ensembl/src/content/app/entity-viewer/shared/rest/rest-data-fetchers/proteinData.ts
```diff
@@ -50,11 +52,10 @@ export const fetchProteinSummaryStats = async (
 ): Promise<ProteinStats | null> => {
   const proteinStatsUrl = `https://www.ebi.ac.uk/pdbe/graph-api/uniprot/summary_stats/${xrefId}`;
 
-  const proteinStatsData:
-    | UniProtSummaryStats
-    | undefined = await apiService.fetch(proteinStatsUrl, {
-    signal
-  });
+  const proteinStatsData: UniProtSummaryStats | undefined =
```
azangru, line 55: I know we missed it before, but the function (line 60) returns `null` rather than undefined; so we should update our type.

azangru (approved):
Looks good

PR #519, opened 2021-07-04
Fix grid layout for Genome Browser and Entity Viewer

azangru:
## Type
- Bug fix

## Importance
Currently broken in production

## Description
PR https://github.com/Ensembl/ensembl-client/pull/516 updated the grid styles of the `Root` component, which didn't sit well with PR https://github.com/Ensembl/ensembl-client/pull/510.

**Bug:** (notice the large grey block extending to the bottom of the page)
image

**After the fix:**
image

## Deployment URL
http://fix-layout.review.ensembl.org/genome-browser

## Views affected
- Genome Browser
- Entity Viewer
(in both cases — it's the interstitial screen when the user has not selected any species)

### DIFF
--- a/src/ensembl/src/content/app/browser/Browser.scss
+++ b/src/ensembl/src/content/app/browser/Browser.scss
@@ -2,6 +2,7 @@
 
 .browserInnerWrapper {
   display: grid;
+  align-items: start;
   grid-template-rows: 80px 1fr;
 }
 
--- a/src/ensembl/src/content/app/entity-viewer/EntityViewer.scss
+++ b/src/ensembl/src/content/app/entity-viewer/EntityViewer.scss
@@ -3,5 +3,6 @@
 .entityViewer {
   height: 100%;
   display: grid;
+  align-items: start;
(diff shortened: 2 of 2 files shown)


ens-ma7:
I am a bit confused with the fix, i was expecting the fix to be setting the height for that box. Currently it has min-height shouldn't it not be a fixed height? And we have the same kind of box positioning in species-selector and in this one the box height is fixed. 

image

azangru:
> how did the align-items start made the box obey a fixed height?

The default align-items value is `stretch`

ens-ma7:
is the fix `align-items: start` the correct one or should we set the height of the box instead of using min-height? Since the other box has a set height.

azangru:
Since the box changes its height in response to the body width, I doubt that the fixed height is the right approach here:

image

Screenshot 2021-07-05 at 11 44 49

ens-ma7 (approved):

jyothishnt (approved):

jyothishnt:
This do not need a QA I think we can just merge it in.

PR #520, opened 2021-07-05
Label Ensembl canonical on Transcript Image using data from Thoas

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-520

## Description
The default (topmost) transcript in the transcripts table will be identified with a label:

1) Labelled transcript as 'MANE Select' or 'Ensembl canonical' depending on the evidence used to identify it
2) Added logic to display additional 'MANE Plus Clinical' transcripts if any
3) Tooltip updated from the definition field coming from metadata

~NOTE: Currently it is not the topmost, we will have to tweak the sorting algorithm which will be done in another ticket.~

## Deployment URL
MANE select - http://add-ensembl-canonical.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000139618?view=transcripts

Ensembl canonical (not Mane select) http://add-ensembl-canonical.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000278704?view=transcripts

Multiple MANE labels http://add-ensembl-canonical.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000187098?view=transcripts

## Views affected
Entity Viewer

### Other effects
NA

## Possible complications
NA

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -160,7 +160,11 @@ const QUERY = gql`
             value
             label
             definition
-            description
+          }
+          mane {
+            value
+            label
+            definition
           }
         }
       }
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -155,6 +155,14 @@ const QUERY = gql`
             }
           }
         }
+        metadata {
+          canonical {
+            value
+            label
+            definition
+            description
+          }
+        }
       }
     }
   }
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
@@ -14,15 +14,12 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { Pick2, Pick3 } from 'ts-multipick';
 
 import { getFeatureCoordinates } from 'src/content/app/entity-viewer/shared/helpers/entity-helpers';
-import {
-  transcriptSortingFunctions,
-  defaultSort
-} from 'src/content/app/entity-viewer/shared/helpers/transcripts-sorter';
+import { transcriptSortingFunctions } from 'src/content/app/entity-viewer/shared/helpers/transcripts-sorter';
 
 import { filterTranscriptsBySOTerm } from 'src/content/app/entity-viewer/shared/helpers/transcripts-filter';
 
@@ -58,6 +55,7 @@ type Transcript = DefaultTranscriptListItemProps['transcript'] & {
 } & {
   spliced_exons: Array<Pick3<SplicedExon, 'exon', 'slice', 'location'>>;
 } & Pick2<FullTranscript, 'slice', 'location'>;
+
 type Gene = DefaultTranscriptListItemProps['gene'] & {
   stable_id: FullGene['stable_id'];
   transcripts: Array<Transcript>;
@@ -79,12 +77,6 @@ const DefaultTranscriptslist = (props: Props) => {
   const dispatch = useDispatch();
 
   const { gene } = props;
-
-  //Using this to get the default order of transcripts in which the first one is selected, this might change later with the data coming directly from thoas
-  const defaultTranscriptId = useMemo(() => {
-    const sortedTranscripts = defaultSort(gene.transcripts);
-    return sortedTranscripts[0].stable_id;
-  }, [gene.stable_id]);
 
   const sortingFunction = transcriptSortingFunctions[sortingRule];
   const sortedTranscripts = sortingFunction(gene.transcripts) as Transcript[];
@@ -121,7 +113,6 @@ const DefaultTranscriptslist = (props: Props) => {
           return (
             <DefaultTranscriptsListItem
               key={index}
-              isDefault={transcript.stable_id === defaultTranscriptId}
               gene={gene}
               transcript={transcript}
               rulerTicks={props.rulerTicks}
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
@@ -92,6 +92,6 @@ describe('<DefaultTranscriptListItem />', () => {
     const { container } = renderComponent();
     expect(
       container.querySelector('.transcriptQualityLabel')?.textContent
-    ).toBe(defaultProps.transcript.metadata.mane.label);
+    ).toBe(defaultProps.transcript.metadata.mane?.label);
   });
 });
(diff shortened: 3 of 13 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
```diff
@@ -80,12 +78,6 @@ const DefaultTranscriptslist = (props: Props) => {
 
   const { gene } = props;
 
-  //Using this to get the default order of transcripts in which the first one is selected, this might change later with the data coming directly from thoas
```
ens-ma7, line 83: i think this code need to stay because we need to apply the default sorting first and then display the transcripts. sorry if i confused you when we spoke but we do need this default sorting and when you apply the default sorting, the first transcript will be either "Mane" or "Canonical" or can be empty.

src/ensembl/src/shared/types/thoas/metadata.ts
```diff
@@ -14,13 +14,16 @@
  * limitations under the License.
  */
 
-import { Source } from './source';
+type ValueSetMetadata = {
```
ens-ma7, line 17: according to thoas docs, there is also a description field. Do you not need it? As discussed i have created my branch for the transcript more info from this branch and i am going to add it to the code unless its not needed.

azangru:
For the reference, the alternative way of writing the default sorter function we tried was:

```
export function defaultSort<
  T extends Array<
    IsProteinCodingTranscriptParam & {
      slice: SliceWithOnlyLength;
      so_term: string;
    }
    & Pick2 <FullTranscript, 'metadata', 'canonical' | 'mane'>
  >
>(transcripts: T): T {

  const sortingFunction = (a: T[0], b: T[0]): number => {
    if (b.metadata.canonical) {
      // canonical transcript goes first
      return 1;
    } else if (a.metadata.canonical) {
      return -1;

    } else if (b.metadata.mane && !a.metadata.mane) {
      // transcripts with MANE metadata go second
      return 1;
    } else if (a.metadata.mane) {
      return -1;


    } else if (isProteinCodingTranscript(b) && !isProteinCodingTranscript(a)) {
      // protein-coding transcripts go before non-coding
      return 1;

    } else if (isProteinCodingTranscript(a) && !isProteinCodingTranscript(b)) {
      // protein-coding transcripts go before non-coding
      return -1;

    } else {
      // transcripts within a group are ranked by their length (from longer to shorter)
      return compareTranscriptLengths(a, b);
    }
  };

  return [...transcripts].sort(sortingFunction) as T;
}
```

I was expecting it to be several times faster than the other sorting function (because it was going through the array just once), but it ended up being slower. Still not sure how that happened. But I am just recording the alternative implementation here if we ever wanted to return to it.

src/ensembl/src/shared/types/thoas/metadata.ts
azangru: > Do you not need it?

It isn't populated with any real data at the moment, and we aren't using it in the client anyway. Currently, the definition field is sufficient for us.

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/DefaultTranscriptsList.tsx
ens-ma7: Agreed we dont need this piece of code but will be good to use `useMemo` to apply sorting when the geneId or sortingRule are different

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -155,6 +155,18 @@ const QUERY = gql`
             }
           }
         }
+        metadata {
+          canonical {
+            value
```
azangru, line 160: We don't really use this value anywhere 😃

src/ensembl/src/content/app/entity-viewer/gene-view/components/default-transcripts-list/default-transcripts-list-item/DefaultTranscriptListItem.test.tsx
```diff
@@ -88,10 +88,10 @@ describe('<DefaultTranscriptListItem />', () => {
     expect(queryByTestId('transcriptsListItemInfo')).toBeTruthy();
   });
 
-  it('displays selected transcript', () => {
-    const { container } = renderComponent({ isDefault: true });
+  it('displays Ensembl canonical transcript with the correct label', () => {
```
azangru, line 91: The test isn't really checking what it claims to be checking. The description of the test talks about Ensembl canonical transcript, but the code on line 95 is checking the mane metadata (which means that this also happens to be a MANE select transcript)?

If you really want to test the content of the label, you should probably test the `TranscriptQualityLabel` component, move this test over there, and add some other test cases:

- canonical + mane select => uses transcript.metadata.mane.label
- canonical without mane select => uses transcript.metadata.canonical.label
- non-canonical, with mane metadata => uses transcript.metadata.mane.label
- non-canonical, no mane metadata => returns null

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
```diff
@@ -90,6 +91,23 @@ const QUERY = gql`
...
+          canonical {
+            value
+            label
+            definition
+          }
+          mane {
+            value
+            label
+            definition
+          }
```
azangru, line 109: Where do you use all this metadata for gene external references? You probably only need canonical.value and mane.value for sorting?

src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-sorter.test.ts
```diff
@@ -24,26 +24,32 @@ import {
 
 import { createTranscript } from 'tests/fixtures/entity-viewer/transcript';
 
+const createTranscriptWithEmptyMetadata = () => {
+  const transcript = createTranscript();
+  transcript.metadata.canonical = transcript.metadata.mane = null;
```
azangru, line 29: Umm, I am not sure I like this function, for several reasons, one of which being that it claims to create a transcript with empty metadata, but only empties two metadata fields.

Why don't we assign nulls to all metadata fields directly in the `createTranscript` function?

src/ensembl/src/content/app/entity-viewer/shared/components/default-transcript-label/TranscriptQualityLabel.tsx
```diff
@@ -17,32 +17,43 @@
 import React from 'react';
 
 import QuestionButton from 'src/shared/components/question-button/QuestionButton';
+import { TranscriptMetadata } from 'ensemblRoot/src/shared/types/thoas/metadata';
```
imransl, line 20: We need to separate this from `QuestionButton` import which is a component

src/ensembl/src/content/app/entity-viewer/shared/components/default-transcript-label/TranscriptQualityLabel.tsx
```diff
@@ -17,32 +17,43 @@
...
+      label: props.metadata.canonical.label,
+      definition: props.metadata.canonical.definition
+    };
+  } else if (props.metadata.mane) {
+    return {
+      label: props.metadata.mane.label,
+      definition: props.metadata.mane.definition
+    };
+  }
+};
```
imransl, line 45: How about destructuring the `metadata` properties so that we don't need to repeat `props.metadata` everywhere?

```const { canonical, metadata, mane } = props.metadata;```

ens-ma7 (approved):

ens-ma7:
@jyothishnt i have approved from what i can understand, please review the suggested changes by @azangru and @imransl and wait for their approval before merging. Dont rely on my approval :)

src/ensembl/src/content/app/entity-viewer/shared/helpers/tests/transcripts-sorter.test.ts
jyothishnt: updated code, please check.

src/ensembl/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
jyothishnt: yeah you are right. It is used for default sort in this component. updated

azangru (approved):

PR #522, opened 2021-07-08
Fix layout for the Species page

azangru:
## Type
- Bug fix

## Importance
Bug in production

## Description

**Before the fix:**

https://user-images.githubusercontent.com/6834224/125003793-eab99080-e04f-11eb-9758-1a470e933728.mp4

**After the fix:**

https://user-images.githubusercontent.com/6834224/125003829-fc9b3380-e04f-11eb-8e10-ab6ecf91d419.mp4

## Deployment URL
http://fix-layout.review.ensembl.org

## Views affected
Species page

### DIFF
--- a/src/ensembl/src/content/app/species/SpeciesPage.scss
+++ b/src/ensembl/src/content/app/species/SpeciesPage.scss
@@ -1,6 +1,8 @@
 @import 'src/styles/common';
 
 .speciesPage {
+  display: grid;
+  grid-template-rows: auto 1fr;
   height: 100%;
 }
 
--- a/src/ensembl/src/content/app/species/SpeciesPage.scss
+++ b/src/ensembl/src/content/app/species/SpeciesPage.scss
@@ -1,5 +1,9 @@
 @import 'src/styles/common';
 
+.speciesPage {
+  height: 100%;
+}
+
 .topbar {
   display: grid;
   grid-template-columns: 1fr auto;
--- a/src/ensembl/src/content/app/species/SpeciesPage.scss
+++ b/src/ensembl/src/content/app/species/SpeciesPage.scss
@@ -2,7 +2,7 @@
 
 .speciesPage {
   display: grid;
-  grid-template-rows: auto 1fr;
+  grid-template-rows: max-content 1fr;
   height: 100%;
 }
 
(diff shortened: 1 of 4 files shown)


src/ensembl/src/content/app/species/SpeciesPageContent.tsx
```diff
@@ -59,7 +59,7 @@ const SpeciesPageContent = () => {
   }, [genomeId]);
 
   return (
-    <>
+    <div className={styles.speciesPage}>
```
ens-ma7, line 62: Do we need to add any style/class to it, just a wrapper div will be enough to make it obey the parent grid layout. will the height 100% makes a difference when there is less content. maybe we should have a min-height instead of 100%

src/ensembl/src/content/app/species/SpeciesPage.scss
```diff
@@ -1,5 +1,9 @@
 @import 'src/styles/common';
 
+.speciesPage {
+  height: 100%;
```
jyothishnt, line 4: This have introduced an extra scrollbar to the right which is always visible

azangru: The scrollbar turns out to be the result of the not overridden margin bottom of the species component.  

image

azangru: Made some updates to the layout that should account for that.

ens-ma7:
I am still seeing the double scrollbar on the right and also when we have multiple species, it seems broken:
image

azangru:
Not sure what you mean @ens-ma7 . The second scrollbar appears when the height of the window gets too small.

https://user-images.githubusercontent.com/6834224/125275380-d8b54780-e306-11eb-953f-2973a6597bbf.mp4

That's because we set the minimum height for the StandardAppLayout component to 550px or thereabouts. It is not specific to the Species page, but is everywhere the StandardAppLayout component is used. Users weren't supposed to view Ensembl on such small screens anyway, as Andrea likes to point out.

ens-ma7:
@azangru the other thing that i wanted to point out is the issue with the species tag/button in the top bar on the species homepage. When we have more species than the window width can handle, it doesnt look like its wrapping all of them properly. But this is also an issue on 2020.ensembl.org. Its up to you if you want to fix it as part of this PR or can create another ticket.

https://user-images.githubusercontent.com/6040010/125257153-0beedb00-e2f5-11eb-88ed-611015c2988d.png

azangru:
@ens-ma7 Good catch. Check again; it should be fixed now.

ens-ma7 (approved):
Thanks for fixing all these @azangru , looks good now.

PR #525, opened 2021-07-11
Update dependencies

azangru:
## Description
Routine update of dependencies.

### Notable changes:
- typescript updated to 4.3.x
- redux-toolkit updated to version 1.6 (includes RTK Query)

### Other important notes
The latest version of `d3` is 7, in which they switched to using ESM modules. This breaks our tests, because Jest, with `babel-jest` to transpile our typescript files, on the one hand, tries to avoid the transpilation of the files in the `node_modules` folder (because that's slow), but on the other hand fails to import proper ES6 modules. I've tried updating `jest.config.js` with the following settings:

```diff
   ],
   testEnvironment: 'jsdom',
   transform: {
-    '.+\\.tsx?$': 'babel-jest',
+    '.+\\.[t|j]sx?$': 'babel-jest',
     '.+\\.(css|scss|png|jpg|svg|gif|eot|ttf|otf|woff|woff2)$':
       'jest-transform-stub'
-  }
+  },
+  transformIgnorePatterns: [
+    'node_modules/(?!(d3.*)/)'
+  ]
 };
```

but that only moved the problem one step further, because all d3 modules, in their turn, depend on some other modules, which are also in ES6 module format. Don't know how to deal with this situation.

## Deployment URL
http://update-packages.review.ensembl.org


imransl:
Noticed some issues with this update:

1. Sass can't be installed in the latest stable version of Node.js, however it can be with the latest LTS version.
2. The fonts don't load properly
3. The external link for `Training` in the `Help` page doesn't work

azangru:
@imransl 

> Sass can't be installed in the latest stable version of Node.js, however it can be with the latest LTS version.

I am running the LTS, but I also tested the latest current release (16.5.0) after you mentioned that there's a problem with one of the dependencies. Worked fine (`npm ci` completed without a hitch).



> The fonts don't load properly

Looks fine over here 🤷‍♂️ 

image

> The external link for Training in the Help page doesn't work

Was fixed in https://github.com/Ensembl/ensembl-client/pull/531

> Sequence download doesn't work locally (not sure if this a known issue or not)

1) The client-side problem is the whole reason PR https://github.com/Ensembl/ensembl-client/pull/524 exists (and is now merged into dev)
2) Refget-proxy on staging is broken, and since we normally proxy our requests to the apis on staging in development, sequence download may also feel broken. Pointing the proxy at production apis should result in a successful download.

I've rebased the branch on dev, so all new commits from dev should now be in this branch as well.

imransl:
@azangru Yes can confirm all of them work now after I pulled in the latest code. As for the build and fonts failure, maybe it was related to my Node.js version. I was using 16.3.0 while testing this PR. After I updated to 16.5.0 and tested again, those issues disappeared.

PR #530, opened 2021-07-16
Explicitly set line length for prettier in the prettier config

azangru:
Hopefully after this prettier will behave identically on our machines.

Reference: [prettier docs](https://prettier.io/docs/en/options.html#print-width)

### DIFF
--- a/src/ensembl/prettier.config.js
+++ b/src/ensembl/prettier.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   singleQuote: true,
   arrowParens: 'always',
-  trailingComma: "none"
+  trailingComma: "none",
+  printWidth: 80
 };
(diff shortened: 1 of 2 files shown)


src/ensembl/prettier.config.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   singleQuote: true,
   arrowParens: 'always',
-  trailingComma: "none"
+  trailingComma: "none",
```
imransl, line 4: Use single quotes instead?

imransl:
Strange that the CI/CD is breaking despite this change not seeming to be one to do so

src/ensembl/prettier.config.js
azangru: I only added a comma on this line...

azangru: Updated

imransl: Thanks. Yes I know it isn't related to PR but thought we'll just change it here itself 😄

imransl (approved):

PR #535, opened 2021-07-26
Update the Ensembl help link

imransl:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1232](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1232)

## Description
Changed the text of the help link/button in the homepage. In order to do this, had to alter the `HomePopupButton` component.

**Note**: The related articles are not sent in the API response. However, it is sent for other slugs like `genome-browser` and `entity-viewer`.  Also a couple of slugs are not working i.e. `species-selector` and `species-homepage`.

## Deployment URL
http://update-help-link.review.ensembl.org/

## Views affected
The help link of all pages are affected by this change


imransl:
Noticed while going through the review site that the homepage help overlay styling is not right. I'll fix this right away and push the changes.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
```diff
@@ -28,7 +29,16 @@ import { SlugReference } from './types';
...
-type Props = SlugReference;
+type ClassNames = Partial<{
+  label: string;
+  button: string;
+}>;
+
+type Props = {
+  classNames?: ClassNames;
+  text: string;
+  slug: SlugReference;
```
azangru, line 40: What's the benefit of changing the field type from `slug: string` to `slug: { slug: string }`?

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
```diff
@@ -28,7 +29,16 @@ import { SlugReference } from './types';
 
 import styles from './HelpPopupButton.scss';
 
-type Props = SlugReference;
+type ClassNames = Partial<{
+  label: string;
+  button: string;
+}>;
```
azangru, line 35: According to the [XD](https://xd.adobe.com/view/56379b1c-f153-4172-a085-6ad119262f49-0364/screen/1be73a31-3fae-42e6-b82f-1f055e464ed5?fullscreen), the button remains the same across all screens of the site, which suggests that there is no need for additional classnames passed through props.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
```diff
@@ -28,7 +29,16 @@ import { SlugReference } from './types';
...
+
+type Props = {
+  classNames?: ClassNames;
+  text: string;
+  slug: SlugReference;
+};
```
azangru, line 41: Let's have a think about the props.

The component seems to exist in two states:
1) Only the button itself is clickable
2) Both the button and the label next to it are clickable

When the label text is grey, like here:



I wouldn't expect it to be clickable, but in this PR, it is.

So I would expect the Props to have a shape that is approximately like this:

```
type Props = {
  slug: string;
  label: string;
  isLabelClickable: boolean;
}
```

If the label is clickable, it will be styled as blue and have a pointer cursor. If it isn't it will be styled as grey and not change the cursor to a pointer.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
imransl: Not sure if there is a benefit. I was thinking that maybe we will need to extend `SlugReference` to pass other details.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
imransl: The grey text in other pages is still clickable though, isn't it? It's just the homepage's help text that is in blue colour as far as I know.

azangru: It is on internal, that's true; but that would violate our general style guide, where grey isn't clickable. It's possible that Andrea had a different opinion on this particular instance of grey, and as a result I made it clickable. Probably should ask her again to be sure.

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
azangru: Let's change it if we discover that we need to, not before 🙂

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
imransl: Got rid of `ClassNames` type and added only `labelClass` to `props`

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
imransl: Changed the props to:

```
type Props = {
  slug: string;
  labelClass?: string;
  text: string;
}
```

imransl: ✅

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
imransl: ✅

src/ensembl/src/shared/components/help-popup/HelpPopupButton.tsx
azangru: Can I also ask to change the name of the `text` property to `label`, to make it obvious that it refers to the same thing as the `labelClass`?

imransl: That's done now ✅

azangru (approved):
Looks good. Time to resolve the conflicts and QA?

imransl:
@azangru Merged dev and fixed the conflicts. Should be ready for QA now.

PR #537, opened 2021-07-28
Use sequence field instead of sequence_checksum in payloads from Thoas

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1160

## Importance
N/A

## Description
Use new sequence payload from thoas. sequence_checksum is deprecated and now using the sequence object which has checksum

## Deployment URL
http://sequence-payload.review.ensembl.org

## Views affected
Where there is sequence download (Entity viewer, Genome browser)

### Other effects
N/A

## Possible complications
None


src/ensembl/src/shared/types/thoas/product.ts
```diff
@@ -42,6 +42,10 @@ export enum ProductType {
   PROTEIN = 'Protein'
 }
 
+export type Sequence = {
+  checksum: string;
+};
```
azangru, line 47: I think `Sequence` deserves its own file. There's nothing in it that is specific to Product.

ens-ma7: Done. I wasn't too sure if it merits its own file since it will just be 3 lines.

src/ensembl/src/shared/components/instant-download/instant-download-fetch/fetchSequenceChecksums.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import { gql } from '@apollo/client';
+import { Sequence } from 'ensemblRoot/src/shared/types/thoas/sequence';
```
imransl, line 18: Separate this import from vendor imports i.e. `gql`?

ens-ma7: Done. @azangru and @imransl can you please review. Thanks

src/ensembl/src/shared/types/thoas/cdna.ts
```diff
@@ -14,9 +14,11 @@
...
  */
 
+import { Sequence } from './sequence';
+
 export type CDNA = {
   start: number;
   end: number;
   length: number;
-  sequence_checksum?: string;
+  sequence?: Sequence;
```
azangru, line 23: Given how we treat the rest of the fields in this type, and in the CDS type, `sequence` should be declared as a mandatory, not an optional field.

Both here and in the CDS type file.

src/ensembl/src/shared/types/thoas/product.ts
```diff
@@ -52,7 +53,7 @@ export type Product = {
   length: number;
   protein_domains: ProteinDomain[];
   external_references: ExternalReference[];
-  sequence_checksum?: string;
+  sequence?: Sequence;
```
ens-ma7, line 56: @azangru what about the sequence here, should it be optional or compulsory as well to keep in line with cds and cdna?

src/ensembl/src/shared/types/thoas/cdna.ts
ens-ma7: Done

src/ensembl/tests/fixtures/entity-viewer/product.ts
```diff
@@ -40,7 +40,10 @@ export const createProduct = (fragment: Partial<Product> = {}): Product => {
     length: length,
     protein_domains: createProteinDomains(length),
     external_references: times(2, () => createExternalReference()),
-    ...fragment
+    ...fragment,
```
azangru, line 43: fragment should be the last, to be able to override any of the above properties

ens-ma7: Done

azangru (approved):
Looks good.

I updated the title of the PR.

PR #538, opened 2021-07-28
Add the start screen for contact us

azangru:
## Description
Add the start screen for the "Contact us" section of the communication overlay. 

## Jira ticket
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1227

## Deployment url
http://add-contact-us-screen.review.ensembl.org


imransl:
@azangru I've pushed the code after adding the contact overview section content to `ContactUs`. Styling doesn't seem to be right, so I'll fix that tomorrow.

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -29,13 +36,63 @@ const ContactUs = () => {
...
+          <p>
+            Each app has a Help icon (top right) that will show
+            context-sensitive help for that view.
+          </p>
+          <p>All Help articles and videos can be found in the Help app.</p>
+        </section>
+        <section>
+          <h4>Mailing lists</h4>
+          <p> We have two public mailing lists: </p>
+          <dl>
```
azangru, line 75: Why do you use the description list tags? Semantically, description lists are intended for things like glossaries, etc (see [html5 spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-dl-element.html))

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -29,13 +36,63 @@ const ContactUs = () => {
...
+            service status updates.
+          </p>
+          <p>
+            Ensembl Blog <BlogIcon className={styles.icon} />{' '}
+            <FacebookIcon className={styles.icon} />{' '}
+            <TwitterIcon className={styles.icon} />
+          </p>
+        </section>
+        <section>
+          <h4>Online Help</h4>
```
azangru, line 65: I'm in general very confused about these heading tags. Couldn't find any guidance on the web for how to think about headings, but... why are they heading 4's?

On the one hand, the purpose of the headings is semantics and we shouldn't be influenced by how they are automatically styled by the browser, because we can easily style them for our specific needs.

On the other hand, I find it hard to use an h1 here.

Still... Why an h4? 😄

imransl: Hmm that's true. I thought these headings don't have a high importance in the whole page. Probably `h2` would be preferable here?

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
imransl: I went by the MDN description of them:

```The <dl> HTML element represents a description list. The element encloses a list of groups of terms (specified using the <dt> element) and descriptions (provided by <dd> elements). Common uses for this element are to implement a glossary or to display metadata (a list of key-value pairs).```

Since we are specifying key-value pairs here, thought `dl` is better than other list types.

Comment (azangru) [reply src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx:75]: Hm. All right then.

File: src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx (hunk 2 above)
Comment (imransl) [reply src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx]: I've changed these heading tags to `h2`

File: src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+              />
+            </dt>
+            <dd>
+              a low-traffic list for release announcements and web status
+              updates
+            </dd>
+            <dt>
+              <ExternalLink
+                linkText="dev"
+                to=""
```
azangru, line 96: to where? 😃

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <p>All Help articles and videos can be found in the Help app.</p>
+        </section>
+        <section>
+          <h2>Mailing lists</h2>
+          <p> We have two public mailing lists: </p>
+          <dl>
+            <dt>
+              <ExternalLink
+                linkText="announce"
+                to="https://www.ensembl.info/category/01-release/"
```
azangru, line 85: This isn't a mailing list though.

We have links to the mailing lists at https://www.ensembl.org/info/about/contact/index.html; don't know if these are what we should be using though.

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <div className={styles.wrapper}>
+      <div>
+        <p>
+          Please contact us if you have a problem with the website or need help
+        </p>
+        <SecondaryButton onClick={() => setShouldShowForm(!shouldShowForm)}>
+          Contact us
+        </SecondaryButton>
+      </div>
+      <div>
```
azangru, line 47: do we even need this div?

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
imransl: I left it blank back then since I didn't know where to link it to. Should've followed this up though. My bad.

imransl: ✅

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
imransl: Corrected the link ✅

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
imransl: Yes we don't need it. Remove it ✅

src/ensembl/src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <p>
+          Each app has a Help icon (top right) that will show context-sensitive
+          help for that view.
+        </p>
+        <p>All Help articles and videos can be found in the Help app.</p>
+      </section>
+      <section>
+        <h2>Mailing lists</h2>
+        <p> We have two public mailing lists: </p>
+        <dl>
```
msmanoj, line 78: I remember we used `dl`, `dt` tags on the track panel but then we changed it later on. Should we continue to use it?

azangru: We discussed it with Imran in an [earlier comment](https://rapid.ensembl.org/index.html).

There's nothing inherently wrong with `dl` etc in itself. HTML tags are used to convey meaning. The meaning of `dl` etc is to represent a list of descriptions. The prime example of such a list is a glossary, where a term (marked with `dt`) is explained in a description (marked with `dd`).

Imran felt that this:



looks like such a description list, where "announce" and "dev" are the terms, and the paragraphs under them are their descriptions. It wouldn't have occurred to me to think about it this way, but I have to agree that it kind of does look like it. So I agreed that the `dl` etc. were used here as they are intended to be used.

msmanoj (approved):

PR #540, opened 2021-07-30
in-app design review

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1251

## Description
Design review of in-app-gene search

## Deployment URL
http://in-app-search-design-review.review.ensembl.org/

## Views affected
In-app gene search

### Other effects
NA

## Possible complications
NA

### DIFF
--- a/src/ensembl/src/shared/components/question-button/QuestionButton.scss
+++ b/src/ensembl/src/shared/components/question-button/QuestionButton.scss
@@ -17,6 +17,8 @@
   width: 16px;
   height: 16px;
   padding: 0;
+  height: 14px;
+  width: 14px;
 }
 
 .inline {
--- a/src/ensembl/src/shared/components/question-button/QuestionButton.scss
+++ b/src/ensembl/src/shared/components/question-button/QuestionButton.scss
@@ -14,8 +14,6 @@
 }
 
 .in-input-field {
-  width: 16px;
-  height: 16px;
   padding: 0;
   height: 14px;
   width: 14px;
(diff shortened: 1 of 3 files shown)


ens-ma7:
The review URL isn't working.

src/ensembl/src/shared/components/question-button/QuestionButton.scss
```diff
@@ -17,6 +17,8 @@
   width: 16px;
   height: 16px;
   padding: 0;
+  height: 14px;
```
ens-ma7, line 20: Looks like you didnt remove/overwrite, the width and height above. width and height are specified twice :)

ens-ma7 (commented):
Ignore this. I press the wrong button

src/ensembl/src/shared/components/question-button/QuestionButton.scss
ens-ma7: sorry @jyothishnt just realised that the default height and width is 14px so we dont need to specify the width and height here.  do you agree?

ens-ma7: sorry ignore the above comment, i was looking at the wrong question icon but i notice that you can specify just the height or the width to 14px and it works, you dont have to specify both but probably best to specify both to be safe.

ens-ma7 (approved):

ens-ma7:
@jyothishnt i have approved the code, it looks fine to me. I leave setting the width and height for you to decide. either way is working fine.

azangru (approved):

PR #541, opened 2021-08-02
Add contact us form

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1228

## Description
- Add the initial version of the contact us form to the communication panel.
- Only can submit if all fields are filled in
- Allows attachment of multiple files
- The combined size of attached files cannot exceed 10MB
- Attached files can be deleted
- Upon submission, shows a success screen
- Default styles of the Upload component updated (made it larger by default, according to Andrea's instructions)

## Left to be done
- Add the Messages and Contact us elements (tabs) to the header of the popup
- Disable the Upload component if one of the fields of the form is empty
- Prevent body scroll when the popup is open
- Handle file upload cancellation (currently, this gets registered as onChange event, but passes an `undefined` to the event handler)
- Panel animation on unmount
- Close the panel upon a press on the close button
- The whole state management thing

## Deployment URL
http://contact-us-form.review.ensembl.org


src/ensembl/package.json
```diff
@@ -45,6 +45,7 @@
     "dotenv": "10.0.0",
     "ensembl-genome-browser": "https://raw.githubusercontent.com/Ensembl/ensembl-genome-browser-assets/master/assets-80f51620ed443c640cdfd6b5aebd505b.tar.gz",
     "express": "4.17.1",
+    "filesize": "7.0.0",
```
azangru, line 48: This is a small library for formatting file size into a human-readable string. Looks more robust than what I would have written myself.

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
```diff
@@ -0,0 +1,247 @@
...
+    areMandatoryFieldsFilled(formState) &&
+    !exceedsAttachmentsSizeLimit(formState)
+  );
+};
+
+const areMandatoryFieldsFilled = (formState: State) => {
+  return (['name', 'email', 'subject', 'message'] as const).every(
+    (field) => formState[field]
+  );
+};
```
imransl, line 235: Don't we also need to validate if the `email` is valid or not?

msmanoj:
Steps to reproduce the following error:

> TypeError: Cannot read property 'size' of undefined

- Click on the upload button
- Select a file to upload
- Click on the upload button again
- Click cancel

azangru:
> TypeError: Cannot read property 'size' of undefined

@msmanoj yes, I know; it's an error in the Upload component 😄 . I've fixed it in the newer branch:

http://contact-us-form-contd.review.ensembl.org

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
azangru: Not sure yet; also not sure how we will be showing this in the UI. Let's ignore this for now.

src/ensembl/src/shared/components/uploaded-file/UploadedFile.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.size {
+  flex: 0 0 auto;
+  white-space: nowrap;
+  font-weight: $light;
+}
+
+.delete {
+  flex: 0 0 auto;
+  margin-left: 6px;
```
msmanoj, line 26: Please check if this margin is required. PR looks good otherwise.

azangru: In Andrea's designs, the space between the file size and the close button is visibly larger than between the file size and the file name. I am not sure whether gap + margin-left, like I did here, is any better than just explicitly setting left and right margin on the fileSize element

azangru: I checked with Andrea and updated the distances in this element. Changed the gap + margin-left combo to margin-left + margin-right on the fileSize element.

PR #549, opened 2021-08-10
Add submit slider and email validation

azangru:
## Type
- New feature

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1258
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1269

## Description
- Add the slider as a challenge for form submission
- Add email validation using browser's inbuilt functionality

## Deployment URL
http://submit-slider.review.ensembl.org

### DIFF
--- a/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
+++ b/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
@@ -21,6 +21,7 @@ import React, {
   useCallback,
   useRef
 } from 'react';
+import classNames from 'classnames';
 import noop from 'lodash/noop';
 
 import { submitForm } from '../submitForm';
@@ -276,11 +277,17 @@ const ContactUsInitialForm = () => {
               isDisabled={!isFormValid}
               onClick={noop}
             >
-              Submit
+              Send
             </PrimaryButton>
           ) : (
             <>
-              <span className={commonStyles.sliderLabel}>Slide, then send</span>
+              <span
+                className={classNames(commonStyles.sliderLabel, {
+                  [commonStyles.sliderLabelDisabled]: !isFormValid
+                })}
+              >
+                Slide, then send
+              </span>
               <SubmitSlider
                 className={commonStyles.submitSlider}
                 isDisabled={!isFormValid}
--- a/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
+++ b/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
-import React, { useState, useReducer, useCallback, useRef } from 'react';
+import React, {
+  useState,
+  useEffect,
+  useReducer,
+  useCallback,
+  useRef
+} from 'react';
 import noop from 'lodash/noop';
 
 import { submitForm } from '../submitForm';
@@ -25,6 +31,7 @@ import ShadedTextarea from 'src/shared/components/textarea/ShadedTextarea';
 import Upload from 'src/shared/components/upload/Upload';
 import UploadedFile from 'src/shared/components/uploaded-file/UploadedFile';
 import { PrimaryButton } from 'src/shared/components/button/Button';
+import SubmitSlider from '../submit-slider/SubmitSlider';
 
 import commonStyles from '../ContactUsForm.scss';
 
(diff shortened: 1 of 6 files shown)


imransl:
@azangru Noticed two things while reviewing this PR:

1. Email validation passes when there is no TLD e.g. jack_ryan@email
2. The error message is not aligning horizontally with the send button

image

azangru:
@imransl We discussed both of these points during the code review:

> Email validation passes when there is no TLD e.g. jack_ryan@email

I'm relying on the browser's native method for validating emails. Apparently, email addresses without the top-level domain are valid 🤷‍♂️ See the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#basic_validation).

> The error message is not aligning horizontally with the send button

We now have the slider that takes most of the available width, don't we. So the old space to left of what used to be a submit button is now occupied by the slider and its label. I understand that it is possible to slide the slider, see the "Send" button, and then bring the form in an invalid state, in which case the error message will still be displayed above the button; but am hesitant to make any decision on whether it is still the right place for the error messages or whether they should be displayed to the left of the button. I think we should wait for Andrea's decision on that.

imransl:
> I'm relying on the browser's native method for validating emails. Apparently, email addresses without the top-level domain are valid 🤷‍♂️ See the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#basic_validation).

I dread what kind of email inputs we get because of this. But if the common consensus is that this isn't a big issue, then so be it.

> We now have the slider that takes up most of the available width, don't we. So the old space to the left of what used to be a submit button is now occupied by the slider and its label. I understand that it is possible to slide the slider, see the "Send" button, and then bring the form in an invalid state, in which case the error message will still be displayed above the button; but am hesitant to make any decision on whether it is still the right place for the error messages or whether they should be displayed to the left of the button. I think we should wait for Andrea's decision on that.

Yes, agree that we should wait for Andrea's decision.

imransl (approved):

ens-ma7:
@azangru i manage to break the upload i think. steps to reproduce:
* Go to the contact us form (from any page, i have tried homepage)
* fill in the form
* upload a file
* then remove the file
* try uploading the same file again, it doesnt upload the file
* But if you try a file with a different file name, it works

azangru:
@ens-ma7 great, bug confirmed, thanks!

azangru:
@ens-ma7 I pushed the fix for the bug you noticed.

ens-ma7 (approved):

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
```diff
@@ -131,6 +162,12 @@ const ContactUsInitialForm = () => {
     }
   }, []);
 
+  const validateEmail = useCallback(() => {
+    if (emailFieldRef.current) {
+      setEmailFieldValid(emailFieldRef.current?.checkValidity());
```
jyothishnt, line 167: wonder why a missing [dot] is considered valid

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submit-slider/SubmitSlider.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const recordInitialPointerX = (event: TouchEvent | MouseEvent) => {
+    initialPointerX.current = getCurrentPointerX(event);
+  };
+
+  const getDistance = (event: TouchEvent | MouseEvent) => {
+    const currentPointerX = getCurrentPointerX(event);
+    const trackRect = sliderRectsRef.current?.trackRect as DOMRect;
+    const sliderRect = sliderRectsRef.current?.sliderRect as DOMRect;
+    const minValue = 0;
+    const maxValue = trackRect.width - sliderRect.width;
```
jyothishnt, line 188: Should we really check if the user have dragged to the far end?  Isn't it better to automatically complete the slide once the user drag reaches, say 75% for a better user experience?

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
```diff
@@ -209,15 +255,46 @@ const ContactUsInitialForm = () => {
         </div>
 
         <div className={commonStyles.submit}>
-          {exceedsAttachmentsSizeLimit(state) && (
-            <span className={commonStyles.errorText}>
-              Attachment(s) exceed 10 MB
-            </span>
+          {!isFormValid && (
+            <div className={commonStyles.formErrors}>
```
jyothishnt, line 259: The div that shows the error is there in the DOM until the form is enabled.
When the form is enabled this is removed and you will see the bottom div moving upwards due to the margin on this div (12px) https://github.com/Ensembl/ensembl-client/pull/549/files#diff-1d4823aa7281cf982010dd4c2118669983f65c0ec6a2c85f24f5875a4aabdd75R70

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
jyothishnt: Decided to stay with basic browser validation

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
azangru: I've updated the styles. The vertical shift of the slide button now only happens when there is an actual error to display above it. It's quite likely that Andrea will update the design to deal with the errors.

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submit-slider/SubmitSlider.tsx
azangru: Decided that a slide all the way to the right margin is good for now; will hear from Andrea and Lucy if they want it to be changed.

PR #550, opened 2021-08-10
Use related articles for About pages

azangru:
## Type
- New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1270

## Description
If an article among the About pages has a list of related articles, show this list in the right-hand aside column. If it doesn't, show the second-level menu instead.

## Deployment URL
http://related-articles-in-about.review.ensembl.org

### DIFF
--- a/src/ensembl/src/content/app/about/About.tsx
+++ b/src/ensembl/src/content/app/about/About.tsx
@@ -21,6 +21,7 @@ import useApiService from 'src/shared/hooks/useApiService';
 
 import {
   TextArticle,
+  RelatedArticles,
   HelpArticleGrid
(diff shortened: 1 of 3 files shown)


ens-ap5 (changes requested):
Is it possible to list the related articles in the order they're given in the YAML? In the review sandbox, they look to be sorted alphabetically by URL.

azangru:
Thanks for noticing. It was a problem in the help&docs api; fixed in this commit: https://github.com/Ensembl/ensembl-website-help-and-docs/commit/ea2f2960a8e33b332875eb4ad20c9d8b65cdfaad

PR #551, opened 2021-08-12
Submit form

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1277

## Description
- Fetch a token for form submission
- Submit the form
- Show the spinner while the form is being submitted

## Deployment URL
http://submit-form.review.ensembl.org

### DIFF
--- a/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submitForm.ts
+++ b/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submitForm.ts
@@ -33,6 +33,18 @@ export const submitForm = async (form: Form) => {
       Authorization: `Bearer ${token}`
     },
     body: formData
+  }).then(async (response) => {
+    let responseMessage;
+    try {
+      responseMessage = await response.json();
+    } catch {
+      throw new Error();
+    }
+    if (!response.ok) {
+      throw new Error(responseMessage);
+    } else {
+      return responseMessage;
+    }
   });
 };
 
--- a/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submitForm.ts
+++ b/src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submitForm.ts
@@ -18,9 +18,22 @@ type FormFieldType = string | File | File[] | null;
 
 type Form = Record<string, FormFieldType>;
 
-export const submitForm = (form: Form) => {
+type TokenSuccessResponse = {
+  token: string;
+};
+
+export const submitForm = async (form: Form) => {
   const formData = buildFormData(form);
-  console.log('formData', formData); // eslint-disable-line no-console
+  const url = '/api/support/ticket';
+  const token = await getToken();
+
+  return fetch(url, {
+    method: 'POST',
+    headers: {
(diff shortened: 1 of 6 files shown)


src/ensembl/src/shared/components/loading-button/ControlledLoadingButton.tsx
```diff
@@ -27,24 +27,32 @@ import { LoadingState } from 'src/shared/types/loading-state';
 
 import styles from './LoadingButton.scss';
 
-type Props = {
+export type Props = {
```
msmanoj, line 30: Why do we need to export this here?

ens-ma7:
I have submitted a message with some special characters to see if it works and i can confirm it works (i can see the special characters in the RT ticket). I also had a look at the code and i couldn't find anything wrong but its best to leave someone else to review/approve the code.

ens-ma7 (approved):

ens-ma7:
I have submitted the form on the help app but it doesnt seem to have submitted the message. The response from the server is `{success:false}` not sure what this means but i also check RT and there was no ticket created.

image

azangru:
We talked with Kamal and agreed that responses with status code 201 and body `{ success: false }` are not legitimate. He will update the backend to use the 500 status code when there are server-side problems in the api.

src/ensembl/src/shared/components/loading-button/ControlledLoadingButton.tsx
azangru: I was trying to derive the props of the `LoadingButton` from the props of `ControlledLoadingButton`, but then gave up on the idea. This was the remnant of that attempt. I've removed the export now.

imransl:
@ens-ma7 @azangru  I just tested the contact form now and the status code returned is 201 with the response body `{ success: false }`

azangru:
I still strongly believe that this is something that needs to be fixed server-side. The server shouldn't respond to the client with illegitimate payloads.

imransl:
@azangru Sorry for the confusion. The response body actually returned as `{success:true}`.

src/ensembl/src/shared/components/communication-framework/contact-us/contact-us-form/submitForm.ts
```diff
@@ -18,9 +18,34 @@ type FormFieldType = string | File | File[] | null;
...
 
-export const submitForm = (form: Form) => {
+type TokenSuccessResponse = {
+  token: string;
+};
+
+export const submitForm = async (form: Form) => {
   const formData = buildFormData(form);
-  console.log('formData', formData); // eslint-disable-line no-console
+  const url = '/api/support/ticket';
```
imransl, line 27: Shouldn't we specify the base URL `/api/support` as a constant to reuse it?

azangru: 🤷‍♂️ We could, but at the moment, this api is used only in two functions, and both these functions are in the same file. I don't know if extracting `/api/support` into a constant is going to help much.

msmanoj (approved):

PR #556, opened 2021-08-19
Show the conversation icon in production

azangru:
## Description
In order to be able to access the Contact Us form, the user must click on the conversation icon. We have been hiding it in production environment until now. Time to make it visible.

_(not including a review url, because you wouldn't be able to see the change in the review branch anyway)_


ens-ma7 (approved):

ens-ma7:
I have approved the code change but are we sure the contact us form is ready for use. The state part isn't live yet. Do we not want to wait for the state part to be merged in order to make it visible?

azangru:
> The state part isn't live yet. Do we not want to wait for the state part to be merged in order to make it visible?

I don't think saving the form state is a blocker; the form is quite functional even without that.

PR #552, opened 2021-08-12
Replace using genome search endpoints for track panel with thoas data

imransl:
## Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement to existing feature

## Related JIRA Issue(s)
[ENSWBSITES-1256](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1256)
[ENSWBSITES-1191](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1191)

## Description
There are two tickets covered by this PR.

- ENSWBSITES-1256 replaces the genome search REST endpoints used to retrieve data needed for the track panel list. Thoas data is now used for this purpose.
- ENSWBSITES-1191 shows the MANE Select/Ensembl canonical transcript as thoas supports this

## Deployment URL
http://track-panel-thoas.review.ensembl.org/

## Views affected
Genome browser -> track panel

PR #563, opened 2021-08-23
Use RTK-query to get gene data for track panel

azangru:
## Description
Proposed as an alternative for https://github.com/Ensembl/ensembl-client/pull/552

- Uses redux-toolkit query rather than Apollo client to fetch gene data from Thoas
- The result of the query is currently used to display the tracks for the gene and its transcript; but there are other components on the genome browser page that can use this query (specifically, the top bar with a brief gene summary).

## Questions
- Should the query for the gene fetch the extra information that will appear in the drawer, or will we send a dedicated request when the user opens the drawer
- What is the best name for the typescript type that describes the gene data that has been fetched? I've called it `TrackPanelGene`, but I don't particularly like this name, because the fetched data will be used in several components, not just in the track panel.

## Future steps (in subsequent PRs)
- Remove the `bio_type` and the  `strand` fields from following the `EnsObjectGene` type
- Review and simplify `EnsObjectTrack` type
- Figure out how to properly pass information about transcript from the track panel to the `TranscriptSummary` component in the drawer (we are currently losing this information)
- Possibly rework the props of the `TrackPanelListItem` component to move from biological field names defined in the `track` prop to presentational field names, so that this component can support different track types
- Remove data fetching via Apollo client and move to RTK-query

## Deployment URL
http://track-panel-alternative.review.ensembl.org/

## Views affected
Genome browser

### DIFF
--- a/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
+++ b/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
@@ -232,22 +232,31 @@ const TranscriptSummary = () => {
         <div className={styles.label}>Transcript</div>
         <div className={styles.value}>
           <div className={styles.featureDetails}>
-            <span className={styles.featureSymbol}>{stableId}</span>
-            <span className={styles.label}>
-              <TranscriptQualityLabel metadata={metadata} />
-            </span>
+            <div className={styles.featureDetail}>
+              <span className={styles.featureSymbol}>{stableId}</span>
+              <div className={styles.transcriptQuality}>
+                <TranscriptQualityLabel metadata={metadata} />
+              </div>
+            </div>
+
             {metadata.biotype && (
-              <>
+              <div className={styles.featureDetail}>
                 <span>{metadata.biotype.label}</span>
                 <div className={styles.questionButton}>
                   <QuestionButton helpText={metadata.biotype.definition} />
                 </div>
-              </>
+              </div>
             )}
             {transcript.slice.strand.code && (
-              <span>{getStrandDisplayName(transcript.slice.strand.code)}</span>
+              <div className={styles.featureDetail}>
+                <span>
+                  {getStrandDisplayName(transcript.slice.strand.code)}
+                </span>
+              </div>
             )}
-            <span>{getFormattedLocation(ensObjectGene.location)}</span>
+            <div className={styles.featureDetail}>
+              <span>{getFormattedLocation(ensObjectGene.location)}</span>
+            </div>
           </div>
         </div>
       </div>
@@ -330,6 +339,7 @@ const TranscriptSummary = () => {
                 }}
                 gene={{ id: gene.unversioned_stable_id }}
                 theme="light"
+                layout="vertical"
               />
               <CloseButton
                 className={styles.closeButton}
--- a/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
+++ b/src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
@@ -25,7 +25,6 @@ import {
   getDisplayStableId,
   buildFocusIdForUrl
 } from 'src/shared/state/ens-object/ensObjectHelpers';
-import { getBrowserActiveEnsObject } from 'src/content/app/browser/browserSelectors';
 import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
 import { getGeneName } from 'src/shared/helpers/formatters/geneFormatter';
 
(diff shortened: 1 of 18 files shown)


src/ensembl/src/shared/state/ens-object/ensObjectActions.ts
```diff
@@ -40,84 +40,99 @@ export const fetchEnsObjectAsyncActions = createAsyncAction(
...
+  async (dispatch, getState: () => RootState) => {
+    if (typeof payload === 'string') {
+      payload = parseEnsObjectId(payload);
+    }
+    const state = getState();
+    const ensObjectId = buildEnsObjectId(payload);
+    const ensObjectLoadingStatus = getEnsObjectLoadingStatus(
+      state,
+      ensObjectId
     );
```
imransl, line 56: Prettier again!

imransl:
@azangru I have a doubt after reviewing this PR. Aren't we going to rename `EnsObject` and/or move it to Genome Browser directory?

azangru:
> Aren't we going to rename EnsObject and/or move it to Genome Browser directory?

Not in this PR! I'd suggest moving towards the goal in multiple small steps (some of which are outlined in the `Future steps` section in the description for this PR).

src/ensembl/src/content/app/browser/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -156,21 +158,27 @@ const GENE_AND_TRANSCRIPT_QUERY = gql`
 
 const TranscriptSummary = () => {
   const ensObjectGene = useSelector(getBrowserActiveEnsObject) as EnsObjectGene;
+  const [shouldShowDownload, showDownload] = useState(false);
 
-  const transcriptTrack = ensObjectGene?.track?.child_tracks?.[0];
+  const { data: geneData } = useGetTrackPanelGeneQuery({
```
msmanoj, line 163: Would it be better to name it something like `useGetGeneDataForTrackPanel` instead?

src/ensembl/src/store.ts
```diff
@@ -35,18 +37,19 @@ const rootReducer = createRootReducer(history);
 export type RootState = StateType<typeof rootReducer>;
 
 const middleware = [
-  ...getDefaultMiddleware(),
   routerMiddleware(history),
   epicMiddleware,
-  analyticsMiddleWare
+  analyticsMiddleWare,
+  genomeBrowserApiSlice.middleware
```
msmanoj, line 43: Should we considering [code splitting](https://redux-toolkit.js.org/rtk-query/usage/code-splitting) in this initial PR so that we can avoid adding individual middleware for each app?

src/ensembl/src/shared/state/ens-object/ensObjectActions.ts
```diff
@@ -40,84 +40,99 @@ export const fetchEnsObjectAsyncActions = createAsyncAction(
...
-    dispatch(fetchEnsObject({ genomeId, type, objectId: id }));
-  });
-};
+    exampleFocusObjects.forEach(({ id, type }) => {
+      dispatch(fetchEnsObject({ genomeId, type, objectId: id }));
+    });
+  };
+
+// FIXME: many fields here are unnecessary for an ensObject
+const buildGeneObject = (params: any): EnsObjectGene => {
```
msmanoj, line 114: Does it have to be `any` here?

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelListItem.test.tsx
```diff
@@ -135,7 +135,14 @@ describe('<TrackPanelListItem />', () => {
     });
 
     it('toggles the expanded/collapsed state of the track when clicked on the expand button', async () => {
-      const { container } = wrapInRedux();
+      store = mockStore(mockState);
+      const { container } = render(
+        <Provider store={store}>
+          <TrackPanelListItem {...defaultProps}>
+            <TrackPanelListItem {...defaultProps} />
+          </TrackPanelListItem>
```
msmanoj, line 143: `TrackPanelListItem` nested within `TrackPanelListItem` doesn't look right here. Please check.

azangru: Yep, @imran is right; this is how this PR is handling transcript tracks inside of the gene track.

src/ensembl/src/store.ts
azangru: Code splitting is nice; but in the docs they are setting the `baseQuery` parameter, which defines both the root of the api endpoints and the protocol used to talk to these endpoints, in a single field in the empty api slice. We could probably hack around that, but it feels like a separate area of exploration.

src/ensembl/src/shared/state/ens-object/ensObjectActions.ts
azangru: Added the type definition.

src/ensembl/src/shared/state/ens-object/ensObjectActions.ts
```diff
@@ -40,84 +40,99 @@ export const fetchEnsObjectAsyncActions = createAsyncAction(
...
+    genome_id: params.genomeId,
+    label: params.gene.symbol,
+    location: {
+      chromosome,
+      start,
+      end
+    },
+    stable_id: params.gene.unversioned_stable_id,
+    versioned_stable_id: params.gene.stable_id,
+    bio_type: '',
```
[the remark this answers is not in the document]
azangru: Eventually, ensObject should not store biotype data; the component that needs it should get it directly from the fetched gene. But since I have left this change for a subsequent PR, I have added the biotype back to the ens object for the time being.

msmanoj (approved):

imransl (approved):

PR #564, opened 2021-08-24
Fix invalid query param breaking Entity Viewer

imransl:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1274](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1274)

## Description
When we substitute the `view` query parameter in the URL to access Entity viewer, the user gets redirected to the broken error page. However, we should be instead redirect the user to the default transcripts view in Entity Viewer. This PR fixes that.

## Deployment URL
http://ev-query-param.review.ensembl.org/

## Views affected
Entity Viewer

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -345,7 +345,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
 };
 
 const ensureGeneViewExists = (view: string) =>
-  !!Object.values(View).some((value) => value === view);
+  Object.values(View).some((value) => value === view);
 
 const useGeneViewRouting = () => {
   const dispatch = useDispatch();
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -345,7 +345,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
 };
 
 const ensureGeneViewExists = (view: string) =>
-  !!Object.values(View).find((value) => value === view);
+  !!Object.values(View).some((value) => value === view);
 
 const useGeneViewRouting = () => {
   const dispatch = useDispatch();
--- a/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -344,6 +344,9 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
   );
 };
 
+const ensureGeneViewExists = (view: string) =>
+  !!Object.values(View).find((value) => value === view);
+
 const useGeneViewRouting = () => {
   const dispatch = useDispatch();
   const params: { [key: string]: string } = useParams();
@@ -361,7 +364,7 @@ const useGeneViewRouting = () => {
 
   useEffect(() => {
     if (view && viewInRedux !== view) {
-      if (view in View) {
+      if (ensureGeneViewExists(view)) {
         dispatch(updateView(view as View));
       } else {
         const url = urlFor.entityViewer({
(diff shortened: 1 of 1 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -361,7 +361,17 @@ const useGeneViewRouting = () => {
 
   useEffect(() => {
     if (view && viewInRedux !== view) {
-      dispatch(updateView(view as View));
+      if (view in View) {
```
azangru, line 364: I've searched on StackOverflow for how this is supposed to work, and the [answer](https://stackoverflow.com/a/43805090/3925302) is confusing. Are we sure we even want an enum here? Would just a simple [array of strings](https://thoughtbot.com/blog/the-trouble-with-typescript-enums#another-option-since-typescript-34) work better for us? Just curious.

imransl: Yes agree that was overkill and also it wasn't working properly. I've changed this in df97ef9

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -344,6 +344,9 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
   );
 };
 
+const ensureGeneViewExists = (view: string) =>
+  !!Object.values(View).find((value) => value === view);
```
azangru, line 348: `!Object.values(View).find` is identical to `.some`

imransl: I don't know how I ended up with `find` when I actually wanted to use `some`. Thanks for pointing that out 😀

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -344,6 +344,9 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
   );
 };
 
+const ensureGeneViewExists = (view: string) =>
+  !!Object.values(View).some((value) => value === view);
```
azangru, line 348: `.some` returns a boolean. No need to typecast it to boolean again.

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -344,6 +344,9 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
   );
 };
 
+const ensureGeneViewExists = (view: string) =>
```
azangru, line 347: `ensureGeneViewExists` -> `isViewParameterValid`

(the answer to the question in the name of this function would be yes/no, which corresponds to a boolean)

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -361,7 +364,17 @@ const useGeneViewRouting = () => {
 
   useEffect(() => {
     if (view && viewInRedux !== view) {
-      dispatch(updateView(view as View));
+      if (ensureGeneViewExists(view)) {
```
azangru, line 367: Let's try to get rid of the nested conditionals.

```ts
if (view && isViewParameterValid(valid) && viewInRedux !== view) {
  dispatch(updateView(view as View));
} else {
  // I believe the code as it stands will already do the right thing,
  // because initially the view in redux is "transcripts"
}
```

imransl: ✅

src/ensembl/src/content/app/entity-viewer/gene-view/GeneView.tsx
imransl: ✅

azangru (approved):

PR #566, opened 2021-08-25
new general error screen

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1216

## Description
This will add new general error screen with options to reload or reset the page.

## Deployment URL
http://error-screen.review.ensembl.org/
Url with wrong view param that can display the new error page 
http://error-screen.review.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000125952?view=transcriptssdf

## Views affected
Could be any view

### Other effects
NA

## Possible complications
Hopefully nothing

### DIFF
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -67,7 +67,6 @@
     display: flex;
     width: 195px;
     height: 195px;
-    display: flex;
     align-items: center;
     border-radius: 50%;
     background-color: rgba(183, 192, 200, 0.24);
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -1,11 +1,5 @@
 @import 'src/styles/common';
 
-$generalErrorImageWidth: 170px;
-$generalErrorImageGap: 60px;
-$generalErrorImagesTotalWidth: calc(
-  3 * #{$generalErrorImageWidth} + 2 * #{$generalErrorImageGap}
-);
-
 .newTechErrorWrapper {
   padding-top: 10vh;
   display: flex;
@@ -56,7 +50,7 @@ $generalErrorImagesTotalWidth: calc(
 }
 
 .reloadButton {
-  padding-top: 20px;
+  padding-top: 40px;
 }
 
 .moreOptions {
@@ -123,23 +117,5 @@ $generalErrorImagesTotalWidth: calc(
   img {
     width: 170px;
     height: 170px;
-  }
-
-  @media screen and (min-width: $generalErrorImagesTotalWidth) {
-    // margin-top: 10vh;
-    margin-bottom: calc(#{$generalErrorImageGap} / 2);
-
-    img:not(:last-child) {
-      margin-right: $generalErrorImageGap;
-    }
-  }
-
-  @media screen and (max-width: $generalErrorImagesTotalWidth) {
-    display: inline-flex;
-    flex-direction: column;
-
-    img:not(:last-child) {
-      margin-bottom: calc(#{$generalErrorImageGap} / 2);
-    }
   }
 }
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -67,47 +67,38 @@
     display: flex;
     width: 195px;
     height: 195px;
+    display: flex;
+    align-items: center;
+    border-radius: 50%;
+    background-color: rgba(183, 192, 200, 0.24);
+    position: relative;
+  }
 
-    .vennCircleLeft {
-      display: flex;
-      align-items: center;
-      border-radius: 50%;
-      background-color: rgba(183, 192, 200, 0.24);
-      position: relative;
-      left: 32px;
-      text-align: right;
+  .vennCircleLeft {
+    left: 32px;
+    text-align: right;
 
-      span {
-        margin-right: 83px;
-        margin-left: 30px;
-      }
+    span {
+      margin-right: 83px;
+      margin-left: 30px;
     }
-    .vennCircleRight {
-      display: flex;
-      align-items: center;
-      border-radius: 50%;
-      background-color: rgba(183, 192, 200, 0.24);
-      position: relative;
-      right: 32px;
-      text-align: left;
-      span {
-        margin-right: 15px;
-        margin-left: 30px;
-      }
-    }
-    .infoIcon {
-      padding-left: 19px;
-      svg {
-        fill: $red;
-        width: 24px;
-        height: 24px;
-      }
+  }
+
+  .vennCircleRight {
+    right: 32px;
+    text-align: left;
+    span {
+      margin-right: 15px;
+      margin-left: 30px;
     }
   }
   
-
-  img {
-    width: 170px;
-    height: 170px;
+  .infoIcon {
+    padding-left: 19px;
+    svg {
+      fill: $red;
+      width: 24px;
+      height: 24px;
+    }
   }
 }
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -36,23 +36,97 @@ $generalErrorImagesTotalWidth: calc(
   flex: 0 0 auto;
 }
 
+.generalErrorTopMessage {
+  font-weight: bold;
+  font-size: 13px;
+}
+
 .generalErrorBody {
   flex-grow: 1;
   display: flex;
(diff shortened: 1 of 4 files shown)


src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,92 @@ $generalErrorImagesTotalWidth: calc(
...
+.resetOption {
+  color: $red;
+  button {
+    margin-top: 24px;
+  }
+}
+.generalErrorImage {
+  display: flex;
+  font-size: 12px;
+  .vennCircle1 {
```
azangru, line 75: In XD, text alignment is right in the left circle and left in the left circle.

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
```diff
@@ -14,34 +14,93 @@
...
+              direction={chevronDirection()}
+              animate={true}
+              onClick={toggleChevron}
+            ></Chevron>
+          </div>
+          {moreOptionExpanded && (
+            <div className={styles.resetOption}>
+              <div>Unfortunately we need to reset everything</div>
+              <div>
+                All your species, configuration if views &amp; history will be
```
azangru, line 92: 1) `if` -> `of`

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,92 @@ $generalErrorImagesTotalWidth: calc(
   flex: 0 0 auto;
 }
 
+.generalErrorTopMessage {
+  font-weight: bold;
+  font-size: 13px;
```
azangru, line 35: this is the default font size; so this rule isn't doing anything

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
```diff
@@ -14,34 +14,93 @@
...
+          </div>
+        </div>
+        <div className={styles.reloadButton}>
+          <PrimaryButton isDisabled={moreOptionExpanded} onClick={reloadPage}>
+            Reload
+          </PrimaryButton>
+        </div>
+        <div className={styles.generalErrorBottomMessage}>
+          <div className={styles.moreOptions}>
+            <a onClick={toggleChevron}>If the site still fails to load</a>
```
azangru, line 81: We have a special component for this job: https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/shared/components/show-hide/ShowHide.tsx

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
```diff
@@ -14,34 +14,93 @@
...
+            ></Chevron>
+          </div>
+          {moreOptionExpanded && (
+            <div className={styles.resetOption}>
+              <div>Unfortunately we need to reset everything</div>
+              <div>
+                All your species, configuration if views &amp; history will be
+                lost
+              </div>
+              <PrimaryButton isDisabled={false} onClick={resetPage}>
```
azangru, line 95: `isDisabled` must be a remnant of copy-pasting. It isn't needed.

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
```diff
@@ -14,34 +14,93 @@
...
+  };
+
+  const reloadPage = () => {
+    window.location.reload();
+  };
+
+  const resetPage = () => {
+    storageService.clearAll();
+    dispatch(replace(urlFor.home()));
+    reloadPage();
```
azangru, line 53: wouldn't `window.location.href = urlFor.home()` do the work of both above lines at once?

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,92 @@ $generalErrorImagesTotalWidth: calc(
...
+  .vennCircle2 {
+    width: 195px;
+    height: 195px;
+    border-radius: 50%;
+    background-color: rgba(183, 192, 200, 0.24);
+    right: 32px;
+    position: relative;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
azangru, line 100: most of the styles for the two vennCircle classes are identical. Better move repeated styles into a common class and add two classes (vennCircleLeft, vennCircleRight) specifically for the styles that are different

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,92 @@ $generalErrorImagesTotalWidth: calc(
...
-    margin-top: 10vh;
-    margin-bottom: calc(#{$generalErrorImageGap} / 2);
+.reloadButton {
+  padding-top: 40px;
+}
+
+.moreOptions {
+  padding: 50px 0 20px 0;
+  a {
+    padding: 0 10px;
```
imransl, line 59: Can't we use `margin-right: 10px` instead for better centering?

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
jyothishnt: Loll its not from copy-pasting atleast this time. :-)  It was shown on the props (though optional). I thought to add it.  I will remove it

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
jyothishnt: yes it will do, but @imransl suggested to use replace(). Wondering which is better

azangru: `replace` is probably fine (I've never used it, so don't know); but the way the code is written here what gets called is `window.location.reload()`, which looks excessive. `window.location.replace(urlFor.home())` would probably be sufficient?

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,86 @@ $generalErrorImagesTotalWidth: calc(
...
+.generalErrorTopMessage {
+  font-weight: bold;
+}
+
 .generalErrorBody {
   flex-grow: 1;
   display: flex;
   flex-direction: column;
-  justify-content: center;
+  justify-content: flex-start;
```
azangru, line 41: Is there any point in keeping `flex-grow: 1;` and `justify-content: flex-start;` at all anymore? The element doesn't seem to grow to the full length of the flex container, nor does it seem to need to:

image

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,86 @@ $generalErrorImagesTotalWidth: calc(
...
+        margin-right: 83px;
+        margin-left: 30px;
+      }
+    }
+    .vennCircleRight {
+      display: flex;
+      align-items: center;
+      border-radius: 50%;
+      background-color: rgba(183, 192, 200, 0.24);
+      position: relative;
```
azangru, line 92: there are still lots of CSS rules in common between `.vennCircleRight ` and `.vennCircleLeft`; they should probably go in `.vennCircle`

jyothishnt: Not sure about this. I have spent some time to play with it but I couldn't move any of these. May be I am wrong, have you tried it?

src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -36,36 +30,75 @@ $generalErrorImagesTotalWidth: calc(
...
+}
+.generalErrorImage {
+  display: flex;
+  font-size: 12px;
+
+  .vennCircle {
+    display: flex;
+    width: 195px;
+    height: 195px;
+    display: flex;
```
azangru, line 70: duplicated `display: flex` line (cf. line 67)

src/ensembl/src/shared/components/error-screen/GeneralErrorScreen.tsx
```diff
@@ -14,34 +14,84 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import storageService from 'ensemblRoot/src/services/storage-service';
```
azangru, line 17: - 🙅‍♂️ `ensemblRoot`
- we put npm imports (react, classNames) first

jyothishnt: Oh no! again!
Btw I have already reordered the imports :-)

azangru (approved):

PR #570, opened 2021-08-26
Update SearchField styles

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1176

## Description
- Moved the box-shadow to the SearchField component.
- Used the same box-shadow and height as for the ShadedInput component

## Deployment URL
http://search-field-styles.review.ensembl.org/

## Views affected
- Species Selector
- Interstitial screen for Genome Browser and Entity Viewer
- InAppSearch in the sidebar modal for Genome Browser and Entity Viewer

### DIFF
--- a/src/ensembl/src/shared/components/search-field/SearchField.scss
+++ b/src/ensembl/src/shared/components/search-field/SearchField.scss
@@ -2,8 +2,6 @@
 
 .searchField {
   box-shadow: $form-field-shadow;
-  border: none;
-  padding: 0 15px;
   height: 30px;
   display: inline-flex;
   padding: 0 12px;
--- a/src/ensembl/src/shared/components/search-field/SearchField.scss
+++ b/src/ensembl/src/shared/components/search-field/SearchField.scss
@@ -1,9 +1,16 @@
 @import 'src/styles/common';
 
 .searchField {
-  border: 1px solid $grey;
+  box-shadow: $form-field-shadow;
+  border: none;
+  padding: 0 15px;
+  height: 30px;
   display: inline-flex;
-  padding: 0 0.6em;
+  padding: 0 12px;
+
+  input {
+    background-color: transparent;
+  }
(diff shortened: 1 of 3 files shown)


src/ensembl/src/shared/components/search-field/SearchField.scss
```diff
@@ -1,9 +1,16 @@
...
+  padding: 0 15px;
+  height: 30px;
   display: inline-flex;
-  padding: 0 0.6em;
+  padding: 0 12px;
+
+  input {
+    background-color: transparent;
+  }
 }
```
imransl, line 14: Do we need to specify `border` and `padding-right` in `.searchField`? It doesn't seem to make any change to the search field. Same with `background-color` in `input`.

Also noticed that padding is specified twice in `.searchField`.

azangru: Background-color actually does make a difference:

https://user-images.githubusercontent.com/6834224/131754484-52c82ec8-bfe4-412d-b906-067c0e4d0eeb.mp4

Instead of the background colour, this could potentially be solved with a bit of vertical padding. Don't know if that would be better.

I'm removing the extra padding rule in `.searchField`, as well as the border and padding-right 👍  I also wonder whether `background: $white;` should be moved from the `.speciesSearchField` in `SpeciesSearchField.scss` to the `.searchField` class.

jyothishnt:
Looks good Andrey. Noticed the padding and height different on search fields in species selector and interstitial pages. Is that part of the design?

azangru:
> Noticed the padding and height different on search fields in species selector and interstitial pages. Is that part of the design?

I just reproduced what is currently on 2020. I am sure Andrea will tell us to change field sizes during design review.

imransl (approved):

PR #573, opened 2021-08-27
Fix species list layout in species selector and species homepage

imransl:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1299](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1299)

## Description
~~The species list in the app bar doesn't appear properly when it overflows after adding many species.~~

image

~~So I've made the species list to appear in one line like how it does in other places like genome browser and entity viewer.~~

image

However I am not sure if this was what was intended in the design as I couldn't find anything similar to it.

**Edit:** As Andrey pointed in the PR comment below, it was the wrong solution. Instead the task required to wrap the species in multi lines. I have updated the PR accordingly. The species list now looks like this.

image

## Deployment URL
http://fix-species-list.review.ensembl.org/

## Views affected
Species selector -> App bar
Species homepage -> App bar

### DIFF
--- a/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
@@ -78,7 +78,9 @@ const SelectedSpeciesList = (props: Props) => {
     />
   ));
 
-  return <SpeciesTabsWrapper speciesTabs={selectedSpecies} />;
+  return (
+    <SpeciesTabsWrapper isWrappable={false} speciesTabs={selectedSpecies} />
+  );
 };
 
 const mapStateToProps = (state: RootState) => ({
--- a/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
+++ b/src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
@@ -78,9 +78,7 @@ const SelectedSpeciesList = (props: Props) => {
     />
   ));
 
-  return (
-    <SpeciesTabsWrapper isWrappable={false} speciesTabs={selectedSpecies} />
-  );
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
```diff
@@ -78,7 +78,9 @@ const SelectedSpeciesList = (props: Props) => {
     />
   ));
 
-  return <SpeciesTabsWrapper speciesTabs={selectedSpecies} />;
+  return (
+    <SpeciesTabsWrapper isWrappable={false} speciesTabs={selectedSpecies} />
+  );
```
azangru, line 83: Whoa! The idea was that both in the species selector and on the species page, the selected species lozenges are allowed to wrap over multiple lines in order to show the full names of the selected species. In other apps, such as genome browser or entity viewer, species lozenges can only ever occupy a single line, and thus can shrink and crop off parts of species names.

It should be up to the rest of the Species Selector and Species page to adapt to the growing height of the selected species bar.

imransl: @azangru Thanks for pointing this out. I had misunderstood this task. I've pushed the changes after fixing this properly and also have vertically aligned the app bar aside.

azangru:
Ok, this fixes the bug, so I guess we can merge this.

One problem that I foresee though is that Andrea would most probably want us to fix the position of the launchbar and maybe of th selected species during scroll (she made me do [this](https://internal-2020.ensembl.org/about/articles/software-license) for the Help and the About app). Currently if your window is too small, the scrolling of the window vertically will scroll away the launchbar:

https://user-images.githubusercontent.com/6834224/131882314-17117433-145c-4a5c-8387-db5f5f8aeeec.mp4

We should probably ask... I have no idea how this can work properly. But if Andrea says yes, it will be a different task.

azangru (approved):

PR #575, opened 2021-09-06
ENSWBSITES-642: filter and sorting state update

ens-ma7:
Only remember state for filtering and sorting when viewing gene or switching from app back to same gene
Do not remember filtering and sorting state for previously viewed objects
Resetting filter and sorting when clicking on previously viewed objects

## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-642

## Importance
N/A

## Description
Currently we are storing the filters and sorting state on the gene and when switching app or browsing away we still keep the filter and sorting state on that gene. 
New way of doing it (more detailed description in Jira) is to only store the state on the gene thats being viewed, once you navigate to another gene, the state is reset.

## Deployment URL
http://filter-reset.review.ensembl.org

## Views affected
Entity Viewer

### Other effects
N/A

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
@@ -31,7 +31,6 @@ import {
 import { getPreviouslyViewedEntities } from 'src/content/app/entity-viewer/state/bookmarks/entityViewerBookmarksSelectors';
 
 import { closeSidebarModal } from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSlice';
-import { clearFilterSorting } from 'src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice';
 
 import TextLine from 'src/shared/components/text-line/TextLine';
 
@@ -47,11 +46,6 @@ type PreviouslyViewedLinksProps = {
 
 export const PreviouslyViewedLinks = (props: PreviouslyViewedLinksProps) => {
   const dispatch = useDispatch();
-
-  const onClick = () => {
-    dispatch(clearFilterSorting());
-    dispatch(closeSidebarModal());
-  };
 
   const activeEntityStableId = parseEnsObjectId(props.activeEntityId).objectId;
   const previouslyViewedEntitiesWithoutActiveEntity =
@@ -73,7 +67,7 @@ export const PreviouslyViewedLinks = (props: PreviouslyViewedLinksProps) => {
 
           return (
             <div key={index} className={styles.linkHolder}>
-              <Link to={path} onClick={onClick}>
+              <Link to={path} onClick={() => dispatch(closeSidebarModal())}>
                 <TextLine
                   text={previouslyViewedEntity.label}
                   className={styles.label}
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
```diff
@@ -47,6 +48,11 @@ type PreviouslyViewedLinksProps = {
 export const PreviouslyViewedLinks = (props: PreviouslyViewedLinksProps) => {
   const dispatch = useDispatch();
 
+  const onClick = () => {
```
jyothishnt, line 51: minor, should we call this handleClick() or name as per the functionality?

ens-ma7: I am not sure what we do in 2020 code, i dont see any problem with using either. Lets see what others say.

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
```diff
@@ -47,6 +48,11 @@ type PreviouslyViewedLinksProps = {
...
+    dispatch(clearFilterSorting());
```
azangru, line 52: The way I am reading the ticket, including the comments, every change to the active gene should reset the filters and sorting settings associated with it. This suggests that the proper place for clearing filters and sorting is within a redux action that sets the active gene rather than in any of the UI elements.

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
ens-ma7: @jyothishnt this is not relevant anymore because I have changed the way we are doing this. Please have a look at the changes again.

src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.tsx
ens-ma7: What @azangru said makes sense and we also spotted a scenario where the filtering was being applied where it shouldn't. After a pair programming session, we have updated the code to make the filter state reset/clear in the EntityViewerGeneralSlice where the genomeId and entityId change is happening.

src/ensembl/src/content/app/entity-viewer/state/general/entityViewerGeneralSlice.ts
```diff
@@ -90,6 +91,9 @@ export const setDataFromUrl =
           })
         );
       }
+      if (activeGenomeId === genomeIdFromUrl && entityId !== activeEntityId) {
```
jyothishnt, line 94: Is the first check necessary? If the entityId has changed we clear filters.

ens-ma7: @jyothishnt it is necessary. The reason for this is this scenario:
Human BRCA2, you apply filter -> switch to another species -> come back to human again
The filter should still be applied and the check make sure it does because when switching the species it was reloading the component and hence losing the filters.

jyothishnt: got it.. thanks

jyothishnt (approved):

azangru:
Heard during sprint planning that we are not going ahead with this PR?

ens-ma7:
Awaiting final decision from Andrea once she and lucy had a look at it.

ens-ma7:
The final decision is that this PR is not relevant anymore since the current behaviour on the live site is fine but we want to extend the state to be persistent on browser refresh/close. Creating another ticket for this.

PR #587, opened 2021-09-21
Convert track panel groups to accordions

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-395

## Description
Updated the track panel groups to use accordions

Created another ticket to update the track API and remove the following sections:
- Experiment Design and References and Evidence from Genomic
- Experiment Design from Variation 
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1354

## Deployment URL
http://track-panel-accordions.review.ensembl.org

## Views affected
Genome browser -> track panel

### DIFF
--- a/src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.scss
+++ b/src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.scss
@@ -22,3 +22,45 @@
   margin-left: 0;
   padding: 4px 10px;
 }
+
+
+
+.accordionContainer {
+  width: 100%;
+
+  .trackPanelAccordion {
+    border: none;
+    margin-top: 30px;
+  }
+  
+  .trackPanelAccordionHeader {
+    background-color: $light-grey;
+    border-radius: 5px;
+  }
+  
+  .trackPanelAccordionButton {
+    color: $blue;
+    padding: 5px 15px 5px 17px;
+    border-radius: 5px;
+    font-size: 12px;
+    &:before,
+    &:after {
+      margin-top: -3px;
+    }
+  
+    &Disabled {
+      color: $grey;
+    }
+  }
+  
+  .trackPanelAccordionItem:not(:first-child) {
+    border-color: transparent;
+  }
+  
+  .trackPanelAccordionItemContent {
+    background-color: transparent;
+    border: none;
+    padding: 10px 0 20px;
+  }
+
+}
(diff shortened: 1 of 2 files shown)


azangru:
- Surely, according to the design, the sections should be open by default.
- I wonder if it is sufficiently safe to represent all track panel sections as a single Accordion component, or whether it will be safer to represent them as multiple single-section accordions. 

Also, this is for the future, but I am sure Andrea will start asking us to maintain the expanded/collapsed state of accordions when user switches between apps or species 🙈

msmanoj:
> * Surely, according to the design, the sections should be open by default.

That's done

> * I wonder if it is sufficiently safe to represent all track panel sections as a single Accordion component, or whether it will be safer to represent them as multiple single-section accordions.

I think we can use a single Accordion for now and change it later if it needs to be.

src/ensembl/src/content/app/browser/track-panel/track-panel-list/TrackPanelList.scss
```diff
@@ -22,3 +22,45 @@
...
+  
+  .trackPanelAccordionItem:not(:first-child) {
+    border-color: transparent;
+  }
+  
+  .trackPanelAccordionItemContent {
+    background-color: transparent;
+    border: none;
+    padding: 10px 0 20px;
+  }
```
azangru, line 64: Is there any real reason to nest all these classes inside `.accordionContainer`? Nesting adds specificity, and increases the size of CSS by adding the parent class name to every nested CSS rule. Do we need it here?

msmanoj: No, it doesn't have to be. I have removed it now. Is it not a good practice to use nesting?

azangru: > Is it not a good practice to use nesting?

Don't know about good practice, but the nesting in SASS:

```css
.parent {
  .child {
    foo: bar;
  }
}
```

is expanded into 

```css
.parent .child {
  foo: bar;
}
```

which, on the one hand, makes the compiled CSS slightly larger, and on the other hand, makes the `.child` selector more specific than it needs to be.

I'd say the least amount of specificity we apply to target an element, the better. With our unfortunate CSS modules setup, if we need to override a class of a child component via a classname passed from the parent, then it may be a good idea to nest the class that is passed to the child in order to increase its specificity:

```css
/* Parent component: */

.parent {
  .overridingClass {
    color: blue;
  }
}

/* Child component */

.originalClass {
  color: red;
}
```

azangru (approved):

PR #591, opened 2021-09-22
Enable click handling, as well as hover, on QuestionButton

azangru:
## Type

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1343

## Description
- Add click handling to `QuestionButton`. A click on `QuestionButton` will show the tooltip; the next click will hide it, and so on.
- Also, keep the previous behaviour on hover (wait for a while and show the tooltip)
- Update some CSS to prevent text selection and other nuisance on repeat clicks

## Deployment URL
http://question-button.review.ensembl.org

### DIFF
--- a/src/ensembl/src/shared/components/question-button/QuestionButton.test.tsx
+++ b/src/ensembl/src/shared/components/question-button/QuestionButton.test.tsx
@@ -46,6 +46,24 @@ describe('<QuestionButton />', () => {
       });
       expect(queryByText(helpMessage)).toBeTruthy();
     });
+
+    it('hides the tooltip on mouseleave', async () => {
+      const { container, queryByText } = render(
+        <QuestionButton helpText={helpText} />
+      );
+      const questionButton = container.querySelector(
+        '.questionButton'
+      ) as HTMLElement;
+
+      userEvent.hover(questionButton);
+
+      act(() => {
+        jest.advanceTimersByTime(TOOLTIP_TIMEOUT + 10);
+      });
+
+      userEvent.unhover(questionButton);
+      expect(queryByText(helpMessage)).toBeFalsy();
+    });
   });
 
   describe('on click', () => {
@@ -85,6 +103,25 @@ describe('<QuestionButton />', () => {
         // give the Tooltip component a little time to complete its async tasks
         jest.advanceTimersByTime(10);
       });
+      expect(queryByText(helpMessage)).toBeTruthy();
+    });
+
+    it('does not hide the tooltip on mouseleave', () => {
+      const { container, queryByText } = render(
+        <QuestionButton helpText={helpText} />
+      );
+      const questionButton = container.querySelector(
+        '.questionButton'
+      ) as HTMLElement;
+
+      userEvent.click(questionButton);
+
+      act(() => {
+        // give the Tooltip component a little time to complete its async tasks
+        jest.advanceTimersByTime(TOOLTIP_TIMEOUT + 10);
+      });
+
+      userEvent.unhover(questionButton); // <-- should have no effect on the tooltip
       expect(queryByText(helpMessage)).toBeTruthy();
     });
   });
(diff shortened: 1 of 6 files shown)


src/ensembl/src/shared/components/question-button/QuestionButton.tsx
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 import classNames from 'classnames';
 
-import useHover from 'src/shared/hooks/useHover';
+import { useQuestionButtonBehavior } from './useQuestionButtonBehavior';
```
imransl, line 20: Aren't we sticking with UK spellings for variables?

src/ensembl/src/shared/components/question-button/useQuestionButtonBehavior.tsx
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useReducer, useEffect } from 'react';
+import useHover from 'src/shared/hooks/useHover';
```
imransl, line 18: Separate these imports?

msmanoj (approved):

src/ensembl/src/shared/components/question-button/QuestionButton.tsx
jyothishnt: we may need to change this too https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/entity-viewer/gene-view/components/proteins-list/proteins-list-item/ProteinsListItem.tsx#L102 😆

jyothishnt: I just did a quick check in the code and found we are sticking on to British English. Sorry @azangru  :-)

azangru: We can't change that, Jyo; it's the browser API, and browsers speak American English:

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

This is why I would love us to use American English in code. All the other examples of the word `behaviour` in our code are for display labels; which is totally fine, we could write labels in Chinese hyerogliphs for all we care. But the code itself would be much more consistent if it stuck to American spelling 😞

jyothishnt: oh oh!

jyothishnt:
When I click to show tooltip, should it close only when I click outside?
It closes when I hover the icon for few seconds and move away. Just checking if this is the correct behaviour.

src/ensembl/src/shared/components/question-button/QuestionButton.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+      );
+      const questionButton = container.querySelector(
+        '.questionButton'
+      ) as HTMLElement;
+
+      userEvent.hover(questionButton);
+      expect(questionButton.querySelector('.pointerBox')).toBeFalsy();
+
+      act(() => {
+        jest.advanceTimersByTime(TOOLTIP_TIMEOUT + 10);
```
jyothishnt, line 45: TOOLTIP_TIMEOUT = 800  Is it in milliseconds? If so, does it wait for 8 seconds here when we run this test?

src/ensembl/src/shared/components/question-button/QuestionButton.tsx
azangru: It is with heavy heart that I have changed behavior for behaviour 😞 

✅

azangru:
> Currently if the tooltip is activated by click and we do a hover action (hover in and out), it closes the tooltip.

Thanks for spotting this, @jyothishnt . I updated the code to prevent this from happening, and added a test to confirm that this is now the case.

jyothishnt (approved):

PR #594, opened 2021-09-27
Update 404 error page

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1325

## Description
- Updated the 404 error page to match with the new design

## Subsequent tasks
- Log the 404s in the service monitoring system being developed by BE (https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1360)

## Deployment URL
http://update-404-page.review.ensembl.org/hello

### DIFF
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -20,7 +20,7 @@
 }
 
 .generalErrorScreen {
-  height: 100vh;
+  height: 100%;
   display: flex;
   flex-direction: column;
 }
--- a/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
+++ b/src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
@@ -46,7 +46,7 @@
   font-size: 12px;
 }
 
-.reloadButton {
+.reloadButton, .homeButton {
   padding-top: 40px;
 }
 
@@ -78,8 +78,8 @@
     text-align: right;
 
     span {
-      margin-right: 83px;
-      margin-left: 30px;
+      margin-right: 77px;
+      margin-left: 25px;
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/ensembl/src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -46,7 +46,7 @@
   font-size: 12px;
```
azangru, line 46: You will find above that the `.generalErrorScreen` class has a height of `100vh`. This is too much for the 404 page when it is rendered client-side, with the top bar and the launchbar. Currently, when the 404 page is server-rendered, it does not have the topbar and the launchbar on the top; so the 100vh works. I suppose we will want to include the topbar and the launchbar into the server-rendered payload, but we don't yet have clarity on that.

Anyhow, just wanted to let you know. Not sure what the proper solution should be; there are several options.

msmanoj: Changed `100vh` to `100%` as discussed during code review

jyothishnt:
All good. One doubt - When I go to http://update-404-page.review.ensembl.org/help and then change the url to http://update-404-page.review.ensembl.org/helpx  I see the help icon still selected in the launchbar. Is this how it is?

jyothishnt (approved):

azangru:
> and then change the url

The problem is a bit more general: if you enter an invalid url starting with the same string as any of our existing apps (`helpx`, `genome-browser-foo`, or whatever), a button in the launchbar will light up as selected. This is because we are using a regex in `LaunchbarButton` to check whether an app is matched ([link](https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/header/launchbar/LaunchbarButton.tsx#L39-L40)). We could instead look inside redux, where we store the current app name. This can be fixed in this or in a separate PR.

azangru (approved):

PR #595, opened 2021-09-28
Console log GA events on dev environments

msmanoj:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1359

## Description
- Enabled the debug option on `react-ga` to log all the events to the console only for dev environments

## Deployment URL
http://console-log-ga.review.ensembl.org


src/ensembl/src/services/analytics-service.ts
```diff
@@ -15,12 +15,11 @@
...
 
 import ReactGA from 'react-ga';
-import { AnalyticsOptions, CustomDimensions } from 'src/analyticsHelper';
 
+import { AnalyticsOptions, CustomDimensions } from 'src/analyticsHelper';
 import config from 'config';
 
-const { googleAnalyticsKey } = config;
-
+const { googleAnalyticsKey, isDevelopment } = config;
```
azangru, line 22: I'm confused by what's happening here.

`isDevelopment` that is coming from the config is supposed to only refer to the build environment (i.e. what the NODE_ENV variable was set to). You can see [here](view-source:http://console-log-ga.review.ensembl.org/species-selector) that it is set to `production` in review deployment. And yet I am seeing the console logs in the deployed branch. Why?

msmanoj: I am getting the "[react-ga] gaTrackingID is required in initialize()" warning now when it was working fine previously.

azangru: `[react-ga] gaTrackingID is required in initialize()` this is expected; we are not passing the id to the service. We may be a bit smarter about all that and not try to initialise the service if there is no google analytics key available.

src/ensembl/src/services/analytics-service.ts
```diff
@@ -38,6 +49,17 @@ class AnalyticsTracking {
 
   // Track an event
   public trackEvent(ga: AnalyticsOptions) {
+    if (this.isTestModeEnabled === undefined) {
```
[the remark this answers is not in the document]
imransl: Ah I didn't think of that scenario. Let's leave it like this then.

PR #597, opened 2021-09-28
Enable dedicated click handlers for ViewInApp component

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1344

## Description
Turns out, we need to support use cases when user clicks on a `ViewInApp` button, and the app doesn't change, but we need to handle this click in some clever way anyhow. Who knew!

This PR adds click handlers that can be both unique to individual app buttons, or common for all buttons. The immediate use case is hiding the tooltip when user clicks on a search match in the InAppSearch component. 

## Deployment URL
http://view-in-app-update.review.ensembl.org

## Views affected
InAppSearch in GenomeBrowser and EntityViewer.


msmanoj (approved):
Looks good 👍

ens-ma7 (approved):

imransl (approved):

PR #598, opened 2021-10-02
Use species svg icons via the img tag

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1370

## Description
Switch from fetching species svgs over XHR and inlining them as svg markup into html to using the native `` tag for the display of the images. Apply the CSS invert filter to invert the icon colors in `PopularSpeciesButton`.

Also, delete the InlineSVG component, because it is no longer used anywhere.

## Deployment URL
http://use-species-svg-as-images.review.ensembl.org

## Consequences
For a brief moment (depends on how coordinated the deployment of backend assets will be) we won't be able to display the number `38` and `37` on the human icons. This will be fixed once the updated svg images are available from the backend.

image

## Views affected
- Species selector
- Species page


ens-ma7 (approved):

msmanoj (approved):
Looks good 👍

PR #601, opened 2021-10-04
Entity viewer analytics hook

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1368

## Description
- Updated the analytics service to handle `feature` as a dimension
- Created the analytics hook for entity viewer
- Added the tracking events to track the tab changes

## Deployment URL
http://ev-ga-hook.review.ensembl.org

## Views affected
Entity viewer -> Change tabs

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/hooks/useEntityViewerAnalytics.ts
+++ b/src/ensembl/src/content/app/entity-viewer/hooks/useEntityViewerAnalytics.ts
@@ -1,0 +1,62 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useSelector } from 'react-redux';
+
+import analyticsTracking from 'src/services/analytics-service';
+import { getSpeciesAnalyticsName } from 'src/content/app/species-selector/speciesSelectorHelper';
+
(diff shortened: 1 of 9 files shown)


azangru:
How does setting a custom dimension work? How long is it reported?

In the following scenario, for example: `Species Selector -> Entity Viewer -> Species Selector`, when user returns to species selector from the entity viewer, will all analytics events sent in species selector contain the `feature` dimension set in entity viewer?

msmanoj:
> How does setting a custom dimension work? How long is it reported?
> 
> In the following scenario, for example: `Species Selector -> Entity Viewer -> Species Selector`, when user returns to species selector from the entity viewer, will all analytics events sent in species selector contain the `feature` dimension set in entity viewer?

Currently, the scope for feature dimension is set to `hit` and I thought it will be valid until the next hit but it doesn't seem to be the case. It is using the same value for subsequent hits as well. I guess we need to set and reset it for every hit 😞

src/ensembl/src/services/analytics-service.ts
```diff
@@ -49,6 +50,9 @@ class AnalyticsTracking {
       transport: 'xhr',
       value: ga.value
     });
+
+    typeof ga.species === 'string' && this.setSpeciesDimension('');
+    typeof ga.feature === 'string' && this.setFeatureDimension('');
```
azangru, line 55: 1) Do you remember why we write `typeof something === 'string' && ...` rather than simply `something && ...`. It would make sense if we were considering empty string to be a meaningful input, which would fail the `something && ...` check; but we aren't passing empty strings to this method.

2) Does unsetting a field mean setting it to empty string or to `undefined`? The docs are very confusing 😞

src/ensembl/src/content/app/entity-viewer/hooks/useEntityViewerAnalytics.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { parseEnsObjectId } from 'ensemblRoot/src/shared/state/ens-object/ensObjectHelpers';
+import {
+  getEntityViewerActiveGenomeId,
+  getEntityViewerActiveEntityId
+} from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSelectors';
+import { getCommittedSpeciesById } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
+
+import { RootState } from 'src/store';
+import { useEffect } from 'react';
```
imransl, line 29: Move this import to the top?

azangru (approved):

PR #602, opened 2021-10-04
Include biotype definition into the payload for GeneSummary component

azangru:
## Type
- Bug fix

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1373

## Description
Add the forgotten `definition` field to a graphql query.

Before:

image

After:
image

## Deployment URL
http://fix-missing-biotype-definition.review.ensembl.org

## Views affected
Genome browser gene view


ens-st3 (approved):
Tooltip works

ens-ma7:
@azangru tooltip works but it is transcript related even when you are hovering on a gene biotype.

azangru:
> Or is that something that needs fixing from the backend?

Yes. The tooltip is simply showing the text that is coming from the backend.

ens-ma7 (approved):

PR #613, opened 2021-10-18
Change from faker.random to faker.lorem in tests of ExternalReference component

azangru:
## Description
Tests for `ExternalReference` component are using the functions in the `faker.random` namespace to generate class names, which sometimes produces invalid class names that fail the tests.

[Example of the latest test failure](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-client/-/jobs/600157):

image

This PR changes `faker.random` to `faker.lorem` for `ExternalReference` tests.


imransl:
`faker.random.word` is used in other places as well to generate random class names e.g. [`SlideToggle.test.tsx`](faker.random.word). Does this mean we'll need change in those files too?

PR #629, opened 2021-11-03
ENSWBSITES-1336: Extend filter and sort state to local storage

ens-ma7:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1336

## Importance
N/A

## Description
Extending the filtering and sorting functionality to local storage. This is now stored against the previously viewed entity in local storage.

## Deployment URL
http://filterstate-extension.review.ensembl.org

## Views affected
Entity Viewer

### Other effects
N/A

## Possible complications
Redux code is getting uglier :)

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/state/bookmarks/entityViewerBookmarksSlice.ts
+++ b/src/ensembl/src/content/app/entity-viewer/state/bookmarks/entityViewerBookmarksSlice.ts
@@ -14,13 +14,25 @@
  * limitations under the License.
  */
 
-import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { createSlice, createAsyncThunk, PayloadAction } from '@reduxjs/toolkit';
+import merge from 'lodash/merge';
+
+import { getPreviouslyViewedEntities } from './entityViewerBookmarksSelectors';
+
 import entityViewerBookmarksStorageService from 'src/content/app/entity-viewer/services/bookmarks/entity-viewer-bookmarks-storage-service';
+
+import {
+  Filters,
+  SortingRule
+} from 'src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice';
+import { RootState } from 'src/store';
 
 type PreviouslyViewedEntity = {
   entity_id: string;
   label: string | string[];
   type: 'gene';
+  filters?: Filters;
+  sortingRule?: SortingRule;
 };
 
 export type PreviouslyViewedEntities = {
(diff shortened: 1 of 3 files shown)


src/ensembl/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.test.tsx
```diff
@@ -39,13 +39,13 @@ const mockState = {
...
       activeEntityIds: {
-        human: 'gene:brca2'
+        human: 'human:gene:brca2'
       }
     },
     geneView: {
       transcripts: {
         human: {
-          'gene:brca2': {
+          'human:gene:brca2': {
```
jyothishnt, line 48: It is already structured under human, why is human added again in the key?

jyothishnt:
I feel it would be good to discuss this PR in code review? will be much easier to understand. :-)

src/ensembl/src/content/app/entity-viewer/state/bookmarks/entityViewerBookmarksSlice.ts
```diff
@@ -40,6 +52,77 @@ type UpdatePreviouslyViewedPayload = {
...
+    const previouslyViewedEntities = getPreviouslyViewedEntities(
+      state,
+      genomeId
+    );
+
+    const alreadySavedGene = previouslyViewedEntities.find(
+      (entity) => entity.entity_id === gene.unversioned_stable_id
+    );
+
+    const newPreviouslyViewedEntity = {
```
imransl, line 71: Is this object necessary if `alreadySavedGene` exists?

src/ensembl/src/content/app/entity-viewer/state/bookmarks/entityViewerBookmarksSlice.ts
```diff
@@ -40,6 +52,77 @@ type UpdatePreviouslyViewedPayload = {
...
+      entity_id: gene.unversioned_stable_id,
+      label: gene.symbol ? [gene.symbol, gene.stable_id] : gene.stable_id,
+      type: 'gene' as const
+    };
```
imransl, line 75: Is this object necessary if `alreadySavedGene` exists?

ens-ma7: From my understanding thats needed when it doesn't exist, its a new gene being added. But @azangru can confirm this if i am saying the right thing here.

jyothishnt:
This no more considers the 20 previously viewed limit 
Screenshot 2021-11-09 at 09 42 06

jyothishnt:
When I go to EV on an example gene, I see that gene already visible in previously viewed.

ens-ma7:
As discussed in the code review on 10/11/2021, we will try the other solution which is to have a separate storage for Entityviewer filters. This will be as a separate PR and then we can decide which solution to go with.

There is also a third solution which is to separate the redux state and the storage and having redux retrieve from storage first and then update the redux state.

PR #632, opened 2021-11-05
Use faker.lorem.word instead of faker.random.word

ens-ma7:
To prevent generation of invalid css classnames (faker.random.word sometimes can generate special characters)

## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1401

## Importance
N/A

## Description
use faker.lorem.word instead of faker.random.word because faker.random.word can generate special characters which can break css classnames.

## Deployment URL
N/A since its only test that has been updated.

## Views affected
N/A

### Other effects

## Possible complications
N/A

### DIFF
--- a/src/ensembl/src/shared/components/external-reference/ExternalReference.test.tsx
+++ b/src/ensembl/src/shared/components/external-reference/ExternalReference.test.tsx
@@ -22,7 +22,7 @@ import ExternalReference, { ExternalReferenceProps } from './ExternalReference';
 
 const defaultProps: ExternalReferenceProps = {
   label: faker.random.word(),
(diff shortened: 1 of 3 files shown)


azangru (approved):

azangru:
Would it be ok to remove the ticket number from the name of the PR? We already refer to the jira ticket in the description of the PR; so the ticket number will only creates noise in the commit message.

imransl (approved):

jyothishnt (approved):

PR #634, opened 2021-11-06
Update Node to latest LTS and remove unused packages

azangru:
## Description
- Update Node to the latest LTS release (16.13.0)
- Remove unused packages
  - `@types/http-proxy-middleware` is no longer required (type definitions provided by `http-proxy-middleware`)
  - `@types/classnames` no longer required (`classnames` provides its own type definitions)
  - `worker-loader` is no longer used (forgot to remove it when webpack was updated to version 5)
- `package-lock.json` was regenerated (notice that as a result it removed most of the `initegrity` fields, which seems to be the correct behaviour)

## Deployment URL
http://cleanup-dependencies.review.ensembl.org


kamaldodiya (approved):
Gitlab-ci and Dockerfile looks good.

imransl (approved):
Looks good locally too

msmanoj (approved):

PR #635, opened 2021-11-08
Add tests for useBrowserRouting hook

azangru:
Included in this PR:

- tests for useBrowserRouting hook
- minor update to a test of ZmenuContent (check what is being passed to the function rather than just the fact of a function being called)
- fix mock functions in `BrowserRegionField.test.tsx` (mocks of redux actions should return actions)


msmanoj:
Nicely written @azangru. The only thing I wanted to check is if we need some tests for the function `changeGenomeId` as well?

azangru:
> The only thing I wanted to check is if we need some tests for the function changeGenomeId as well?

I was on the fence about them, but added them now.

msmanoj (approved):

PR #636, opened 2021-11-10
Add analytics on help page

jyothishnt:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1408

## Description
Create a hook for help app analytics and to add tracking events to the help app mega nav that are listed in the following spreadsheet:

https://docs.google.com/spreadsheets/d/1CnW-rx212UwAMTXnjZXCPfZ72Kva7_kRiG6X6pfSN94/edit#gid=2090835159

- [x] - Track clicks on the 3 links on the meganav
- [x] - Do not track the click which closes the menu
- [x] - Consider Overview as a meganav item and not an article even though it doesnt open a nav menu

## Deployment URL
http://help-app-ga.review.ensembl.org

## Views affected
Help app mega nav and submenu links

### Other effects
NA

## Possible complications
NA

### DIFF
--- a/src/content/app/help/components/help-menu/HelpMenu.tsx
+++ b/src/content/app/help/components/help-menu/HelpMenu.tsx
@@ -20,12 +20,14 @@ import { useDispatch } from 'react-redux';
 import { push } from 'connected-react-router';
 
 import { isEnvironment, Environment } from 'src/shared/helpers/environment';
+import useHelpAppAnalytics from '../../hooks/useHelpAppAnalytics';
 
 import Chevron from 'src/shared/components/chevron/Chevron';
 import HelpMenuLink from './HelpMenuLink';
 
 import {
   Menu as MenuType,
+  MenuArticleItem,
   MenuItem
 } from 'src/shared/types/help-and-docs/menu';
 
@@ -40,6 +42,8 @@ const HelpMenu = (props: Props) => {
   const [submenuItems, setSubmenuItems] = useState<MenuItem[] | null>(null);
   const clickedMenuRef = useRef<number | null>(null);
 
+  const { trackTopLevelMenu } = useHelpAppAnalytics();
+
   if (isEnvironment([Environment.PRODUCTION])) {
     return (
       <div className={styles.helpMenu}>
@@ -48,7 +52,11 @@ const HelpMenu = (props: Props) => {
     );
   }
 
-  const toggleMegaMenu = (items: MenuItem[], menuIndex: number) => {
+  const toggleMegaMenu = (
+    items: MenuItem[],
+    menuIndex: number,
+    menuName: string
+  ) => {
     let nextValue = null;
     if (
       clickedMenuRef.current === null ||
@@ -58,11 +66,16 @@ const HelpMenu = (props: Props) => {
       clickedMenuRef.current = menuIndex;
       nextValue = items;
     } else {
-      // this means a repeated click on the same menu iteem
+      // this means a repeated click on the same menu item
       clickedMenuRef.current = null;
     }
-
     setSubmenuItems(nextValue);
+    nextValue && trackTopLevelMenu(menuName);
+  };
+
+  const handleHelpMenuLinkClick = (menuName: string) => {
+    closeMegaMenu();
+    trackTopLevelMenu(menuName);
   };
 
   const closeMegaMenu = () => {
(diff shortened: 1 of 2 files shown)


src/ensembl/src/content/app/help/components/help-menu/HelpMenu.tsx
```diff
@@ -40,6 +42,8 @@ const HelpMenu = (props: Props) => {
   const [submenuItems, setSubmenuItems] = useState<MenuItem[] | null>(null);
   const clickedMenuRef = useRef<number | null>(null);
 
+  const { trackMegaNavOpen } = useHelpAppAnalytics();
```
azangru, line 45: If we are also tracking clicks on the links that do not open the mega nav (such as "Overview") with this function, then let's at least not confuse ourselves with its name – it is not tracking the opening of the mega nav, but rather the clicking on the top-level menu item.

src/ensembl/src/content/app/help/components/help-menu/HelpMenu.tsx
```diff
@@ -58,32 +62,41 @@ const HelpMenu = (props: Props) => {
...
-      <HelpMenuLink {...commonProps} to={item.url} onClick={closeMegaMenu}>
-        {item.name}
-      </HelpMenuLink>
+
+    return (
+      <div
+        key={index}
+        className={className}
+        onClick={() => handleMenuItemClick(item.name)}
+      >
```
azangru, line 89: Mmm, I don't like that in order to add a common onclick behaviour the elements are wrapped in another DOM element. Looks like it would be sufficient simply to modify the click handlers.

src/ensembl/src/content/app/help/components/help-menu/HelpMenu.tsx
```diff
@@ -58,32 +62,41 @@ const HelpMenu = (props: Props) => {
...
+        {item.type === 'collection' ? (
+          <span key={index} onClick={() => toggleMegaMenu(item.items, index)}>
```
imransl, line 91: How about creating a click handler above and then using it here like this:

```
// create the click handler
const clickHandler = () => toggleMegaMenu(item.items, index);

// use the click handler
<span key={index} onClick={clickHandler}>
```

src/ensembl/src/content/app/help/components/help-menu/HelpMenu.tsx
```diff
@@ -114,25 +127,29 @@ const Submenu = (props: SubmenuProps) => {
...
     // might need extra logic if we can have external urls in the menu
     props.onLinkClick();
-    dispatch(push(url));
+    dispatch(push(item.url));
+    trackArticleLinkClicked(item.name, item.type);
   };
 
   const renderedMenuItems = props.items.map((item, index) => {
     const className = classNames(styles.submenuItem);
     const props: Record<string, unknown> = {};
```
imransl, line 146: I know this was not added in this PR. However naming this variable as `props` in a React component file is misleading. Can we name this something else like `eventObj`?

azangru: But this variable here does represent the collections of props — the ones that are passed to the child.

src/ensembl/src/content/app/help/components/help-menu/HelpMenu.tsx
jyothishnt: I can, but then I will have to check if the item type is collection above when we create that function too. Do we really need this?

src/ensembl/src/content/app/help/hooks/useHelpAppAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+const useHelpAppAnalytics = () => {
+  const trackTopLevelMenu = (navLabel: string) => {
+    analyticsTracking.trackEvent({
+      category: 'help_mega_nav',
+      label: navLabel,
+      action: 'opened'
+    });
+  };
+
+  const trackArticleLinkClicked = (linkLabel: string, type: string) => {
```
azangru, line 28: - `clicked` -> `click` (see useEntityViewerAnalytics for comparison)
- maybe `trackMegaNavItemClick`? The reasons being that
  - this would indicate that the clicked thingy is specifically in the meganav (as opposed to any other place)
  - the clicked thingy can be a link to an article or a link to a video

ens-ma7:
@azangru this is ready for another review. @jyothishnt made some changes

azangru (approved):
Looks good to me.

PR #640, opened 2021-11-18
Remove ensemblRoot alias

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1423

## Description
- `ensemblRoot` alias path to the project root didn't work out, because VS Code was always picking it up during autocompletion of import paths; this PR removes it
- add `webpackDir` alias to the webpack folder (obviously can't just call the alias `webpack`, because we `webpack` is the name of one of our dependencies)


msmanoj:
Can we remove `dotenv`, `copy-dotenv` & `dotenv-webpack` from package.json unless they are still used?

azangru:
I thought maybe do so in a separate PR, as it's a different concern. If we remove the dotenv, it means we won't be using the `.env` file (which we aren't currently using anyway), and therefore should remove the `.env.example` as well. Are you happy with that? If so, I can remove everything dotenv-related in this PR.

msmanoj:
> I thought maybe do so in a separate PR, as it's a different concern. If we remove the dotenv, it means we won't be using the `.env` file (which we aren't currently using anyway), and therefore should remove the `.env.example` as well. Are you happy with that? If so, I can remove everything dotenv-related in this PR.

I'm happy if it is done in this PR but if you prefer to keep that for a separate PR, that's fine too.

azangru:
Removed dotenv packages and `.env.example`

msmanoj (approved):

PR #641, opened 2021-11-18
Hide current entity from previously viewed

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1422

## Description
- Fixed the issue with current entity getting displayed as previously viewed

## Deployment URL
http://fix-ev-previously-viewed.review.ensembl.org

## Views affected
- Entity viewer previously viewed

### Other effects

### DIFF
--- a/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.test.tsx
+++ b/src/ensembl/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-modal/modal-views/EntityViewerBookmarks.test.tsx
@@ -45,6 +45,9 @@ const exampleEntities = [
   }
 ];
 
+const currentEntityId = `braf`;
+const currentEntityLabel = `BRAF`;
+
 const previouslyViewedEntities = [
   {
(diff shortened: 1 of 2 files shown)


jyothishnt:
Noticed that bookmarks functionality logic  is different in GB and EV.

In GB, it doesn't register a previouslyViewed entity in localstorage on the initial entry to the page whereas the same gets registered when visit a gene for the first time in EV. 
The condition is checked here for GB I think https://github.com/Ensembl/ensembl-client/blob/dev/src/ensembl/src/content/app/browser/track-panel/trackPanelActions.ts#L142

I can explain this tomorrow if you didnt understand.

jyothishnt:
Also now we are back to the initial issue... the list is getting jumbled

azangru:
What Andrea wants from us is that when user clicks on previously viewed links, they do not change their position (which they do in this branch):

https://user-images.githubusercontent.com/6834224/142857070-206a5154-2761-4256-9a46-7322b2178f56.mp4

But I don't understand how we can both "hide the current entity from the list of previously viewed entities", and avoid visible changes in the list of links to previously viewed entities. These are two mutually incompatible requirements.

azangru (approved):

PR #651, opened 2021-12-02
Update WASM 0.1.2 & fix initial load jump

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature

## Description
- Updated the WASM to 0.1.2 
- Updated the change location logic to prevent an extra jump when the genome browser loads
To reproduce the extra jump issue on 2020.ensembl.org, open the genome browser for brca2, move away to a different location, and refresh the browser. You will notice that the genome browser first loads the location of the focused object and then moves to the previous location.  The fix in this PR will directly load the genome browser with the previous location. 

## Deployment:
http://update-wasm-012.review.ensembl.org

## Views affected
Genome browser

### Other effects
- [x] The PR has as its final commit a WASM/JS update commit if any Rust files were updated.


src/content/app/browser/browser-region-editor/BrowserRegionEditor.test.tsx
```diff
@@ -265,10 +265,10 @@ describe('<BrowserRegionEditor />', () => {
         userEvent.click(submitButton);
         expect(mockChangeBrowserLocation).toHaveBeenCalled();
 
-        const { ensObjectId, chrLocation } =
+        const { focusId, chrLocation } =
           mockChangeBrowserLocation.mock.calls[0][0];
         const [, start, end] = chrLocation;
-        expect(ensObjectId).toBe(null);
+        expect(focusId).toBe(undefined);
```
azangru, line 271: This suggests that the argument to `changeBrowserLocation` should have a `focusId` field; but the object passed to `changeBrowserLocation` doesn't have such a field. If you want to make sure that you are passing an object without this field, you should probably save the argument that `mockChangeBrowserLocation` was called with and inspect it:

```
const newBrowserLocationData = mockChangeBrowserLocation.mock.calls[0][0];
expect('focusId' in newBrowserLocationData).toBe(false);

or

expect(Object.keys(newBrowserLocationData).includes('focusId')).toBe(false);
```

or, alternatively, don't even check this field if you don't care that it's absent.

msmanoj: I've removed this check

src/content/app/browser/hooks/useGenomeBrowser.ts
```diff
@@ -155,23 +158,26 @@ const useGenomeBrowser = () => {
...
     chrLocation: ChrLocation;
   }) => {
     if (!genomeBrowser) {
       return;
     }
 
-    const { genomeId, chrLocation, ensObjectId } = locationData;
+    const { genomeId, chrLocation, focusId } = locationData;
+
+    const { objectId = null } = focusId ? parseFocusIdFromUrl(focusId) : {};
```
azangru, line 170: So `useGenomeBrowser` sends to the genome browser only the stable id of a feature, not the whole "gene:stable_id" thing? Does this perhaps suggest that the focus  url parameter should have been parsed in `useBrowserRouting`, and that the `changeBrowserLocation` function should have received just the gene stable id in the `focusId` field?

msmanoj: >So useGenomeBrowser sends to the genome browser only the stable id of a feature, not the whole "gene:stable_id" thing?
Yes, it just sends the stable id so it makes more sense to parse it in `useBrowserRouting` itself. That's done.

azangru (approved):

PR #655, opened 2021-12-07
Migrate global and genome state to redux toolkit

msmanoj:
## Type

- [ ] Bug fix
- [ ] New feature
- [x] Maintenance

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1402

## Description
- Merged header state with global state
- Migrated genome and global state to use redux toolkit

## Deployment URL
http://global-and-genome-redux-toolkit.review.ensembl.org


src/global/globalSlice.ts
```diff
@@ -43,3 +51,24 @@ export const updateBreakpointWidth: ActionCreator<
       dispatch(setBreakpointWidth(viewportWidth));
     }
   };
+
+const globalSlice = createSlice({
+  name: 'global',
+  initialState: defaultState as GlobalState,
```
azangru, line 57: `defaultState` already knows that it's of type `GlobalState`

azangru (approved):
`genomeSlice` has lots of unnecessary booleans; but we they already were there... We can improve that later; hopefully, we'll migrate this to RTK-query some time.

src/global/globalSlice.ts
msmanoj: Removed the casting

imransl (approved):

PR #656, opened 2021-12-08
Don't open transcript/protein accordion by default if user closed all

ens-ma7:
### Intended behaviour
1. If the user visits the list of transcripts, or of proteins, for a given gene for the first time, expand the default (canonical) transcript.
2. If the user collapses the default transcript, keep its info panel hidden until the user refreshes the page (should work while user switches between different entity viewer views, and between apps)
3. If the user has selected a particular set of filters, or a non-default sorting rule, only expand the default transcript if it is included in the list of displayed transcripts. For example, if the canonical transcript happens to be protein-coding, and the user has filtered out protein-coding transcripts, then when he refreshes the page, no transcript should expand by default.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1351

## Deployment URL
http://transcript-accordion.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/gene-view/hooks/useExpandedDefaultTranscript.ts
+++ b/src/content/app/entity-viewer/gene-view/hooks/useExpandedDefaultTranscript.ts
@@ -37,7 +37,7 @@ type Params = {
 // the canonical transcript is always put on the first place.
 // Knowing this, and also knowing that every gene will always have a canonical transcript,
 // we can simply find the canonical transcript among gene's transcripts, and expand it in the initial view
-const useExpandedCanonicalTranscript = (params: Params) => {
+const useExpandedDefaultTranscript = (params: Params) => {
   const { transcripts, skip = false } = params;
   const dispatch = useDispatch();
   const haveTranscriptsBeenExpanded = useSelector(
@@ -62,4 +62,4 @@ const useExpandedCanonicalTranscript = (params: Params) => {
   }, [params.geneStableId]);
 };
 
-export default useExpandedCanonicalTranscript;
+export default useExpandedDefaultTranscript;
--- a/src/content/app/entity-viewer/gene-view/hooks/useExpandedDefaultTranscript.ts
+++ b/src/content/app/entity-viewer/gene-view/hooks/useExpandedDefaultTranscript.ts
@@ -30,6 +30,7 @@ type TranscriptWithCanonicalMetadata = {
 type Params = {
   geneStableId: string;
   transcripts: TranscriptWithCanonicalMetadata[];
+  skip?: boolean;
 };
 
 // When the default sorting rule is applied to the list of gene's transcripts,
@@ -37,17 +38,17 @@ type Params = {
 // Knowing this, and also knowing that every gene will always have a canonical transcript,
 // we can simply find the canonical transcript among gene's transcripts, and expand it in the initial view
 const useExpandedCanonicalTranscript = (params: Params) => {
+  const { transcripts, skip = false } = params;
   const dispatch = useDispatch();
   const isExpandedTranscriptsListModified = useSelector(
     getExpandedTranscriptsModified
   );
 
   useEffect(() => {
-    if (isExpandedTranscriptsListModified) {
+    if (isExpandedTranscriptsListModified || skip) {
       return; // nothing to do
     }
 
-    const { transcripts } = params;
     const canonicalTranscript = transcripts.find((transcript) =>
       Boolean(transcript.metadata.canonical)
     );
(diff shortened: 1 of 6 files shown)


imransl:
If I follow these steps I still see the transcript accordion opened:

1. Close all transcript accordions
2. Go to `Gene function`
3. Click on a protein accordion
4. Head back to `Transcripts` and see the accordion of the selected protein's transcript opened

I've also attached a screen recording of this below.

https://user-images.githubusercontent.com/32512909/145229481-ca0eb054-6a97-4615-a2f5-4e06893094be.mov

ens-ma7:
> If I follow these steps I still see the transcript accordion opened:
> 
> 1. Close all transcript accordions
> 2. Go to `Gene function`
> 3. Click on a protein accordion
> 4. Head back to `Transcripts` and see the accordion of the selected protein's transcript opened
> 
> I've also attached a screen recording of this below.
> 
>  Screen.Recording.2021-12-08.at.14.51.06.mov

Hi @imransl , thats the right behavior. As soon as you interact with the accordion in either view (Transcript list or protein view) the same view is reflected in the other view. So for example if you have no accordion open in the transcript view and then head to protein view then you wont see any accordion open.  And if you head back to the transcript view you will still see no accordion open.
But now if you interact with the view on protein and decide to open one or two accordion then when you go back to the transcript view the same transcript accordion will be open.

Hope that makes sense.

imransl:
@ens-ma7 Yes that makes sense. Thanks for the explanation.

azangru:
Interesting and clever, but too hacky. It injects a fake transcript id into the list of transcript ids and relies on the knowledge that no real transcript will have an id that matches this fake id. For anyone reading this code, the implementation of this behaviour will be a surprise, while it is generally advisable to have as few surprises in the codebase as possible.

ens-ma7:
> Interesting and clever, but too hacky. It injects a fake transcript id into the list of transcript ids and relies on the knowledge that no real transcript will have an id that matches this fake id. For anyone reading this code, the implementation of this behaviour will be a surprise, while it is generally advisable to have as few surprises in the codebase as possible.

There wont be a transcriptID whose name is called `noselection` :) I could add a comment there to explain for others to understand :)  Otherwise what solution will you suggest?

src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts
```diff
@@ -171,6 +173,14 @@ export const toggleTranscriptInfo =
...
       })
     );
+
+    dispatch(
+      transcriptsSlice.actions.updateAccordionUserInteraction({
+        activeGenomeId,
+        activeEntityId,
+        accordionUserInteraction: true
+      })
+    );
```
azangru, line 183: It doesn't look like we need this action at all. The boolean flag field can be modified directly in the `updateExpandedTranscripts` reducer.

src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts
```diff
@@ -67,6 +68,7 @@ export type Filters = Record<string, Filter>;
 
 const defaultStatePerGene: TranscriptsStatePerGene = {
   expandedIds: [],
+  accordionUserInteraction: false,
```
azangru, line 71: I would prefer a better name (here and for the corresponding selector). For one, `accordionUserInteraction` doesn't suggest that this is going to be a boolean; but just as importantly, this flag isn't even tracking user interactions (you run `toggleTranscriptInfo` automatically in `useEffect` to expand the first transcript by default). Maybe `expandedTranscriptsModified`?

src/content/app/entity-viewer/gene-view/hooks/useExpandedDefaultTranscript.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (canonicalTranscript?.stable_id) {
+      // a bit of defensive programming:
+      // there's something very wrong with our data if a gene doesn't have a canonical transcript;
+      // but at least the code won't crash here
+      dispatch(toggleTranscriptInfo(canonicalTranscript.stable_id));
+    }
+  }, [params.geneStableId]);
+};
+
+export default useExpandedCanonicalTranscript;
```
msmanoj, line 65: I think this file needs to be renamed to `useExpandedCanonicalTranscript.ts`?

msmanoj: Also, would be better to say `Expand` instead of `Expanded`?

azangru: > Also, would be better to say Expand instead of Expanded?

I'd like a noun after the `use`, so that it reads grammatically. This hook is letting me use an expanded default transcript :-)

azangru: > I think this file needs to be renamed to useExpandedCanonicalTranscript.ts?

Hmm, I think I'd rather rename the function from `useExpandedCanonicalTranscript` to `useExpandedDefaultTranscript` 🤔

msmanoj (approved):

PR #659, opened 2021-12-14
Use redux-toolkit-query for entity viewer

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1384

## Description
After some consideration, we decided that requx-toolkit query may be a better data fetching option than apollo client for ensembl-client, because:
- we use different kinds of apis, whereas apollo focuses on graphql apis
- we are already using redux for global state management

This PR is switching the Entity Viewer from apollo client to redux-toolkit query. 

## Deployment URL
http://redux-toolkit-for-entity-viewer.review.ensembl.org

## Views affected
Entity viewer pages

### DIFF
--- a/src/content/app/entity-viewer/EntityViewer.tsx
+++ b/src/content/app/entity-viewer/EntityViewer.tsx
@@ -15,12 +15,10 @@
  */
 
 import React, { useEffect } from 'react';
-import { ApolloProvider } from '@apollo/client';
 import { useSelector, useDispatch } from 'react-redux';
 import { replace } from 'connected-react-router';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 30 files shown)


msmanoj:
Can [this](https://github.com/Ensembl/ensembl-client/blob/redux-toolkit-for-entity-viewer/src/content/app/entity-viewer/state/pageMeta/entityViewerPageMetaSlice.ts#L40) be replaced with `useGeneSummaryQuery` from `src/content/app/entity-viewer/state/api/entityViewerThoasSlice`?

msmanoj (approved):

PR #663, opened 2021-12-21
Add nav buttons to help app

imransl:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
[ENSWBSITES-1386](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1386)

## Description
Add navigation buttons to the Help app

## Deployment URL
http://help-app-nav-buttons.review.ensembl.org

## Views affected
Help app -> any article

### Other effects

### DIFF
--- a/src/content/app/help/Help.tsx
+++ b/src/content/app/help/Help.tsx
@@ -28,7 +28,7 @@ import {
   resetNavHistory
 } from './state/helpSlice';
 import {
-  getCurrentItem,
+  getCurrentHistoryItem,
   getNextItem,
   getPreviousItem
 } from './state/helpSelectors';
@@ -69,7 +69,7 @@ const Help = () => {
     endpoint: `/api/docs/article?url=${encodeURIComponent(pathname)}`,
     skip: isIndexPage
   });
-  const currentHistoryItem = useSelector(getCurrentItem);
+  const currentHistoryItem = useSelector(getCurrentHistoryItem);
   const dispatch = useDispatch();
 
   useEffect(() => {
--- a/src/content/app/help/Help.tsx
+++ b/src/content/app/help/Help.tsx
@@ -14,10 +14,25 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useLocation } from 'react-router';
+import { push } from 'connected-react-router';
+import { useDispatch, useSelector } from 'react-redux';
+import classNames from 'classnames';
 
 import useApiService from 'src/shared/hooks/useApiService';
+
+import {
+  setCurrentItemIndex,
+  resetNavHistory,
+  addHelpItem
+} from './state/helpSlice';
+import {
+  getCurrentItemIndex,
+  getNavHistory,
+  getNextItem,
+  getPreviousItem
+} from './state/helpSelectors';
 
 import ConversationIcon from 'src/shared/components/communication-framework/ConversationIcon';
 import HelpMenu from './components/help-menu/HelpMenu';
@@ -30,16 +45,21 @@ import {
 } from 'src/shared/components/help-article';
 import Breadcrumbs from 'src/shared/components/breadcrumbs/Breadcrumbs';
 
+import { ReactComponent as BackIcon } from 'static/img/browser/navigate-left.svg';
+import { ReactComponent as ForwardIcon } from 'static/img/browser/navigate-right.svg';
+
 import {
   Menu as MenuType,
   MenuItem
 } from 'src/shared/types/help-and-docs/menu';
 import {
+  RelatedArticleData,
   TextArticleData,
   VideoArticleData
 } from 'src/shared/types/help-and-docs/article';
 
 import styles from './Help.scss';
+import navButtonsStyles from 'src/shared/components/help-popup/HelpPopupBody.scss';
 
 type ArticleData = TextArticleData | VideoArticleData;
 
@@ -53,8 +73,16 @@ const Help = () => {
     endpoint: `/api/docs/article?url=${encodeURIComponent(location.pathname)}`,
     skip: isIndexPage
   });
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    return () => {
+      dispatch(resetNavHistory());
+    };
+  }, []);
 
   let breadcrumbs: string[] = [];
+
   if (menu) {
     breadcrumbs = buildBreadcrumbs(menu, { url: location.pathname });
   }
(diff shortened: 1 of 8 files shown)


src/content/app/help/Help.tsx
```diff
@@ -92,14 +120,82 @@ const AppBar = () => {
   );
 };
 
+const NavButtons = () => {
```
azangru, line 123: As we discussed at code review, this looks like it contains what can be a shared component with the following interface:

```ts
type HistoryButtonsProps = {
  onHistoryBack: () => void;
  onHistoryForward: () => void;
  hasPrevious: boolean
  hasNext: boolean
}

const HistoryButtons = () => { ... }
```

The Help component, then, will contain a component that will wrap the `HistoryButtons` component; for example `HelpHistory` or something. This component will read data from the redux state and pass it to the `HistoryButtons` component.

azangru:
During code review, we talked about two different ways of keeping track of the history.

One, implemented in this PR, consists of an array and a pointer:



addition of another history item will both extend the array and move the pointer:



and moving back and forward in history would mean to move the pointer around:



Another way of thinking about the history is as consisting of two stacks: a stack of previous pages, and a stack of next pages. When you add a new page, the stack of the previous pages grows, whereas the stack of the next pages is empty. 



Moving backwards and forwards in history means popping a history item from one stack and pushing it into another:



The code for this alternative implementation will look as follows:

```ts

// redux state

{
  navHistory: {
    previous: [],
    next: []
  };
}

const helpSlice = createSlice({
  name: 'species-page-general',
  initialState,
  reducers: {
    addToHistory(state, action: PayloadAction<string>) {
      const { navHistory, currentItemIndex } = state;
      state.navHistory.previous.push(action.payload)
      state.navHistory.next.length && state.navHistory.next = [];
    },
    moveBackInHistory() {
      const historyItem = state.navHistory.previous.pop();
      state.navHistory.next.push(historyItem);
    },
    moveForwardInHistory() {
      const historyItem = state.navHistory.next.pop();
      state.navHistory.previous.push(historyItem);
    }
})
```

I don't know whether this alternative implementation is any better that the one with the pointer. It looks a bit more elegant to me; but in the end, if you like the one with the pointer more, then that's ok too.

src/content/app/help/state/helpSlice.ts
```diff
@@ -0,0 +1,55 @@
...
+      const newNavHistory = navHistory
+        .slice(0, currentItemIndex + 1)
+        .concat(action.payload);
+
+      state.navHistory = newNavHistory;
+      state.currentItemIndex = newNavHistory.length - 1;
+    },
+    setCurrentItemIndex(state, action: PayloadAction<number>) {
+      state.currentItemIndex = action.payload;
+    },
```
azangru, line 44: I think rather than passing the index from the component to the redux state, it would be cleaner to have two declarative action creators (methods on the `reducers` object), titled `moveForwardInHistory` and `moveBackInHistory`, or something similar. The reducer should be able to handle these actions without any extra directions from the component.

src/content/app/help/Help.tsx
imransl: I have pushed the changes for this ✅

src/content/app/help/state/helpSlice.ts
imransl: This is done as well ✅

azangru:
Further comments:

1) Instead of listening to clicks, we should look at:

```ts
const location = useLocation();

useEffect(() => {
  if (location.pathname !== currentHistoryItem) {
    addPageToHistory()
  }
}, [location.pathname])
```
 
1.1) This useEffect should be running in the Help component (in order to be able to return to the index page)


2) Clicks on the nav buttons should use `goBack` and `goForward` rather than `push`

imransl:
@azangru I have made these changes with the latest commit b70526b

src/content/app/help/Help.tsx
```diff
@@ -78,10 +78,17 @@ const Help = () => {
     };
   }, []);
 
+  useEffect(() => {
+    if (pathname !== currentHistoryItem) {
+      dispatch(addPageToHistory(pathname));
+    } else {
+    }
```
azangru, line 85: ❓

src/content/app/help/state/helpSelectors.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RootState } from 'src/store';
+
+export const getNavHistory = (state: RootState) => state.help.navHistory;
```
azangru, line 19: Doesn't seem to be used anywhere?

src/content/app/help/state/helpSelectors.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getCurrentItem = (state: RootState) => {
```
azangru, line 21: We should do something to the names of these selectors. "Item", as a name, is way too general. We should at least call it "history item"; or perhaps you could think of a better name.

src/content/app/help/Help.tsx
imransl: Forgot to remove this one. Added `no-empty` rule to ESLint to spot these here after.

imransl: ✅

src/content/app/help/state/helpSelectors.ts
imransl: Removed this ✅

src/content/app/help/state/helpSelectors.ts
imransl: Renamed this to `currentHistoryItemIndex`

imransl: ✅

azangru: The same should apply to getNextItem and getPreviousItem.

(And, by, the way, is there any reason to check that `currentHistoryItemIndex > -1`? Wouldn't `return navHistory[currentHistoryItemIndex] ?? null` always work?

imransl: This is done as well ✅

imransl: >(And, by, the way, is there any reason to check that currentHistoryItemIndex > -1? Wouldn't return navHistory[currentHistoryItemIndex] ?? null always work?

In that case it will always result in true even when it is -1:

azangru: > In that case it will always result in true even when it is -1:

I am confused. I am not talking about returning `currentHistoryItemIndex ?? null`, but about returning `navHistory[currentHistoryItemIndex] ?? null`. `navHistory[-1]` will be undefnied.

imransl: > > In that case it will always result in true even when it is -1:
> 
> I am confused. I am not talking about returning `currentHistoryItemIndex ?? null`, but about returning `navHistory[currentHistoryItemIndex] ?? null`. `navHistory[-1]` will be undefnied.

Ah yes that should work I think. Got carried away focussing too much on the -1 😁

imransl: That's done now ✅

src/content/app/help/Help.tsx
```diff
@@ -92,14 +123,52 @@ const AppBar = () => {
...
+
+  const isHelpContent = article.type === 'article' || article.type === 'video';
+
+  if (!isHelpContent) {
     return null;
   }
+
+  const onRelatedItemClick = (reference: RelatedArticleData) => {
+    dispatch(push(reference.url));
+  };
```
azangru, line 170: This function no longer serves any purpose, right?

imransl: It still does as we need to push the URL in the `RelatedArticles` component.

azangru: Why would you need to explicitly push a url on click? Isn't it what the `useEffect` that we added is doing here?

https://github.com/Ensembl/ensembl-client/blob/26ebadbafc76d71c5aa26c2cfee8a7ba406fe0b8/src/content/app/help/Help.tsx#L81-L85

imransl: It is adding that to the help navigation history and not the browser history. For `back` and `forward` navigation we use `goBack` and `goForward` from `connected-react-router`.

https://github.com/Ensembl/ensembl-client/blob/26ebadbafc76d71c5aa26c2cfee8a7ba406fe0b8/src/content/app/help/Help.tsx#L134

imransl: Had a discussion with @azangru and seems like I misunderstood the issue here. I need to get rid of `onRelatedItemClick` as it is not needed anymore. The link will be pushed to the browser history in the `RelatedArticles` components.

imransl: This is done now

✅

azangru (approved):

PR #664, opened 2021-12-22
Add sequence input components for BLAST in Storybook

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1430

## Description
1. Adds components (developed in Storybook) for inputting sequences for the Blast tool. The components behave as follows:
    - accept bare sequences and FASTA-formatted sequences
    - allow file upload
    - detect when a single input contains multiple sequences, and create new input boxes, one per sequence
    - show a button for adding another sequence when a sequence is committed
    - have a "clear all" control to delete all added sequences
    - a delete button on individual input boxes, to delete just these input boxes

2. Adds individual components for PlusButton and DeleteButton
3. Adds sequence input parser and updates the fasta formatter 

## Not yet implemented (to be added in subsequent PRs)
- sequence type guessing (nucleotide vs amino acid sequence)
- connect to redux (in order to coordinate components on the page, and to persist the contents of the input while the user switches between apps)
- update the Upload component (new design; possibly split into specialised Upload components, such as UploadText, UploadFile, etc.)
- any kind of errors

## Deployment URL
https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-blast-blastinputsequences--blast-input-sequences-story


msmanoj:
> detect when a single input contains multiple sequences, and create new input boxes, one per sequence

When the new input boxes are created, should we retain the original input box as it is?

msmanoj:
>show a button for adding another sequence when a sequence is committed 
Do we consider removing the focus from the input as committing?

msmanoj:
1. Enter a random sequence
2. Click on `Add another sequence`
3. Focus on the previous input field
4. Remove the focus or try to focus on the new input field
You will notice that the new input field will go away immediately after removing the focus

azangru:
> Do we consider removing the focus from the input as committing?

In the current implementation, yes; the input is committed when the user pastes a sequence, uploads a file with sequence(s), or unfocuses from the textarea. I really wanted to avoid passing the sequence to the parent component on every typed in character, because the parent component formats the sequences, and it will be a hell to reconcile this with what the user is typing.

The resulting behaviour of the "Add another sequence" sequence button may be a bit disorienting; but I am not yet sure how best to address this.

azangru:
> You will notice that the new input field will go away immediately after removing the focus

Yes, this is another behaviour I uncertain about. It's a result of contradicting requirements: on the one hand, we want the input box to disappear if the user has deleted a sequence in it; on the other hand, we don't want a clean new input box to just randomly disappear.

I agree that the current behaviour is surprising and annoying; but I don't think it's a blocker for this PR. It's something that can be refined further on.

azangru:
> When the new input boxes are created, should we retain the original input box as it is?

Not sure what you mean. Could you demonstrate what, in the current behaviour of the inputs, has raised this question?

msmanoj:
> > When the new input boxes are created, should we retain the original input box as it is?
> 
> Not sure what you mean. Could you demonstrate what, in the current behaviour of the inputs, has raised this question?

When I try to enter two different sequences in a single input box and commit it, it gets split and put into two new input boxes, and keeps the original input box as it is. Is this behavior expected?

azangru:
> When I try to enter two different sequences in a single input box and commit it, it gets split and put into two new input boxes, and keeps the original input box as it is. Is this behavior expected?

When I enter this sequence, with two FASTA headers, in a clear box:

```
>ENST00000350908.9 Ensembl Transcript chromosome:GRCh38:7:1:159345973:1
ACAGACATGAAAGCTAACCGAGGACTTGAGAGACTCAAACTGGTGCTTTTGTCTCTCTCT
CTCTGTCTTTCTCTCTCTCACACACACACTCACACACTCACACACATGCACACACACACA
TACACACACACAAAAATGAAGCACTTACTTTAGAAAGATTATGGTAAGCATGCTGGCTCA
GTCTTGAACCTTTGTCACCCCTCACGTTGCACACCAAAGACATACCCTAGTGATTAAATG
CTGATTTTGTGTACGATTGTCCACGGACGCCAAAACAATCACAGAGCTGCTTGATTTGTT
>ENST00000350908.9 Ensembl Transcript chromosome:GRCh38:7:1:159345973:1
TACACACACACAAAAATGAAGCACTTACTTTAGAAAGATTATGGTAAGCATGCTGGCT
```

I am getting the following:

image

Is this what you are seeing? From what you are describing, you are getting something different? Could you post a sequence that you are testing and the result that you are getting?

imransl:
I noticed some strange behaviour when I copy paste similar or two different sequences in the same input box and then keep doing that between two different input boxes as depicted in the video.

https://user-images.githubusercontent.com/32512909/147236316-48e791ec-377a-4238-be3a-eb9904c6b61d.mov

azangru:
I've pushed an ugly hack that seems to fix the issue with multiple sequences in an input box.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  const onClear = () => {
+    canSubmit.current = false; // lock against the sequence submission on blur
+    setInput('');
+    sequence.value && props.onRemoveSequence?.(index);
+    setTimeout(() => (canSubmit.current = true), 0);
+  };
+
+  const textareaClasses = classNames(styles.textarea);
```
msmanoj, line 100: Can we directly use `styles.textarea`?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
```diff
@@ -0,0 +1,118 @@
...
+  useEffect(() => {
+    if (!sequences.length) {
+      setShouldAppendEmptyInput(true);
+    }
+  });
+
+  const onSequenceAdded = (input: string, index: number | null) => {
+    const parsedSequences = parseBlastInput(input).filter((result) =>
+      Boolean(result.value)
+    ) as ParsedInputSequence[]; // <-- this would make a header without a sequence disappear; it is a good idea?
```
msmanoj, line 41: I don't think it's a good idea. 

If I start by typing the header and then I try to copy and paste the sequence from somewhere else, I lose the header.

src/content/app/tools/blast/utils/blastInputParser.ts
```diff
@@ -0,0 +1,95 @@
...
+    // this is a FASTA header; return it as is
+    return line;
+  }
+
+  // remove all digits and spaces from the line
+  line = line.replace(/[\d\s]+/g, '');
+  return line.toUpperCase();
+};
+
+const endsInEmptySequenceObject = (sequences: ParsedInputSequence[]) => {
```
msmanoj, line 87: How is the name of this function related to what it does?

src/content/app/tools/blast/utils/blastInputParser.ts
```diff
@@ -0,0 +1,95 @@
...
+      const newSequence = createEmptySequence();
+      sequences.push(newSequence);
+    } else {
+      if (index === 0) {
+        // this is the first line of a bare sequence; create the sequence object first
+        const newSequence = createEmptySequence();
+        sequences.push(newSequence);
+      }
+      const sequenceObject = sequences[sequences.length - 1];
+      sequenceObject.value += line;
```
msmanoj, line 67: Is this used anywhere?

src/shared/components/plus-button/PlusButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import React, { type HTMLAttributes } from 'react';
+import { ReactComponent as PlusIcon } from './plus-icon.svg';
+
+import styles from './PlusButton.scss';
+
+type Props = Omit<HTMLAttributes<HTMLButtonElement>, 'children'>;
+
+const DeleteButton = (props: Props) => {
```
imransl, line 24: I am a bit confused about the naming of this button as `DeleteButton`. Isn't this supposed to be the `PlusButton`?

azangru: Ouch! 🤦 My bad! Must have copy-pasted code from the DeleteButton component that I had created earlier. I was sure I renamed it!

src/content/app/tools/blast/utils/blastInputParser.ts
azangru: If you refer to the `parseBlastInput` function, you will notice that there is a `sequences` object there that is an array of parsed sequence objects.

As a string passed to the `parseBlastInput` gets parsed line by line:



the `sequences` object grows. If the input contains an empty line, the parser will add a new empty sequence object to the sequences array:



You can see this happening on line 56 of this file.

This is what the `endsInEmptySequenceObject` function is for. It checks if the `sequences` array has an empty sequence object in the end, and acts accordingly — either populates this empty sequence object with data or creates a new empty sequence object and adds data to it.

If there is a better way of keeping track of when/how to add parsed sequences to the array of sequences, I am open to suggestions.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
azangru: ✅

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
azangru: Yes, I had similar apprehensions. I removed the filter. Probably should filter this out at submission step.

src/shared/components/plus-button/PlusButton.tsx
azangru: ✅

src/content/app/tools/blast/utils/blastInputParser.ts
msmanoj: I tried the following and it seems to work:
```javascript
const sections = input.split('\n\n');
  sections.forEach((section, index) => {
    const lines = section.split('\n');
    lines.forEach(line => {
      if (line.startsWith('>')) {
        line = line.replace(/^\>\s*/, '');
        sequences[index] = { header: line, value: '' }
      } else if(sequences[index]){
        const sequenceObject = sequences[index];
        sequenceObject.value += line;
      } else {
        sequences[index] = { value: line };
      }
    });
  });

```

msmanoj: Sorry, it's this one which passes the tests:

```javascript
  input = input.trim();
  const sequences: ParsedInputSequence[] = [];
  // Replace single line separator with double line 
  input = input.replaceAll('\n\n>', '\n>').replaceAll('\n>', '\n\n>');
  const sections = input.split('\n\n');
  sections.forEach((section, index) => {
    const lines = section.split('\n');
    lines.forEach(line => {
      line = prepareLine(line);
      if (line.startsWith('>')) {
        line = line.replace(/^\>\s*/, '');
        sequences[index] = { header: line, value: '' }
      } else if(sequences[index]){
        const sequenceObject = sequences[index];
        sequenceObject.value += line;
      } else {
        sequences[index] = { value: line };
      }
    });
  });
```

jyothishnt:
Few things noticed on the UI

- Allows duplicate sequences whereas current site doesn't
- As soon as I paste sequences with new lines it splits into multiple (not onblur) but if I am typing a sequence (which rarely happens I think) it allows me to add multiple sequences on new lines. Is this the correct behaviour? Is it looking for a copy paste command or onblur? 
- Deleting sequences either randomly hides the entire box or clears the textarea and leave the box as it is. I guess this is a bug? 
- height of body changes as soon as you start typing the sequence because we remove the file upload. But when I delete my sequence file upload appears and height increases
- "Add another sequence" link flickers on blur positioned top right on the header. But when a seqeunce is added it appears outside the box.

azangru:
@jyothishnt regarding your points:

These three I wouldn't worry too much about:

> Allows duplicate sequences whereas current site doesn't

Yes, that's correct. I am not sure how to handle the duplicates in the UI; making the boxes just disappear feels too abrupt and surprising to me. I suggest we address this at a later stage, when we consider error handling. Is there _really_ that much value beyond removing duplicates? Will allowing duplicates actually be a problem?

> As soon as I paste sequences with new lines it splits into multiple (not onblur) but if I am typing a sequence (which rarely happens I think) it allows me to add multiple sequences on new lines. Is this the correct behaviour? Is it looking for a copy paste command or onblur?

It listens to both. It reacts immediately to a paste command, but if you enter a sequence manually or edit a sequence, it waits for a blur event.

> height of body changes as soon as you start typing the sequence because we remove the file upload. But when I delete my sequence file upload appears and height increases

The current appearance of the upload button is temporary; Andrea was working of a new consistent design for upload buttons, which I will add in a subsequent PR

---

> Deleting sequences either randomly hides the entire box or clears the textarea and leave the box as it is. I guess this is a bug?

I was trying to follow this logic:
- for any input boxes that have been added after the first one, if the user deletes its content and clicks out of the textarea, thus leaving the textarea empty, remove the box;
- if there is only one input box on the screen, then if the user deletes its content, just reset the input box to its initial state

Do you think it's the wrong logic, or is the code not working according to this logic?

> "Add another sequence" link flickers on blur positioned top right on the header. But when a seqeunce is added it appears outside the box.

I'll see if I can do anything about the flicker.

jyothishnt:
> I am not sure how to handle the duplicates in the UI; making the boxes just disappear feels too abrupt and surprising to me. I suggest we address this at a later stage, when we consider error handling. Is there _really_ that much value beyond removing duplicates? Will allowing duplicates actually be a problem?

jyothishnt:
> @jyothishnt regarding your points:
> 
> These three I wouldn't worry too much about:
> 
> > Allows duplicate sequences whereas current site doesn't
> 
> Yes, that's correct. I am not sure how to handle the duplicates in the UI; making the boxes just disappear feels too abrupt and surprising to me. I suggest we address this at a later stage, when we consider error handling. Is there _really_ that much value beyond removing duplicates? Will allowing duplicates actually be a problem?

The current site shows a message, so may be we can include a message along with splitting the sequences logic? If you include duplicate sequence on submit, it will unnecessarily increase the number of jobs and eventually longer runtimes. 

> 
> > As soon as I paste sequences with new lines it splits into multiple (not onblur) but if I am typing a sequence (which rarely happens I think) it allows me to add multiple sequences on new lines. Is this the correct behaviour? Is it looking for a copy paste command or onblur?
> 
> It listens to both. It reacts immediately to a paste command, but if you enter a sequence manually or edit a sequence, it waits for a blur event.
OK
> 
> > height of body changes as soon as you start typing the sequence because we remove the file upload. But when I delete my sequence file upload appears and height increases
> 
> The current appearance of the upload button is temporary; Andrea was working of a new consistent design for upload buttons, which I will add in a subsequent PR
OK
> 
> > Deleting sequences either randomly hides the entire box or clears the textarea and leave the box as it is. I guess this is a bug?
> 
> I was trying to follow this logic:
> 
> * for any input boxes that have been added after the first one, if the user deletes its content and clicks out of the textarea, thus leaving the textarea empty, remove the box;
> * if there is only one input box on the screen, then if the user deletes its content, just reset the input box to its initial state
> 
> Do you think it's the wrong logic, or is the code not working according to this logic?
I think the logic sounds good, we dont want to leave empty boxes in between if there are multiple sequences. But unfortunately it doesnt work according to the logic.
> 
> > "Add another sequence" link flickers on blur positioned top right on the header. But when a seqeunce is added it appears outside the box.
> 
> I'll see if I can do anything about the flicker.
Thanks

src/content/app/tools/blast/utils/blastInputParser.ts
azangru: I removed the `endsInEmptySequenceObject` checker, but I didn't use `input.replaceAll('\n\n>', '\n>').replaceAll('\n>', '\n\n>');`

azangru:
@jyothishnt I tried to address your points about the flicker and about the incorrect behaviour upon deletion. Could you please check?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
azangru: ✅

jyothishnt:
Thanks it looks ok now.

jyothishnt (approved):
Functionality tested and bugs reported which are now fixed. Looks fine to me 
I hope others are looking at the code

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.body {
+  display: grid;
+  grid-template-columns: 1fr auto;
+  padding: 10px 65px 12px 28px;
+  background-color: $light-grey;
+  border: $border;
+
+  // FIXME: this is a hack
+  [class*='defaultUpload'] {
```
msmanoj, line 27: Can this not be passed as a prop?

msmanoj: We don't have to worry about this now as the styles for the upload component will be changing soon.

msmanoj (approved):

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.scss
azangru: This is a temporary hack anyway. The upload component that is targeted by this selector will be updated by Andrea; so I expect this css rule to be deleted then.

PR #665, opened 2021-12-25
Add empty Blast pages

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1436

## Description
- Added the `src/content/app/tools` folder, with `blast` as a subfolder
- Added an empty Blast Search page to the `/blast` route
- Added an empty Blast Jobs page to the `/blast/jobs` route
- Added a Blast button to the launchbar, currently only available for development

## ALSO
- Because of a new set of app icons, which have a different padding within the svg canvas, created individual AppIcon components to capture the ratio between the size of the icon box and the size of the image inside of the box. This change had ramifications in various other components.

## Deployment URL
http://add-blast-pages.review.ensembl.org


src/content/app/tools/blast/BlastPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </Helmet>
+      {hasMounted && (
+        <Switch>
+          <Route exact path="/blast">
+            <BlastSearch />
+          </Route>
+          <Route exact path="/blast/jobs">
+            <BlastJobs />
+          </Route>
+        </Switch>
```
azangru, line 49: Note to self: remember that this is react-router 5; will need to update this for react-router 6.

src/global/globalConfig.ts
```diff
@@ -36,16 +36,15 @@ export enum AppName {
   GENOME_BROWSER = 'Genome browser',
   SPECIES_SELECTOR = 'Species selector',
   CUSTOM_DOWNLOADS = 'Custom downloads',
-  ENTITY_VIEWER = 'Entity viewer'
+  ENTITY_VIEWER = 'Entity viewer',
+  TOOLS = 'Tools' // this is the name of a group of apps
 }
```
azangru, line 41: It seems that the only purpose of this enum is to be used as a title of an app passed to the AppBar component. Perhaps consider removing it?

src/shared/components/app-icon/AppIcon.scss
```diff
@@ -0,0 +1,61 @@
...
+  background-color: var(--app-icon-bg-color, transparent);
+
+  svg {
+    fill: white;
+  }
+}
+
+.globalSearch {
+  svg {
+    width: 62.5%;
```
ens-ma7, line 16: why are we using %? why are we not specifying the width in px? % should be avoided isn't it?

src/shared/components/app-icon/GenomeBrowserIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import classNames from 'classnames';
+
+import { ReactComponent as SVGIcon } from 'static/img/launchbar/browser.svg';
```
ens-ma7, line 20: for consistency purpose, should the svg not be called genome-browser.svg

src/shared/components/app-icon/AppIcon.scss
azangru: > why are we not specifying the width in px?

Because what I am after is the ratio between the size of the `.iconWrapper` class and the size of the svg inside of it. This allows the svg to respond to the dimensions of the `iconWrapper`. Consider the help page, where the same icons are rendered at different sizes (smaller in the grey panel below): 

image

src/shared/components/app-icon/GenomeBrowserIcon.tsx
azangru: 🤷‍♂️  
We'll figure it out in [this ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1456), when we organise our svg icons. Andrea's name of the file was different still (`icon_launchbar_browser.svg`). We'll worry about that then.

src/global/globalConfig.ts
ens-ma7: Maybe the mapping should `BLAST = 'Blast'` we dont need an entry for tools isn't it?

src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const speciesTabs = speciesList.map((species, index) => (
+    <SelectedSpecies key={index} species={species} onClick={noop} />
+  ));
+  const speciesSelectorLink = useMemo(() => {
+    return <Link to={urlFor.speciesSelector()}>Change</Link>;
+  }, []);
+
+  const wrappedSpecies = (
+    <SpeciesTabsWrapper
+      isWrappable={false}
```
ens-ma7, line 43: Are we sure we want it to be wrappable? I thought i heard Andrea saying that it will be different for tools?

ens-ma7:
There is a very weird behavior happening with the "About using ensembl" link on the homepage and i think it might be related to the css rules issue with webpack. so the way i have been able to replicate this:
1) Go to deploy URL (no species selected)
2) then click on genome browser link in the app bar
3) Then click on the home icon (top left icon)
4) The "About using ensembl" link will not be bold.

image

src/global/globalConfig.ts
azangru: This is what the mockup currently shows. Could be Andrea's mistake; I didn't ask 🤷‍♂️ :

image

As far as I can tell, it is the only use of this enum. If so, we don't really need it that much at all.

azangru:
> There is a very weird behavior happening with the "About using ensembl" link on the homepage and i think it might be related to the css rules issue with webpack. so the way i have been able to replicate this:

Yep, this is already present in production. Calls for a separate ticket.

These two classes in the same css file built by webpack compete with each other:





I don't understand how it is possible that the earlier defined class ever wins; but yeah, they both have the same specificity, which is a problem.

src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar.tsx
azangru: I believe the intention was to make the list of species wrappable, i.e. that the list would grow vertically rather than squeeze species lozenges to remain on a single line. Which means that `isWrappable={false}` is incorrect; I've pushed a commit to remove it.

src/global/globalConfig.ts
jyothishnt: Is that launchbar icon representing tools or blast?

src/content/app/tools/blast/BlastSearch.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import ToolsAppBar from 'src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar';
+import ToolsTopBar from 'src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar';
+
+const BlastSearch = () => {
```
jyothishnt, line 22: I like this name but still was wondering if it is better to call BlastForm or BlastInputForm?

src/global/globalConfig.ts
azangru: The icon is representing blast.

jyothishnt:
Looks ok to me.

ens-ma7:
> Yep, this is already present in production. Calls for a separate ticket.
Created a separate ticket for this:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1461

src/content/app/tools/blast/BlastSearch.tsx
azangru: Renamed BlastSearch to BlastForm

ens-ma7 (approved):

msmanoj (approved):

PR #666, opened 2021-12-30
Update to react-router v6

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1414

## Description
- Update React Router to version 6
  - See [the docs](https://reactrouter.com/docs/en/v6/upgrading/v5#remove-activeclassname-and-activestyle-props-from-navlink-) about the changes required for the update
- Remove connected-react-router. Maintainers of react-router [advise](https://v5.reactrouter.com/web/guides/deep-redux-integration) against integrating react-router into redux; unfortunately, we didn't listen to them when we started work on ensembl-client
  - As a consequence, the code for reporting analytics on page navigation, which was inside a redux middleware, has to be moved into a hook
- Update all dependencies

## Important changes
- React router 6 has [dropped](https://reactrouter.com/docs/en/v6/faq#what-happened-to-regexp-routes-paths) a powerful `path-to-regexp` library that they used for path matching. Which means that it no longer allows optional parameters (e.g. `/genome-browser/:genomeId?`). While not a deal-breaker, it's definitely a bummer.
- `StaticRouter` in react-router 5 [had a dedicated `context` property](https://v5.reactrouter.com/web/api/StaticRouter/context-object), which was removed in v6. 

## Deployment URL
http://update-react-router.review.ensembl.org


src/content/app/App.tsx
```diff
@@ -41,47 +40,52 @@ const AppContainer = () => {
...
-            key={index}
-            path={route.path}
-            exact={route.exact}
-            render={(props) => <route.component {...props} />}
-          />
-        ))}
-        <Route component={NotFound} />
-      </Switch>
+      {routes}
+      {/*  remember to handle the 404's (update commented code below) */}
```
msmanoj, line 64: I guess this is for this PR itself?

src/shared/hooks/useUrlParams.ts
```diff
@@ -0,0 +1,52 @@
...
+  }, {}) as Partial<Record<ParamKey, string>>;
+  // if (Object.keys(params).length) {
+  //   return params as Record<ParamKey, string>;
+  // } else {
+  //   return null;
+  // }
+};
+
+// exported as a React hook
+export const useUrlParams = getPathParameters;
```
msmanoj, line 52: Why do we need to export it as a hook here?

azangru: Hmm, the way it's written, no real reason.

I can do one of two things:

1) Expand it into an actual hook that uses the `useLocation` hook inside, so that it no longer needs to receive the `pathname` from the React component. So, while it will still use the `getPathParameters` function, it will only need to receive the `patternOrPatterns` from the React component.
2) Or, alternatively, I can remove `useUrlParams`, and move the file (which will no longer contain any hooks) to `src/routes`.

Not sure which is better yet, but leaning slightly towards option 1 🤔

src/shared/hooks/useUrlParams.ts
```diff
@@ -0,0 +1,52 @@
...
+      : patternOrPatterns;
+  return patterns.reduce((params, pattern) => {
+    const matchResult = matchPath(pattern, pathname);
+    return { ...params, ...matchResult?.params };
+  }, {}) as Partial<Record<ParamKey, string>>;
+  // if (Object.keys(params).length) {
+  //   return params as Record<ParamKey, string>;
+  // } else {
+  //   return null;
+  // }
```
azangru, line 48: Note to self: remove the commented code.

src/content/app/App.tsx
azangru: Yeah, the 404s are being handled; I'll need to clean up this file.

src/shared/hooks/useUrlParams.ts
msmanoj: option 1 sounds good to me as well

src/content/app/App.tsx
```diff
@@ -41,47 +40,52 @@ const AppContainer = () => {
...
-    </Status>
-  );
-};
+// const NotFound = () => {
+//   return (
+//     <Status code={404}>
+//       <NotFoundErrorScreen />
+//     </Status>
+//   );
+// };
```
imransl, line 89: Will this code get uncommented later?

src/root/rootReducer.ts
```diff
@@ -28,15 +27,14 @@ import speciesSelector from 'src/content/app/species-selector/state/speciesSelec
...
-const createRootReducer = (history: any) =>
+const createRootReducer = () =>
   combineReducers({
     browser,
     customDownload,
     communication,
     inAppSearch,
     genome,
     global,
-    router: connectRouter(history),
```
imransl, line 39: I am wondering how this will affect #663 as it depends on `goBack` and `goForward` in `connected-react-router`.

imransl: Okay found the answer to this and also now recall we discussed about this before Christmas last year: https://github.com/remix-run/react-router/blob/5730e28a7472e3a1b967d95c5c4cc643c570d5c7/docs/upgrading/v5.md

src/shared/hooks/useUrlParams.ts
imransl: Agree with @msmanoj. Option 1 looks better.

azangru: Updated the `useUrlParams` hook.

src/content/app/App.tsx
azangru: No. I removed it.

src/content/app/entity-viewer/EntityViewer.tsx
```diff
@@ -70,12 +63,26 @@ const EntityViewer = () => {
...
+const EntityViewerForGene = () => {
+  const isSidebarOpen = useSelector(isEntityViewerSidebarOpen);
   const isSidebarModalOpen = Boolean(
     useSelector(getEntityViewerSidebarModalView)
   );
-
-  const params: EntityViewerParams = useParams();
-  const { genomeId, entityId } = params;
+  const viewportWidth = useSelector(getBreakpointWidth);
+  const { genomeId, entityId } = useParams<'genomeId' | 'entityId'>();
```
msmanoj, line 84: Here and elsewhere, why do we need to use `<'genomeId' | 'entityId'>`? `check-types` doesn't seem to complain if I remove them.

azangru: > check-types doesn't seem to complain if I remove them.

Hover in your devtools over the destructured variables with and without the string types passed to the useParams. Is there any difference?

msmanoj: It seems to be the same for me:

Screenshot 2022-01-10 at 21 34 59
Screenshot 2022-01-10 at 21 35 12

azangru: If you pass to the expected keys as a union type, it will know which keys to expect on the returned object:



as opposed to:



Which means that now it gives you autocomplete:

image

as opposed to nothing:



and will complain if you try to access an undeclared field:



as opposed to:

msmanoj (approved):

PR #668, opened 2022-01-09
Change the implementation of help history

azangru:
## Description
This is an experimental PR that simplifies the code that keeps track of the history of navigation within the Help app, and better synchronises it with the browser history.

This PR fixes two problems present in dev:
- Currently, pressing the browser back button will confuse the help history buttons (browser back will be interpreted as adding a new page to the help history stack).
- Pressing and holding the browser back button, and selecting a more distant history item (two pages back, three pages back, etc.) will definitely confuse the help history buttons.

There are still problems that remain with the implementation suggested in this PR. For example, refreshing the page or switching to a different app and returning back to the help app will confuse the history buttons. But these problems also exist with the implementation currently in dev.

NOTE: this PR depends on PR https://github.com/Ensembl/ensembl-client/pull/666 being merged first.

## Deployment URL
http://use-help-history.review.ensembl.org

## Views affected
Help page


imransl:
This implementation is neater than using Redux for this. Also, didn't see any issues while testing this with the back and forward browser buttons. However, noticed an issue which definitely would have been with the previous implementation as well. This occurs when clicking on a video help link - the navigation buttons disappear. Probably something to ask from Andrea.

https://user-images.githubusercontent.com/32512909/149005267-995138c2-5405-403e-9982-1a0f35eeffc2.mov

azangru:
> However, noticed an issue which definitely would have been with the previous implementation as well. This occurs when clicking on a video help link - the navigation buttons disappear.

Pushed an update to address this.

imransl (approved):

msmanoj (approved):
Looks good

PR #669, opened 2022-01-10
Detect unsupported browsers using client-side strategy

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1341

## Description
This PR attempts to detect unsupported browsers and redirect the user to a dedicated page informing them that their browser is not supported.

### How it works
- old browsers that do not support ES modules and do not recognize `script type="nomodule"` will get redirected immediately on the client side
- newer browsers that also have unsupported features will be redirected by running a script that checks the availability of relevant features

### Notes
- the unsupported page is server-side-rendered with React, but the entry point is separate from the rest of the app, and thus the javascript is much simpler

## Deployment URL
http://browser-fallback.review.ensembl.org

## Google's endorsement
While Google, generally, doesn't object to server-side user agent sniffing, it seems to prefer client-side feature detection. For example, in a [blog post](https://developers.google.com/search/blog/2019/10/updating-user-agent-of-googlebot) about an update to Googlebot's user agent string, they wrote:

> Sites that follow our recommendations to use feature detection and progressive enhancement instead of user agent sniffing should continue to work without any changes.


ens-ma7:
I have tested this in IE11, IE10 and IE9. It look fine in IE11 but styling start to be broken in IE10 and in IE9 its quite bad. Not sure we can do much.

image

src/content/unsupported-browser/components/supported-web-browser/SupportedWebBrowser.scss
```diff
@@ -0,0 +1,18 @@
+@import 'src/styles/common';
+
+.container {
+  display: flex;
```
ens-ma7, line 4: these wont be supported in older browser such as IE9 (see screenshot in one of the comment). Probably best to use the old float styling or display: block and aligning the images.

src/content/unsupported-browser/UnsupportedBrowser.scss
```diff
@@ -0,0 +1,45 @@
...
+
+html,
+body,
+[class='ens-app'] {
+  min-width: initial !important; // !important seems to be the only way to override the min-width set in styles/main.scss
+}
+
+.container {
+  height: 100%;
+  display: flex;
```
ens-ma7, line 11: same here, `display: flex` wont be supported in old browser (IE10 and below) and the styling will be broken.

ens-ma7 (approved):

ens-ma7:
Tested in IE11, IE10, IE9 and IE8 and the page displays fine.

msmanoj (approved):

PR #672, opened 2022-01-14
Circular chromosome indicator

jyothishnt:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-265

## Description
The navigation panel should have an icon when browsing species that have a circular genome. For others we can show the name of the chromosome. Will need the BE to indicate whether we have a circular chromosome or not

Design at:
https://xd.adobe.com/view/37352bdd-fdc5-43f2-93a8-47cba4ee7039-6266/screen/b0fdc853-de27-494e-b07e-ca393e827cab?fullscreen

## Deployment URL
http://circular-chromosome-indicator.review.ensembl.org/

## Views affected
Genome browser

### Other effects
NA

## Possible complications
NA

### DIFF
--- a/src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.test.tsx
+++ b/src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.test.tsx
@@ -20,6 +20,7 @@ import { render } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import userEvent from '@testing-library/user-event';
 import configureMockStore from 'redux-mock-store';
+import set from 'lodash/fp/set';
 
 import { getCommaSeparatedNumber } from 'src/shared/helpers/formatters/numberFormatter';
 
@@ -78,6 +79,17 @@ describe('BrowserLocationIndicator', () => {
       const { container } = renderComponent();
       const renderedName = container.querySelector('.chrCode');
       expect(renderedName?.textContent).toBe(chrName);
+    });
+
+    it('displays circular chromosome', () => {
+      const newstate = set(
+        'genome.genomeKaryotype.genomeKaryotypeData.human[0].is_circular',
+        true,
+        mockState
+      );
+      const { container } = renderComponent(newstate);
+      const circularIndicator = container.querySelector('.circularIndicator');
+      expect(circularIndicator).toBeTruthy();
     });
 
     it('displays location', () => {
--- a/src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.test.tsx
+++ b/src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.test.tsx
@@ -49,6 +49,13 @@ const mockState = {
       },
       activeGenomeId: 'human'
     }
+  },
+  genome: {
+    genomeKaryotype: {
+      genomeKaryotypeData: {
+        human: [{ is_circular: false }]
+      }
+    }
   }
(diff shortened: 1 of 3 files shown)


src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.tsx
```diff
@@ -51,7 +54,11 @@ export const BrowserLocationIndicator = (props: Props) => {
     <div className={className}>
       <div className={styles.chrLabel}>Chromosome</div>
       <div className={styles.chrLocationView} {...onClickProps}>
-        <div className={styles.chrCode}>{chrCode}</div>
+        {genomeKaryotype && genomeKaryotype[0].is_circular ? (
```
azangru, line 57: We can do better than this. Rather than inspecting the first chromosome from the karyotype, we should look at the one that is actually shown in the genome browser. You can find it in the redux store at browser → browserGeneral → actualChromosomeLocations:

image

One problem though: e. coli is currently broken in genome browser; so the genome browser doesn't show anything in the canvas and can't report anything about its location in the E.coli genome. See https://github.com/Ensembl/ensembl-dauphin-style-compiler/issues/23 I suspect it may block the work on this PR.

ens-ds23: e coli should only be broken if there's a focus object as I understand it. This shouldn't block testing the indicator, I don't think.

jyothishnt: Thanks Dan and Andrey. I have updated the code.

src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.test.tsx
```diff
@@ -73,6 +81,17 @@ describe('BrowserLocationIndicator', () => {
...
+    it('displays circular chromosome', () => {
+      const newstate = set(
+        'genome.genomeKaryotype.genomeKaryotypeData.human[0].is_circular',
+        true,
+        mockState
+      );
+      const { container } = renderComponent(newstate);
+      const circularIndicator = container.querySelector('.circularIndicator');
+      expect(circularIndicator).toBeTruthy();
+    });
```
azangru, line 93: This test seems to be failing at the moment. From CI/CD logs:

image

src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.tsx
```diff
@@ -51,7 +52,11 @@ export const BrowserLocationIndicator = (props: Props) => {
     <div className={className}>
       <div className={styles.chrLabel}>Chromosome</div>
       <div className={styles.chrLocationView} {...onClickProps}>
-        <div className={styles.chrCode}>{chrCode}</div>
+        {chrCode === 'Chromosome' ? (
```
azangru, line 55: You need to check the `isCircular` field from the karyotype response to check whether the chromosome should be circular. `chrCode` here is just the chromosome name, which should be used to check the response from the karyotype endpoint:

image

jyothishnt: Aah gotcha. Will update it soon

PR #675, opened 2022-01-17
Refactor checkboxes and radio buttons

imransl:
## Related JIRA Issue(s)
[ENSWBSITES-1411](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1411)
[ENSWBSITES-1489](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1489)

## Description
1. Update visual appearance of checkboxes and radio buttons in accordance with the spec ([see XD](https://xd.adobe.com/view/c25e6b69-74b8-4f31-bc41-3f9d5685223a-970d/?fullscreen)). Add a `theme` property to capture the CSS rules that go together.
2. Refactor the styling of the checkbox and the radio buttons to enable more flexible and reliable customisation. Instead of multiple class names passed to the component from the parent as parameters, treat the checkbox and the radio button components as one would treat custom html elements:
  - expose relevant customisable properties as CSS variables
  - allow a single `className` property to pass a class that gets attached to the outermost element of the component, for necessary styling
3. Enable both vertical and horizontal orientation of radio buttons.

### Also changed
Removed the click handler from the pointer box, which wasn't doing anything but prevented click default behaviour and propagation. This doesn't seem to be necessary any longer (we have a dedicated `useOutsideClick` for what this click handler was probably trying to achieve), and was actually breaking the default click handling by labels wrapping a checkbox input.

## TODO in different PRs
- The `disabled` property for the radio group needs updating. Currently, it belongs to the whole group, which doesn't make sense. It should, instead, be moved into the `option` data structure.
- Explore if it's possible to update the type of radio group props to use a generic type so that typescript could infer the type of the value passed to the callback function.

## Deployment URL
http://refactor-checkboxes.review.ensembl.org/

## Views affected
- Genome browser:
  - track settings
  - sequence download in zmenu and drawer
- Entity viewer
  - downloads: instant transcript and protein download
  - transcripts filter and sorting panel
- Custom download (unimportant)

### DIFF
--- a/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
+++ b/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
@@ -29,6 +29,11 @@
   margin-bottom: 5px;
 }
 
+.checkbox {
+  height: 14px;
+  width: 14px;
+}
+
 .collapseButtonWrapper {
   position: relative;
   width: 20px;
--- a/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
+++ b/src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
@@ -29,11 +29,6 @@
   margin-bottom: 5px;
 }
 
-.checkbox {
-  height: 14px;
-  width: 14px;
-}
-
 .collapseButtonWrapper {
   position: relative;
   width: 20px;
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -31,6 +31,7 @@
   display: grid;
   grid-template-columns: repeat(3, 210px);
   height: auto;
+  --checkbox-margin-top: 9px;
 }
 
 .header {
@@ -44,9 +45,5 @@
 }
 
 .radio {
-  margin-top: 9px;
-}
-
-.checkboxDefault {
   margin-top: 9px;
 }
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -45,13 +45,6 @@
 
 .radio {
   margin-top: 9px;
-
-  &Checked {
-    &:after {
-      width: 6px;
-      height: 6px;
-    }
-  }
 }
 
 .checkboxDefault {
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -30,7 +30,7 @@
 .filterContent {
   display: grid;
   grid-template-columns: repeat(3, 210px);
-  height: auto;
+  padding-left: 12px;
 }
 
 .header {
@@ -47,6 +47,6 @@
   display: block;
 }
 
-.radio {
-  margin-top: 9px;
+.radioGroup {
+  margin-left: 12px;
 }
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -2,11 +2,11 @@
 @import 'src/content/app/entity-viewer/gene-view/styles/constants';
 
 .container {
-  grid-area: 3/1/3/4;  
+  grid-area: 3/1/3/4;
   position: sticky;
   height: 100%;
   top: -5px;
-  z-index: 1;     
+  z-index: 1;
   padding-bottom: 9px;
 }
 
@@ -24,7 +24,7 @@
 }
 
 .sortContainer {
-    padding-left: 160px;
+  padding-left: 160px;
 }
 
 .filterContent {
@@ -43,15 +43,7 @@
   display: block;
 }
 
-.label {
-  color: white;
-  font-size: 12px;
-}
-
 .radio {
-  border-color: $dark-grey;
-  height: 12px;
-  width: 12px;
   margin-top: 9px;
 
   &Checked {
@@ -62,18 +54,6 @@
   }
 }
 
-.checkboxUnchecked {
-  background-color: transparent;
-  border-color: $dark-grey;
-  height: 12px;
-  width: 12px;
-  margin-top: 9px;
-}
-
-.checkboxChecked {
-  background-color: $green;
-  border-color: $dark-grey;
-  height: 12px;
-  width: 12px;
+.checkboxDefault {
   margin-top: 9px;
 }
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -31,7 +31,6 @@
   display: grid;
   grid-template-columns: repeat(3, 210px);
   height: auto;
-  --checkbox-margin-top: 9px;
 }
 
 .header {
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
@@ -43,6 +43,10 @@
   display: block;
 }
 
+.checkbox {
+  display: block;
+}
+
 .radio {
   margin-top: 9px;
 }
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
@@ -175,11 +175,10 @@ const TranscriptsFilter = (props: Props) => {
           <div className={styles.sortContent}>
             <RadioGroup
               classNames={{
-                label: styles.label,
                 radio: styles.radio,
-                radioChecked: styles.radioChecked,
                 wrapper: styles.buttonWrapper
               }}
+              theme="dark"
               options={radioData}
               onChange={onSortingRuleChange}
               selectedOption={sortingRule}
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
@@ -155,10 +155,9 @@ const TranscriptsFilter = (props: Props) => {
       <Checkbox
         key={key}
         classNames={{
-          unchecked: styles.checkboxUnchecked,
-          checked: styles.checkboxChecked
+          default: styles.checkboxDefault
         }}
-        labelClassName={styles.label}
+        theme="dark"
         checked={filter.selected}
         label={filter.label}
         onChange={(selected) => onFilterChange(key, filter, selected)}
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
@@ -176,6 +176,7 @@ const TranscriptsFilter = (props: Props) => {
               options={radioData}
               onChange={onSortingRuleChange}
               selectedOption={sortingRule}
+              className={styles.radioGroup}
             />
           </div>
         </div>
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
@@ -155,6 +155,7 @@ const TranscriptsFilter = (props: Props) => {
       <Checkbox
         key={key}
         theme="dark"
+        className={styles.checkbox}
         checked={filter.selected}
         label={filter.label}
         onChange={(selected) => onFilterChange(key, filter, selected)}
--- a/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
@@ -154,9 +154,6 @@ const TranscriptsFilter = (props: Props) => {
     const checkboxes = filtersForColumn.map(([key, filter]) => (
       <Checkbox
         key={key}
-        classNames={{
-          default: styles.checkboxDefault
-        }}
         theme="dark"
         checked={filter.selected}
         label={filter.label}
@@ -174,10 +171,6 @@ const TranscriptsFilter = (props: Props) => {
           <div className={styles.header}>Sort by</div>
           <div className={styles.sortContent}>
             <RadioGroup
-              classNames={{
-                radio: styles.radio,
-                wrapper: styles.buttonWrapper
-              }}
               theme="dark"
               options={radioData}
               onChange={onSortingRuleChange}
--- a/src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
+++ b/src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
@@ -14,6 +14,10 @@
     margin: 8px 0px;
     padding: 0px 18px;
 
+    &:first-child {
+      padding-left: 30px;
+    }
+
     .subLabel {
       line-height: 30px;
       font-size: 12px;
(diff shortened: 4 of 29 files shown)


azangru:
A general consideration: Let's try to model the appearance of these basic components in Storybook. We already have some checkbox stories, and a radio options story; let's build upon those to easily see the different variants of our components.

imransl:
@azangru I've fixed the tests

src/shared/components/checkbox/Checkbox.scss
```diff
@@ -5,31 +5,27 @@
...
-  height: 14px;
-  width: 14px;
-  border: 1px solid $grey;
-  background-color: $white;
   transition: background-color 0.1s ease-in-out;
   margin-top: 7px;
+  background-color: var(--checkbox-background-color, #{$white});
+  border: 1px solid var(--checkbox-border-color, #{$grey});
+  height: var(--checkbox-dimension-size, 13px);
+  width: var(--checkbox-dimension-size, 13px);
```
azangru, line 16: Isn't "dimension-size" excessive? `--checkbox-side`? `--checkbox-side-length`?

src/shared/components/radio-group/RadioGroup.scss
```diff
@@ -11,14 +11,14 @@
...
-  border: 1px solid $grey;
-  background-color: none;
-  margin-top: 7px;
   cursor: pointer;
+  justify-self: center;
+  margin-top: 7px;
+  background-color: var(--checkbox-background-color, #{$white});
+  border: 1px solid var(--radio-border-color, #{$grey});
+  height: var(--radio-dimension-size, 13px);
+  width: var(--radio-dimension-size, 13px);
```
azangru, line 21: Isn't "dimension-size" excessive? `--radio-diameter`?

src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
```diff
@@ -29,6 +29,11 @@
   margin-bottom: 5px;
 }
 
+.checkbox {
+  height: 14px;
+  width: 14px;
+}
```
azangru, line 35: So, now that we should be able to customise the size of the checkbox through CSS variables, this class becomes unnecessary. The parent, `checkboxGridContainer`, can define the value of the variable that controls the checkbox size. Only I have a suspicion that we shouldn't customise checkbox size here at all; but simply use the default size of the checkbox. Since it's in the custom download section, and Andrea doesn't care about custom download any longer (she says it will likely morph into some other tool), we can just remove this class from here (and from the tsx that's using it).

src/shared/components/checkbox/Checkbox.tsx
```diff
@@ -44,27 +48,43 @@ const isWithLabel = (props: CheckboxProps): props is WithLabelProps => {
...
 
+const getThemeClasses = (theme: Theme = 'lighter') => {
+  if (theme === 'dark') {
+    return styles.themeDark;
+  }
+
+  if (theme === 'light') {
+    return styles.themeLight;
+  }
+};
```
azangru, line 59: Would it be to horrible of me to ask that the default theme be "light" rather than "lighter"? It's just odd to that pale variant as the default.

src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.scss
```diff
@@ -15,37 +15,16 @@
   margin-left: 15px;
 }
 
-.checkboxUnchecked {
-  background-color: transparent;
-  border-color: $grey;
+.checkboxDefault {
+  height: 14px;
+  width: 14px;
 }
```
azangru, line 21: - is this something that can be controlled through a CSS variable on the parent CSS class?
- does Andrea really want 14 pixels now? doesn't it contradict her style guide? please check with her

src/shared/components/radio-group/RadioGroup.scss
```diff
@@ -11,14 +11,14 @@
...
-  height: 15px;
-  width: 15px;
   border-radius: 50%;
-  border: 1px solid $grey;
-  background-color: none;
-  margin-top: 7px;
   cursor: pointer;
+  justify-self: center;
+  margin-top: 7px;
+  background-color: var(--checkbox-background-color, #{$white});
```
azangru, line 18: no longer a checkbox

src/shared/components/checkbox/Checkbox.scss
imransl: If we use `--checkbox-side-length` then how about the other dimension (width or height)? Also isn't `--checkbox-side` ambiguous?

src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.scss
imransl: Yes we should be using the CSS variables. As for the different size, I'll check with her and see. We probably can remove this class like you suggested for the `CustomDownloads` checkbox.

src/shared/components/checkbox/Checkbox.tsx
imransl: No it isn't 😁 

I chose that since it was the only theme that used the default styling at that time. Perhaps I can change the default styling to be light instead.

src/shared/components/radio-group/RadioGroup.scss
imransl: ✅

src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.scss
imransl: ✅

src/shared/components/checkbox/Checkbox.tsx
imransl: ✅

src/content/app/custom-download/components/checkbox-grid/CheckboxGrid.scss
imransl: ✅

src/shared/components/radio-group/RadioGroup.scss
imransl: ✅

src/shared/components/checkbox/Checkbox.scss
azangru: > how about the other dimension (width or height)

Well, our hope with checkboxes is that they are squares. If they are squares (you can confirm with Andrea whether they are), then their height should equal their width, and we should only expose a single variable that controls this size.

`--checkbox-side-length` or `--checkbox-side-size` look ok.

src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
```diff
@@ -155,10 +155,9 @@ const TranscriptsFilter = (props: Props) => {
       <Checkbox
         key={key}
         classNames={{
-          unchecked: styles.checkboxUnchecked,
-          checked: styles.checkboxChecked
+          default: styles.checkboxDefault
```
azangru, line 158: Does the `styles.checkboxDefault` class actually exist? What does it do?

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
```diff
@@ -48,3 +48,8 @@
     vertical-align: middle;
   }
 }
+
+.radio {
+  height: 15px;
+  width: 15px;
+}
```
azangru, line 55: What's going on here? Why are these not set through the CSS variable, and why are they set at all? Do they need to be different from the default size?

src/shared/components/checkbox/Checkbox.scss
```diff
@@ -5,31 +5,27 @@
...
-
-.checked {
+.checkboxChecked {
   background-color: $green;
 }
 
-.disabled {
+.checkboxDisabled {
   background-color: $light-grey;
   cursor: default;
```
azangru, line 30: we may want to add `pointer-events: none`

src/shared/components/checkbox/Checkbox.scss
```diff
@@ -38,16 +34,31 @@
...
   line-height: 30px;
-  font-size: 12px;
   padding-left: 8px;
   padding-right: 2px;
   max-width: calc(100% - 17px);
   word-wrap: break-word;
 }
 
+.themeLighter {
+  --checkbox-label-font-size: 12px;
```
azangru, line 49: I have a suspicion that this might make the variable harder to override from a parent component. What if instead of defining the variable here, we instead write `font-size: var(--checkbox-label-font-size, 13px);`? Then the parent can still change the `--checkbox-label-font-size` variable if it needs to.

Same for other properties.

src/shared/components/checkbox/Checkbox.tsx
```diff
@@ -19,8 +19,11 @@ import classNames from 'classnames';
...
 import styles from './Checkbox.scss';
 
+type Theme = 'lighter' | 'light' | 'dark';
+
 type ClassNames = Partial<{
   checkboxHolder: string;
+  default: string;
   checked: string;
   unchecked: string;
   disabled: string;
```
azangru, line 29: A big question for me now is whether we need to pass css class names to the check component at all, or its relevant styles can be overridden entirely through css variables.

src/shared/components/radio-group/RadioGroup.scss
```diff
@@ -28,18 +28,33 @@
...
+  font-weight: var(--radio-label-font-weight, #{$light});
   line-height: 30px;
-  font-size: 14px;
   padding-right: 2px;
   word-wrap: break-word;
   cursor: pointer;
 }
+
+.themeLight {
+  --radio-label-font-size: 13px;
```
azangru, line 50: Same as the comment in the checkbox scss.

src/shared/components/checkbox/Checkbox.tsx
imransl: I agree. This would make it easier to customise the checkboxes. I'll replace these classes with CSS variables.

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
imransl: I missed these when checking where radio buttons were used. I'll look into this and see if it should be any different.

src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
imransl: Yes it does for adding a margin top: https://github.com/Ensembl/ensembl-client/blob/57af4751ad6bf3ba8325fe3b133281da7175c758/src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss#L50

src/shared/components/checkbox/Checkbox.scss
imransl: That makes sense. I'll check with Andrea if we might have any outliers for checkboxes.

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
imransl: We don't need this class so I will remove it

src/shared/components/checkbox/Checkbox.scss
imransl: So if we do this here, won't we need to do this in the other two theme classes as well?

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.scss
imransl: ✅

src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.tsx
imransl: ✅

src/shared/components/checkbox/Checkbox.scss
imransl: ✅

src/shared/components/checkbox/Checkbox.tsx
imransl: ✅

src/shared/components/checkbox/Checkbox.scss
imransl: I've changed this to `--checkbox-side-size` ✅

src/shared/components/radio-group/RadioGroup.scss
imransl: ✅

src/shared/components/checkbox/Checkbox.scss
imransl: ✅

stories/shared-components/checkbox/Checkbox.stories.tsx
```diff
@@ -72,30 +96,30 @@ LabelledCheckboxStory.storyName = 'with label';
 export const LongLabelledCheckboxStory = (args: DefaultArgs) => (
```
azangru, line 96: Need a storyName field on this story, similar to other stories. Otherwise Storybook uses the name of the function for the story name:

azangru:
Looking at the checkbox and the radio button in the browser:

1. Radio button

1.1. The radio button against a dark background looks different than in the XD 



2. Both radio button and checkbox

Note the vertical position of a label against the checkbox or against a radio button in Storybook. I would expect the label to be centred against a checkbox or a radio button; but this isn't the case.

azangru:
Just thinking out loud here: one thing we might consider is that the Checkbox component have a single className property (string) that we would put on the outermost component (the checkbox holder) if we want to add margins or something else to it. `--checkbox-holder-margin-top` as a variable looks wrong.

But this isn't a blocker and is something to discuss.

stories/shared-components/checkbox/Checkbox.stories.tsx
imransl: ✅

src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter.scss
```diff
@@ -24,13 +24,14 @@
...
 .sortContainer {
-    padding-left: 160px;
+  padding-left: 160px;
 }
 
 .filterContent {
   display: grid;
   grid-template-columns: repeat(3, 210px);
   height: auto;
+  --checkbox-margin-top: 9px;
```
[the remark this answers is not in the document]
imransl: ✅

imransl:
> 1. Radio button
> 
> 1.1. The radio button against a dark background looks different than in the XD

I am not sure what the difference is. I checked the XD spec again in the ticket and that matches the styling for both the radio button and label.

> 2. Both radio button and checkbox
> 
> Note the vertical position of a label against the checkbox or against a radio button in Storybook. I would expect the label to be centred against a checkbox or a radio button; but this isn't the case.

I've fixed this one with the latest PR ✅

azangru:
> I am not sure what the difference is. I checked the XD spec again in the ticket and that matches the styling for both the radio button and label.

@imransl Notice that for the selected radio button with the dark theme, the green dot is against the white background in the XD:



but not in the storybook or in the review branch.

imransl:
@azangru Thanks for clarifying this. I looked into this and found out that the dark theme class was overriding the checked class. So I had to put in the checked class for both the radio and checkbox buttons in the latest commit 8fb3b68. A bit uncertain if this is a good fix for this problem.

src/shared/components/checkbox/Checkbox.scss
```diff
@@ -1,53 +1,84 @@
...
-  grid-template-columns: 27px 1fr;
+.wrapper {
+  display: inline-block;
 }
 
-.defaultCheckbox {
+.grid {
+  display: inline-grid;
+  grid-template-columns: auto 1fr;
+  column-gap: var(--checkbox-distance-to-label, 12px);
```
msmanoj, line 10: Why are we defining these var?

src/shared/components/instant-download/instant-download-transcript/InstantDownloadTranscript.scss
```diff
@@ -87,6 +51,7 @@
 
     .geneSection {
       grid-area: gene;
+      --checkbox-holder-margin-left: 6px;
```
msmanoj, line 54: Unused definition?

src/shared/components/checkbox/Checkbox.scss
azangru: If we ever need to tweak the distance between the checkbox and the label. I thought this is within the possibilities of what we may be asked to do.

msmanoj:
I'm getting the following warning when I try to run it locally:

src/shared/components/instant-download/instant-download-transcript/InstantDownloadTranscript.scss
azangru: Deleted.

azangru:
> It looks like the column layout for RadioGroup wouldn't wrap if there isn't enough space. 

Andrea said she didn't expect radio buttons to wrap; but I added the wrapping behaviour and added an example to the "radios in a row" story.

src/shared/components/checkbox/Checkbox.scss
azangru: Removed the var 🤷‍♂️ 

_(Also, the distance between the checkbox and the label is now defined via label padding)_

msmanoj (approved):

PR #679, opened 2022-01-25
Update upload button

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1450

## Description
This is a refactoring of the Upload component to achieve the following goals:
- Improve the types, such that typescript could infer the type of the parameter of the callback function that is called when the file is uploaded. Currently, in the dev branch, the Upload component is equally happy with a callback function that accepts `File` or `File[]` regardless of the `allowMultiple` prop; and cannot distinguish between a file parsed as an array buffer or as text. The stricter types added in this PR allow such granular distinction (see tests and stories).
- Move the drag&drop code and the code for reading file contents out of the Upload component (see the `useFileDrop` hook and the upload helpers). This allows other components, not just the `Upload`, to also serve as a file drop zone.
- Update the design to match new specs.

## Deployment URL
- ensembl-client site: http://update-upload-button.review.ensembl.org
- storybook: https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-upload--default-story

## Views affected
- Feedback form
- Blast input boxes
- Custom download components


src/shared/components/upload/Upload.test.tsx
```diff
@@ -15,130 +15,176 @@
...
+      const onUpload = (result: FileTransformedToString) =>
+        (textFromFile = result.content as string);
+
+      const { container } = render(
+        <Upload onUpload={onUpload} transformTo="text" />
+      );
+
+      await act(async () => {
+        fireEvent.change(container.querySelector('input') as HTMLElement, {
+          target: { files: [file1, file2] }
```
msmanoj, line 104: Was it intentional to upload two files here?

src/shared/components/upload/_uploadConstants.scss
```diff
@@ -0,0 +1,8 @@
+@import 'src/styles/common';
+
+$highlightedUploadBackground: $ice-blue;
+$uploadButtonTallerHeight: 102px;
+$uploadButtonShallowerHeight: 90px;
```
msmanoj, line 5: `uploadButtonShorterHeight`?

src/shared/components/upload/helpers/uploadHelpers.ts
```diff
@@ -0,0 +1,120 @@
...
+  text: 'readAsText'
+};
+
+export const transformFiles = async <
+  T extends Options & OptionsWithDefinedTransform
+>(
+  files: File[],
+  options: T
+): Promise<Result<T>> => {
+  const { allowMultiple, transformTo } = options;
```
msmanoj, line 43: Why should transformFiles worry about `allowMultiple` instead of just transforming the files it gets?

azangru: Oof. Fair point. I think I did it in order to make the types work. I'll see if typescript will object if I just pass a file or an array of files.

src/shared/components/upload/Upload.test.tsx
azangru: Kinda. I didn't have a strong opinion one way or the other. On the one hand, realistically, the event won't contain more than one file in the array, because the input element won't allow the user to select more than one file. On the other hand, we are proving to ourselves that even if, miraculously, the event did contain more than one file, only the first file would be processed. I can only keep one file in the array if you prefer 🤷‍♂️

msmanoj: I was wondering what if the user drops multiple files but we only handle the first file, ignoring the rest?

azangru: That's the current behaviour of the component: it will accept all files if the "allowMultiple" parameter is set to true; but will only process the first file if the allowMultiple parameter is set to false. I don't know what the correct behaviour should otherwise be if the allowMultiple is false, but the user drops multiple files.

src/shared/components/upload/_uploadConstants.scss
azangru: ✅

src/shared/components/upload/helpers/uploadHelpers.ts
azangru: Updated the signature of `transformFiles`. It didn't upset typescript and simplified the code.

msmanoj (approved):

PR #680, opened 2022-01-25
Add Contact us label to conversation icon on help and about app pages

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1455
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1454

## Description
A clickable text link needs to appear in the same style as in other apps in the Help and About us page

## Deployment URL
http://contact-icon-label.review.ensembl.org/

## Views affected
Help and About us page

### Other effects
NA

## Possible complications
NA

### DIFF
--- a/src/shared/components/communication-framework/ConversationIcon.scss
+++ b/src/shared/components/communication-framework/ConversationIcon.scss
@@ -1,3 +1,13 @@
+@import 'src/styles/common';
+
 .conversationIcon {
+  width: 40px;
+}
+
+.conversationIconWrapper {
+  display: grid;
+  grid-template-columns: 75px 40px;
+  align-items: center;
   cursor: pointer;
+  color: $blue;
 }
--- a/src/shared/components/communication-framework/ConversationIcon.tsx
+++ b/src/shared/components/communication-framework/ConversationIcon.tsx
@@ -25,7 +25,11 @@ import CommunicationPanel from 'src/shared/components/communication-framework/Co
 
 import styles from './ConversationIcon.scss';
 
-const ConversationIcon = () => {
+type Props = {
+  withLabel?: boolean;
+};
+
+const ConversationIcon = (props: Props) => {
   const dispatch = useDispatch();
 
   const { trackContextualHelpOpened } = useCommonAnalytics();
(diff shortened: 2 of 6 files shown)


src/shared/components/communication-framework/ConversationIcon.tsx
```diff
@@ -35,13 +39,15 @@ const ConversationIcon = () => {
...
+  const wrapperClass = props.withLabel ? styles.conversationIconWrapper : '';
   return (
     <>
       <CommunicationPanel />
-      <ConversationImageIcon
-        onClick={onClick}
-        className={styles.conversationIcon}
-      />
+      <div className={wrapperClass} onClick={onClick}>
+        {props.withLabel && 'Contact Us'}
```
azangru, line 48: XD says "Contact us" (lower-cased u)

src/shared/components/communication-framework/ConversationIcon.scss
```diff
@@ -1,3 +1,13 @@
+@import 'src/styles/common';
+
 .conversationIcon {
+  width: 40px;
+}
+
+.conversationIconWrapper {
+  display: grid;
+  grid-template-columns: 75px 40px;
```
azangru, line 9: `display: grid` -> `display: flex`
`grid-template-columns: 75px 40px;` -> `gap: 16px;`

Not only is this more suitable for the effect you are after, but it will allow you to always apply this class to the wrapper div rather than conditionally, as you are now doing in `ConversationIcon.tsx`

azangru:
We noticed that when you switch between the help page and the about page, the communication icon shifts vertically. The reason is that the app bar on the help page has a height set to 86 px, whereas on the about page (an on other pages as well) it is set to 80px. Could you change the height of the app bar on the help page to 80px as well?

image

Another thing we noticed is that on the home page the help and communication icons are positioned a bit lower than on the rest of the pages. Andrea says that the vertical position of these icons should be the same on all pages. Could you take a look at that as well, or do you think it should go in a separate ticket?

jyothishnt:
> Another thing we noticed is that on the home page the help and communication icons are positioned a bit lower than on the rest of the pages. Andrea says that the vertical position of these icons should be the same on all pages. Could you take a look at that as well, or do you think it should go in a separate ticket?

Could you please check if the vertical heights are ok?

azangru:
Help page is great. For the home page, I think I am measuring a 3-pixel difference (the buttons start 3 pixels lower on the home page than on the species selector page):

image

image

jyothishnt:
Hi @azangru I think it is the problem with species selector page species lozenges margin. Can this be done in a design review ticket?

azangru (approved):

PR #681, opened 2022-01-31
Refactor Zmenu and fix issue with multiple transcripts data

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1439

## Description
- Updated GB version to 0.2.0
- Fixed the zmenu bug with multiple transcripts 

### GB 0.2.0
  * Round sizes correctly in observer to avoid excessive CPU usage
  * Don't crash if window size becomes zero
  * New "variety" key to reduce zmenu ambiguity.
  * Deduplicate varieties
  * Update current style compiler programs to use "variety key" (BE version 6->7, code version 0.1.x->0.2.x)
  * Allow an element to be passed in the config instead of an element id (if desired)
  * Replace ad-hoc approaches to logging verbosity on console with single, unified and configurable approach
  * Build system work:
    * Use dockerignore to reduce build time of docekr builds
    * Use docker buildkit for extra features
    * Menuize build scripts
    * Bust docker cacheing to pick up github changes, where requested
  * Code quality work in callbacks:
    * Simpler scheduler
    * New, self-contained classes which manage unloading correctly for
      * timer
      * requestanimationframe
      * custom events

## Deployment URL
http://refactor-and-zmenu-fix.review.ensembl.org/genome-browser/homo_sapiens_GCA_000001405_28?focus=gene:ENSG00000198182&location=19:37684675-37731456 

### Other effects


azangru:
Consider these entries in webpack configs:
- https://github.com/Ensembl/ensembl-client/blob/refactor-and-zmenu-fix/webpack/client/development.ts#L53-L57 — probably no longer needed?
- https://github.com/Ensembl/ensembl-client/blob/dev/webpack/server/development.ts#L49-L53 — also likely no longer needed
- https://github.com/Ensembl/ensembl-client/blob/dev/webpack/server/common.ts#L131-L135 (this is why webpack wasn't complaining previously about the incorrect common js bundle) — should be updated to new package name?

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +59,68 @@ const Zmenu = (props: ZmenuProps) => {
...
-    setZmenus && setZmenus(pickBy(zmenus, (value, key) => key !== props.id));
+    setZmenus &&
+      setZmenus(pickBy(zmenus, (value, key) => key !== props.zmenuId));
   };
 
   const direction = chooseDirection(props);
   const toolboxPosition =
     direction === Direction.LEFT ? ToolboxPosition.LEFT : ToolboxPosition.RIGHT;
 
+  const { content, variety } = props.payload;
```
azangru, line 70: I am not sure I understand what is happening below. I am also not sure I understand the distribution of responsibilities between components anymore.

We now find ourselves with the following arrangement:
1. Genome browser sends a zmenu message
2. There is some code that listens to the message, creates a new zmenu per message, keeps track of the created zmenus, and removes them when they are no longer needed. Previously this code lived in ZmenuController; but by now it is shared between ZmenuController and useGenomeBrowser in a way that I don't quite understand
3. Next, a newly instantiated zmenu needs to decide what kind of zmenu it is. We expect that for transcript, variant, and so on, zmenus are going to be different; this is what the `variety` field suggests. So, by inspecting the variety key, we should be able to select the appropriate zmenu component to display the content of the message. I would expect us to have a `ZmenuTranscript` component for handling the `gene-and-one-transcript` zmenu type.

Here are the questions for us:
- Does the `ZmenuController` component have any purpose, or can its logic be moved into a hook (e.g. `useZmenus`). By the way, does the zmenu-related code belong in the useGenomeBrowser hook at all; or is it better placed in such useZmenus hook?
-  What's the responsibility of the `Zmenu` component? Should it only inspect the `variety` property and choose the appropriate zmenu component for handling this variety, or is it also responsible for the zmenu positioning logic? Or should the positioning logic be extracted into a hook?

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +59,68 @@ const Zmenu = (props: ZmenuProps) => {
...
+    (f) => f.metadata.type === ZmenuFeatureType.TRANSCRIPT
+  ) as ZmenuContentTranscript[];
+  const genes = content.filter(
+    (f) => f.metadata.type === ZmenuFeatureType.GENE
+  ) as ZmenuContentGene[];
+
+  const features: (ZmenuContentTranscript | ZmenuContentGene)[] = [];
+
+  // get the first variety type
+  const firstVarietyType = variety[0].type;
```
azangru, line 82: Didn't @ens-ds23 say that the contract is that only one variety is going to have a `type` field on it? If so, then we want to do `const zmenuType = variety.find(variety => Boolean(variety.type))?.type as string`

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +59,68 @@ const Zmenu = (props: ZmenuProps) => {
...
 
   const direction = chooseDirection(props);
   const toolboxPosition =
     direction === Direction.LEFT ? ToolboxPosition.LEFT : ToolboxPosition.RIGHT;
 
+  const { content, variety } = props.payload;
+
+  const transcripts = content.filter(
+    (f) => f.metadata.type === ZmenuFeatureType.TRANSCRIPT
+  ) as ZmenuContentTranscript[];
```
azangru, line 74: You are doing the same on line 89, using the more appropriate `find` method

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +59,68 @@ const Zmenu = (props: ZmenuProps) => {
...
+  const genes = content.filter(
+    (f) => f.metadata.type === ZmenuFeatureType.GENE
+  ) as ZmenuContentGene[];
```
azangru, line 77: You are doing the same on line 98 using the more appropriate `find` method (a gene can be found as a feature of type gene whole id equals the gene id of the transcript)

src/server/helpers/getConfigForClient.ts
```diff
@@ -25,7 +25,8 @@ const getBaseApiUrls = (): BaseApiUrls => {
     docsBaseUrl: process.env.BROWSER_DOCS_BASE_URL ?? '/api/docs',
     refgetBaseUrl: process.env.REFGET_BASE_URL ?? '/api/refget',
     genomeBrowserBackendBaseUrl:
-      process.env.GENOME_BROWSER_BACKEND_BASE_URL ?? '/api/browser/data',
+      process.env.GENOME_BROWSER_BACKEND_BASE_URL ??
+      'http://127.0.0.1:3333/api/data',
```
azangru, line 29: ❓

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: My understanding is that we will still have multiple genes and transcripts until gene bumping is out.

msmanoj:
> Consider these entries in webpack configs:
> 
> * https://github.com/Ensembl/ensembl-client/blob/refactor-and-zmenu-fix/webpack/client/development.ts#L53-L57 — probably no longer needed?

Removed it

> * https://github.com/Ensembl/ensembl-client/blob/dev/webpack/server/development.ts#L49-L53 — also likely no longer needed

Removed it

> * https://github.com/Ensembl/ensembl-client/blob/dev/webpack/server/common.ts#L131-L135 (this is why webpack wasn't complaining previously about the incorrect common js bundle) — should be updated to new package name?

Could you remind me what this is for?

src/server/helpers/getConfigForClient.ts
msmanoj: Reverted it

azangru:
> > https://github.com/Ensembl/ensembl-client/blob/dev/webpack/server/common.ts#L131-L135 (this is why webpack wasn't complaining previously about the incorrect common js bundle) — should be updated to new package name?

>Could you remind me what this is for?

It says not to bundle the genome browser into the server bundle, but to use the code directly from the node_modules instead if needed. Server-side bundle should even need to access ensembl-genome-browser code; so I thought it would be a good thing not to even add it to the bundle.

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
azangru: But we can just grab the first one.

msmanoj: That's done

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: Remove this

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: That's done

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +58,66 @@ const Zmenu = (props: ZmenuProps) => {
...
   };
 
   const direction = chooseDirection(props);
   const toolboxPosition =
     direction === Direction.LEFT ? ToolboxPosition.LEFT : ToolboxPosition.RIGHT;
 
+  const { content, variety } = props.payload;
+  const features: (ZmenuContentTranscript | ZmenuContentGene)[] = [];
+
+  // get the first variety type
```
azangru, line 72: this comment is probably no longer necessary?

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -51,15 +58,66 @@ const Zmenu = (props: ZmenuProps) => {
...
+      return null;
+    }
+
+    if (gene) {
+      features.push(gene);
+    }
+    // FIXME: we need this hack until ENSWBSITES-1375 is fixed
+    unversionedTranscriptId = transcript.metadata.transcript_id.split(
+      '.'
+    )[0] as string;
```
azangru, line 109: Isn't this going to break some species (wheat?)? I thought some species' unversioned transcripts end in a dot and a number. I could have sworn that this is the case for wheat.

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: Removed it

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: I tried using versioned stable id and it worked.

azangru (approved):

PR #684, opened 2022-02-04
Add Blast settings

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1444
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1474

## Description
- Add a config for the Blast parameters (hard-coded for now; will be fetched from the backend in the future)
- Add a React component to display a list of available Blast parameters. The availability of some of the parameters depends on which other parameters have been chosen.

Also:
- Fixed a test of the Upload component to properly wait for the expected value (without hard-coding the timeout duration).

## Deployment URL
http://blast-params.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
@@ -51,14 +51,11 @@ import styles from './BlastSettings.scss';
 const blastSettingsConfig = untypedBlastSettingsConfig as BlastSettingsConfig;
 
 const getParameterData = (name: BlastParameterName) => {
-  return (blastSettingsConfig as unknown as BlastSettingsConfig).parameters[
-    name
-  ];
+  return blastSettingsConfig.parameters[name];
 };
 
 const getPresetsList = () => {
-  const presets = (blastSettingsConfig as unknown as BlastSettingsConfig)
-    .presets;
+  const { presets } = blastSettingsConfig;
   const { label, options } = presets;
   return { label, options };
 };
@@ -104,7 +101,7 @@ const BlastSettings = () => {
     dispatch(
       setSequenceType({
         sequenceType: sequenceType as SequenceType,
-        config: blastSettingsConfig as unknown as BlastSettingsConfig
+        config: blastSettingsConfig
       })
     );
   };
@@ -113,7 +110,7 @@ const BlastSettings = () => {
     dispatch(
       setBlastDatabase({
         database,
-        config: blastSettingsConfig as unknown as BlastSettingsConfig
+        config: blastSettingsConfig
       })
     );
   };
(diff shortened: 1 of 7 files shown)


ens-ma7:
@azangru not sure if you notice but there is one conflict that need resolving.

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -0,0 +1,357 @@
...
+  BlastSelectSetting,
+  BlastBooleanSetting,
+  BlastParameterName,
+  BlastSettingsConfig
+} from 'src/content/app/tools/blast/types/blastSettings';
+
+import styles from './BlastSettings.scss';
+
+const getParameterData = (name: BlastParameterName) => {
+  return (blastSettingsConfig as unknown as BlastSettingsConfig).parameters[
```
msmanoj, line 52: Can we do something like the following so that we don't have to repeatedly do the casting?

```
import blastSettings from './blastSettingsConfig.json';
const  blastSettingsConfig = blastSettings as unknown as BlastSettingsConfig;
```

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -0,0 +1,357 @@
...
+    false: string;
+  };
+  selectedOption: string;
+  onChange: (value: string) => void;
+}) => {
+  const { label, options, selectedOption, onChange } = params;
+  const isChecked = options.true === selectedOption;
+
+  const onCheckboxChange = (isChecked: boolean) => {
+    onChange(options[`${isChecked}`]);
```
msmanoj, line 349: It is slightly confusing to read `isChecked` defined with same name one after the other. Should we rename one of it?

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
azangru: ✅

msmanoj (approved):

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -0,0 +1,354 @@
...
+        </select>
+      </label>
+    </div>
+  );
+};
+
+const buildCheckbox = (params: {
+  label: string;
+  options: {
+    true: string;
```
jyothishnt, line 336: is this the best way to find out the selectedOption?

azangru: For checkboxes, I thought it is. Can you think of a better way?

PR #686, opened 2022-02-07
Fix transcripts loading indicator

msmanoj:
## Type

- [x] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1476

## Importance
- There is a high chance of similar issues occurring in other areas where we use RTK

## Description
According to the RTK docs, `isLoading` :
> When true, indicates that the query is currently loading for the first time, and has no data yet. This will be true for the first request fired off, but not for subsequent requests.

This means that `isLoading` will only be true for the first time.

Also, `data`:
> The latest returned result regardless of hook arg, if present.

This would mean that `data` will still be the value of the previous request (different gene). `currentData` is what will have the result of the latest query

## Deployment URL
http://ev-transcript-loading.review.ensembl.org

## Views affected
Entity viewer - > transcripts

### DIFF
--- a/src/content/app/entity-viewer/components/example-links/ExampleLinks.tsx
+++ b/src/content/app/entity-viewer/components/example-links/ExampleLinks.tsx
@@ -38,7 +38,7 @@ const ExampleLinks = () => {
     getGenomeExampleFocusObjects(state, activeGenomeId || '')
   );
   const exampleGeneId = exampleEntities.find(({ type }) => type === 'gene')?.id;
-  const { isLoading, data, error } = useGeneSummaryQuery(
+  const { isFetching, currentData, error } = useGeneSummaryQuery(
     {
       geneId: exampleGeneId || '',
       genomeId: activeGenomeId || ''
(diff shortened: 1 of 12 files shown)


ens-ma7:
When i first look at the code change, i thought it was just about renaming variables and why was that making a difference until i read your comment @msmanoj in the description. And its a bit scary with this and i think we should remove everywhere else where we are using the wrong variable because it will be a silent bug (hard to notice). I did a search on the code base and there are only 6 other files where these are used. Do you think you can changed all of them in this PR or happy for you to create another ticket with what needs doing.

image

ens-ma7:
@msmanoj  Looks like one of the test is failing. can you have a look.

msmanoj:
> @msmanoj Looks like one of the test is failing. can you have a look.

Hi @ens-ma7 - here is that test that failed which is not related to this PR:
https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/components/upload/Upload.test.tsx#L113
The expected value was the empty string "" instead of 'Lorem ipsum dolor sit amet'

I tried to rerun the pipeline and it worked fine.

ens-ma7 (approved):

PR #689, opened 2022-02-08
Add sequence guessing code

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1481

## Description
Guess sequence type when user commits a sequence, according to the following rules:
- The guess is only made on the basis of the first sequence for consistency
- The guessing algorithm checks if there are protein-specific characters in the input sequence; and if not, whether unambiguous nucleotide codes make up more than 90% of the sequence
- If the user deletes the first sequence such that the first input box now contains the sequence of a different type, we automatically change the sequence type
- If the user has chosen sequence type manually, we do not attempt to make any guesses about the sequence type
- However, if the user presses the "Clear all" button or clears the only input box on the page, we reset the sequence type to initial (i.e. nucleotide)

### Other changes
This PR adds `@testing-library/react-hooks` to our list of dependencies.

## Deployment URL
http://sequence-type-predictor.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
@@ -18,11 +18,13 @@ import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
 import {
+  getSelectedSequenceType,
   getSequences,
   getEmptyInputVisibility
 } from 'src/content/app/tools/blast/state/blast-form/blastFormSelectors';
 
 import {
+  setSequenceType,
   setSequences,
   updateEmptyInputDisplay,
   switchToSpeciesStep
@@ -30,6 +32,14 @@ import {
 
 import { SecondaryButton } from 'src/shared/components/button/Button';
 import PlusButton from 'src/shared/components/plus-button/PlusButton';
+import RadioGroup from 'src/shared/components/radio-group/RadioGroup';
+
+import untypedBlastSettingsConfig from 'src/content/app/tools/blast/components/blast-settings/blastSettingsConfig.json';
+
+import type {
+  SequenceType,
+  BlastSettingsConfig
+} from 'src/content/app/tools/blast/types/blastSettings';
 
 import styles from './BlastInputSequences.scss';
 
@@ -37,12 +47,24 @@ export type Props = {
   compact: boolean;
 };
 
+const blastSettingsConfig = untypedBlastSettingsConfig as BlastSettingsConfig;
+
 const BlastInputSequencesHeader = (props: Props) => {
   const { compact } = props;
+  const sequenceType = useSelector(getSelectedSequenceType);
   const sequences = useSelector(getSequences);
   const shouldAppendEmptyInput = useSelector(getEmptyInputVisibility);
 
   const dispatch = useDispatch();
+
+  const onSequenceTypeChange = (sequenceType: SequenceType) => {
+    dispatch(
+      setSequenceType({
+        sequenceType,
+        config: blastSettingsConfig
+      })
+    );
+  };
 
   const onClearAll = () => {
     dispatch(
@@ -67,6 +89,10 @@ const BlastInputSequencesHeader = (props: Props) => {
         <span className={styles.sequenceCounter}>{sequences.length}</span>
         <span className={styles.maxSequences}>of 30</span>
       </div>
+      <SequenceSwitcher
+        sequenceType={sequenceType}
+        onChange={onSequenceTypeChange}
+      />
       <div className={styles.headerGroup}>
         <span className={styles.clearAll} onClick={onClearAll}>
           Clear all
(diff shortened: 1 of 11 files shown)


src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
```diff
@@ -102,4 +101,28 @@ const AddAnotherSequence = (props: AddAnotherSequenceProps) => {
...
+      label: 'Protein',
+      value: 'protein'
+    }
+  ];
+  return (
+    <RadioGroup
+      options={options}
+      direction="row"
+      selectedOption={props.sequenceType}
+      onChange={(val) => props.onChange(val as unknown as SequenceType)}
```
msmanoj, line 123: Can it just be `as SequenceType`?

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+import blastFormReducer, {
+  initialState as initialBlastFormState,
+  type BlastFormState
+} from 'src/content/app/tools/blast/state/blast-form/blastFormSlice';
+
+const getWrapper = (
+  { state }: { state?: Partial<BlastFormState> } = { state: {} }
+) => {
+  const blastFormState = Object.assign({}, initialBlastFormState, state);
+  const rootReducer = combineReducers({
```
msmanoj, line 37: Can this be move outside of the `getWrapper` function?

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      const { result } = renderHook(() => useBlastInputSequences(), {
+        wrapper
+      });
+
+      // initial values
+      expect(getSelectedSequenceType(store.getState() as any)).toEqual(
+        'protein'
+      );
+      expect(getSequenceSelectionMode(store.getState() as any)).toEqual(
+        'manual'
```
msmanoj, line 241: Is it worth testing the initial state here?

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  // to avoid stale closures persisting in functions returned from the hook
+  useEffect(() => {
+    ref.current = { sequenceType, sequenceSelectionMode };
+  });
+
+  const updateSequences = (sequences: ParsedInputSequence[]) => {
+    dispatch(setSequences({ sequences }));
+    updateSequenceTypeAutomatically(sequences);
```
msmanoj, line 61: Would it be better to avoid calling this function if `ref.current.sequenceSelectionMode === 'manual'` here instead of doing the check on L67?

msmanoj:
- I added a protein sequence
- I see SequenceType automatically changes to `protein`
- I removed all input using backspace or by selecting all using ctrl + a
- I added a dna sequence
- I still see SequenceType set to `protein`

Is this expected?

azangru:
Can't reproduce the above. Here is what I am seeing:

https://user-images.githubusercontent.com/6834224/154543389-dadc39ab-c049-4bde-85d9-06f9a053e135.mp4

One thing Andrea could potentially object to is that we aren't immediately switching to nucleotide sequence when deleting input from the box via backspace. But that's because an empty sequence doesn't get committed. I am not quite sure yet how we are going to handle this.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
azangru: It can. Weird, I was sure that when I tried it before, typescript complained of something or other.

✅

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.test.tsx
azangru: Not really; it was just a sanity check. It can go.

✅

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.ts
azangru: I don't think so. We also automatically update the sequence type to default when there are no sequences, sadly, regardless of whether the `sequenceSelectionMode` is automatic or manual. I think this check belongs in the function that is concerned with the updating of sequence type rather than with the updating of the sequences.

src/content/app/tools/blast/components/blast-input-sequences/useBlastInputSequences.test.tsx
azangru: ✅

msmanoj (approved):

PR #692, opened 2022-02-09
blast species lozenges onclick function

ens-ma7:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1481

## Importance
N/A

## Description
select species checkbox when clicking on the species lozenge in blast. Note that if the species checkbox is already selected/checked, nothing happen when clicking the species lozenge. Only action is select the species checkbox if it is not selected.

## Deployment URL
http://blast-specieslozenge.review.ensembl.org

## Views affected
Blast

### Other effects
N/A

## Possible complications
N/A

### DIFF
--- a/src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar.test.tsx
+++ b/src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar.test.tsx
@@ -16,51 +16,61 @@
 
 import React from 'react';
 import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router';
 import { configureStore, combineReducers } from '@reduxjs/toolkit';
 import { render } from '@testing-library/react';
 
 import blastFormReducer, {
-  initialState as initialBlastFormState,
-  type BlastFormState
+  initialState as initialBlastFormState
 } from 'src/content/app/tools/blast/state/blast-form/blastFormSlice';
+
+import speciesSelectorReducer, {
+  SpeciesSelectorState
+} from 'src/content/app/species-selector/state/speciesSelectorSlice';
 
 import ToolsAppBar from './ToolsAppBar';
 
-const renderComponent = (
-  { state }: { state?: Partial<BlastFormState> } = { state: {} }
-) => {
-  const blastFormState = Object.assign({}, initialBlastFormState, state);
+jest.mock(
+  'src/shared/components/communication-framework/ConversationIcon',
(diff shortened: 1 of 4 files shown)


ens-ma7:
Note, not sure if this needs a test to be written. there was no test for ToolsAppBar hence didnt create one but happy to do one if it makes sense to have one.

azangru (approved):
The line `updateSelectedSpecies({ isChecked: true, genomeId })` suggests that the name of the `isChecked` parameter of the `updateSelectedSpecies` was influenced by the UI of the species selector that had checkboxes. That's not very good. I'd rather we had two methods instead: `addSelectedSpecies` and `removeSelectedSpecies`, which just take a genome id.

On the other hand, we'll have to replace this provisional species selector with a different species selector anyway; so I am not super bothered by this.

ens-ma7:
@azangru it also didnt feel right but wasnt sure about the approach but thanks for the hint and i think it makes sense to add these two functions because even though the species selector will be different one thing that i have noticed is the same is the presence of the checkbox and also to select the species using the checkbox. I will make the suggested change. And do you think we should write a test for this (clicking on the species lozenge will add a species to the state?)

jyothishnt (approved):

azangru:
> do you think we should write a test for this (clicking on the species lozenge will add a species to the state?)

Won't hurt 🙂

src/content/app/tools/shared/components/tools-app-bar/ToolsAppBar.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    store
+  };
+};
+
+describe('ToolsAppBar', () => {
+  describe('Species Lozenge click', () => {
+    it('updates the selectedSpecies state', () => {
+      const { container, store } = renderComponent();
+
+      const speciesLozenge = container.querySelector('.inUseInactive');
```
azangru, line 89: `inUseInactive` is a strange class to use to target the species lozenge. We can use the class `.species`; or we can use the `getByText` helper from the same object from which we are pulling the `container` and  find an element using `getByText('Human')`.

ens-ma7: Done.

azangru (approved):

PR #695, opened 2022-02-18
Blast input validation

azangru:
**NOTE:** this PR contains code from https://github.com/Ensembl/ensembl-client/pull/689 and should be merged after 689 is approved and merged.

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1482
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1501

## Description
This PR adds the following functionality:
- Focusing on a blast sequence input:
  - changes the colour of the textarea border to blue
  - hides the upload file component
- Added validation of sequence length (sadly, as per spec, the input box shows invalid input while the user is still typing)
- Added validation of sequence alphabet for given sequence type

## Deployment URL
http://blast-input-validation.review.ensembl.org/blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
@@ -25,10 +25,7 @@ import React, {
 import classNames from 'classnames';
 
 import { toFasta } from 'src/shared/helpers/formatters/fastaFormatter';
-import {
-  hasSufficientLength,
-  hasValidSequenceCharacters
-} from 'src/content/app/tools/blast/utils/sequenceValidators';
+import { isValidSequence } from 'src/content/app/tools/blast/utils/sequenceValidators';
 
 import Textarea from 'src/shared/components/textarea/Textarea';
 import {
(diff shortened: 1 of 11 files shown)


src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -110,10 +128,18 @@ const BlastInputSequence = (props: Props) => {
     }
   };
 
+  const isInputValid = input
+    ? checkSequenceValidity(input, sequenceType)
```
msmanoj, line 132: Wouldn't it be better to validate the sequence `onBlur` instead of every keyups?
I wonder if it would annoy the users if is starts to complain as soon as the use inputs the first character.

azangru: It absolutely would, but this is something that I was arguing about with Andrea during the planning meeting, and she insisted that we do it while typing. I was hoping somebody other than me would argue this point with Andrea as well, but no-one joined 🤷‍♂️

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -128,19 +154,33 @@ const BlastInputSequence = (props: Props) => {
...
-        {!input && <Upload transformTo="text" onUpload={onFileDrop} />}
+        {!input && !isFocused && (
+          <Upload transformTo="text" onUpload={onFileDrop} />
+        )}
       </div>
     </div>
   );
 };
 
+const checkSequenceValidity = (
```
msmanoj, line 173: Can this function be moved to `sequenceValidators.ts`?

azangru: I've added a function that combines the checks `hasSufficientLength` and  `hasValidSequenceCharacters` to `sequenceValidators.ts`; but the `checkSequenceValidity` function above contains the cleanup logic that belongs in the input box.

msmanoj (approved):

PR #696, opened 2022-02-21
Add job name component and job submit component

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1491

## Description
The scope of this PR is only to add a job name field and a button for submitting the Blast job.
- The job name field is added as a part of BlastSettings component (it kind of feels like a Blast setting)
- The job submission button has been created as a separate component. At the moment, the button is always disabled.

## Deployment URL
http://job-name-submit.review.ensembl.org/blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
+++ b/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
@@ -1,0 +1,34 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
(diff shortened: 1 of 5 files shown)


msmanoj (approved):

ens-ma7:
Do we need a test for the blastjobname?

azangru:
> Do we need a test for the blastjobname?

We don't have any tests for the BlastSettings component. Don't know if we should or what they might be (there are too many options to meaningfully test); but there is no point in adding a test just for the blast job name

ens-ma7 (approved):

PR #697, opened 2022-02-21
Add simple select component

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1490

## Description
- Created a simple select component that is based on the browser native select
- Added an entry to the storybook
- Updated BLAST settings to use SimpleSelect

## Deployment URL
http://simple-select.review.ensembl.org

**Storybook:**
https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-simpleselect--simple-select-default-story

## Views affected
- Storybook
- BLAST settings

### Other effects

## Possible complications
- Currently, we have Select & SimpleSelect both having the same purpose. We probably want to replace Select with SimpleSelect at some point


azangru:
The button doesn't always press reliably (tested the right-hand corner)


https://user-images.githubusercontent.com/6834224/155041698-4aa356f4-0ebf-4f8d-b138-a243cefbfafe.mp4

src/shared/components/select/SelectOptionsPanel.tsx
```diff
@@ -161,14 +160,15 @@ const SelectOptionsPanel = (props: Props) => {
   }, []);
 
   useEffect(() => {
-    document.addEventListener('click', handleClickOutside);
-    document.addEventListener('touchend', handleClickOutside);
+    setTimeout(() => document.addEventListener('click', handleClickOutside), 0);
+    // document.addEventListener('click', handleClickOutside);
+    // document.addEventListener('touchend', handleClickOutside);
```
azangru, line 165: ✂️

src/shared/components/simple-select/SimpleSelect.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  if (optionGroups.length === 1) {
+    return (
+      <div className={styles.select}>
+        <select
+          className={styles.selectResetDefaults}
+          defaultValue=""
+          {...rest}
+        >
+          {placeholder && <option value="">{placeholder}</option>}
```
azangru, line 74: [This](https://jsbin.com/nipakatori/edit?html,css,js,output) works surprisingly beautifully in Chrome (not sure about other browsers)

([from SO](https://stackoverflow.com/a/64045344/3925302))

src/shared/components/simple-select/SimpleSelect.scss
```diff
@@ -0,0 +1,41 @@
+@import 'src/styles/common';
+
```
azangru, line 2: Is there anything in this component that we want to make tweakable via custom properties?

msmanoj:
> The button doesn't always press reliably (tested the right-hand corner)

I've put a fix for it

src/shared/components/select/SelectOptionsPanel.tsx
msmanoj: That's done

src/shared/components/simple-select/SimpleSelect.tsx
msmanoj: padding: 3px 10px;

src/shared/components/simple-select/SimpleSelect.scss
msmanoj: I've made the following properties tweakable:
- min-width: var(--select-min-width, 15ch);
- max-width: var(--select-max-width, 30ch);
- background-color: var(--select-background-color, $white);

azangru:
Been playing with the min-width.

Shouldn't we make sure that the arrow down is safely within the right padding area?

image

msmanoj:
> Been playing with the min-width.
> 
> Shouldn't we make sure that the arrow down is safely within the right padding area?

Well spotted! I've increased the right padding

src/shared/components/simple-select/SimpleSelect.tsx
```diff
@@ -0,0 +1,130 @@
...
+              ))}
+            </optgroup>
+          );
+        })}
+      </select>
+    </div>
+  );
+};
+
+const SimpleSelectAdapter = (props: SimpleSelectAdapterProps) => {
```
azangru, line 116: Reading through the code, I don't even see the point in this adapter component. What if you passed the props directly to the Select, and if the props contain an `optionGroups` field, then you build the Select one way, whereas if it contains the `options` field, you build it another way?

BTW, you shouldn't need to write `if ((props as OptionGroupsSelectProps).optionGroups)`. Typescript should be able to infer the type of props if you write `if ('optionGroups' in props)` etc.

msmanoj: That's done

azangru:
Clicks on the arrowhead still don't get registered. So sorry.

https://user-images.githubusercontent.com/6834224/155709483-a6137166-eaa0-4a0e-846a-e4232d818c22.mp4

azangru (approved):

PR #698, opened 2022-02-21
Updated blast config labels

jyothishnt:
## Type

- [ ] Bug fix
- [ ] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1487

## Description
Updated blast labels after review by AndreaW

## Deployment URL
http://update-blast-labels.review.ensembl.org/

## Views affected
Blast

### Other effects
NA

### DIFF
--- a/src/content/app/tools/blast/components/blast-settings/blastSettingsConfig.json
+++ b/src/content/app/tools/blast/components/blast-settings/blastSettingsConfig.json
@@ -47,7 +47,7 @@
       ]
     },
     "alignments": {
-      "label": "Alignments",
+      "label": "Max. alignments",
       "description": "Maximum number of match alignments reported in the result output.",
       "type": "select",
       "options": [
@@ -102,7 +102,7 @@
       ]
     },
     "scores": {
-      "label": "Scores",
+      "label": "Max. scores",
       "description": "Maximum number of match score summaries reported in the result output.",
       "type": "select",
       "options": [
@@ -157,7 +157,7 @@
       ]
     },
     "hsps": {
-      "label": "High-scoring Segment Pair",
+      "label": "HSPs per hit",
       "description": "Maximum number of HSPs (alignments) to keep for any single query-subject pair.",
       "type": "select",
       "options": [
@@ -168,7 +168,7 @@
       ]
     },
     "dropoff": {
-      "label": "Dropoff",
+      "label": "Drop-off",
       "description": "The amount a score can drop before gapped extension of word hits is halted",
       "type": "select",
       "options": [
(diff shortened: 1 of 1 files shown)


azangru (approved):
Labels correspond to the ticket. The label "statistical accuracy" instead of "composition-based statistics" or "composition-based adjustments" makes no sense to me (I've checked the articles referenced in the [docs](https://www.ebi.ac.uk/seqdb/confluence/pages/viewpage.action?pageId=94147939)), but there is nothing I can do. I wish someone like @andrewyatz could review these labels 😞

ens-ma7:
@jyothishnt the review URL isn't working for me. Its giving this error:
`default backend - 404`
can you check if it works for you.

ens-ma7:
> Labels correspond to the ticket. The label "statistical accuracy" instead of "composition-based statistics" or "composition-based adjustments" makes no sense to me (I've checked the articles referenced in the [docs](https://www.ebi.ac.uk/seqdb/confluence/pages/viewpage.action?pageId=94147939)), but there is nothing I can do about it. I wish someone like @andrewyatz could review these labels 😞

I noticed that there is a description field in the json, maybe we can used this with a help icon next to each field. Its just a suggestion, lets see what comes out of the user testing.

ens-ma7:
Probably not related to this PR but i have noticed that HSPs has only one value (100), is that correct? looking at uniprot, the HSPs dropdown has a few options. sorry if it has already been discussed but it feels wrong to have a dropdown with just one value.

azangru:
> Probably not related to this PR but i have noticed that HSPs has only one value (100), is that correct

This is the only (default) value provided by the jDispatcher

> sorry if it has already been discussed but it feels wrong to have a dropdown with just one value

Yes, it is wrong. We'll need to come up with our own list of values; jDispatcher won't give it to us. A similar situation is with the word size, where we are getting three suggested defaults from jDispatcher, but should extend the list with our own values.

azangru:
> I noticed that there is a description field in the json, maybe we can used this with a help icon next to each field.

Yes. But Andrea didn't want question marks next to each field.

ens-ma7:
> > sorry if it has already been discussed but it feels wrong to have a dropdown with just one value
> 
> Yes, it is wrong. We'll need to come up with our own list of values; jDispatcher won't give it to us. A similar situation is with the word size, where we are getting three suggested defaults from jDispatcher, but should extend the list with our own values.

Indeed just noticed there is a jira for this task in the next sprint:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1498

azangru:
> @jyothishnt the review URL isn't working for me. Its giving this error:
> default backend - 404
> can you check if it works for you.

The link to the deployment is wrong. Should be `update-blast-labels`, not `updated`. Same as the name of the branch.

ens-ma7 (approved):

PR #709, opened 2022-03-03
Blast submission and result

msmanoj:
## Type

- [ ] Bug fix
- [x] New feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1504

## Description
- Created a hook to handle submission of the jobs
- A new tab showing the result page will be opened upon submission

## Deployment URL
http://blast-submission-and-result.review.ensembl.org

## Views affected
Blast


src/content/app/tools/blast/state/blast-jobs/blastJobsSelectors.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RootState } from 'src/store';
+
+export const getNextJobId = (state: RootState) =>
```
ens-ma7, line 19: why is it called `getNextJobId`? should it not be `getJobId`

msmanoj: Here we are trying to find the id we can use for the upcoming job.

For example, if we have 3 jobs already stored in the state:
```javascript
{
    jobs: [
        { jobId:0 }, 
        { jobId:1 }, 
        { jobId:2 }
   ]
}
```
`jobs.length` will give us 3, which will be used as the id for the upcoming job.

ens-ma7: I dont think it is for the upcoming job but the current job you just ran. why are we talking about upcoming job? i might have missed something here.

msmanoj: As per the jobs already stored in the store, the next job will be the current job we are trying to run.
I say next job here because the current job is not saved to the store yet.

ens-ma7: I understand your point but still feels it should be just getjobIds. also looking at what its doing, its just returning the length of the object, shouldn't not be returning the value of the jobId instead? (in our case the first one)

src/content/app/tools/blast/hooks/useBlastAPI.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    return;
+  };
+
+  return {
+    submitBlastJob
+  };
+};
+
+export const createBlastSubmissionData = (
```
azangru, line 104: Not that it's important, but I would keep the function that generates the payload in the submit component.

src/content/app/tools/blast/state/blast-jobs/blastJobsSelectors.ts
azangru: I am very surprised to see that you are generating job ids on the client side. I would expect the ids to come from the server.

Again, the new redux slice and selector that are responsible for job results do not _really_ belong in the task that this jira ticket is addressing. If you could cleanly isolate these files into a different branch/PR, it would be better. If not, we can accept the current code and update it in subsequent PRs.

azangru (approved):
I am happy with this PR, but would be even happier if the code concerning job results were moved to a separate PR.

ens-ma7 (approved):

PR #712, opened 2022-03-08
Update svgr loader settings

azangru:
## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1460

## Description
This PR updates the configuration of svgr loader that we use for handling the svg images.

Previously, we used svgr-loader together with `file-loader`, such that a default import of an svg file produced path to the file (to be used in the `img` tag), whereas a `ReactComponent` named import transformed the svg image into a React component.

However, since webpack 5, `file-loader` has been deprecated, and the recommendations have changed. The documentation now recommends using webpack's assets modules, and to mark svgs that you want to import as url paths with a special query in the import.

As a result, after this PR:
- Default import of an svg file will import a React component
- Default import of an svg file marked with `?url` in the end will import a string that is the path to the svg file to be used in the img tag.

## Other changes
- Removed `file-loader` as no longer necessary; regenerated package-lock.json because of a conflict with dev
- Updated server-side webpack config to avoid emitting the assets files, since they aren't going to be used anyway

## Documentation
- SVGR docs https://react-svgr.com/docs/webpack/
- Webpack: Generating assets paths without emitting the assets: https://webpack.js.org/configuration/module/#rulegeneratoremit
- Webpack: using resource queries with require.context: `https://github.com/webpack/webpack/issues/9364#issuecomment-508513772`

## Deployment URL
http://update-svgr-loader.review.ensembl.org

## Checked locally
- unit tests — ✅ 
- typescript types — ✅ 
- development build — ✅ 
- production build — ✅ 
- storybook — ✅

### DIFF
--- a/.storybook/webpack.js
+++ b/.storybook/webpack.js
@@ -44,7 +44,15 @@ module.exports = (config) => {
   });
   config.module.rules.push({
     test: /\.svg$/,
-    use: ['@svgr/webpack', 'file-loader'],
+    oneOf: [
+      {
+        resourceQuery: /url/, // will match all imports that end in `.svg?url`
+        type: 'asset/resource',
+      },
+      {
+        use: ['@svgr/webpack']
+      }
+    ]
   });
 
   config.resolve.extensions.push('.ts', '.tsx');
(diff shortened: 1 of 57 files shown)


jyothishnt (approved):
Thanks for discussing this in the code review. Checked svg icons on different pages. Looks good.

imransl (approved):
Checked locally and everything is working as expected

PR #715, opened 2022-03-11
Update in-app search

azangru:
## Description
This PR adds some improvements around the behaviour of the InAppSearch component

1. Update the component's behaviour while the request is in flight, to both give some feedback to the user and to prevent excessive requests:
  - disable the "Go" button
  - show a spinner
  - these will still show up if the response comes back quickly; but Andrea doesn't mind
  - updated the CircleLoader component to enable two preset sizes: 40x40 and 30x30px; further customisations available via CSS variables

https://user-images.githubusercontent.com/6834224/158028223-9b7e717a-089f-4f73-bef0-ed06bcdf2630.mp4

2. CSS fixes to get rid of the unnecessary vertical scrollbar in the sidebar modal:

image

3. Disabled spellchecking in input fields by default (this can be overridden by parent if necessary) to avoid the annoying red underline (we trust that users will know what they type into the input fields)

image

3. Added typed `useAppDispatch` and `useAppSelector` hooks, which correctly infer the types of the redux state and of return values of redux middleware, as suggested in the [docs](https://redux.js.org/usage/usage-with-typescript#define-typed-hooks).
  Specifically, writing `await dispatch(search(searchParams));` would be impossible without the typed `useAppDispatch`; because the regular `dispatch` does not know that an action creator created by the `createAsyncThunk` function returns a promise when it is `dispatch`ed
5. Using a simple `fetch` instead of the `apiService` for requests to search api.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1495. This PR doesn't fix the problem described in ENSWBSITES-1495 — it's a backend problem — but it adds certain improvements to the behaviour of InAppSearch component, which were made while investigating the problem described in ENSWBSITES-1495.

## Deployment URL(s)
http://in-app-search-fixes.review.ensembl.org

## Views affected
- InAppSearch in the sidebar on the genome browser and entity viewer page
- InAppSearch in the interstitial view on the genome browser and entity viewer page

### DIFF
--- a/src/shared/components/in-app-search/InAppSearch.tsx
+++ b/src/shared/components/in-app-search/InAppSearch.tsx
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 
-import React from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import React, { useState } from 'react';
+import { useSelector } from 'react-redux';
 import classNames from 'classnames';
 import upperFirst from 'lodash/upperFirst';
+
+import { useAppDispatch } from 'src/store';
 
 import {
   search,
@@ -57,19 +59,22 @@ export type Props = {
 
 const InAppSearch = (props: Props) => {
   const { app, genomeId, mode } = props;
+  const [isLoading, setIsLoading] = useState(false);
   const query = useSelector((state: RootState) =>
     getSearchQuery(state, app, genomeId)
   );
   const searchResult = useSelector((state: RootState) =>
     getSearchResults(state, app, genomeId)
   );
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
 
   const onQueryChange = (query: string) => {
     dispatch(updateQuery({ app, genomeId, query }));
   };
 
-  const onSearchSubmit = () => {
+  const onSearchSubmit = async () => {
+    setIsLoading(true);
+
     const searchParams = {
       app,
       genome_id: genomeId,
@@ -77,14 +82,19 @@ const InAppSearch = (props: Props) => {
       page: 1,
       per_page: 50
     };
-    dispatch(search(searchParams));
 
-    if (app === 'entityViewer') {
-      analyticsTracking.trackEvent({
-        category: `${app}_${mode}_search`,
-        action: 'submit_search',
-        label: query
-      });
+    try {
+      await dispatch(search(searchParams));
+
+      if (app === 'entityViewer') {
+        analyticsTracking.trackEvent({
+          category: `${app}_${mode}_search`,
+          action: 'submit_search',
+          label: query
+        });
+      }
+    } finally {
+      setIsLoading(false);
     }
   };
 
@@ -124,7 +134,7 @@ const InAppSearch = (props: Props) => {
         <PrimaryButton
           onClick={onSearchSubmit}
           className={styles.searchButton}
-          isDisabled={!query}
+          isDisabled={!query || isLoading}
         >
           Go
         </PrimaryButton>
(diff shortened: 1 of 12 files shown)


azangru:
**Note for reviewers:** I was considering switching the fetching logic, which is currently located in a thunk in the `inAppSearchSlice` to redux-toolkit query; but in the end, decided against it. In any case, for comparison, this is what an api slice for making such requests via RTK-query might look like:

```ts
import restApiSlice from 'src/shared/state/api-slices/restSlice';

import type { AppName, SearchResults } from './inAppSearchSlice';

type SearchParams = {
  app: AppName;
  genome_id: string;
  query: string;
  page: number;
  per_page: number;
};

type InAppSearchTransformedResponse = {
  app: AppName,
  genome_id: string,
  result: SearchResults
};

// NOTE: this may become part of a larger searchApiSlice, or remain its own slice
const inAppSearchApiSlice = restApiSlice.injectEndpoints({
  endpoints: (builder) => ({
    inAppSearch: builder.query<InAppSearchTransformedResponse, SearchParams>({
      query: (params) => {
        const queryParams = {
          genome_ids: [params.genome_id], // backend expects genome ids as an array
          query: params.query,
          page: params.page,
          per_page: params.per_page
        };

        return {
          url: `/api/search`,
          headers: {
            method: 'POST'
          },
          body: JSON.stringify(queryParams)
        };
      },
      transformResponse: (response: SearchResults, meta, params: SearchParams) => {
        return {
          app: params.app,
          genome_id: params.genome_id,
          result: response
        };
      }
    })
  })
});

```

imransl:
While testing the in-app search in Genome Browser noticed a minor nuisance. The sidebar flashes when I choose to view the searched gene. This doesn't happen in Entity Viewer though. 

https://user-images.githubusercontent.com/32512909/158649446-aa0deaa3-3231-412f-8849-331164ca166c.mov

azangru:
@imransl the flash isn't due to the InAppSearch component, so I suggest addressing it in a different ticket.

I did realise by watching your screengrab though that the `InAppSearch` component in the genome browser was always showing the scroll bar to the right. Turns out that we are using the `SidebarModal` component for the sidebar in the genome browser, we were wrapping it needlessly in another element that was interfering with the height settings. I deleted that element.

src/shared/components/in-app-search/InAppSearch.tsx
```diff
@@ -137,8 +148,12 @@ const InAppSearch = (props: Props) => {
           </div>
         )}
       </div>
-      {searchResult && (
-        <InAppSearchMatches {...searchResult} app={app} mode={mode} />
+      {isLoading ? (
+        <CircleLoader className={styles.spinner} size="small" />
```
jyothishnt, line 152: CircleLoader serves the purpose but would a LoadingButton be better here?

azangru: I discussed this with Andrea when I was considering what to do with the button — whether to use a loading button or to disable it on submit. Andrea preferred a spinner in the results section. As for the `size` prop, she also said she would like us to make the spinner in the interstitial smaller.

https://user-images.githubusercontent.com/6834224/158917844-4a999835-bc77-425b-a5a3-0fb818bfc688.mp4

She said that, in general, that she expected that we would need a larger spinner and a smaller spinner, which is why I created two default spinner sizes. We can ask her again...

jyothishnt: ok no problem.

imransl (approved):

PR #717, opened 2022-03-14
check for abortController for browser support

ens-ma7:
## Description
Adding one more check for browser support. AbortController isn't supported on new browser (chrome 65 and older). See linked Jira ticket for more details.

This can be tested  on browserstack.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1510

## Deployment URL(s)
http://abortapi-check.review.ensembl.org

## Views affected
All

### DIFF
--- a/src/shared/helpers/browserSupport.ts
+++ b/src/shared/helpers/browserSupport.ts
@@ -26,11 +26,11 @@ const hasOutdatedMediaQueryList = () => {
   );
 };
 
-const abortControllerSupport = () => {
+const hasNoAbortController = () => {
   return !('AbortController' in window);
 };
 
-const browserChecks = [hasOutdatedMediaQueryList, abortControllerSupport];
+const browserChecks = [hasOutdatedMediaQueryList, hasNoAbortController];
 
 const ensureBrowserSupport = () => {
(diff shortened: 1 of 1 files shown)


src/shared/helpers/browserSupport.ts
```diff
@@ -26,7 +26,11 @@ const hasOutdatedMediaQueryList = () => {
   );
 };
 
-const browserChecks = [hasOutdatedMediaQueryList];
+const abortControllerSupport = () => {
+  return !('AbortController' in window);
```
azangru, line 30: why `!`?

ens-ma7: Because the check in the code want the function to return true to show the unsupported-browser page. So when abortcontroller in window is not present for old browser it will return false but we want it to return true. I hope i understood this right :) 

```const browserChecks = [hasOutdatedMediaQueryList, abortControllerSupport];

const ensureBrowserSupport = () => {
  if (browserChecks.some((check) => check())) {
    window.location.replace('/unsupported-browser');
  }
};```

azangru (approved):

PR #718, opened 2022-03-23
Blast jobs polling and storage using observables

azangru:
## Description
This is one of several alternative PRs (see also https://github.com/Ensembl/ensembl-client/pull/719 and https://github.com/Ensembl/ensembl-client/pull/722) to enable the following scenario:

1. Submit the blast from
2. Store the result of the submission
  (the response will be of shape `{submissionId: "some_id", jobs: [ { jobId: "ncbi-blah-blah" }, { error: "error message" }, { jobId: "ncbi-blah-blah" } ]})
    - in redux
    - in indexedDB
3. Poll the jobs while their status is `RUNNING`
4. When the status of a job is reported as `FINISHED` or `FAILURE`:
     - update redux state
     - update indexedDB data
     - stop polling

The solution proposed in this PR has combined all async logic in a redux-observable epic. To be compared with https://github.com/Ensembl/ensembl-client/pull/719 and https://github.com/Ensembl/ensembl-client/pull/722.

### Advantages and disadvantages of the approach

| Advantages  | Disadvantages  |
|---|---|
| Using a redux middleware separates the logic for BLAST submission storage and job polling from the React component tree  | RxJS is an extra dependency (although we already have code in the main branch that is using it)  |
|  RxJS is a very powerful library for orchestrating complex asynchronous logic | RxJS has an api that requires learning |
| RxJS has a declarative syntax promoting code composition  |  |

### How to read this PR
Everything starts from the `BlastJobSubmit` component. It calls the `submitBlast`action creator in `blastApiSlice` when the blast form is submitted. When `submitBlast` receives the response, it transforms it into a proper shape, and completes with an action that can be matched via the `submitBlast.matchFulfilled` matcher. This action triggers redux state update in `blastResultsSlice`, and also starts the `blastFormSubmissionEpic`, which saves the data to indexedDB and polls for job status.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1527
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1529

## Deployment URL(s)
http://blast-submission-observables.review.ensembl.org


src/content/app/tools/blast/state/epics/blastEpics.ts
```diff
@@ -0,0 +1,184 @@
...
+    })
+  );
+
+/**
+ * Similar to blastFormSubmissionEpic, but listens to redux action
+ * that reads BLAST submissions back from IndexedDB.
+ * It checks for running jobs among the saved submissions,
+ * and polls their status until their are finished or failed.
+ */
+export const blastSubmissionsResroteEpic: Epic<Action, Action, RootState> = (
```
msmanoj, line 97: `blastSubmissionsResroteEpic` -> `blastSubmissionsRestoreEpic`

src/content/app/tools/blast/state/epics/blastEpics.ts
```diff
@@ -0,0 +1,184 @@
...
+          .filter((job) => job.status === 'RUNNING')
+          .map((job) => ({
+            submissionId,
+            job
+          }))
+      );
+    }),
+    poll(),
+    tap(databaseUpdaterSubject),
+    map((pollingResult) => {
```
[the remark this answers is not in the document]
azangru: > would it be a bad idea to call updateJob within databaseUpdaterSubject to avoid the map on L114?

A redux-observable epic is a stream that starts with some redux action(s) and ends with another redux action. In our case, the final action that comes out of the epic is the return value of`updateJob`.

Updating IndexedDB is a side effect, which is why I put it in a `tap` function. `tap` is an identity function that returns the same value as it receives; but it allows you to perform whatever logic you want inside. Sadly, it's not asynchronous, whereas updating IndexedDB is an asynchronous operation; which is why I created a dedicated subject that can control the rate of indexeddb updates such that the next one will happen only after the first.

src/content/app/tools/blast/state/epics/blastEpics.ts
```diff
@@ -0,0 +1,184 @@
...
+ * An epic that listens to redux action with the data from a BLAST form submission.
+ * It saves the submission to IndexedDB, polls for job statuses,
+ * and when a job is finished or failed, updates its status in indexedDB,
+ * and sends a redux action to update its status in redux
+ */
+export const blastFormSubmissionEpic: Epic<Action, Action, RootState> = (
+  action$
+) =>
+  action$.pipe(
+    filter(submitBlast.matchFulfilled),
```
msmanoj, line 65: Can we use `ofType` instead of `filter`?

azangru: I believe `ofType` won't work, because what `redux-toolkit` gives us for RTK-query endpoints is matcher functions for the thunks dispatched by the endpoint rather that a specific action type. As you can see in RTK docs [here](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints), only `matchFulfilled`, `matchPending` and `matchRejected` functions are available. The `ofType` operator doesn't work with matcher functions :-)

I am a bit skeptical of the whole purpose of the `ofType` operator, to be honest. As you can see from their source code [here](https://github.com/redux-observable/redux-observable/blob/6a6518c1052da17f271ad3a19b28b515d89008e8/src/operators.ts#L16-L48), it's just a filter with some typescript shenanigans, which we already have.

jyothishnt:
Found issues connecting jdispatcher
First time the status endpoint returned a 404 with error message that says wrong JDispatcher endpoint.
second time it returned FAILURE.

Another job submission was good and opened new tab. But a refresh of the page did a poll which returned an error on the status endpoint with blank error message.  

Is this a problem with the jdispatcher that we should inform?

src/content/app/tools/blast/state/epics/blastEpics.ts
azangru: ✅

src/content/app/tools/blast/state/epics/tests/blastEpics.test.ts
```diff
@@ -106,16 +104,16 @@ afterAll(() => server.close());
 
 describe('blast epics', () => {
   let store: ReturnType<typeof buildReduxStore>['store'];
-  let shutdownEpics: ReturnType<typeof buildReduxStore>['shutdownEpics'];
+  let shutdownTestEpic: ReturnType<typeof buildReduxStore>['shutdownEpic'];
 
   beforeEach(() => {
-    const { store: _store, shutdownEpics: _shutdownEpics } = buildReduxStore();
+    const { store: _store, shutdownEpic } = buildReduxStore();
```
msmanoj, line 110: Looks like we can simply do `({ store, shutdownEpic } = buildReduxStore());` to re-assign.

PR #720, opened 2022-03-24
Blast config file update

ens-ma7:
update blast config file with finalised parameters and values and to deal with wordsize limit for each program

## Description
The changes in the config file includes:
Updating the values for the blast parameters which has been finalised in this [google sheet](https://docs.google.com/spreadsheets/d/1FHtfkLIcm0VWaQvpQqZ8ursGAZLMo-W9tPLw7XJgQmc/edit#gid=0)

And also updating the file shape to include the limit for wordsize. Depending on the program (blastn, blastx, tblastx, tblastn,..), the range for the dropdown can be different (see google sheet above). 

The new choosen shape is a separate object which list the parameters limit
`  "programs_parameters_limit": {
    "wordsize": {
      "blastn" : {
        "min": 2,
        "max": 15 
      },
      "blastx" : {
        "min": 2,
        "max": 8
      },
      "blastp" : {
        "min": 2,
        "max": 8
      },
      "tblastn" : {
        "min": 2,
        "max": 8
      },
      "tblastx": {
        "min": 2,
        "max": 6
      }
    }
  },`

For now this has been adapted for wordsize specifically but in the future can be improved if required for other parameters limit.

Tip to review code:
- The changes in `src/content/app/tools/blast/views/blast-form/BlastForm.tsx` is only temporary so that it picks the local config file
- The actual changes start in the config file (check updated values make sense and the new programs_parameters_limit object)
- Then review type update change in here `src/content/app/tools/blast/types/blastSettings.ts` which should match the new shape of the file
- And the code change to deal with the new parameter limit added is in `src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx`

**IMPORTANT NOTE:**
- Need to remove the line using the local blast config file
- Once this PR has been approved, send the new config file to backend (Andres) to update ToolsAPI
- Coordinate the ToolsAPI update with merging this PR in dev and updating master

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1532

## Deployment URL(s)
http://newblast-config.review.ensembl.org

## Views affected
http://newblast-config.review.ensembl.org/blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
@@ -43,6 +43,7 @@ import type {
   BlastSelectSetting,
   BlastBooleanSetting,
   BlastParameterName,
+  ParameterRange,
   BlastSettingsConfig
 } from 'src/content/app/tools/blast/types/blastSettings';
 
@@ -238,6 +239,8 @@ const BlastSettings = ({ config }: Props) => {
           {buildSelect({
             ...(config.parameters['wordsize'] as BlastSelectSetting),
             selectedOption: blastParameters.wordsize as string,
+            limit: config.programs_parameters_limit['wordsize'],
+            blastProgram: blastProgram,
             onChange: (value: string) =>
               onBlastParameterChange('wordsize', value)
           })}
@@ -287,12 +290,25 @@ const buildSelect = (setting: {
   options: { label: string; value: string }[];
   label: string;
   selectedOption: string;
+  limit?: ParameterRange;
+  blastProgram?: BlastProgram;
   onChange: (value: string) => void;
 }) => {
   const onChange = (e: FormEvent<HTMLSelectElement>) => {
     const value = e.currentTarget.value;
     setting.onChange(value);
   };
+
+  const blastProgram = setting.blastProgram as BlastProgram;
+  if (setting.limit && setting.limit[blastProgram]) {
+    setting.options = setting.options.filter(
+      (item) =>
+        parseInt(item.value) >=
+          (setting.limit as ParameterRange)[blastProgram].min &&
+        parseInt(item.value) <=
+          (setting.limit as ParameterRange)[blastProgram].max
+    );
+  }
 
   return (
     <div className={styles.select}>
--- a/src/content/app/tools/blast/types/blastSettings.ts
+++ b/src/content/app/tools/blast/types/blastSettings.ts
@@ -85,6 +85,16 @@ type ProgramConfiguration = {
   programs: BlastProgram[];
 };
 
+export type ParameterRange = {
+  [program: string]: {
+    min: number;
+    max: number;
+  };
+};
+type ProgramParametersLimit = {
+  [BlastParameterName: string]: ParameterRange;
+};
+
 type Defaults = {
   database: string;
 };
@@ -92,6 +102,7 @@ type Defaults = {
 export type BlastSettingsConfig = {
   parameters: Record<BlastParameterName, BlastSetting>;
   programs_configurator: ProgramConfiguration[];
+  programs_parameters_limit: ProgramParametersLimit;
   database_sequence_types: Record<string, SequenceType>;
   defaults: Defaults;
   presets: Presets;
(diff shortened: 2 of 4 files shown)


azangru:
> programs_parameters_limit
> ...
> For now this has been adapted for wordsize specifically but in the future can be improved if required for other parameters limit.

Unless we expect to switch to free text inputs for which we validate user's input, I am not sure I like this idea. There doesn't seem to be anything in the BLAST tool that suggests that in the future we will have other options for which the client will specifically need to generate ranges of values between the upper and the lower limit. It is a very special client-side logic that is only required for the word size.

We could approach it either more generally, by listing the actual values of the wordsize parameter:

```json
"program_parameter_overrides": {
  "wordsize": {
    "blastn" : [
      {
        "label": "2",
        "value": "2"
      },
      {
        "label": "3",
        "value": "3"
      }
    ],
    "blastx": [
      {
        "label": "2",
        "value": "2"
      },
      {
        "label": "3",
        "value": "3"
      },
    ]
  }
}
```
and so on. This would make such config field general enough to be used for any parameter whose values depend on the blast program.

Alternatively, we could have done the opposite and written something very specific:

```json
"word_sizes": {
  "blastn": {
    "min": 2,
    "max": 15
  },
  "blastx": {
    "min": 2,
    "max": 8
  }
}
```

But the program parameter limits field is a strange middle ground between the two approaches. It looks as if it's general, but at thee same time is very specific.

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -287,13 +290,26 @@ const buildSelect = (setting: {
   options: { label: string; value: string }[];
   label: string;
   selectedOption: string;
+  limit?: ParameterRange;
+  blastProgram?: BlastProgram;
```
azangru, line 294: The `buildSelect` doesn't need to know all these. Instead, we should have a dedicated function that transforms the combination of the blast program and the max/min values into an array of options, of the shape defined on line 290.

src/content/app/tools/blast/types/blastSettings.ts
```diff
@@ -85,13 +85,24 @@ type ProgramConfiguration = {
   programs: BlastProgram[];
 };
 
+export type ParameterRange = {
+  [program: string]: {
+    min: number;
+    max: number;
+  };
+};
+type ProgramParametersLimit = {
```
azangru, line 94: If we go with the min/max approach, we should review the use of the word "limit". I am not sure that the word _limit_ (singular) can mean both min and max at the same time.

ens-ma7:
After the PR reivew this morning, we decided to go with Andrey's suggestion using `program_parameter_overrides` as a separate object with all the overwritten values.

azangru (approved):
Looks good to me.

msmanoj (approved):

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
```diff
@@ -34,6 +34,11 @@ import BlastSpeciesSelector from 'src/content/app/tools/blast/components/blast-s
 
 import styles from './BlastForm.scss';
 
+// ToDo: Remove the 2 imports below once PR is approved and revert the testConfig props change on line 53 (Only for development env and review env)
+// Then we need to feedback to backend to update the config file for the tools API and coordinate the master branch update
+import testConfig from 'tests/fixtures/blast/blastSettingsConfig.json';
+import { BlastSettingsConfig } from '../../types/blastSettings';
```
imransl, line 40: Import order is not right here

tests/fixtures/blast/blastSettingsConfig.json
```diff
@@ -308,63 +320,75 @@
...
-          "label": "11",
-          "value": "11"
+          "label": "7",
+          "value": "7"
         },
         {
-          "label": "28",
-          "value": "28"
+          "label": "8",
+          "value": "8"
```
jyothishnt, line 391: I guess you added these because most of the programs wordsize range is 2-8?

jyothishnt (approved):

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
ens-ma7: @imransl thanks for reviewing but the changes to this file is only temporary so that we can test using the local file. Once this PR is approved, the changes will be reverted for this file. Hope thats ok.

tests/fixtures/blast/blastSettingsConfig.json
ens-ma7: @jyothishnt yes the default is 2-8 and then its overwritten further down for other program.

imransl (approved):

PR #721, opened 2022-03-24
Use the latest package of ensembl-genome-browser (0.2.1)

imransl:
## Description
Use the latest ensembl-genome-browser test package, which will be changed to `0.2.1` before merging this branch. The latest package includes a fix to the genome browser initialisation and the latest wasm build from ensembl-dauphin-style-compiler.

The related ensembl-genome-browser PR is: https://github.com/Ensembl/ensembl-genome-browser/pull/7

## Related JIRA Issue(s)
[ENSWBSITES-1511](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1511)

## Deployment URL(s)
http://test-gb-package.review.ensembl.org/

## Views affected
Genome browser

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -144,6 +144,7 @@
     "../ensembl-genome-browser": {
       "name": "@ensembl/ensembl-genome-browser",
       "version": "0.2.1-test.2",
+      "extraneous": true,
       "license": "ISC",
       "devDependencies": {
         "@rollup/plugin-typescript": "8.2.1",
@@ -2294,8 +2295,9 @@
       "dev": true
     },
     "node_modules/@ensembl/ensembl-genome-browser": {
-      "resolved": "../ensembl-genome-browser",
-      "link": true
+      "version": "0.2.1-test.2",
+      "resolved": "https://gitlab.ebi.ac.uk/api/v4/projects/3500/packages/npm/@ensembl/ensembl-genome-browser/-/@ensembl/ensembl-genome-browser-0.2.1-test.2.tgz",
+      "integrity": "sha1-AJ8aiear1touhOgmcZ8BT+LJtqg="
     },
     "node_modules/@eslint/eslintrc": {
       "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -141,6 +141,22 @@
...
+        "@rollup/plugin-typescript": "8.2.1",
+        "@typescript-eslint/eslint-plugin": "4.28.1",
+        "@typescript-eslint/parser": "4.28.1",
+        "eslint": "7.29.0",
+        "rollup": "2.52.6",
+        "rollup-plugin-copy": "3.4.0",
+        "tslib": "2.3.1",
+        "typescript": "4.3.4"
+      }
+    },
```
azangru, line 159: This whole block was created when you were building ensembl-genome-browser locally, and shouldn't be committed to our dev.

imransl: Thanks for pointing this out. Also noticed, the `.../ensembl-genome-browser` is strangely properly resolved yet. Will look into this again.

imransl: Removed it ✅

package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@apollo/client": "3.5.10",
-    "@ensembl/ensembl-genome-browser": "0.2.0",
+    "@ensembl/ensembl-genome-browser": "0.2.1-test.2",
```
imransl, line 34: *NOTE TO SELF*: Change the version to `0.2.1` when the related PR in `ensembl-genome-browser` is merged.

imransl: Done ✅

azangru (approved):
I am seeing server-side errors in the console, which I thought could be linked to some strange behaviour of the genome browser upon unmounting and remounting of a component; but it seems that these errors may also occur while refreshing the genome browser page. So probably nothing to do with 0.2.1 client directly.

Approved.

msmanoj (approved):

PR #719, opened 2022-03-24
Blast jobs polling and storage using React context

azangru:
## Description
This is one of several alternative PRs (see also https://github.com/Ensembl/ensembl-client/pull/718) to enable the following scenario:

1. Submit the blast from
2. Store the result of the submission
  (the response will be of shape `{submissionId: "some_id", jobs: [ { jobId: "ncbi-blah-blah" }, { error: "error message" }, { jobId: "ncbi-blah-blah" } ]})
    - in redux
    - in indexedDB
3. Poll the jobs while their status is `RUNNING`
4. When the status of a job is reported as `FINISHED` or `FAILURE`:
     - update redux state
     - update indexedDB data
     - stop polling

The solution proposed in this PR has extracted the logic for job status polling and for saving and updating blast submissions in indexedDB into a dedicated tools context. To be compared with https://github.com/Ensembl/ensembl-client/pull/718.

### Advantages and disadvantages of the proposed solution

| Advantages  | Disadvantages  |
|---|---|
| No extra dependencies required | The solution required creating a dedicated context component, and therefore is tied to the structure of the React component tree |
| The polling logic can be represented with an async iterator in a rather simple way |  |

### How to read this PR
The specific functions that run when a Blast form is submitted have been extracted into a dedicated React context (see `ToolsContextContainer`). When the user presses the button in the `BlastJobSubmit` component, the `submitBlastForm` function gets called. It submits the form through the `submitBlast` action creator in `blastApiSlice`, waits for the response (of shape `{ submissionId: string; submission: BlastSubmission }`), saves this to indexedDB, and stats polling for the status of individual jobs within the blast submission.

The polling logic (see the `getFinishedJobs` function) is written as an async iterator.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1527
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1529

## Deployment URL(s)
http://save-blast-submission.review.ensembl.org

PR #722, opened 2022-03-30
Blast jobs polling and storage using action listener middleware 

azangru:
## Description
This is one of several alternative PRs (see also https://github.com/Ensembl/ensembl-client/pull/718 and https://github.com/Ensembl/ensembl-client/pull/719) to enable the following scenario:

1. Submit the blast from
2. Store the result of the submission
  (the response will be of shape `{submissionId: "some_id", jobs: [ { jobId: "ncbi-blah-blah" }, { error: "error message" }, { jobId: "ncbi-blah-blah" } ]})
    - in redux
    - in indexedDB
3. Poll the jobs while their status is `RUNNING`
4. When the status of a job is reported as `FINISHED` or `FAILURE`:
     - update redux state
     - update indexedDB data
     - stop polling

The solution proposed in this PR is using the recently released **action listener middleware**, which is a part of redux-toolkit (see [docs](https://redux-toolkit.js.org/api/createListenerMiddleware)) To be compared with https://github.com/Ensembl/ensembl-client/pull/718 and https://github.com/Ensembl/ensembl-client/pull/719.

### Advantages and disadvantages of the proposed solution

| Advantages  | Disadvantages  |
|---|---|
| Job submission and polling logic is not tied to React component tree | It is less powerful than redux-observable. For example, it has no way of comparing the current action with the previous action. Its cancellation api is clunky. |
| The api is smaller and simpler than that of redux-observable; less of a learning curve | Notice how we had to write our own timer logic and fetcher logic. Redux-observable would already have ready-made functions. |
| Redux listener middleware is part of redux-toolkit and is officially supported by the redux team | Since the purpose of this middleware is similar to redux-observable, we'll have to get rid of redux-observable if we adopt this one |

### How to read this PR
This is PR uses action listener middleware, which means that there is a redux middleware responding to actions of certain type. Specifically, it is interested in the `submitBlast` action dispatched by `BlastJobSubmit` component, and in `restoreBlastSubmissions` action dispatched by the root component. The listener middleware (see `blastActionListeners.ts` file) intercepts these actions and starts polling for job statuses. When a job is finished (or failed), its status will get updated in redux and in indexedDB.

The polling logic (see the `pollJobStatuses` function) is written as an async iterator.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1527
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1529

## Deployment URL(s)
http://blast-submission-rtk.review.ensembl.org


src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -27,8 +43,45 @@ const blastApiSlice = restApiSlice.injectEndpoints({
...
+          data: null
+        }));
+        return {
+          submissionId,
+          submission: {
+            submission: payload,
+            results,
+            submittedAt: Date.now()
+          } as unknown as BlastSubmission
+        };
```
azangru, line 80: @msmanoj you pointed out in https://github.com/Ensembl/ensembl-client/pull/719 that you don't like a `submission` field within another `submission` field.

In order to address this, we need to consider the `BlastSubmission` type, which is declared in `blastResultsSlice`:

https://github.com/Ensembl/ensembl-client/blob/6ec22fdea53e303ae768d300bc78252aabd9f406/src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts#L37-L50

You will notice that this type, although called `BlastSubmission`, has a `submission` field inside of it. The purpose of this field is to keep the content of the blast submission form sent to the tools api. If you can think of another name, or of a different shape of this type, I am open for suggestions.

By the way, this type is likely to change at least somewhat after https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1566 is completed.

src/content/app/tools/blast/state/action-listeners/blastActionListeners.ts
```diff
@@ -0,0 +1,150 @@
...
+  ) => {
+    const { submissionId, submission } = action.payload;
+    const { dispatch, signal } = listenerApi;
+
+    await saveBlastSubmission(submissionId, submission);
+
+    const runningJobs = submission.results.map(({ jobId }) => ({
+      submissionId,
+      jobId
+    }));
```
azangru, line 54: @msmanoj here and below, updated the variable name from jobIds to runningJobs, according to your [comment](https://github.com/Ensembl/ensembl-client/pull/719#discussion_r837525133 ) from https://github.com/Ensembl/ensembl-client/pull/719.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
msmanoj: Would it be okay rename the `submission` key on L75 to `submissionData`?

azangru: Meh. It's not like the word `data` adds any extra meaning.

src/content/app/tools/blast/state/action-listeners/blastActionListeners.test.ts
```diff
@@ -0,0 +1,243 @@
...
+import listenerMiddleware, {
+  startAppListening
+} from 'src/root/actionListenerMiddleware';
+
+import {
+  submitBlastListener,
+  resforeBlastSubmissionsListener
+} from 'src/content/app/tools/blast/state/action-listeners/blastActionListeners';
+
+import restApiSlice from 'src/shared/state/api-slices/restSlice';
```
imransl, line 32: Move this import below to the group of state imports?

src/content/app/tools/blast/state/action-listeners/blastActionListeners.ts
```diff
@@ -0,0 +1,150 @@
...
+} from 'src/content/app/tools/blast/state/blast-results/blastResultsSlice';
+
+import {
+  saveBlastSubmission,
+  updateSavedBlastJob
+} from 'src/content/app/tools/blast/services/blastStorageService';
+
+import { POLLING_INTERVAL } from './blastActionListenerConstants';
+
+import type { BlastSubmission } from '../blast-results/blastResultsSlice';
```
imransl, line 34: Move this import to the other slice imports above?

src/content/app/tools/blast/state/action-listeners/blastActionListeners.ts
```diff
@@ -0,0 +1,150 @@
...
+        const { status } = await fetchJobStatus({ jobId: job.jobId, signal });
+        if (['FINISHED', 'FAILURE'].includes(status)) {
+          jobs = jobs.filter(({ jobId }) => jobId !== job.jobId);
+          yield {
+            ...job,
+            fragment: { status }
+          };
+        }
+      } catch {
+        // don't care
```
imransl, line 114: Don't we need to show anything to the user if this fails?

azangru: We are still polling though. From user's perspective, it is no different than if the response was that the job was still running.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
azangru: Changed the field name to `submittedData`.

src/content/app/tools/blast/state/action-listeners/blastActionListeners.ts
azangru: Ah, I see I already have imports from `blastResultsSlice`; I'll move it here. Normally though, if we don't have other imports from the same module, we tend to group purely type imports in the bottom, right above the styles.

src/content/app/tools/blast/state/action-listeners/blastActionListeners.test.ts
azangru: ✅

imransl (approved):

azangru:
Closing in favour of https://github.com/Ensembl/ensembl-client/pull/718

PR #724, opened 2022-04-04
Fix flashing issue in in-app search for GB

imransl:
## Description
When clicking on a search result in the in-app search for Genome browser, the sidebar (track panel) flashes momentarily. This does not happen in Entity viewer so needed fixing. This was first observed in #715.

Steps to reproduce this error:

1. Go to Genome browser
2. Open in-app search by clicking on the search link in the track panel sidebar panel 
3. Do a search with a gene that is not the current gene (e.g. ABO if current gene is BRCA2)
4. Click on the search result and then the Genome browser icon
5. Observe the flashing in the track panel

For a visual depiction of this issue please check the video below: 

https://user-images.githubusercontent.com/32512909/158649446-aa0deaa3-3231-412f-8849-331164ca166c.mov

### How to read this PR
This issue is fixed by loading the `TrackPanelModal` inside the `Browser` component instead of the `TrackPanel` component. There were some checks that needed to be changed for this in `Browser.tsx` and `TrackPanel.tsx` files.

Additionally, the `TrackPanelModal` code was changed to dynamically import the modal views similar to how it happens in `EntityViewer`. An enum (`TrackPanelModalViews`) was also introduced into this code, hence some changes in the Redux code.

## Related JIRA Issue(s)
[ENSWBSITES-1538](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1538)

## Deployment URL(s)
http://fix-sidebar-flashing.review.ensembl.org

## Views affected
Genome browser -> Track panel


azangru:
I see lots of changes in this PR, but where, specifically, was the problem?

imransl:
> I see lots of changes in this PR, but where, specifically, was the problem?

I've updated the description with how to read the PR

src/content/app/genome-browser/Browser.tsx
```diff
@@ -90,13 +95,19 @@ export const Browser = () => {
     </>
   );
 
+  const SideBarContent = isTrackPanelModalOpened ? (
+    <TrackPanelModal />
+  ) : (
+    <TrackPanel />
+  );
```
azangru, line 102: I bet this was the only change strictly necessary to fix the bug 😄

msmanoj:
Without this fix, it doesn't seem to flash if we select a gene we have already visited before.

src/content/app/genome-browser/components/track-panel/components/track-panel-modal/TrackPanelModal.tsx
```diff
@@ -14,58 +14,74 @@
...
     dispatch(closeDrawer());
     dispatch(closeTrackPanelModal());
   };
 
-  const { title, content } = getModalViewData();
   return (
-    <SidebarModal title={title} onClose={onClose}>
-      {content}
-    </SidebarModal>
+    <section>
```
azangru, line 78: I'm unsure of this html element. At least in the EntityViewer it is trying to do some useful job:

image

whereas here it isn't doing anything useful.

We might want to replicate here what the section element is doing in the EntityViewer (height 100%, overflow: auto). Or, failing that, get rid of this element altogether.

imransl:
> Without this fix, it doesn't seem to flash if we select a gene we have already visited before.

Yes it is only the first time that it flashes

msmanoj:
Doing the following inside `TrackPanelModal` seems to fix this bug:

```
return isTrackPanelModalOpened ? <TrackPanelModal /> : (
    shouldRenderContent ? <TrackPanelList /> : <SidebarLoader />
  );
```

src/content/app/genome-browser/components/track-panel/components/track-panel-modal/TrackPanelModal.tsx
imransl: I've removed the `section` element from `TrackPanelModal`. Also removed it from `EntityViewerSidebarModal`. None of the styling in `EntityViewerSidebarModal.scss` was used either so deleted the file.

azangru (approved):

imransl:
As discussed in the Code Review meeting, I have created a [new ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1571) to rename and move `TrackPanelModal` out of `TrackPanel`.

PR #726, opened 2022-04-11
Refactor TrackPanelModal and TrackPanelBar

imransl:
## Description
`TrackPanelModal` and `TrackPanelBar` did not belong inside TrackPanel, so they have been moved out of there into their own directories. `TrackPanelModal` has been renamed to `BrowserSidebarModal` and `TrackPanelBar` to `BrowserSidebarToolstrip`. The redux code of `TrackPanelModal` has also been moved out of `TrackPanel`. Additionally, `TrackPanelStorage` has been renamed to `BrowserSidebarModalStorageService` as the code had nothing to do with the former.

## Related JIRA Issue(s)
[ENSWBSITES-1571](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1571)

## Deployment URL(s)
http://rename-track-panel-modal.review.ensembl.org

## Views affected
Genome browser -> Sidebar/TrackPanel


src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-sidebar-modal-storage-service.ts
```diff
@@ -18,14 +18,14 @@ import storageService, {
...
 } from 'src/services/storage-service';
 
-import { PreviouslyViewedObject } from 'src/content/app/genome-browser/state/track-panel/trackPanelSlice';
+import { PreviouslyViewedObject } from 'src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice';
 
 export enum StorageKeys {
-  BOOKMARKS = 'trackPanel.bookmarks',
-  PREVIOUSLY_VIEWED = 'trackPanel.previouslyViewedObjects'
+  BOOKMARKS = 'browserSidebarModal.bookmarks',
+  PREVIOUSLY_VIEWED = 'browserSidebarModal.previouslyViewedObjects'
```
azangru, line 25: It's really just a UI accident that bookmarks or previously viewed objects are being shown in a modal; the key in the storage shouldn't care. Should think about how better to name this. Perhaps even `genomeBrowser.bookmarks`.

src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-sidebar-modal-storage-service.ts
```diff
@@ -64,4 +64,4 @@ export class TrackPanelStorageService {
   }
 }
 
-export default new TrackPanelStorageService(storageService);
+export default new BrowserSidebarModalStorageService(storageService);
```
azangru, line 67: Does this perhaps look rather like `GenomeBrowserBookmarksService`?

src/content/app/genome-browser/components/browser-sidebar-toolstrip/BrowserSidebarToolstrip.test.tsx
```diff
@@ -83,93 +86,101 @@ describe('<TrackPanelBar />', () => {
...
-            action.type === trackPanelActions.updateTrackPanelForGenome.type
+            action.type ===
+            browserSidebarModalActions.updateBrowserSidebarModalForGenome.type
         );
 
       const expectedPayload = {
         activeGenomeId: fakeGenomeId,
         data: {
-          ...mockState.browser.trackPanel[fakeGenomeId],
-          isTrackPanelOpened: true
```
azangru, line 140: This test was also checking that the track panel itself gets open:

https://user-images.githubusercontent.com/6834224/163131612-9ecfee97-cd8c-40f9-843d-2471483b251f.mp4

src/content/app/genome-browser/services/browser-storage-service.ts
```diff
@@ -99,6 +104,18 @@ export class BrowserStorageService {
...
+    [genomeId: string]: Partial<BrowserSidebarModalState>;
+  } {
+    return this.storageService.get(StorageKeys.SIDEBAR_MODAL) || {};
+  }
+
+  public updateBrowserSidebarModals(browserSidebarModals: {
+    [genomeId: string]: Partial<BrowserSidebarModalStateForGenome> | undefined;
+  }): void {
+    this.storageService.update(StorageKeys.SIDEBAR_MODAL, browserSidebarModals);
+  }
```
azangru, line 117: I forgot — did Andrea really want us to store which modals were open for which species? So that if you reload the browser the modal is still open?

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -0,0 +1,289 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PayloadAction } from '@reduxjs/toolkit';
+import { ThunkAction } from 'redux-thunk';
+import { createSlice } from '@reduxjs/toolkit';
+import { Action } from 'redux';
```
azangru, line 20: 1) When importing types, please use either `import type { Foo }` or `import { type Foo }` syntax. The first when you only import types from a module; the second when you also import some actual variables.
2) I believe all three types can be imported from `@reduxjs/toolkit`. Certainly the `Action` type can; don't remember about `ThunkAction`, but I seem to remember that yes.

azangru: Probably all these four lines can be rewritten as:

`import { createSlice, type PayloadAction, type Action, type ThunkAction } from @reduxjs/toolkit`;

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -0,0 +1,289 @@
...
+  getBrowserActiveGenomeId,
+  getBrowserActiveFocusObject
+} from 'src/content/app/genome-browser/state/browser-general/browserGeneralSelectors';
+import {
+  getActiveGenomePreviouslyViewedObjects,
+  getActiveBrowserSidebarModal
+} from './browserSidebarModalSelectors';
+
+import { ParsedUrlPayload } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
+import { RootState } from 'src/store';
```
azangru, line 36: `import type`

src/content/app/genome-browser/services/browser-storage-service.ts
azangru: So now that Andrea said she doesn't want the open/closed state of the modal to survive browser refresh, what are we storing here, and should we be storing anything at all?

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -0,0 +1,289 @@
...
+  previouslyViewedObjects: PreviouslyViewedObject[];
+}>;
+
+export type BrowserSidebarModalState = Readonly<{
+  [genomeId: string]: BrowserSidebarModalStateForGenome;
+}>;
+
+export const defaultBrowserSidebarModalStateForGenome: BrowserSidebarModalStateForGenome =
+  {
+    isBrowserSidebarModalOpened: false,
```
azangru, line 71: Is there any point in having this key to the state at all? Isn't `browserSidebarModalView` key sufficient to indicate whether the modal is open (it should be if it's not null)?

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -0,0 +1,289 @@
...
+    bookmarks: [],
+    previouslyViewedObjects: [],
```
azangru, line 73: I wonder whether these two should be part of browserSidebarModalSlice; or whether they should be move to a dedicated genomeBrowserBookmarksSlice. Doesn't necessarily need to happen in this PR, but something to think about.

src/content/app/genome-browser/services/browser-storage-service.ts
imransl: We are using this to store the previously viewed objects for Bookmarks. So probably should go into the new bookmarks storage service.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.scss
```diff
@@ -31,7 +31,8 @@
 
 .more {
   margin-top: 25px;
-  span { /* stylelint-disable-line no-descending-specificity */
+  span {
+    /* stylelint-disable-line no-descending-specificity */
```
azangru, line 35: Does stylelint complain if we remove this line?

src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-bookmarks-storage-service.ts
```diff
@@ -18,14 +18,14 @@ import storageService, {
...
 } from 'src/services/storage-service';
 
-import { PreviouslyViewedObject } from 'src/content/app/genome-browser/state/track-panel/trackPanelSlice';
+import type { PreviouslyViewedObject } from 'src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice';
 
 export enum StorageKeys {
-  BOOKMARKS = 'trackPanel.bookmarks',
-  PREVIOUSLY_VIEWED = 'trackPanel.previouslyViewedObjects'
+  BOOKMARKS = 'browserSidebarModal.bookmarks',
+  PREVIOUSLY_VIEWED = 'browserSidebarModal.previouslyViewedObjects'
```
azangru, line 25: storage key: replace the part `browserSidebarModal` with `genomeBrowser`.

src/content/app/genome-browser/services/browser-storage-service.ts
```diff
@@ -27,14 +27,19 @@ import {
   TrackPanelState,
   TrackPanelStateForGenome
 } from 'src/content/app/genome-browser/state/track-panel/trackPanelSlice';
+import {
```
azangru, line 30: import type?

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSelectors.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const getActiveGenomeBrowserBookmarks = (state: RootState) => {
+  const activeGenomeId = getBrowserActiveGenomeId(state);
+  const activeGenomeBrowserBookmarks =
+    activeGenomeId && state.browser.browserBookmarks[activeGenomeId];
+
+  return activeGenomeBrowserBookmarks || defaultBrowserBookmarksStateForGenome;
+};
+
+export const getActiveGenomeBookmarks = (state: RootState) => {
```
azangru, line 30: The name of this function looks confusingly similar to `getActiveGenomeBrowserBookmarks`.

Perhaps there is no need for `getActiveGenomeBrowserBookmarks` at all? It doesn't seem to be doing much.

```ts
export const getActiveGenomeBookmarks = (state: RootState) => {
  const noBookmarks = [];
  return activeGenomeId
    ? state.browser.browserBookmarks[activeGenomeId].bookmarks ?? noBookmarks
    : noBookmarks;
}
```

src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
```diff
@@ -15,17 +15,17 @@
...
+  type ActionCreator,
+  type PayloadAction,
+  type ThunkAction
 } from '@reduxjs/toolkit';
 import { batch } from 'react-redux';
 import pickBy from 'lodash/pickBy';
 
 import browserStorageService from 'src/content/app/genome-browser/services/browser-storage-service';
-import trackPanelStorageService from 'src/content/app/genome-browser/components/track-panel/services/track-panel-storage-service';
+import trackPanelStorageService from 'src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-bookmarks-storage-service';
```
azangru, line 28: variable name

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSelectors.ts
```diff
@@ -0,0 +1,34 @@
...
+export const getActiveBrowserSidebarModal = (state: RootState) => {
+  const activeGenomeId = getBrowserActiveGenomeId(state);
+  const activeBrowserSidebarModal =
+    activeGenomeId && state.browser.browserSidebarModal[activeGenomeId];
+
+  return activeBrowserSidebarModal || defaultBrowserSidebarModalStateForGenome;
+};
+
+export const getIsBrowserSidebarModalOpened = (state: RootState) =>
+  Boolean(getActiveBrowserSidebarModal(state).browserSidebarModalView);
```
azangru, line 31: looks like this can be simplified to: 
```
Boolean(getBrowserSidebarModalView(state))
```

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -0,0 +1,160 @@
...
+        data
+      })
+    );
+  };
+
+const browserSidebarModal = createSlice({
+  name: 'genome-browser-sidebar-modal',
+  initialState: {} as BrowserSidebarModalState,
+  reducers: {
+    setInitialBrowserSidebarModalDataForGenome(
```
azangru, line 129: Where are we using this action creator?

src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-sidebar-modal-storage-service.ts
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-sidebar-modal-storage-service.ts
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-toolstrip/BrowserSidebarToolstrip.test.tsx
imransl: ✅

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
imransl: ✅

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
imransl: ✅

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
imransl: ✅

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.scss
imransl: Yes it still does:
```src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.scss
 34:3  ⚠  Expected selector ".more span" to come before selector ".linkHolder .label span + span"  no-descending-specificity```

File: src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts (hunk 15 above)
Comment (imransl) [reply src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts]: It isn't used anywhere. I've removed it ✅

File: src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSelectors.ts (hunk 14 above)
Comment (imransl) [reply src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSelectors.ts]: ✅

File: src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts (hunk 13 above)
Comment (imransl) [reply src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts]: ✅

File: src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSelectors.ts (hunk 12 above)
Comment (imransl) [reply src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSelectors.ts:21]: ✅

File: src/content/app/genome-browser/services/browser-storage-service.ts (hunk 11 above)
Comment (imransl) [reply src/content/app/genome-browser/services/browser-storage-service.ts]: ✅

File: src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-bookmarks-storage-service.ts (hunk 10 above)
Comment (imransl) [reply src/content/app/genome-browser/components/browser-sidebar-modal/services/browser-bookmarks-storage-service.ts]: ✅

File: src/content/app/genome-browser/services/browserStorageService.ts [hunk 16]
```diff
@@ -15,26 +15,31 @@
...
 
 export enum StorageKeys {
   ACTIVE_GENOME_ID = 'browser.activeGenomeId',
   ACTIVE_ENS_OBJECT_ID = 'browser.activeFocusObjectId',
   CHR_LOCATION = 'browser.chrLocation',
   DEFAULT_CHR_LOCATION = 'browser.defaultChrLocation',
   TRACK_STATES = 'browser.trackStates',
-  TRACK_PANELS = 'browser.trackPanels'
+  TRACK_PANELS = 'browser.trackPanels',
+  BOOKMARKS = 'browser.bookmarks'
```
azangru, line 42: Now that bookmarks are in the bookmarks service, are we still using this key?

Also, would there be a benefit in renaming the first part of these keys to genomeBrowser?

src/content/app/genome-browser/services/browserStorageService.ts
```diff
@@ -99,6 +104,18 @@ export class BrowserStorageService {
...
+    [genomeId: string]: Partial<BrowserBookmarksState>;
+  } {
+    return this.storageService.get(StorageKeys.BOOKMARKS) || {};
+  }
+
+  public updateBrowserBookmarks(browserBookmarks: {
+    [genomeId: string]: Partial<BrowserBookmarksStateForGenome> | undefined;
+  }): void {
+    this.storageService.update(StorageKeys.BOOKMARKS, browserBookmarks);
+  }
```
azangru, line 117: I am confused. What is the purpose of these two methods compared to browserBookmarksStorageService?

imransl: I was not going through all of the functions in these files but just renaming them, hence some legacy code remains. My bad.

We had a `TrackPanelStorageService` which was renamed to `BrowserBookmarksStorageService`: https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/genome-browser/components/track-panel/services/track-panel-storage-service.ts

When going through the code now, found out these two are interchangeably used to update the web storage which isn't right. Think we can consolidate the related code from both these files into `BrowserBookmarksStorageService` instead?

src/content/app/genome-browser/services/browserStorageService.ts
imransl: This is related to my comment below. The bookmarks key can be deleted if we get rid of the code related to it here. 

And about renaming from `browser` to `genomeBrowser`, is this because the keys are related to the genome browser itself (wasm) rather than the app (named `Browser` at the moment)?

Also should we create a new `trackPanelStorageService` for the `TrackPanel` code (like how bookmarks has one for itself)?

src/content/app/genome-browser/services/browserStorageService.ts
imransl: As we discussed during code review, everything related bookmarks should be moved into `BrowsersBookmarksStorageService`.

msmanoj: I agree to that as well @imransl . It would be good to move the code related to bookmarks into `BrowserBookmarksStorageService`. 

Regarding creating a new `trackPanelStorageService`, I don't think we need it at this point or in this PR.

msmanoj: I agree to that as well @imransl . It would be good to move the code related to bookmarks into `BrowserBookmarksStorageService`. 

Regarding creating a new `trackPanelStorageService`, I don't think we need it at this point or in this PR.

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
```diff
@@ -46,7 +45,7 @@ export type PreviouslyViewedObjects = {
 };
 
 export type BrowserBookmarksStateForGenome = {
-  bookmarks: [];
+  bookmarks: PreviouslyViewedObject[];
```
azangru, line 48: Well, technically, we don't know anything about the shape of bookmarks (the original intention was that they would be the bookmarks that the user creates manually). We aren't using this array at all in our code; we are just keeping this as a placeholder and a reminder of what's likely to come. So I would advise against this change.

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
```diff
@@ -131,10 +133,12 @@ export const updatePreviouslyViewedObjectsAndSave =
       ...state.browser.browserBookmarks[activeGenomeId],
       previouslyViewedObjects: previouslyViewedObjectsSlice
     };
+    const persistentTrackProperties = pickPersistentBrowserBookmarksProperties(
```
azangru, line 136: variable name

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
imransl: ✅

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
imransl: ✅

azangru:
Have you checked storage and retrieval of previously viewed objects? Looks like it's broken:

https://user-images.githubusercontent.com/6834224/165339609-796cb5f7-7e13-41fe-a07f-628dbf0ef8e5.mp4

src/content/app/genome-browser/services/browserStorageService.ts
imransl: We decided do this later in a separate PR

imransl:
> Have you checked storage and retrieval of previously viewed objects? Looks like it's broken:

Yes but not with this scenario. Thanks for spotting the issue. Will look into this now.

src/root/useRestoredReduxState.ts
```diff
@@ -23,6 +23,7 @@ import { restoreUI as restoreSpeciesPageUI } from 'src/content/app/species/state
 import { loadInitialState as loadEntityViewerGeneralState } from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSlice';
 import { restoreGeneViewTranscripts } from 'src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice';
 import { loadBrowserGeneralState } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
+import { loadPreviouslyViewedObjects } from 'src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice';
```
azangru, line 26: Probably want to alias this import in this file. This hook deals with multiple apps at once.

```
import { loadPreviouslyViewedObjects  as loadPreviouslyViewedGenomeBrowserObjects } from ....
```
?

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
```diff
@@ -113,85 +93,59 @@ export const updatePreviouslyViewedObjectsAndSave =
...
-    browserBookmarksStorageService.updateBookmarks({
-      [activeGenomeId]: persistentBrowserBookmarksProperties
-    });
-
     dispatch(
-      updateBrowserBookmarksForGenome({
+      bookmarksSlice.actions.updatePreviouslyViewedObjects({
         activeGenomeId,
-        data
+        updatedObjects
```
azangru, line 115: Not `previouslyViewedObjectsSlice`?

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
```diff
@@ -81,13 +62,12 @@ export const updatePreviouslyViewedObjectsAndSave =
...
     const activeFocusObject = getBrowserActiveFocusObject(state);
+
     if (!activeGenomeId || !activeFocusObject) {
       return;
     }
 
-    const previouslyViewedObjects = [
-      ...getActiveGenomePreviouslyViewedObjects(state)
-    ];
+    const previouslyViewedObjects = [...getPreviouslyViewedObjects(state)];
```
azangru, line 70: This spread is probably useless.

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
imransl: Argh yes that should be `previouslyViewedObjectsSlice`. Will fix it now.

src/content/app/genome-browser/state/browser-bookmarks/browserBookmarksSlice.ts
imransl: ✅

src/root/useRestoredReduxState.ts
imransl: ✅

azangru (approved):

tests/fixtures/browser.ts
```diff
@@ -245,22 +247,29 @@ export const createMockBrowserState = () => {
...
+        [fakeGenomeId]: {
+          selectedTrackPanelTab: TrackSet.GENOMIC,
+          isTrackPanelOpened: true,
+          highlightedTrackId: '',
+          collapsedTrackIds: []
+        }
+      },
+      browserSidebarModal: {
+        [fakeGenomeId]: {
+          browserSidebarModalView: BrowserSidebarModalView.SEARCH
```
msmanoj, line 259: Do we need the sidebar search to be shown by default for testing?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.test.tsx
```diff
@@ -86,11 +86,14 @@ const mockState = {
         [genomeId]: geneObjectId
       }
     },
-    trackPanel: {
+    browserSidebarModal: {
```
msmanoj, line 89: Not in this PR but it might be worth using [createMockBrowserState](https://github.com/Ensembl/ensembl-client/pull/726/files#diff-7d407488d78263a491106903c1e06ba1dac48dfcf4cdf13b9ec3f22817b01851R208) instead of rewriting the `mockState` here.

tests/fixtures/browser.ts
imransl: Should this be `null` instead then?

src/content/app/genome-browser/services/browser-bookmarks/browserBookmarksStorageService.ts
```diff
@@ -15,29 +15,31 @@
...
+export class BrowserBookmarksStorageService {
   private storageService: StorageServiceInterface;
 
   public constructor(storageService: StorageServiceInterface) {
     this.storageService = storageService;
   }
 
-  public getBookmarks(): { [genomeId: string]: PreviouslyViewedObject[] } {
+  public getBookmarks(): {
+    [genomeId: string]: [];
```
msmanoj, line 36: Why do we need to change `PreviouslyViewedObject` to []?

imransl: We don't know the shape of the object for bookmarks so unfortunately can't use a non-generic type for it

azangru: I am confused by our use of the word "bookmark".

We said, a long time ago, that we expected to have two types of bookmarks:
- the automatically generated "previously viewed objects"
- the bookmarks that a user can create manually.

So far, we have only been dealing with previously viewed objects. But the method, I see, has the word "bookmarks" in its name. Is this going to be confusing?

The `{ [genomeId: string]: []; }` is an incorrect type; it sets the value to the type `never[]`.

imransl: This is because this method is for the second type and it was explicitly called as bookmarks then. Also we had this method inside the `TrackPanelStorageService`, so less confusion than with having it inside `BrowserBookmarksStorageService`. Or at least I think that's close to what the thinking was back then.

Now that it is confusing, should we change this to `getUserBookmarks`? Open to suggestions for better names than this. As for the type should it then be `{ [genomeId: string]: unknown[]; }`?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.test.tsx
imransl: I was a bit overenthusiastic and did it in this PR itself 😁

tests/fixtures/browser.ts
imransl: ✅

src/content/app/genome-browser/services/browser-bookmarks/browserBookmarksStorageService.ts
imransl: Added comments above the two functions and renamed them ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/BookmarksModal.test.tsx
imransl: ✅

azangru (approved):

msmanoj (approved):

PR #732, opened 2022-04-20
Change browser location in the url from inside BrowserImage?

azangru:
## Description
There is currently a regression on the genome browser page: the genome browser location in the url doesn't get updated. Reading through the code, I am not sure when this happened; or how our code even worked properly — I couldn't find code path that would update the location in the url based on the "target position" of the genome browser.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1579

## Deployment URL(s)
http://fix-genome-browser-location.review.ensembl.org/

## Views affected
Genome browser page


msmanoj:
Changing focus object from search doesn't seem to work as expected. 
For example, changing from `BRCA2` continues to show `BRCA2` in the summary strip.

imransl (approved):
Summary strip and sidebar are getting updated after the fix

msmanoj (approved):

PR #733, opened 2022-04-21
Add spinner and load error indicator to DrawerSequenceView

azangru:
## Description
Add an error message and a loading spinner to the drawer sequence view.

## Deployment URL(s)
http://gb-sequence-view-errors.review.ensembl.org


msmanoj:
Do we not need to show the spinner for at least a second?

msmanoj:
The failed state doesn't get cleaned up when we change the sequence type:
Kapture 2022-04-21 at 10 38 27

imransl:
I don't see the loading failure message locally. Can only see it in the review branch site. The spinner is visible in both places though.

imransl (approved):

msmanoj (approved):

PR #737, opened 2022-04-26
Scroll BlastForm to top when changing between views

msmanoj:
## Description
Scrolling the Blast form container to top when we switch between species and sequence view

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1549

## Deployment URL(s)
http:///blast-scroll-to-top.review.ensembl.org

## Views affected
BLAST -> species & sequences view on small screens

### DIFF
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
@@ -39,6 +39,7 @@ import type { SequenceType } from 'src/content/app/tools/blast/types/blastSettin
 
 import styles from './BlastInputSequences.scss';
 import sequenceBoxStyles from './BlastInputSequence.scss';
+
 export type Props = {
   compact: boolean;
 };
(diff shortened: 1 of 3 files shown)


azangru:
If you put the scrolling logic in `useEffect`, which reacts to the change in the step (sequences or species) in [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/tools/blast/views/blast-form/BlastForm.tsx#L84), wouldn't it be both more idiomatic for React, and simpler to implement. You wouldn't need a document.querySelector; a useRef could give you exactly what you want there.

msmanoj:
> If you put the scrolling logic in `useEffect`, which reacts to the change in the step (sequences or species) in [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/tools/blast/views/blast-form/BlastForm.tsx#L84), wouldn't it be both more idiomatic for React, and simpler to implement. You wouldn't need a document.querySelector; a useRef could give you exactly what you want there; the main container will be within the immediate reach of your code.

That's done. A lot better now. Thanks

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
```diff
@@ -39,7 +39,6 @@ import type { SequenceType } from 'src/content/app/tools/blast/types/blastSettin
 
 import styles from './BlastInputSequences.scss';
 import sequenceBoxStyles from './BlastInputSequence.scss';
-
```
azangru, line 42: could you undo this please? `git checkout dev src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx`?

msmanoj: That's done

azangru (approved):

PR #738, opened 2022-04-27
blast parameter layout for different screen size

ens-ma7:
## Description
Styling the blast settings and parameters (the dropdowns and the checkbox), also renaming job list button as well as some labels.

I had to reorganise some of the elements to fit the new layout and mostly css changes.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1498

## Deployment URL(s)
http://blastparameters-layout.review.ensembl.org

## Views affected
Blast

### DIFF
--- a/src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar.scss
+++ b/src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar.scss
@@ -3,7 +3,7 @@
 .topBar {
   min-height: 74px;
   background-color: $light-grey;
-  padding-left: 36px; // TODO: does this value repeat across Genome Browser, Entity Viewer, and Tools?
+  padding: 24px 10px 20px 30px; // TODO: does this value repeat across Genome Browser, Entity Viewer, and Tools?
   box-shadow: 0 3px 5px $global-box-shadow;
   margin-bottom: 21px;
 }
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
@@ -1,27 +1,50 @@
 @import 'src/styles/common';
+.topLevelContainer {
+  display: grid;
+  grid-template-columns: minmax(100px, 1660px) 140px;
+}
 
 .topLevel {
   display: flex;
-  gap: 1rem;
+  flex-wrap: wrap;
+  row-gap: 20px;
+}
+
+.header {
+  font-weight: $bold;
+  margin-right: 45px;
+  font-size: 17px;
+}
+
+.subHeader {
+  font-weight: $bold;
+  margin-right: 55px;
+  margin-top: 5px;
+}
+
+.previousJobs {
+  height: 30px;
+  width: 120px;
+  margin-left: 20px;
 }
 
 .mainSettings {
-  display: flex;
-  gap: 1rem;
-}
-
-.mainSettings > * {
-  flex: 0 0 auto;
-}
-
-.mainSettings select {
-  flex: 0 0 auto;
+  margin-left: 20px;
 }
 
 .select label {
   display: grid;
   grid-template-columns: repeat(2, auto);
   column-gap: 10px;
+}
+
+.parametersColumn .select label {
+  grid-template-columns: 140px 70px;
+  text-align: right;
+}
+
+.parametersColumn .matrixSetting .select label {
+  grid-template-columns: 140px 110px;
 }
 
 .select label span {
@@ -32,8 +55,12 @@
 .bottomLevel {
   display: flex;
   flex-wrap: wrap;
-  gap: 1rem;
-  padding-top: 0.6rem;
+  gap: 15px;
+  padding-top: 20px;
+  margin-left: 104px;
+  @media (max-width: 1610px) {
+    margin-left: -10px;
+  }
 }
 
 .bottomLevel > * {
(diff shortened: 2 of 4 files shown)


azangru:
A curious error occurring specifically if you expand the parameters after they have been moved to the next row:

https://user-images.githubusercontent.com/6834224/165710490-d6cfe2be-4e47-4ac2-8168-9f5cc264f49c.mp4

azangru:
Also, dropdowns got broken:

https://user-images.githubusercontent.com/6834224/165711283-64cde1a3-75c6-4d4b-9097-e682175f11d3.mp4

ens-ma7:
> A curious error occurring specifically if you expand the parameters ~after they have been moved to the next row~ sometimes (I haven't quite figured out under which conditions):
> 
>  Blast.parameters.error.mp4

@azangru i couldn't replicate it on my localhost, to be honest i have played with this a lot on my localhost and it didnt break for me at all, i am just wondering if it is something with the review URL. can you try on your localhost and see if you can replicate the issue.

azangru:
> @azangru i couldn't replicate it on my localhost, to be honest i have played with this a lot on my localhost and it didnt break for me at all, i am just wondering if it is something with the review URL. can you try on your localhost and see if you can replicate the issue.

Ok; I am starting to suspect that it is again a network problem with our deployment — some requests go to a container that serves an outdated config.

azangru:
Didn't you want to fix the "Filter" label?

image

ens-ma7:
> Didn't you want to fix the "Filter" label?
> 
> image

I did fix it and it was updated by Andres as well (i can see my changes on localhost).... i think its like you say something wrong with the config file not being deployed properly.

ens-ma7:
i managed to get the file when the error was triggered and it is an outdated file. so looks like some container has the outdated file. i will check with Andres

azangru:
Several things that came up during a daily scrum:

1) The font-size of the "Blast" title is currently 17px. It is 17px in the dev spec; but Andrea says it was 15px in her XD. She finally decided that she wants it to be in between 15px and 17px, that is 16px 😄 
2) Also, let's use an h1 rather than a div tag for this title
3) The "Previous jobs" button should be padded from the right on smaller screens. See [this XD](https://xd.adobe.com/view/4bd9ad02-5287-4868-9cfe-1bd59f362519-1e8a/screen/f44788d5-59e4-4a9a-891c-af146ddf9abf), for example.
4) The text of the "Previous jobs" button should be changed to "Jobs list" to match [this XD](https://xd.adobe.com/view/f523ac5e-275c-4835-9d91-33910bed4ef3-a070/screen/f891b13a-5150-4ff5-b013-ff8aa1d94a0d/?fullscreen).

azangru:
Notice that because of the header 1, which has its own margins, the word Blast got shifted up and is no longer on the same basline as "Run a job".

image

In any case, I think for the class `topLevel`, we should start with `flex-align: center`, and then adjust the margins of the elements that don't look quite right when centered.

Example:

image

`topLevel` class — `align-items: center`; `.header` class – `margin: 0 45px 3px 0`.

Another thing to consider is maybe to use the `column-gap` property on the `topLevel` class instead of the `margin-left` property on the `mainSettings` class.

src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar.scss
```diff
@@ -3,7 +3,7 @@
 .topBar {
   min-height: 74px;
   background-color: $light-grey;
-  padding-left: 36px; // TODO: does this value repeat across Genome Browser, Entity Viewer, and Tools?
+  padding: 24px 10px 20px 30px; // TODO: does this value repeat across Genome Browser, Entity Viewer, and Tools?
```
imransl, line 6: Is this still a TODO?

imransl:
@azangru @ens-ma7 Will this PR's styling affect #740? The `Jobs List` button is not going to be in `BlastSettings` component once that PR is merged. So do we need any styling related to it?

imransl:
@azangru @ens-ma7 Will this PR's styling affect #740? The `Jobs List` button is not going to be in `BlastSettings` component once that PR is merged. So do we need any styling related to it?

src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar.scss
ens-ma7: I didn't add this comment so not sure but from what i can see this is now specific for blast and probably for tools. GB and entity viewer has different padding

azangru:
> Will this PR's styling affect https://github.com/Ensembl/ensembl-client/pull/740?

I don't believe it will affect it directly; but we will eventually use the `ButtonLink` component from 740 in the `BrowserSettings` component to display the "Jobs list" text. I expect #738 will get merged first, and then #740 will use the `ButtonLink` button in the `BrowserSettings` component.

src/content/app/tools/shared/components/tools-top-bar/ToolsTopBar.scss
azangru: It is; we should figure out whether the distances in this component are the same across all apps. May be worth creating a ticket out of this.

azangru (approved):
Looks good. One remaining nitpick is maybe to change the class name `.checkboxLeftMargin ` to, say, `.checkboxes` or `.checkboxesColumn`; but that's not significant.

imransl (approved):

PR #739, opened 2022-04-28
Use updated endpoint for blast form submission

azangru:
## Description
A BLAST form submission will now create multiple jobs, each for every sequence:genome_id pair. This PR updates the client to use the new shape of payloads in the request and the response.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1554

## Deployment URL(s)
http://add-multiple-jobs-submission.review.ensembl.org/blast

### DIFF
--- a/src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
+++ b/src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
@@ -37,11 +37,13 @@ export type JobStatus =
 export type BlastSubmission = {
   submittedData: {
     species: Species[];
-    sequences: string[]; // Or perhaps parsed sequences?
+    sequences: { id: number; value: string }[]; // TODO: consider whether to have strings or parsed sequences
     parameters: Partial<Record<BlastParameterName, string>>;
   };
   results: Array<{
     jobId: string;
+    sequenceId: number;
+    genomeId: string;
     status: JobStatus;
     seen: boolean;
     data: null; // TODO: add data type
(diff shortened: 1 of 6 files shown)


imransl:
I can't submit a job since the `Run` button is disabled. Tried playing around with the options but with no luck.

azangru:
@imransl you can't just blast a sequence against noting. You need to select a species.

imransl:
> @imransl you can't just blast a sequence against nothing. You need to select a species.

Yes I hadn't selected one indeed! My bad. It worked alright after selecting a species.

Anyway there is an issue with the `HSPs per hit` select input as can be seen in the video. Is this something already looked into in a different PR?

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
```diff
@@ -37,11 +37,13 @@ export type JobStatus =
 export type BlastSubmission = {
   submittedData: {
     species: Species[];
-    sequences: string[]; // Or perhaps parsed sequences?
+    sequences: { id: number; value: string }[]; // TODO: consider whether to have strings or parsed sequences
```
msmanoj, line 40: What do you mean by `parsed sequences`?

azangru: https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/tools/blast/types/parsedInputSequence.ts

msmanoj: I think `string[]` is fine unless we convert it to `parsed sequence` later on demand, multiple times.

msmanoj: As discussed, storing it as parsed sequence will be better but it can be done in a different PR.
I guess we can retain this todo comment just so we don't miss it.

msmanoj (approved):

imransl (approved):

PR #740, opened 2022-05-02
Add a page for unviewed BLAST submissions

azangru:
## Description
When a BLAST form is submitted, open a page for unviewed BLAST submissions.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1560

## Deployment URL(s)
http://add-unviewed-results-page.review.ensembl.org


src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
```diff
@@ -30,6 +30,8 @@ import { getSelectedSpeciesIds } from 'src/content/app/tools/blast/state/blast-f
 
 import { toFasta } from 'src/shared/helpers/formatters/fastaFormatter';
 
+import { PrimaryButton } from 'src/shared/components/button/Button';
```
imransl, line 33: Group this import with `LoadingButton` above?

src/content/app/tools/blast/components/blast-views-navigation/BlastViewsNavigation.scss
```diff
@@ -0,0 +1,34 @@
+.blastViewsNavigation {
+  display: grid;
+  grid-template-columns: [left] auto [right] 1fr;
+  height: 74px; // this isn't ideal; but 100% won't work because the parent container only has a min-height
```
imransl, line 4: Am confused with the `100 % won't work` part? Does it mean this will never work or won't work sometimes?

src/content/app/tools/blast/views/blast-unviewed-submissions/BlastUnviewedSubmissions.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const BlastUnviewedSubmission = () => {
+  return (
+    <div>
+      <BlastAppBar view="unviewed-submissions" />
+      <ToolsTopBar>
+        <BlastViewsNavigation />
+      </ToolsTopBar>
+      <Main />
+    </div>
```
imransl, line 31: Can't we use the `main` element here instead of `div`?

src/content/app/tools/blast/components/blast-views-navigation/BlastViewsNavigation.scss
azangru: Compare [this jsfiddle](https://jsfiddle.net/9fdnwvmh) with [this one](https://jsfiddle.net/9fdnwvmh/1/). In the first fiddle, the parent container (blue) has a fixed `height` property; and the child component (pink), having a height of 100%, is as high as the parent. In the second fiddle, the parent only has a `min-height` property; and the child's "100%" has nothing to reference; so the child component remains small.

It's the same as we are seeing here.

One way around this would be to change the parent component's (in our case, `TopBar`) display property to flex, so that its child stretches vertically. Perhaps this is what I should do; but then, maybe it's for a different PR.

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
azangru: I am confused. This file doesn't import `LoadingButton`.

src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.tsx
```diff
@@ -50,13 +62,27 @@ const ToolsAppBar = () => {
...
-  const speciesTabs = speciesList.map((species, index) => (
+  const enabledSpecies = speciesList.map((species, index) => (
     <SelectedSpecies
       key={index}
       species={species}
       onClick={() => speciesLozengeClick(species)}
     />
   ));
+
+  // TODO: this is not a proper way of expressing disabled species lozenges in this context
```
ens-ma7, line 73: Lets remove the todo and create a jira ticket for this...it wont get done if it doesnt go in a jira.

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
imransl: I made the comment before the component was removed. This no longer applies then 😁

src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.tsx
azangru: Removed the TODO

src/content/app/tools/blast/views/blast-unviewed-submissions/BlastUnviewedSubmissions.tsx
azangru: Changed `div` to `main`

ens-ma7 (approved):

ens-ma7:
@azangru i have approved the code changes but one thing that i have noticed and not sure if it is part of this PR or should be a separate ticket is:
1) Go to http://add-unviewed-results-page.review.ensembl.org/blast and fill in sequence and select species
2) click Run and you get send to the unviewed jobs which is fine
3) Now when you click new job button, it sends you back to the blast form but the form is still filled with the last job that i ran, should i not clear itself?

PR #742, opened 2022-05-03
Add the ButtonLink component

azangru:
## Description
Adding a component that looks like a button, but functions as a link.

To be used in a design like this:

image

NOTE: the component can be inspected in Storybook.

### DIFF
--- a/src/shared/components/button-link/ButtonLink.scss
+++ b/src/shared/components/button-link/ButtonLink.scss
@@ -17,3 +17,10 @@
   pointer-events: none;
   cursor: default;
 }
+
+.buttonLinkDisabled {
+  border: var(--button-link-disabled-border, 1px solid #{$grey});
+  background-color: var(--button-link-disabled-background-color, #{$grey});
+  pointer-events: none;
+  cursor: default;
+}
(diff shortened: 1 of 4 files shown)


msmanoj:
If a disabled button-link is currently active, it is shown as active instead disabled. I am assuming this is the expected behaviour.

msmanoj:
Clicking on a disabled button works as expected but I am able to tab into a disabled button and press `enter` to make it active. I guess we should prevent it?

azangru:
> Clicking on a disabled button works as expected but I am able to tab into a disabled button and press enter to make it active. I guess we should prevent it?

Great point. I changed the component to use a `span` instead of `NavLink` for disabled buttons.

> If a disabled button-link is currently active, it is shown as active instead disabled. I am assuming this is the expected behaviour.

Not sure I follow... Is this something you are still seeing in the latest commit? If so, could you show what you are seeing?

msmanoj:
> > If a disabled button-link is currently active, it is shown as active instead disabled. I am assuming this is the expected behaviour.
> 
> Not sure I follow... Is this something you are still seeing in the latest commit? If so, could you show what you are seeing?

If you try changing `/foo` to `/bar` [here](https://github.com/Ensembl/ensembl-client/pull/742/files#diff-332c125fe79c1b2f8498e6b19a5099b8ef5cb818756ee224b784c70960c04378R29), you'll notice that the disabled button will be displayed like an active button in the storybook entry. I was just wondering if it is the expected behaviour.

msmanoj (approved):

PR #746, opened 2022-05-10
Update BLAST param guessing logic

msmanoj:
## Description
- Updated the parameter guessing / overriding logic to match the flow specified in the XD attached to the JIRA
- Added some tests to cover the scenarios from the XD

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1565

## Deployment URL(s)
http://blast-param-guessing.review.ensembl.org

## Views affected
BLAST settings & sequence input

### DIFF
--- a/src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
+++ b/src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
@@ -110,58 +110,26 @@ const blastFormSlice = createSlice({
     switchToSpeciesStep(state) {
       state.step = 'species';
     },
-    setSequenceType(
+    updateSettings(
       state,
       action: PayloadAction<{
-        sequenceType: SequenceType;
-        isAutomatic: boolean;
-      }>
-    ) {
-      const { sequenceType, isAutomatic } = action.payload;
-
-      state.settings.sequenceType = sequenceType;
-      state.settings.sequenceSelectionMode = isAutomatic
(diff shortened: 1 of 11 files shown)


src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
```diff
@@ -54,6 +58,8 @@ const initialBlastFormSettings: BlastFormSettings = {
   jobName: '',
   sequenceType: 'dna',
   sequenceSelectionMode: 'automatic',
+  databaseSelectionMode: 'automatic',
+  programSelectionMode: 'automatic',
```
azangru, line 62: I don't understand the purpose of this flag. The program is always a computed variable derived from a combination of input sequence type and database sequence type. Why does this flag need to exist?

azangru:
Mmm, I think I preferred it when the logic for updating different bits of redux state was focused in the reducer. I understand your point about a single function in the reducer responsible for changing different pieces of the redux state; but I would say it's perfectly normal when a change in one parameter results in simultaneous updates to several pieces of the state. I would even say that it may be easier to reason about than with a chain of updates that you are now having in the hook, where an onSomethingChange function triggers an onSomethingElseChange function, and so on.

azangru:
Superseded by https://github.com/Ensembl/ensembl-client/pull/748

PR #747, opened 2022-05-10
Track label state

jyothishnt:
## Description

Maintain states of track config selections on browser refresh which means we need to store track config selections on the browser storage.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-924
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-923
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1567

## Deployment URL(s)
http://track-label-state.review.ensembl.org/

## Views affected
Genome browser

HOW TO READ THE PR

New components: BrowserTrackConfig component is now split into 3 separate components for Gene, Transcript and Regular tracks (just like track panel items)

All the methods in BrowserTrackConfig has been moved to a hook useBrowserTrackConfig.ts

trackConfigSlice is where you can see the updated state slice

browser-track-config-storage-service is the service to store the state in local storage

Track config state now has an updated structure as below

Screenshot 2022-05-20 at 12 33 13

The same is stored in the local storage as below

Screenshot 2022-05-20 at 12 33 46


src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -21,43 +21,73 @@ import {
...
+export const getTrackType = (trackId: string) => {
+  if (trackId.startsWith('gene')) {
+    return TrackType.GENE;
+  } else {
+    return TrackType.REGULAR;
+  }
+};
+
+export type GeneTrackConfig = {
+  showMoreTranscripts: boolean;
```
azangru, line 53: This is probably better expressed as `showSeveralTranscripts`. See https://genomes-ebi.slack.com/archives/C01QD9623KK/p1652706953368469

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -21,43 +21,73 @@ import {
...
-    allTrackLabelsOn: boolean;
+    isSelected?: boolean;
   };
-  browserCogList: number;
-  browserCogTrackList: CogList;
+  browserCogList: CogList;
   selectedCog: string | null;
-  trackConfigNames: { [key: string]: boolean };
-  trackConfigLabel: { [key: string]: boolean };
+  tracks: TrackInfo;
```
azangru, line 77: The name `TrackInfo` reads as if it means is some info about some track; but the way you declare it, it looks like a map of track id to track config. We need a better name for this. `TrackConfigs`? `TrackConfigsMap`?

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -21,43 +21,73 @@ import {
...
+  trackType: TrackType.REGULAR;
+};
+
+export type TrackConfig = GeneTrackConfig | RegularTrackConfig;
+
+export type TrackInfo = {
+  [trackId: string]: TrackConfig;
+};
+
+export type TrackConfigStateForObject = Readonly<{
```
azangru, line 71: I don't understand this name at all. What is it intended to mean? Does it refer only to tracks related to an "ens object" (in which case we are trying to call them entities, I believe)? The word "state" doesn't make the intent any clearer.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: I didn't like "several" tbh. "more" better sync with what we show on the label I feel. Lets see what others say about it. :-)

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: Ok, will take `TrackConfigs`

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: This is not the ensobject, it is the configuration for a single track. any suggestions?

imransl: Shouldn't this just be `TrackConfig`?

jyothishnt: Final types after discussion is below
```
type GenomeTrackConfigs = {
  [genomeId]: TrackConfigsPerGenome
}

type TrackConfigsPerGenome = {
  {
    applyToAllTracks: boolean
    tracks: TrackConfigs
  }
}

type TrackConfigs = {
  [trackId]: TrackConfig
}

type TrackConfig = {
  showTrackName: boolean
}```

File: src/content/app/genome-browser/state/track-config/trackConfigSlice.ts [hunk 4]
```diff
@@ -21,43 +21,73 @@ import {
...
+  REGULAR = 'regular'
+}
+
+export const getTrackType = (trackId: string) => {
+  if (trackId.startsWith('gene')) {
+    return TrackType.GENE;
+  } else {
+    return TrackType.REGULAR;
+  }
+};
```
azangru, line 50: While this will work for now, this is something we should discuss with the backend:
- How will we know on the frontend which settings panel to show for which track? Will the tracks we receive from the track api need to have a type field?
- What's the proper way for us to know the id of the focus track? It doesn't come from the track api. Is this something we just hard-code on the client? Will the id of the focus track change when a variation track is in focus? All this is something to discuss with Dan; possibly during the genome browser meeting tomorrow.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
   browserStorageService.saveTrackStates(trackStates);
 };
 
+export const getTracksStateInfo = (
```
azangru, line 101: Confused about the name of this function. "tracksStateInfo" doesn't correspond to what we talked about yesterday

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+      const trackId = track.track_id.replace('track:', '');
+      const trackType = getTrackType(trackId);
+      if (trackType === TrackType.GENE) {
+        tracksInfo[trackId] = {
+          showSeveralTranscripts: false,
+          showTranscriptIds: false,
+          showTrackName: false,
+          showFeatureLabel: true,
+          trackType
+        };
```
azangru, line 126: here and on line 128 — you probably should create functions that return default track configs:

```
const getDefaultGeneTrackConfig = (): GeneTrackConfig => ({ ... });
const getDefaultRegularTrackConfig = (): RegularTrackConfig => ({ ... })

```

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
   browserStorageService.saveTrackStates(trackStates);
 };
 
+export const getTracksStateInfo = (
+  genomeId: string,
```
azangru, line 102: What does the genomeId do in this function? Can't find where it's used. And if it isn't, how did eslint allow you to commit this?

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+  trackCategories: GenomeTrackCategory[]
+) => {
+  const tracksInfo: TrackConfigs = {};
```
azangru, line 105: First, variable name.

And second, it seems that what you are doing in this function is just creating a map of default track configs. Why don't you return it as `{ [trackName]: TrackConfig }`; why do you create a `{ tracks: { [trackName]: TrackConfig } }` object before returning it? Let the function that calls this one do this job.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+        ...getPersistentTrackConfigsPerGenome(genomeId)
+      }
+    : defaultTrackConfigsPerGenome;
+};
+
+export const getPersistentTrackConfigsPerGenome = (
+  genomeId: string
+): Partial<TrackConfigsPerGenome> => {
+  const trackConfigState =
+    browserTrackConfigStorageService.getTrackConfigState();
```
azangru, line 155: `getTrackConfigState` -> `getTrackConfigs`?

tests/fixtures/browser.ts
```diff
@@ -43,36 +43,27 @@ import { LoadingState } from 'src/shared/types/loading-state';
...
-  'track:gene-pc-rev': true,
-  'track:variant': true
+export const createTracksInfo = () => ({
+  'gene-focus': {
+    showMoreTranscripts: false,
+    showTranscriptIds: false,
+    showTrackName: true,
+    showFeatureLabel: true,
+    trackType: 'gene'
+  },
```
azangru, line 58: I mentioned creation of functions that return default configs for different track types in a comment to trackConfigSlice. You may want to reuse them here.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+      }>
+    ) {
+      const { genomeId, trackCategories } = action.payload;
+      if (!state[genomeId]) {
+        state[genomeId] = getTrackConfigsPerGenome(genomeId, trackCategories);
+      }
+      browserTrackConfigStorageService.setTrackConfigState({
+        genomeId: genomeId,
+        fragment: state[genomeId]
+      });
```
azangru, line 180: Here and throughout the rest of the slice — I am very conflicted about writing data to browser storage from inside the reducers. Reducers are supposed to be pure functions, while interacting with the browser storage is a side effect. Not quite sure yet what to do with this; but it's worth thinking about.

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public setTrackConfigState(params: {
+    genomeId: string;
+    fragment: Partial<TrackConfigsPerGenome>;
+  }) {
+    const { genomeId, fragment } = params;
+    this.storageService.update(StorageKeys.TRACK_CONFIG_STATES, {
+      [genomeId]: fragment
+    });
+  }
```
azangru, line 48: - when a genome gets deleted, we will also need to delete the values stored against it

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
```diff
@@ -0,0 +1,51 @@
...
+import storageService, {
+  StorageServiceInterface
+} from 'src/services/storage-service';
+import {
+  GenomeTrackConfigs,
+  TrackConfigsPerGenome
+} from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+
+export enum StorageKeys {
+  TRACK_CONFIG_STATES = 'browser.trackConfigStates'
```
azangru, line 26: Throughout this file — the word "state" is redundant

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -15,163 +15,42 @@
  */
 
 import React from 'react';
-import { useDispatch, useSelector } from 'react-redux';
-import { OutgoingActionType } from '@ensembl/ensembl-genome-browser';
+import { useSelector } from 'react-redux';
+
+import RadioGroup, {
+  RadioOptions
+} from 'src/shared/components/radio-group/RadioGroup';
```
azangru, line 22: Just curious which one is better: to keep this `this track/all tracks` toggle in this component, or to copy it across the individual track config panel components?

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    analyticsTracking.trackEvent({
+      category: 'track_settings',
+      label: selectedCog,
+      action: 'feature_label_' + (isTrackLabelShown ? 'on' : 'off')
+    });
+  };
+
+  const handleRadioChange = (value: OptionValue) => {
```
azangru, line 136: This name is almost meaningless. The reader of the code will wonder, what radio. The radio is in a different file now, and ultimately, is just an implementation detail.

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  const dispatch = useDispatch();
+
+  const { toggleTrackName, toggleTrackLabel } = useGenomeBrowser();
+
+  const updateTrackName = (isTrackNameShown: boolean) => {
+    if (!activeGenomeId) {
+      return;
+    }
+    if (shouldApplyToAllRef.current) {
```
azangru, line 64: Is this not the decision that should be made in the reducer? The reducer should know whether the "shouldApplyToAll" is selected, and would apply the new setting to all relevant tracks, I would have thought.

By the way, is the "apply to all" state shared between all configs? I.e. if I switch the radio button to "apply to all" in one track config panel, and then open another track config panel, does the radio button there show as selected?

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,170 @@
...
+        trackId: selectedCog,
+        shouldShowTrackName: isTrackNameShown
+      });
+    }
+
+    analyticsTracking.trackEvent({
+      category: 'track_settings',
+      label: selectedCog,
+      action: 'track_name_' + (isTrackNameShown ? 'on' : 'off')
+    });
```
azangru, line 93: Wondering if we don't want these pesky splrinklings of analytics code to be moved into its own file (either hook or not), as we did in EntityViewer. We encapsulated analytics functions in a dedicated hook there, which was also pulling genome id from redux and adding it to the analytics event.

src/content/app/genome-browser/hooks/useBrowserRouting.ts
```diff
@@ -171,6 +173,17 @@ const useBrowserRouting = () => {
...
+  useEffect(() => {
+    if (genomeId && trackCategoriesForGenome) {
+      dispatch(
+        setInitialTrackConfigsPerGenome({
+          genomeId,
+          trackCategories: trackCategoriesForGenome
+        })
+      );
+    }
+  }, [genomeId, trackCategoriesForGenome]);
```
azangru, line 185: We shouldn't be doing this here. It doesn't have anything to do with the routing. If we want to restore the data saved to the browser storage, we should probably run this in the Root component (see `useRestoredReduxState` hook)

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+    showSeveralTranscripts: false,
+    showTranscriptIds: false,
+    showTrackName: false,
+    showFeatureLabel: true,
+    trackType: TrackType.GENE
+  };
+
+  trackCategories.forEach((category) => {
+    category.track_list.forEach((track) => {
+      const trackId = track.track_id.replace('track:', '');
```
azangru, line 117: Why are we doing this? We do not control track ids that come from the backend. I see you are modifying the track ids and use the modified ids when saving the state to browser storage. Why?

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
```diff
@@ -0,0 +1,70 @@
+/**
```
azangru, line 1: Do we not want to call them config panels (or settings panels) instead of just configs?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -151,6 +152,78 @@ const useGenomeBrowser = () => {
     }
   };
 
+  const restoreTrackConfigStates = () => {
+    if (!activeGenomeId || !genomeBrowser) {
```
azangru, line 156: No need to use the question mark after `genomeBrowser` in the code below then.

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -151,6 +152,78 @@ const useGenomeBrowser = () => {
...
+      return;
+    }
+
+    type TrackConfigStates = {
+      on: string[];
+      off: string[];
+    };
```
azangru, line 163: we don't declare types inside functions

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -151,6 +152,78 @@ const useGenomeBrowser = () => {
     }
   };
 
+  const restoreTrackConfigStates = () => {
+    if (!activeGenomeId || !genomeBrowser) {
+      return;
+    }
+
+    type TrackConfigStates = {
```
azangru, line 160: What is this name telling us?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -30,22 +30,22 @@ import { BROWSER_CONTAINER_ID } from 'src/content/app/genome-browser/constants/b
...
 import { TrackStates } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
 import { Status } from 'src/shared/types/status';
-import { ChrLocation } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
 
 const useGenomeBrowser = () => {
   const activeFocusObjectId = useSelector(getBrowserActiveFocusObjectId);
   const activeGenomeId = useSelector(getBrowserActiveGenomeId);
-  const { allTrackLabelsOn, allTrackNamesOn } =
-    useSelector(getApplyToAllConfig);
+  const allTracksInfo = useSelector(getAllTrackConfigs);
```
azangru, line 48: What's this info?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -151,6 +152,78 @@ const useGenomeBrowser = () => {
     }
   };
 
+  const restoreTrackConfigStates = () => {
```
azangru, line 155: Doesn't this function, with all the complexities inside it, belong in the `useBrowserTrackConfig` hook?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -178,27 +251,68 @@ const useGenomeBrowser = () => {
     genomeBrowser.send(action);
   };
 
+  const toggleTrackName = (params: {
```
azangru, line 254: Don't these toggle functions also belong in the useBrowserTrackConfig hook?

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -29,27 +29,40 @@ import BrowserCog from './BrowserCog';
...
 import {
   CogList,
-  updateCogTrackList,
+  updateCogList,
   updateSelectedCog
 } from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+import {
+  getBrowserActiveFocusObjectId,
+  getBrowserActiveGenomeId
+} from '../../state/browser-general/browserGeneralSelectors';
```
imransl, line 42: Change the import path to `src/content/app/genome-browser/state/browser-general/browserGeneralSelectors`?

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -29,27 +29,40 @@ import BrowserCog from './BrowserCog';
...
 import styles from './BrowserCogList.scss';
 
 export const BrowserCogList = () => {
   const browserCogList = useSelector(getBrowserCogList);
-  const browserCogTrackList = useSelector(getBrowserCogTrackList);
   const selectedCog = useSelector(getBrowserSelectedCog);
 
+  const genomeId = useSelector(getBrowserActiveGenomeId);
+  const genomeIdRef = useRef(genomeId);
+  const objectId = useSelector(getBrowserActiveFocusObjectId);
```
imransl, line 52: Group all selectors together?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -119,7 +119,8 @@ const useGenomeBrowser = () => {
     Object.values(mergedTrackStates).forEach((trackStates) => {
       Object.keys(trackStates).forEach((trackId) => {
         const track_id =
-          trackId === 'track:gene-feat' || trackId === 'track:transcript-feat-1'
+          trackId === 'track:gene-focus' ||
+          trackId === 'track:transcript-feat-1'
             ? 'focus'
             : trackId.replace('track:', '');
```
jyothishnt, line 125: chop this in genomeSlice.ts and remove this

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -73,34 +95,38 @@ export const BrowserCogList = () => {
...
-  });
+  }, [genomeBrowser, genomeId, objectId]);
 
-  const transformStyle = {
-    transform: 'translate(0,' + browserCogList + 'px)'
-  };
+  const cogs =
+    browserCogList &&
+    Object.entries(browserCogList).map(([name, pos]) => {
+      const posStyle = { top: pos + 'px' };
```
imransl, line 103: We can write this as a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) but this is also alrigt

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -73,34 +95,38 @@ export const BrowserCogList = () => {
...
+            cogActivated={selectedCog === name}
+            trackId={name}
+            updateSelectedCog={(trackId: string | null) =>
+              dispatch(
+                updateSelectedCog({
+                  genomeId: genomeIdRef.current as string,
+                  selectedCog: trackId
+                })
+              )
+            }
```
imransl, line 120: Why not write this function as a function variable above and specify here, like how we do elsewhere?

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.test.tsx
```diff
@@ -15,40 +15,81 @@
...
 import userEvent from '@testing-library/user-event';
 import { Provider } from 'react-redux';
-import thunk from 'redux-thunk';
 
-import { createMockBrowserState } from 'tests/fixtures/browser';
 import MockGenomeBrowser from 'tests/mocks/mockGenomeBrowser';
 
 import * as trackConfigActions from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+import * as browserGeneralActions from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
+import BrowserTrackConfig from './BrowserTrackConfig';
```
imransl, line 27: Separate this import?

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -15,163 +15,42 @@
...
 import {
-  getTrackConfigNames,
-  getTrackConfigLabel,
   getApplyToAllConfig,
-  getBrowserCogTrackList,
   getBrowserSelectedCog
-} from 'src/content/app/genome-browser/state/track-config/trackConfigSelectors';
-
-import analyticsTracking from 'src/services/analytics-service';
+} from '../../state/track-config/trackConfigSelectors';
```
imransl, line 34: Change the import path to `src/content/app/genome-browser/state/track-config/trackConfigSelectors`?

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -15,163 +15,42 @@
...
 
+const getTrackConfigComponent = (trackType: TrackType) => {
+  switch (trackType) {
+    case TrackType.GENE:
+      return <GeneTrackConfig />;
+    case TrackType.REGULAR:
+      return <RegularTrackConfig />;
+  }
+};
+
```
imransl, line 48: How about this instead?

```
const getTrackConfigComponent = (trackType: TrackType) => trackType === TrackType.GENE ? <GeneTrackConfig /> : <RegularTrackConfig />;
```

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -21,43 +21,73 @@ import {
...
+  trackType: TrackType.REGULAR;
+};
+
+export type TrackConfig = GeneTrackConfig | RegularTrackConfig;
+
+export type TrackConfigs = {
+  [trackId: string]: TrackConfig;
+};
+
+export type TrackConfigsPerGenome = Readonly<{
```
imransl, line 71: Wonder if `TrackConfigsOfGenome` be better. @azangru @ens-ma7 Any opinion on this?

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +98,255 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+
+export const getPersistentTrackConfigsPerGenome = (
+  genomeId: string
+): Partial<TrackConfigsPerGenome> => {
+  const trackConfigState =
+    browserTrackConfigStorageService.getTrackConfigState();
+  if (!genomeId || !trackConfigState[genomeId]) {
+    return {};
+  }
+  return trackConfigState[genomeId] || {};
```
imransl, line 159: Won't this do?

```return trackConfigState[genomeId] ?? {};```

File: src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx (hunk 34 above)
Comment (azangru) [reply src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx:48]: There will likely be more variations of the config panel than just these two. It will be difficult to express as a ternary.

File: src/content/app/genome-browser/state/track-config/trackConfigSlice.ts (hunk 35 above)
Comment (azangru) [reply src/content/app/genome-browser/state/track-config/trackConfigSlice.ts:71]: Or "For"? 😄

File: src/content/app/genome-browser/state/track-config/trackConfigSlice.ts (hunk 36 above)
Comment (azangru) [reply src/content/app/genome-browser/state/track-config/trackConfigSlice.ts:159]: Yes, this is probably the more appropriate syntax for expressing this idea these days.

File: src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx [hunk 37]
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { RootState } from 'src/store';
```
imransl, line 18: This is not a third party import. So move it above `styles` import since it is a type? Also use `type` for this import?

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { RootState } from 'src/store';
```
imransl, line 18: Not a third party import. Also type (see below comment).

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import storageService, {
+  StorageServiceInterface
```
imransl, line 18: type import here as well

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
```diff
@@ -0,0 +1,51 @@
...
+} from 'src/services/storage-service';
+import {
+  GenomeTrackConfigs,
+  TrackConfigsPerGenome
+} from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
```
imransl, line 23: And more type imports

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -15,163 +15,42 @@
  */
 
 import React from 'react';
-import { useDispatch, useSelector } from 'react-redux';
-import { OutgoingActionType } from '@ensembl/ensembl-genome-browser';
+import { useSelector } from 'react-redux';
+
+import RadioGroup, {
+  RadioOptions
```
imransl, line 21: Another type

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -29,27 +29,40 @@ import BrowserCog from './BrowserCog';
 
 import {
   getBrowserCogList,
-  getBrowserCogTrackList,
   getBrowserSelectedCog
 } from 'src/content/app/genome-browser/state/track-config/trackConfigSelectors';
 import {
   CogList,
```
imransl, line 35: type import

src/content/app/genome-browser/components/browser-track-config/track-config-views/TranscriptTrackConfig.tsx
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { RootState } from 'src/store';
```
imransl, line 18: Same import issues as mentioned above

src/content/app/genome-browser/components/browser-track-config/track-config-views/TranscriptTrackConfig.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { useSelector } from 'react-redux';
+
+import SlideToggle from 'src/shared/components/slide-toggle/SlideToggle';
+
+import { TrackType } from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
```
imransl, line 23: Type import

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect, useRef } from 'react';
+import { RootState } from 'src/store';
```
imransl, line 18: type import

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+import { useSelector, useDispatch } from 'react-redux';
+
+import { OptionValue } from 'src/shared/components/radio-group/RadioGroup';
```
imransl, line 21: type import

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+
+import { getBrowserActiveGenomeId } from '../../state/browser-general/browserGeneralSelectors';
```
imransl, line 23: the import path

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+import {
+  getTrackConfigForTrackId,
+  getApplyToAllConfig,
+  getBrowserCogList,
+  getBrowserSelectedCog
+} from '../../state/track-config/trackConfigSelectors';
```
imransl, line 29: Import path here as well as below

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+import {
+  updateTrackName as updateTrackConfigTrackName,
+  updateFeatureLabel as updateTrackConfigFeatureLabel,
+  updateShowSeveralTranscripts as updateTrackConfigShowSeveralTranscripts,
+  updateShowTranscriptIds as updateTrackConfigShowTranscriptIds,
+  updateApplyToAll,
+  TrackType
```
imransl, line 36: type import

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: Yes 'For' is better

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -0,0 +1,258 @@
...
+} from '../../state/track-config/trackConfigSlice';
+
+import analyticsTracking from 'src/services/analytics-service';
+import useGenomeBrowser from '../../hooks/useGenomeBrowser';
+
+const useBrowserTrackConfig = () => {
+  const applyToAllConfig = useSelector(getApplyToAllConfig);
+  const browserCogList = useSelector(getBrowserCogList);
+  const selectedCog = useSelector(getBrowserSelectedCog) || '';
```
imransl, line 45: Think we should not give a default value as empty string here and instead check in `selectedTrackConfigInfo` whether `selectedCog` is `null` before calling `getTrackConfigForTrackId`.

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
```diff
@@ -15,163 +15,42 @@
...
+      return <GeneTrackConfig />;
+    case TrackType.REGULAR:
+      return <RegularTrackConfig />;
+  }
+};
+
 export const BrowserTrackConfig = () => {
   const applyToAllConfig = useSelector(getApplyToAllConfig);
-  const browserCogTrackList = useSelector(getBrowserCogTrackList);
   const selectedCog = useSelector(getBrowserSelectedCog) || '';
```
imransl, line 51: I've mentioned about giving a default empty string value below. Please see that comment.

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
```diff
@@ -0,0 +1,96 @@
...
+  getBrowserSelectedCog,
+  getTrackConfigForTrackId
+} from 'src/content/app/genome-browser/state/track-config/trackConfigSelectors';
+
+import useBrowserTrackConfig from '../useBrowserTrackConfig';
+
+import styles from '../BrowserTrackConfig.scss';
+
+export const GeneTrackConfig = () => {
+  const selectedCog = useSelector(getBrowserSelectedCog) || '';
```
imransl, line 34: Same here too about the empty string as default

src/content/app/genome-browser/components/browser-track-config/track-config-views/TranscriptTrackConfig.tsx
```diff
@@ -0,0 +1,70 @@
...
+  getBrowserSelectedCog,
+  getTrackConfigForTrackId
+} from 'src/content/app/genome-browser/state/track-config/trackConfigSelectors';
+
+import useBrowserTrackConfig from '../useBrowserTrackConfig';
+
+import styles from '../BrowserTrackConfig.scss';
+
+export const TranscriptTrackConfig = () => {
+  const selectedCog = useSelector(getBrowserSelectedCog) || '';
```
imransl, line 34: Or maybe we need to add another selector in `trackConfigSelectors` to extract `selectedTrackConfigInfo` rather than calling these two selectors everywhere.

src/content/app/genome-browser/hooks/useBrowserRouting.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import { useEffect, useRef, useCallback } from 'react';
+import { RootState } from 'src/store';
```
imransl, line 18: type import

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -178,27 +251,109 @@ const useGenomeBrowser = () => {
...
-    genomeBrowser?.send(action);
+    });
+
+    const trackInfo = allTracksInfo && allTracksInfo[trackId];
 
-    if (allTrackLabelsOn && isTurnedOn) {
+    if (trackInfo && 'showFeatureLabel' in trackInfo && isTurnedOn) {
       genomeBrowser?.send({
-        type: allTrackLabelsOn
+        type: trackInfo?.showFeatureLabel
```
imransl, line 356: The `?` operator is not necessary here

src/content/app/genome-browser/state/track-config/trackConfigSelectors.ts
```diff
@@ -15,21 +15,44 @@
  */
 
 import { RootState } from 'src/store';
```
imransl, line 17: type import

src/content/app/genome-browser/state/track-config/trackConfigSelectors.ts
```diff
@@ -15,21 +15,44 @@
...
-  state.browser.trackConfig.trackConfigLabel;
-
-export const getApplyToAllConfig = (state: RootState) =>
-  state.browser.trackConfig.applyToAllConfig;
+export const getApplyToAllConfig = (state: RootState) => {
+  const genomeId = getBrowserActiveGenomeId(state);
+  if (!genomeId || !state.browser.trackConfig[genomeId]) {
+    return defaultTrackConfigsPerGenome.applyToAllConfig;
+  }
+  return state.browser.trackConfig[genomeId]?.applyToAllConfig;
```
imransl, line 57: `?` not needed here

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -21,43 +21,73 @@ import {
...
 import browserStorageService from 'src/content/app/genome-browser/services/browserStorageService';
+import browserTrackConfigStorageService from '../../components/browser-track-config/services/browser-track-config-storage-service';
+import { BrowserTrackStates } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
 
 import { updateTrackStates } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
 import { getBrowserTrackStates } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSelectors';
 
-import { BrowserTrackStates } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
-import { RootState } from 'src/store';
+import { GenomeTrackCategory } from 'src/shared/state/genome/genomeTypes';
```
imransl, line 33: A few more type imports here as well

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: Looking this at the second time we don't even need a default value as that's already done in line 157. Simply `return trackConfigState[genomeId];` should do.

tests/fixtures/browser.ts
```diff
@@ -43,36 +43,27 @@ import { LoadingState } from 'src/shared/types/loading-state';
 import { BreakpointWidth } from 'src/global/globalConfig';
```
imransl, line 43: type import

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { useEffect, memo } from 'react';
+import React, { useEffect, memo, useRef } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
```
imransl, line 18: We now need to use `useAppDispatch` and `useAppSelector` from `src/store` here and everywhere else

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: Discussed with BE. We need to create JIRA tickets from the action points from the meeting

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: because I destructure the array where it is called. So its either here or there we need to add. It makes sense we add it here.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,52 +98,190 @@ export const updateTrackStatesAndSave: ActionCreator<
...
-      const { selectedCog, isTrackLabelShown } = action.payload;
-      state.trackConfigLabel[selectedCog] = isTrackLabelShown;
+      const { genomeId, selectedCog, isTrackLabelShown } = action.payload;
+      const trackConfigState = state[genomeId].tracks[selectedCog];
+
+      if (trackConfigState.trackType !== TrackType.GENE) {
+        return;
+      }
+
+      // TODO: check if trackConfigState is a reference and the actual state is not getting updated
```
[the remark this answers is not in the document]
jyothishnt: redux toolkit uses immer library. So this is a copy and not the original state.

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.tsx
jyothishnt: I am not sure which one is good. It is a shared component across all types of track configs. Will this change in the future? May be we can ask Andrea?

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
jyothishnt: yes, "apply to all" is global as far as I know

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
imransl: ✅

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
imransl: ✅

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
imransl: ✅

src/content/app/genome-browser/components/browser-cog/BrowserCogList.test.tsx
```diff
@@ -19,38 +19,30 @@ import configureMockStore from 'redux-mock-store';
...
+
+import { updateCogList } from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
 
 const mockGenomeBrowser = jest.fn(() => new MockGenomeBrowser() as any);
 
 jest.mock(
   'src/content/app/genome-browser/hooks/useGenomeBrowser',
   () => () => ({
     genomeBrowser: mockGenomeBrowser()
+    // toggleTrackName: () => jest.fn()
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: ✅

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: ✅

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: ✅

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
imransl: ✅

tests/fixtures/browser.ts
imransl: ✅

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
imransl: ✅

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
imransl: ✅

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
imransl: ✅

src/content/app/genome-browser/hooks/useBrowserRouting.ts
imransl: ✅

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
imransl: ✅

package.json
```diff
@@ -30,7 +30,7 @@
     "coverage": "jest --coverage"
   },
   "dependencies": {
-    "@ensembl/ensembl-genome-browser": "0.2.1",
+    "@ensembl/ensembl-genome-browser": "0.3.0-test",
```
azangru, line 33: `0.3.0-test` no longer exists; should merge dev, which contains `0.3.0`

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -15,49 +15,109 @@
...
+  [trackId: string]: TrackConfig;
+};
+
+export type TrackConfigsForGenome = Readonly<{
   applyToAllConfig: {
-    isSelected: boolean;
-    allTrackNamesOn: boolean;
-    allTrackLabelsOn: boolean;
+    isSelected?: boolean;
   };
```
azangru, line 76: Did we discuss whether it should be `applyToAll : boolean` rather than `applyToAllConfig: {...}`?

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +128,247 @@ export const updateTrackStatesAndSave: ActionCreator<
   browserStorageService.saveTrackStates(trackStates);
 };
 
+export const prepareTrackConfigs = (trackCategories: GenomeTrackCategory[]) => {
+  const defaultTrackConfigs: TrackConfigs = {};
```
azangru, line 132: The word `default` is probably unnecessary here.

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +128,247 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+
+  defaultTrackConfigs['gene-focus'] = {
```
azangru, line 134: Remember there was that function you created? ;-)

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +128,247 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+};
+
+export const getPersistentTrackConfigsForGenome = (
+  genomeId: string
+): Partial<TrackConfigsForGenome> => {
+  const trackConfigs = browserTrackConfigStorageService.getTrackConfigs();
+
+  if (!genomeId || !trackConfigs[genomeId]) {
+    return {};
+  }
```
azangru, line 177: this check probably can go: `genomeId` is defined as string.

`return trackConfigs[genomeId] ?? {};`

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +128,247 @@ export const updateTrackStatesAndSave: ActionCreator<
...
-      const { selectedCog, isTrackLabelShown } = action.payload;
-      state.trackConfigLabel[selectedCog] = isTrackLabelShown;
+      const { genomeId, selectedCog, isTranscriptIdsShown } = action.payload;
+      const trackConfigState = state[genomeId].tracks[selectedCog];
+
+      if (trackConfigState.trackType !== TrackType.GENE) {
+        return;
+      }
+
+      // TODO: check if trackConfigState is a reference and the actual state is not getting updated
```
azangru, line 345: ✂️

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -73,34 +95,41 @@ export const BrowserCogList = () => {
...
+      }
+
+      const handleCogSelect = (trackId: string | null) => {
+        dispatch(
+          updateSelectedCog({
+            genomeId: genomeIdRef.current as string,
+            selectedCog: trackId
+          })
+        );
+      };
```
azangru, line 116: why is the code for the function creation inside the map loop? why isn't the function created once outside the loop?

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -14,42 +14,55 @@
...
+  const genomeId = useAppSelector(getBrowserActiveGenomeId);
+  const objectId = useAppSelector(getBrowserActiveFocusObjectId);
+  const dispatch = useAppDispatch();
 
+  const genomeIdRef = useRef(genomeId);
   const { genomeBrowser } = useGenomeBrowser();
 
+  useEffect(() => {
+    genomeIdRef.current = genomeId;
+  }, [genomeId]);
```
azangru, line 59: What‘s the purpose of assigning genome id to a ref?

If you need access to the genome id as it existed during the previous render cycle, then we have a `usePrevious` hook. But looking at the code, I don't even understand the intention behind this change.

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -73,34 +95,41 @@ export const BrowserCogList = () => {
...
+      const posStyle = { top: `${pos}px` };
+
+      if (!genomeIdRef.current || !objectId) {
+        return;
+      }
+
+      const handleCogSelect = (trackId: string | null) => {
+        dispatch(
+          updateSelectedCog({
+            genomeId: genomeIdRef.current as string,
```
azangru, line 112: line 105 makes sure it's a string

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.test.tsx
```diff
@@ -15,40 +15,82 @@
...
+  const fragment = {
+    selectedCog: selectedTrackId,
+    tracks: {
+      [selectedTrackId]: {
+        showSeveralTranscripts: false,
+        showTranscriptIds: false,
+        showTrackName: false,
+        showFeatureLabel: false,
+        trackType: trackConfigActions.TrackType.GENE
+      }
```
azangru, line 49: Do I remember a certain function that was intended to build this? 😉

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.test.tsx
```diff
@@ -15,40 +15,82 @@
...
 import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { Provider } from 'react-redux';
-import thunk from 'redux-thunk';
 
-import { createMockBrowserState } from 'tests/fixtures/browser';
 import MockGenomeBrowser from 'tests/mocks/mockGenomeBrowser';
 
 import * as trackConfigActions from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+import * as browserGeneralActions from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
```
azangru, line 26: It's so odd we are calling these two imports "actions". As a result, you write below `browserGeneralActions.defaultBrowserGeneralState`, `trackConfigActions.defaultTrackConfigsForGenome`, `trackConfigActions.TrackType.GENE`, etc., which has nothing to do with actions.

Why don't we call them slices?

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.storageService.update(StorageKeys.TRACK_CONFIGS, {
+      [genomeId]: fragment
+    });
+  }
+
+  public deleteTrackConfigs(genomeId: string): void {
+    this.setTrackConfigs({
+      genomeId,
+      fragment: undefined
+    });
```
azangru, line 54: We don't need to store the undefined. Genome id is being deleted; we should forget it ever existed.

```
this.storageService.update(StorageKeys.TRACK_CONFIGS, genomeId);
```

src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx
```diff
@@ -0,0 +1,57 @@
...
+    getTrackConfigForTrackId(state, selectedCog)
+  );
+  const shouldShowTrackName = selectedTrackConfigInfo?.showTrackName ?? false;
+
+  const { updateTrackName } = useBrowserTrackConfig();
+
+  return (
+    <div className={styles.section}>
+      <div className={styles.subLabel}>Show</div>
+      <div>
```
azangru, line 43: is there a reason for this div?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -45,7 +45,7 @@ type TrackPanelGeneProps = {
 };
 
 // TODO: figure out proper gene and transcript track naming conventions
-const GENE_TRACK_ID = 'track:gene-feat';
+const GENE_TRACK_ID = 'track:gene-focus';
```
azangru, line 48: Here and elsewhere.

There's code around that does `trackId.replace('track:', '')`. We should get rid of it. If the backend is sending us track names that start with the word "track", we should remove this prefix [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/state/genome/genomeSlice.ts#L128), and then ask Andres to remove this from the backend.

src/content/app/genome-browser/components/browser-track-config/services/browser-track-config-storage-service.ts
azangru: Also, I am not seeing this method actually used anywhere? We want to run it when a species is deleted.

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
azangru: Decided against it.

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -58,49 +83,53 @@ export const BrowserCogList = () => {
...
-        cogList[trackSummary['switch-id']] = Number(trackSummary.offset);
+        const trackId =
+          trackSummary['switch-id'] === 'focus'
+            ? 'gene-focus'
+            : trackSummary['switch-id'];
+        cogList[trackId] = Number(trackSummary.offset);
       }
     });
 
     if (cogList) {
```
azangru, line 94: How can it not be defined? It was defined on line 78.

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
azangru: Removed the ref.

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
jyothishnt: created ticket

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
jyothishnt: We will do this in another ticket.

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
jyothishnt: created ticket

ens-ma7 (approved):

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
```diff
@@ -15,41 +15,64 @@
...
-  updateCogTrackList,
+  type CogList,
+  updateCogList,
   updateSelectedCog
 } from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+import {
+  getBrowserActiveFocusObjectId,
+  getBrowserActiveGenomeId
+} from 'src/content/app/genome-browser/state/browser-general/browserGeneralSelectors';
+import { TrackId } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
```
imransl, line 46: This enum import is in the wrong place

src/content/app/genome-browser/components/browser-cog/useBrowserCogList.ts
```diff
@@ -0,0 +1,111 @@
...
+import {
+  setInitialTrackConfigsForGenome,
+  getDefaultGeneTrackConfig,
+  getDefaultRegularTrackConfig,
+  getTrackType,
+  TrackType,
+  type TrackConfigs,
+  type TrackConfigsForGenome
+} from 'src/content/app/genome-browser/state/track-config/trackConfigSlice';
+import { TrackId } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
```
imransl, line 38: Move this import to the types group below?

azangru (approved):
As I pointed out on Slack, the existence of [this code](https://github.com/Ensembl/ensembl-client/blob/7c7abc0c72f8071d638f5cdc73809444ec120364/src/content/app/genome-browser/hooks/useGenomeBrowser.ts#L178-L180) makes the [TrackId enum](https://github.com/Ensembl/ensembl-client/blob/7c7abc0c72f8071d638f5cdc73809444ec120364/src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts#L64-L66) rather meaningless. Not a blocker though. We can either clean this up in this PR or address this in a subsequent PR.

src/content/app/genome-browser/components/browser-track-config/BrowserTrackConfig.test.tsx
```diff
@@ -58,55 +105,71 @@ describe('<BrowserTrackConfig />', () => {
...
       await userEvent.click(toggle);
-
-      expect(trackConfigActions.updateTrackConfigNames).toHaveBeenCalledTimes(
-        1
-      );
-      expect(trackConfigActions.updateTrackConfigNames).toHaveBeenCalledWith({
-        isTrackNameShown: false,
-        selectedCog: mockState.browser.trackConfig.selectedCog
+      const updatedState = store.getState();
+      expect(trackConfigSlice.updateTrackName).toHaveBeenCalledTimes(1);
```
imransl, line 139: This is redundant

src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <div className={styles.toggleWrapper}>
+          <label>Track name</label>
+          <SlideToggle
+            isOn={shouldShowTrackName}
+            onChange={() => updateTrackName(!shouldShowTrackName)}
```
imransl, line 48: Use a function expression instead of an inline function?

src/content/app/genome-browser/components/browser-track-config/track-config-views/TranscriptTrackConfig.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <div className={styles.section}>
+      <div className={styles.subLabel}>Show</div>
+      <div>
+        <div className={styles.toggleWrapper}>
+          <label>Track name</label>
+          <SlideToggle
+            isOn={shouldShowTrackName}
+            onChange={() => updateTrackName(!shouldShowTrackName)}
```
imransl, line 53: Same as above here and below

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -178,27 +266,108 @@ const useGenomeBrowser = () => {
...
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  const toggleTranscriptIds = (params: {
+    trackId: string;
+    shouldShowTranscriptIds: boolean;
+  }) => {
+    const { trackId, shouldShowTranscriptIds } = params;
+    const trackIdWithoutPrefix = trackId.replace('track:', '');
+    const trackIdToSend =
+      trackIdWithoutPrefix === GENE_TRACK_ID ? 'focus' : trackIdWithoutPrefix;
+
```
imransl, line 337: Can we put a comment here on why this bit of the code is commented out?

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -68,54 +112,154 @@ export const updateTrackStatesAndSave: ActionCreator<
...
+    cogList: {},
+    selectedCog: null
+  },
+  configs: {}
+};
+
 const browserTrackConfigSlice = createSlice({
   name: 'genome-browser-track-config',
-  initialState: defaultTrackConfigState,
+  initialState: initialState,
```
imransl, line 150: Just `initialState` would do. No need to specify key and value separately when both have the same name.

src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx
jyothishnt: you mean have a separate function to call this? Sorry, I didnt get you.

src/shared/state/genome/genomeSlice.ts
```diff
@@ -82,7 +82,7 @@ export const fetchGenomeData =
   async (dispatch) => {
     await Promise.all([
       dispatch(fetchGenomeInfo(genomeId)),
-      dispatch(fetchGenomeTrackCategories(genomeId)),
+      dispatch(fetchGenomeTrackCategories(genomeId)), // <--
```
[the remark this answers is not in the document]
azangru: Ah, that's probably a remainder from our mob programming session on Tuesday 😳  Please feel free to delete it.

src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx
imransl: Write a function like this above and use it (`handleChange`) here:

```const handleChange = () => updateTrackName(!shouldShowTrackName)```;

jyothishnt: OK, thanks. So basically the main function called inside `handleChange`. Wondering the benefits of this as there are 4 functions in that component that would require this change.

src/content/app/genome-browser/components/browser-cog/BrowserCogList.tsx
jyothishnt: Would a new line before this line work?

src/content/app/genome-browser/components/browser-track-config/track-config-views/RegularTrackConfig.tsx
imransl: Let it be like this for now

PR #748, opened 2022-05-11
Add automatic switching for database selector

azangru:
## Description
_This is an alternative proposal for https://github.com/Ensembl/ensembl-client/pull/746. It also supersedes  https://github.com/Ensembl/ensembl-client/pull/749._

### Done in this PR
This PR moves the logic for updating the redux state responsible for BLAST form from the `useBlastInputSequences` React hook into `blastFormSlice`. As a consequence:
- `useBlastInputSequences` hook and its tests have been deleted.
- The logic for updating redux state, including the automatic guessing of input sequence type and the updates to the form settings, has been moved to `blastFormSlice`. Tests have been added to capture the logic of state updates.
- A new hook, `useForm`, has been created to make sure that redux actions for updating the state that depend on the BLAST config are only called only if the config has been fetched from the server.

Also included in this PR:
- Added the button for filling in the BLAST form with a sequence from another app — for example, from the sequence view in the genome browser drawer. The name of the component is `BlastSequenceButton`.

### How to review this PR
The central part of this PR are the updates to `blastFormSlice`, and the tests for `blastFormSlice` that capture the desired behaviour; so it's reasonable to start with those. The `useBlastForm` hook is a connector between `blastFormSlice` and the React components that talk to it. Finally, the `BlastSequenceButton` component adds new functionality.

## Related JIRA Issues
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1565
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1569

## Deployment URL(s)
http://database-selector-updated.review.ensembl.org


ens-ma7:
@azangru could you add the Jira tickets that are related to this PR in the description please.

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+type Props = {
+  label?: string;
+  species: Species;
+  sequence?: string; // if the sequence isn't provided, the button will be disabled
+  sequenceType: SequenceType;
+  className?: string;
+};
+
+const BlastSequenceButton = (props: Props) => {
+  const defaultLabel = 'Blast whole sequuence';
```
ens-ma7, line 50: `sequuence`  -> `sequence` (typo with double u)

azangru: Yes, this is my famous crappy keyboard that randomly puts in duplicate letters 😞

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
```diff
@@ -27,6 +27,10 @@
   grid-area: aside-top;
 }
 
+.blastSequenceButton {
+  --blast-sequence-button-margin: 0 0 0 25px;
```
ens-ma7, line 31: I am confused as to how we are initialising this css variable and how it is working.
It feels the wrong file to initialise it. Should we not initialise it in `src/shared/components/blast-sequence-button/BlastSequenceButton.scss` where it feels like the root css file. 
And how is it working, how it is setting the margin-left to 25px? should we not have something like this: 
`margin: var(--blast-sequence-button-margin, 0 0 0 25px)`

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
```diff
@@ -44,10 +48,12 @@
 
 .sequenceTypeSelection {
   margin-top: 50px;
+  --radio-label-font-weight: #{$light};
```
ens-ma7, line 51: same here, where else are we using this variable?

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
azangru: `BlastSequenceButton.scss` says: I am going to use the variable `--blast-sequence-button-margin` if it's available, or I will fall back to a default value.

This allows us to set the CSS variable anywhere up the DOM tree relative to BlastSequenceButton, and BlastSequenceButton will see it.

In this case, DrawerSequenceView component wants to move BlastSequenceButton around.

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.tsx
```diff
@@ -38,6 +44,7 @@ const sequenceLabelsMap: Record<SequenceType, string> = {
 
 // TODO: we probably also want to pass a sequence header in order to be able to blast it
 type Props = {
+  genomeId: string;
```
ens-ma7, line 47: we dont seems to be using this anywhere in the component? why has this been added in the props?

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
azangru: Now that I think of it, maybe `BlastSequenceButton` doesn't need the margin variable. Maybe, since it never defines a margin itself, it is sufficient for BlastSequenceButton to accept a css class that would have the margin defined.

_(This is different from the `--radio-label-font-weight` though. There, the radio component really needs to define the font-weight of its label.)_

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+import type { SequenceType } from 'src/content/app/tools/blast/types/blastSettings';
+
+import styles from './BlastSequenceButton.scss';
+
+// TODO: think about the possible header for the sequence
+type Props = {
+  label?: string;
+  species: Species;
+  sequence?: string; // if the sequence isn't provided, the button will be disabled
+  sequenceType: SequenceType;
```
ens-ma7, line 45: should this not be optional as well if `sequence` is optional?

azangru: I think the parent that is using this component will always know the sequence type, even if it may be fetching the sequence.

If this changes, we may change this prop later.

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+      sequence: { value: sequence },
+      species,
+      sequenceType
+    });
+    dispatch(addSelectedSpecies(species));
+
+    navigate(urlFor.blastForm());
+  };
+
+  const componentClasses = classNames(styles.container, className);
```
ens-ma7, line 79: `componentClasses` -> `containerClasses` ?

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
ens-ma7: I am confused... so even if `props.sequence` isn't provided as mentioned in the comment above, we still want to know about the sequenceType?

azangru: So far, I've only seen that the parent component may not know about the sequence (while it is still fetching it), but has a pretty clear idea what it is that it is fetching.

I do not know which other data can be legitimately unavailable to the parent that wants to render the blast button.

src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
azangru: Removed the `--blast-sequence-button-margin` variable.

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
azangru: I don't know. I was trying to express the idea that this is the topmost class that applies to the whole component.

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
azangru: ✅

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -35,7 +37,6 @@ import {
   getBlastJobName
 } from 'src/content/app/tools/blast/state/blast-form/blastFormSelectors';
 import {
-  setBlastDatabase,
   setBlastProgram,
```
ens-ma7, line 40: This might not be related to this change but more of a general question for blast, should the setBlastProgram also be moved to the blast hook? How do we know which one to move to the hook, its just a worry about there will be a mixture between hooks and redux function.

ens-ma7: so after looking at the hook useBlastForm, it feels like all of them has moved to the hooks and we should use them instead, is that correct?

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
ens-ma7: To me componentClasses are the one we see in the dom when inspecting it with the actual react component name so for this one something like `_BlastSequenceButton_` :shrug:

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -88,22 +89,24 @@ const BlastSettings = ({ config }: Props) => {
...
 
-  const onDatabaseChange = (database: string) => {
-    dispatch(
-      setBlastDatabase({
-        database,
-        config
-      })
-    );
+  const onDatabaseChange = (
+    database: string,
```
imransl, line 103: Can't this be a type instead?

src/content/app/tools/blast/hooks/useBlastForm.ts
```diff
@@ -0,0 +1,184 @@
...
+  setSequenceForGenome as setSequenceForGenomeAction,
+  setBlastProgram as setBlastProgramAction,
+  setSequenceType as setSequenceTypeAction,
+  setBlastDatabase as setBlastDatabaseAction,
+  updateEmptyInputDisplay as updateEmptyInputDisplayAction,
+  changeSensitivityPresets as changeSensitivityPresetsAction,
+  setHasUncommittedSequence as setHasUncommittedSequenceAction,
+  clearBlastForm as clearBlastFormAction,
+  type Species
+} from 'src/content/app/tools/blast/state/blast-form/blastFormSlice';
```
imransl, line 32: Why is it necessary to distinguish these functions as actions?

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
```diff
@@ -142,50 +239,32 @@ const blastFormSlice = createSlice({
...
-      const parameters = config.presets.settings[programName][presetName];
-      state.settings.program = programName;
-      state.settings.preset = presetName;
-      state.settings.parameters = { ...parameters, database };
+      const { database, config, isAutomatic } = action.payload;
+
+      state.settings.parameters = { ...state.settings.parameters, database };
+      if (!isAutomatic) {
+        state.settings.databaseSelectionMode = 'manual';
+        // NOTE: disabling the automatic guessing of sequence type in response to a manual choice of the database is deeply irrational;
```
imransl, line 263: 👀

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
```diff
@@ -253,6 +333,25 @@ const blastFormSlice = createSlice({
...
+        config: BlastSettingsConfig;
+      }>
+    ) {
+      const { sequence, species, sequenceType, config } = action.payload;
+      resetSettings(state, config);
+      state.sequences = [sequence];
+      state.selectedSpecies = [species];
+      state.shouldAppendEmptyInput = false;
+      state.settings.sequenceType = sequenceType;
+      state.settings.sequenceSelectionMode = 'manual'; // FIXME: does this suggest the need for a better flag name?
```
imransl, line 352: How about

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
azangru: How do you mean?

src/content/app/tools/blast/hooks/useBlastForm.ts
azangru: It isn't necessary. I am trying to avoid the naming clash with the functions declared inside the useBlastForm hook, and renaming the imports seemed like an easy way of doing so.

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
azangru: yep

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
azangru: I was more concerned with the word "manual" here, which we take as a signal to prevent automatic updates to the sequence type or the database. In this reducer, we are setting sequence selection mode to "manual" although the user has not manually selected the sequence type. I was wondering to myself here whether this suggests that instead of 

```
{
  sequenceSelectionMode: 'automatic' | 'manual';
  databaseSelectionMode: 'automatic' | 'manual';
}
```

 we should instead use something like:

```
{
  autoUpdateSettings: {
    sequenceType: boolean;
    database: boolean;
  }  
}
```

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
```diff
@@ -97,16 +101,109 @@ const getDatabaseSequenceType = ({
...
+const autoUpdateSettings = (
+  state: BlastFormState,
+  config: BlastSettingsConfig
+) => {
+  if (
+    state.sequences.length &&
+    state.settings.sequenceSelectionMode === 'automatic'
+  ) {
+    const firstSequence = state.sequences[0].value;
+    const sequenceType = guessSequenceType(firstSequence); // FIXME — should this only be done if there were no sequences before?
```
ens-ma7, line 116: is the comment still valid? Because we still want to guess the sequence type if lets say the user either change the existing one or paste a different sequence?

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
```diff
@@ -97,16 +101,109 @@ const getDatabaseSequenceType = ({
...
     ) {
-      const { sequences } = action.payload;
+      const { sequences, config } = action.payload;
+      updateEmptyInputVisibility(state, sequences);
+
       state.sequences = sequences;
+
+      if (!sequences.length) {
+        resetSettings(state, config);
+        autoUpdateSettings(state, config); // FIXME is this required?
```
ens-ma7, line 200: it does feel its not required since the resetSettings have been updated. Did you try removing it and see if it breaks anything?

src/shared/components/blast-sequence-button/BlastSequenceButton.tsx
ens-ma7: I couldn't find another alternative to componentClasses so happy to go with it.

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
ens-ma7: Probably best to rephrase the comment:
```Disabling the automatic guessing of sequence type from UX team request. ```

jyothishnt:
Just so that you dont forget, white space wrapping need to be fixed on the "Blast whole sequence"

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
azangru: Added a couple of more functions to useBlastForm. There are still several places where we are updating blastFormSlice by dispatching actions directly from inside components; but trying to wrap those actions inside the useBlastForm hook felt unnecessary and clumsy.

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
azangru: Decided to keep this as it is. Removing the FIXME.

src/content/app/tools/blast/state/blast-form/blastFormSlice.ts
azangru: But we still need to set the individual parameters based on the sequence type, database, and sensitivity preset. This code is currently in `autoUpdateSettings`, and it isn't clear to me whether it can safely be extracted in a dedicated function.

I'll keep the `autoUpdateSettings`, however silly it may look 😞

azangru:
> Just so that you dont forget, white space wrapping need to be fixed on the "Blast whole sequence"

Done.

ens-ma7:
I also notice the copy link isn't copying the sequence to clipboard. Could it be a regression?

imransl:
@azangru Had another look at the PR and haven't got anything much to add. Looks good to me.

ens-ma7 (approved):
@azangru minor issue, should the cursor be default when the "blast whole sequence" button is disable? otherwise all good

azangru:
> should the cursor be default when the "blast whole sequence" button is disable?
It should; but I think it should be done by the `ImageButton` component itself. I took a quick glance, and it looks like ImageButton needs some proper improvements done to it, including the use of CSS variables and investigation of whether we can avoid the use of multiple classnames passed to it. As a part of this refactoring, ImageButton should learn to disable its cursor when it's inactive. You can see that the attempt to have this behaviour has [already been made](https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/components/image-button/ImageButton.scss#L59).

imransl (approved):

PR #750, opened 2022-05-12
Blast icon status

imransl:
## Description
Added the blast status indicator based on the unviewed job submissions list after branching out from `unviewed-submissions-list`. You should be able to simulate the red (running jobs) and green (all jobs finished running). However, the indicator does not disappear after all jobs finished and viewed as the code doesn't support it yet (or does it?).

## Related JIRA Issue(s)
[ENSWBSITES-1519](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1519)

## Deployment URL(s)
http://blast-icon-status.review.ensembl.org 

## Views affected
Blast


imransl:
@azangru @ens-ma7 @jyothishnt I've made the change to the branch as well as code and forced pushed

src/header/launchbar/Launchbar.scss
```diff
@@ -91,3 +91,24 @@
...
+.jobStatus {
+  border-radius: 10px;
+  height: 18px;
+  width: 18px;
+  position: absolute;
+  top: -9px;
+  left: 23px;
+
+  &Running {
+    background-color: #d90000;
```
ens-ma7, line 108: should we not use the variables here $red?

src/header/launchbar/Launchbar.scss
```diff
@@ -91,3 +91,24 @@
...
+  }
+
+  &Finished {
+    background-color: #47d147;
```
ens-ma7, line 112: same here, global variable $green?

ens-ma7:
@imransl I am not sure if I understand the code right but will the icon colour change dynamically meaning without refreshing the page will the status change from red to green?  Shouldn't we have a `useeffect` somewhere?

ens-ma7:
> However, the indicator does not disappear after all jobs finished and viewed as the code doesn't support it yet (or does it?).

I think this should disappear once you dont have any `unviewed jobs` but will be good to double check this.

jyothishnt:
Hi @imransl, the first time I run a job (on https), I get a 402 error as below. But it runs successfully the second time. 
Screenshot 2022-05-16 at 10 48 54
Another question. I am on a different app while my job is running and I see the icon turned green, I click the blast launchbar icon. But to see the my finished jobs, I have to click JobList  (which shows empty or viewed jobs) and then click on Unviewed Jobs to view my results. Not sure if this workflow is correct. Something for Andrea may be

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { getUnviewedBlastSubmissions } from 'src/content/app/tools/blast/state/blast-results/blastResultsSelectors';
+
+import LaunchbarButton from './LaunchbarButton';
+import { BlastIcon } from 'src/shared/components/app-icon';
+
+import styles from './Launchbar.scss';
+
+const BlastLaunchbarButton = () => {
+  const unviewedSubmissions = useAppSelector(getUnviewedBlastSubmissions);
+  const isAnyJobRunning = useRef(false);
```
azangru, line 31: What's the purpose of using the ref heree?

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+        ? styles.jobStatusRunning
+        : styles.jobStatusFinished
+    );
+
+  const WrappedBlastIcon = () => (
+    <div className={styles.toolsIconWrapper}>
+      <BlastIcon />
+      {!!unviewedSubmissions.length && <div className={getStatusClass()}></div>}
+    </div>
+  );
```
azangru, line 52: Several things here.

For one, we are practically guaranteed to have a similar notification system for other tools buttons in the launchbar (check with Andrea if in doubt); so we might benefit from a more general component that knows how to display a dot (`LaunchbarButtonWithNotification` or a shorter name if you can think of one).

Second, the `LaunchbarButton` component doesn't work with BLAST. Currently, `LaunchbarButton` can only create a single link to `/blast`; but we know that this is insufficient. The most likely behaviour that we will need is for the blast button to keep track which blast view the user was on last, and to take the user to that view when he clicks on the button. I don't know if this should be considered part of this PR; but it suggests that we might rethink the interface of the `LaunchbarButton` component.

Something for us to discuss, perhaps?

src/header/launchbar/Launchbar.scss
```diff
@@ -91,3 +91,24 @@
...
+  position: relative;
+}
+
+.jobStatus {
+  border-radius: 10px;
+  height: 18px;
+  width: 18px;
+  position: absolute;
+  top: -9px;
+  left: 23px;
```
azangru, line 105: Since you are putting the dot in the top right corner, the `right` property would be more appropriate. It won't depend on the width of the button, as it currently seems to be.

imransl: ✅

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: Removed it :white_check_mark

imransl:
> Hi @imransl, the first time I run a job (on https), I get a 402 error as below. But it runs successfully the second time. Screenshot 2022-05-16 at 10 48 54 Another question. I am on a different app while my job is running and I see the icon turned green, I click the blast launchbar icon. But to see the my finished jobs, I have to click JobList (which shows empty or viewed jobs) and then click on Unviewed Jobs to view my results. Not sure if this workflow is correct. Something for Andrea may be

@jyothishnt I've now made the Blast app to show the last visited page when a user revisits it

src/header/launchbar/Launchbar.scss
imransl: ✅

src/header/launchbar/Launchbar.scss
imransl: ✅

src/content/app/tools/blast/BlastPage.tsx
```diff
@@ -38,8 +42,20 @@ The emphasis of this tool is to find regions of sequence similarity, which will
...
 const BrowserPage = () => {
+  const lastVisitedBlastView = useAppSelector(getLastVisitedBlastView);
+  const dispatch = useAppDispatch();
+  const navigate = useNavigate();
+  const location = useLocation();
   const hasMounted = useHasMounted();
 
+  useEffect(() => {
+    navigate(lastVisitedBlastView);
+  }, []);
```
azangru, line 53: Unless this is a temporary change that you wanted to update in a subsequent PR, we should really be setting the path to the relevant BLAST view in the launchbar button. That's what I meant by saying that the `app` property of the launchbar button isn't adequate to the job anymore.

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+import LaunchbarButtonWithNotification from './LaunchbarIconWithNotification';
+import { BlastIcon } from 'src/shared/components/app-icon';
+
+import { JobStatus } from 'src/content/app/tools/blast/state/blast-results/blastResultsSlice';
+
+import styles from './Launchbar.scss';
+
+const BlastLaunchbarButton = () => {
+  const unviewedSubmissions = useAppSelector(getUnviewedBlastSubmissions);
+  let isAnyJobRunning = false;
```
azangru, line 32: why are we declaring it as a `let` here instead of declaring it as a const on line 34?

src/header/launchbar/LaunchbarIconWithNotification.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import styles from './Launchbar.scss';
+
+type LaunchbarButtonWithNotificationProps = {
+  buttonConfig: LaunchbarButtonProps;
+  notificationConfig: {
+    jobStatus: JobStatus;
+    shouldShowNotification: boolean;
+  };
+};
```
azangru, line 32: Something not quite right about these props. A notification launchbar button shouldn't know anything about jobs. What it needs to know is whether it wants to show a notification, and what kind. I suggest we do something like this:

```
type Notification = 'red' | 'green';

type Props = LaunchbarButtonProps & {
  notification: Notification | null;
};
```

(Or you might think of something better than 'red' and 'green'.)

src/header/launchbar/LaunchbarIconWithNotification.tsx
```diff
@@ -0,0 +1,61 @@
+/**
```
azangru, line 1: `LaunchbarIcon` -> `LaunchbarButton`

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: Ah I failed to check this one after getting rid of a `useEffect` that I had for checking jobs that are running. Thanks for spotting this.

src/header/launchbar/LaunchbarIconWithNotification.tsx
imransl: ✅

src/header/launchbar/LaunchbarIconWithNotification.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const getStatusClass = () =>
+    classNames(
+      styles.jobStatus,
+      jobStatus === 'RUNNING'
+        ? styles.jobStatusRunning
+        : styles.jobStatusFinished
+    );
+
+  const WrappedIcon = () => {
+    const { icon: ToolsIcon } = props.buttonConfig;
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/tools/blast/BlastPage.tsx
imransl: I've removed the redux code for this and moved the concern of path change to `BlastLaunchbarButton` ✅

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: ✅

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: ✅

imransl:
> @imransl I am not sure if I understand the code right but will the icon colour change dynamically meaning without refreshing the page will the status change from red to green? Shouldn't we have a `useeffect` somewhere?

Sorry I couldn't answer this question earlier. We don't need `useEffect` to change the colour dynamically. `unviewedSubmissions` from redux gives us enough info to figure out which colour needs to be used.

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const BLAST_APP_ROOT_PATH = '/blast';
+
+const BlastLaunchbarButton = () => {
+  const unviewedSubmissions = useAppSelector(getUnviewedBlastSubmissions);
+  const location = useLocation();
+  const [blastAppPath, setBlastAppPath] = useState(BLAST_APP_ROOT_PATH);
+
+  useEffect(() => {
+    if (location.pathname.includes(BLAST_APP_ROOT_PATH)) {
```
azangru, line 37: `includes` -> `startsWith`?

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { BlastIcon } from 'src/shared/components/app-icon';
+
+import styles from './Launchbar.scss';
+
+const BLAST_APP_ROOT_PATH = '/blast';
+
+const BlastLaunchbarButton = () => {
+  const unviewedSubmissions = useAppSelector(getUnviewedBlastSubmissions);
+  const location = useLocation();
+  const [blastAppPath, setBlastAppPath] = useState(BLAST_APP_ROOT_PATH);
```
azangru, line 34: Ooh, clever! I didn't think we could pull this off without redux. Of course our hope is that the launchbar will never be unmounted over the lifetime of the page, i.e. that we will not ever have a screen that doesn't have the launchbar on it.

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (location.pathname.includes(BLAST_APP_ROOT_PATH)) {
+      setBlastAppPath(location.pathname);
+    }
+  }, [[location.pathname]]);
+
+  const isAnyJobRunning = unviewedSubmissions
+    .flatMap((submission) =>
+      submission.results.map((job) => job.status === 'RUNNING')
+    )
+    .some(Boolean);
```
azangru, line 46: You seem to be doing a filtering operation here. Wouldn't the following be more appropriate:

```
unviewedSubmissions
    .filter((submission) =>
      submission.results.some((job) => job.status === 'RUNNING')
    )
    .length > 0;
```

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const getNotificationIndicatorColour = () => {
+    if (Boolean(unviewedSubmissions.length)) {
```
azangru, line 49: this check seems to be redundant; it's implied in the `isAnyJobRunning` variable

src/header/launchbar/BlastLaunchbarButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  };
+
+  return (
+    <div className={styles.category}>
+      <LaunchbarButtonWithNotification
+        path={blastAppPath}
+        description="BLAST"
+        icon={BlastIcon}
+        enabled={true}
```
azangru, line 62: why isn't this the default?

src/header/launchbar/LaunchbarButtonWithNotification.tsx
```diff
@@ -0,0 +1,57 @@
...
+import classNames from 'classnames';
+
+import LaunchbarButton, { LaunchbarButtonProps } from './LaunchbarButton';
+
+import styles from './Launchbar.scss';
+
+type NotificationIndicatorColour = 'red' | 'green';
+
+type LaunchbarButtonWithNotificationProps = LaunchbarButtonProps & {
+  notificationIndicatorColour: NotificationIndicatorColour | null;
```
azangru, line 27: I still feel that calling this property `notification` (red/green/null) would be more concise and cleaner than calling it `notificationIndicatorColour`. First, `notificationIndicatorColour` is longer. Secondly, it implies that there are some attributes of notification other than the colour that get passed to this component, which is not the case. Third, it makes you check on line 47 for the presence of "colour" rather than of the notification. Ugh 🤢

src/header/launchbar/LaunchbarButtonWithNotification.tsx
```diff
@@ -0,0 +1,57 @@
...
+        props.notificationIndicatorColour === 'green'
+    });
+
+  const WrappedIcon = () => {
+    const { icon: Icon } = props;
+
+    return (
+      <div className={styles.toolsIconWrapper}>
+        <Icon />
+        {Boolean(props.notificationIndicatorColour) && (
```
azangru, line 47: This Boolean looks excessive.

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: Am not sure which attribute you are asking about. Is it `enabled`?

src/header/launchbar/LaunchbarButtonWithNotification.tsx
imransl: I was of this as a `notificationIndicator` more than the notification itself, a bit similar to the notification LED that we have on smartphones. So that's why thought this would make more sense. But as you describe it, seems this itself is the `notification`. Will rename it to that.

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: Yes this looks neater. Thanks!

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: `isAnyJobRunning` is binary, however `notification` can be `null` as well when there are no `unviewedSubmissions`. I've moved the `isAnyJobRunning` variable inside this check. Is this any better?

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: Okay I assume it is `enabled` and added it into `defaultProps`

src/header/launchbar/LaunchbarButtonWithNotification.tsx
imransl: ✅

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: ✅

src/header/launchbar/BlastLaunchbarButton.tsx
imransl: ✅

azangru (approved):
Looks good. The only remaining thing that I think I would have improved is the styles of the notification circle. Instead of positioning it by passing to it the precise number of pixels, I'd use a transform: translate, so that the correctness of its position doesn't depend on the size of the circle:

```
.notification {
  position: absolute;
  top: 0;
  right: 0;
  transform: translate(50%, -50%);
}
```

But that's minor. If you'd rather keep the pixels, it's fine. The reason I thought of that was because the circle seemed too large to me, and I wondered what would happen if we had to make it smaller.

azangru (approved):

ens-ma7 (approved):

PR #751, opened 2022-05-18
Added alert button in storybook

ens-ma7:
## Description
As per the XD for showing errors to users in blast:
https://xd.adobe.com/view/47c05579-d7aa-414a-a0ca-e2b5e2c88636-1293/

we need an alert icon which user can hover or click to display tooltip and this will be reused in other places.  Adding this to storybook and then will add it to blast in a different PR.

We already have a similar icon with the same behaviour (Question icon next to MANE select):
http://2020.ensembl.org/entity-viewer/homo_sapiens_GCA_000001405_28/gene:ENSG00000135040?view=transcripts

some of the code was reused and as part of this PR I have moved the hook `useQuestionBehavior` to `shared/hooks` and rename it to `useShowTooltip`

**Note:**
We also have another occurence of the alert icon here: https://internal-2020.ensembl.org/unsupported-browser but it is just an icon with no tooltip and for this usecase we are just importing the SVG icon directly so not including it as a usecase for the new component created.

**Tip to review PR:**
- You need to checkout the branch locally so that you can run storybook and see the change
- hooks `useQuestionButtonBehaviour` renamed to `useShowTooltip` and moved to `src/shared/hooks`
- changes to `QuestionButton.tsx` are related to the above renaming
- new component `AlertButton` created in `src/shared/components/alert-button` where the changes for the new button are
- The test script `AlertButton.test.tsx` is the same as `QuestionButton.test.tsx` as we are testing the same behavior but different element (not even sure if we need a test script for it) 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1609

## Deployment URL(s)
Not applicable as this is a change to storybook

## Views affected
Storybook


src/shared/components/alert-button/AlertButton.scss
```diff
@@ -0,0 +1,16 @@
+@import 'src/styles/common';
+
+.alertButton {
+  display: inline-block;
+  position: relative;
+  top: 4px;
```
azangru, line 6: I really doubt `position: relative;  top: 4px;` should belong in this component. If it is to be reusable, it shouldn't know anything about the context in which it is put.

src/shared/components/alert-button/AlertButton.scss
```diff
@@ -0,0 +1,16 @@
...
+  cursor: pointer;
+  padding: 2px;
+  font-size: 0;
+
+  > svg {
+    width: 18px;
+    height: 18px;
```
azangru, line 13: Width and height should be configurable through CSS variables (we can use 18px as a fallback default).

src/shared/components/alert-button/AlertButton.scss
```diff
@@ -0,0 +1,16 @@
+@import 'src/styles/common';
+
+.alertButton {
+  display: inline-block;
+  position: relative;
+  top: 4px;
+  cursor: pointer;
+  padding: 2px;
+  font-size: 0;
```
azangru, line 9: Does the component need this?

src/shared/components/alert-button/AlertButton.test.tsx
```diff
@@ -0,0 +1,139 @@
+/**
```
azangru, line 1: Hmm, I think this whole test (as well as the one in the QuestionButton) now belongs to the `useShowTooltip` hook. It's the hook's logic that's being tested.

src/shared/components/alert-button/AlertButton.scss
```diff
@@ -0,0 +1,16 @@
...
+
+  > svg {
+    width: 18px;
+    height: 18px;
+    fill: $red;
```
azangru, line 14: We probably want this configurable through props? Two different alert levels — "error" and "warning", with error shown in red and warning shown in amber.

src/shared/components/alert-button/AlertButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+import AlertIcon from 'static/icons/icon_alert_circle.svg';
+
+import { useShowTooltip } from 'src/shared/hooks/useShowTooltip';
+
+import Tooltip from 'src/shared/components/tooltip/Tooltip';
+
+import styles from './AlertButton.scss';
+
+type Props = {
+  helpText: React.ReactNode;
```
azangru, line 28: `helpText ` -> `description`?

src/shared/components/alert-button/AlertButton.scss
azangru: Also, maybe add `user-select: none`

ens-ma7: remove font-size and added user-select

src/shared/components/alert-button/AlertButton.scss
ens-ma7: So if i understand this right and looking at how question button works, we wrap the component within a div which can then be styled where its being used. if thats case i have moved the style out including the display as well.

src/shared/components/alert-button/AlertButton.tsx
ens-ma7: change it to `tooltipText`  which is a bit clearer

azangru: The `text` part is confusing. By reading it, one would expect it to be a string; but it can be any React node, really. If Andrea changes her mind back, it may contain a button 🤷‍♂️ 

`tooltipContent`?

ens-ma7: Agreed with the new name. why should it be an optional prop? does that mean we can reuse the same component for showing just the icon?

azangru: > why should it be an optional prop? does that mean we can reuse the same component for showing just the icon?

Yep.

azangru: 🤷‍♂️  we might. Would be nice if we did.

ens-ma7: ok will do it as part of this PR, will be  a good test of using the shared component.

src/shared/components/alert-button/AlertButton.scss
ens-ma7: Done

src/shared/components/alert-button/AlertButton.scss
```diff
@@ -0,0 +1,24 @@
+@import 'src/styles/common';
+
+.alertButton {
+  padding: 2px;
+  user-select: none;
+
+  > svg {
+    width: var(--alert-icon-width, 18px);
+    height: var(--alert-icon-height, 18px);
+    cursor: pointer;
```
azangru, line 10: Not always. Only if has a tooltip to show on click.

src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -91,21 +92,20 @@
...
-    .infoText{
+    .infoText {
       font-size: 18px;
       font-weight: bold;
     }
-    
+
     .infoIcon {
       svg {
         fill: $red;
```
azangru, line 105: looks like you don't need this

src/shared/components/error-screen/ErrorScreen.scss
```diff
@@ -91,21 +92,20 @@
...
-        width: 24px;
-        height: 24px;
+        --alert-icon-width: 20px;
+        --alert-icon-height: 20px;
```
azangru, line 107: Make the AlertButton accept the `className` property that gets assigned to the outmost component. Then you'll be able to do `<AlertButton className={styles.alertIcon}>`, and you won't need the `.infoIcon` wrapper.

src/shared/components/alert-button/AlertButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { useShowTooltip } from 'src/shared/hooks/useShowTooltip';
+
+import Tooltip from 'src/shared/components/tooltip/Tooltip';
+
+import styles from './AlertButton.scss';
+
+type Props = {
+  tooltipContent?: React.ReactNode;
+  warningIcon?: boolean;
```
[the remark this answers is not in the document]
ens-ma7: Done...i have added red and amber for now as i remembered Andrea mentioning amber as the next one for warning. We can change this when we have to work on this.

src/shared/components/error-screen/ErrorScreen.scss
ens-ma7: Done

src/shared/components/alert-button/AlertButton.tsx
```diff
@@ -27,16 +27,19 @@ import styles from './AlertButton.scss';
 
 type Props = {
   tooltipContent?: React.ReactNode;
-  warningIcon?: boolean;
+  level?: 'red' | 'amber';
+  className: string;
```
azangru, line 31: This should be an optional property.

azangru (approved):

stories/shared-components/alert-button/AlertButton.stories.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-right: 1rem;
+}
+
+.alertIcon {
+  display: inline-block;
+  position: relative;
+  top: 4px;
+}
+
+.largeAlertIcon {
```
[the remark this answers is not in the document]
azangru: I know that it is I who is probably guilty of introducing this particular style; but I would personally advise in favour of keeping the syntax as close to plain css as possible. That way — who knows — we might one happy day in the future dump scss entirely and replace it with plain CSS.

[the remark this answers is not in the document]
imransl: Let's stick with this then

imransl (approved):

PR #753, opened 2022-05-19
increase sequence limit to 50

ens-ma7:
## Description
Currently for blast we are limiting the maximum number of sequences that can be added to the form to 30; after user testing it was agreed to increase the limit to 50. This is updating the `MAX_BLAST_SEQUENCE_COUNT` from 30 to 50 and also changing the sequence counter background to red when this number is exceeded.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1555

## Deployment URL(s)
http://sequence-limit.review.ensembl.org/blast

## Views affected
Blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
@@ -17,7 +17,6 @@
 import React from 'react';
 import classNames from 'classnames';
 import { useSelector, useDispatch } from 'react-redux';
-import classNames from 'classnames';
 
 import {
   getEmptyInputVisibility,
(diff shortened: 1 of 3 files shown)


src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
```diff
@@ -81,12 +82,18 @@ const BlastInputSequencesHeader = (props: Props) => {
     ? sequences.length + 1
     : sequences.length;
 
+  const sequenceCounterClass = classNames(styles.sequenceCounter, {
+    [styles.sequenceCounterError]: sequences.length > MAX_BLAST_SEQUENCE_COUNT
+  });
```
imransl, line 87: Why do we need this when the `Add another sequence` button is disabled when the maximum sequence count reaches 50?

ens-ma7: we discussed this in the daily scrum this morning, you might have missed it. I confirmed this with Andrea and Lucy, its just another visual indication of whats wrong.

imransl: Yes but this isn't shown at all right? I tried adding 50 sequences but there was no visual indicator to show this.

Screenshot 2022-05-19 at 17 50 59

imransl (approved):

azangru (approved):
I haven't checked the UI; but the code changes look good.

PR #755, opened 2022-05-24
Add different track configs for different types of tracks

jyothishnt:
## Description
Currently we only need 2 versions of the track panel (see spec)

Spec:  https://www.ebi.ac.uk/seqdb/confluence/display/ENSWEB/Genome+browser+settings+panels

(As Andrea is on leave, I decided to go with "Show 5 transcripts", which can be later discussed and changed.)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1567

## Deployment URL(s)
http://more-track-settings.review.ensembl.org/

## Views affected
Track config in GB


jyothishnt:
Merging this so that this can be reviewed together with https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-924

PR #757, opened 2022-05-30
Update dependencies

azangru:
## Description
Routine dependencies update.

### Notable changes
- Typescript has been updated to version 4.7. It discovered a couple of type errors that 4.6 didn't detect
- Faker has removed default export; and now has to be imported only as a named import

### Bugfixes
- The tests for TranscriptSequenceView were broken in various ways:
  - I was not awaiting for the result of the function inside the `waitFor`
  - Mock transcripts got generated with either forward or reverse strand; which had an impact on whether a strand was reverse-complemented in the response. I removed this randomness; and now the generated mock transcripts always are on the forward strand. It will be up to developer to overwrite the strand if needed.
- Didn't need to wrap `userEvent.click` in an `act`, because it already is wrapped in it. Saw console warnings because of that.

## Deployment URL(s)

## Checklist
- npm run test — ✅ 
- npm run check-types — ✅ 
- npm run storybook — ✅ 
- development build — ✅ 
- production build — ✅


imransl:
There is a test that fails randomly when run locally:

Screenshot 2022-05-31 at 11 25 54

src/content/app/genome-browser/components/drawer/components/sequence-view/TranscriptSequenceView.test.tsx
```diff
@@ -173,30 +173,28 @@ describe('<TranscriptSequenceView />', () => {
...
     });
   });
 
   describe('reverse complement', () => {
     it('only shows the reverse complement checkbox for the genomic sequence', async () => {
       const { getByLabelText, queryByLabelText } = renderTranscriptSequenceView(
-        { transcript: proteinCodingTranscript }
+        {
+          transcript: proteinCodingTranscript
+        }
```
imransl, line 197: Has prettier changed its default rules again? 👀

azangru: Prettier doesn't seem to allow thiis:

```
      const { getByLabelText, queryByLabelText } = renderTranscriptSequenceView({
          transcript: proteinCodingTranscript
      });
```
because `      const { getByLabelText, queryByLabelText } = renderTranscriptSequenceView({` is exactly 81 characters.

I hate prettier! Almost as much as my Mac's keyboard.

PR #758, opened 2022-05-30
Remove the extra button wrapper in BlastLaunchbarButton

azangru:
## Description
Regression introduced in https://github.com/Ensembl/ensembl-client/pull/750

The BLAST launchbar button is twice wrapped in a category class:

image

This resulted in extra white space to the right of the button:

image

After the fix:

image

## Deployment URL(s)
http://fix-launchbar-button.review.ensembl.org


imransl (approved):
Thanks for fixing this. I should've checked that extra `div` before adding it in.

PR #761, opened 2022-06-13
GB version update - 0.3.1

jyothishnt:
## Description
Genome browser upgrade to 0.3.1-test
Once approved, we can change the version to 0.3.1

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1627

## Deployment URL(s)
http://version-update-0.3.1.review.ensembl.org/

## Views affected
Genome Browser

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "coverage": "jest --coverage"
   },
   "dependencies": {
-    "@ensembl/ensembl-genome-browser": "0.3.0",
+    "@ensembl/ensembl-genome-browser": "0.3.1-test",
     "@loadable/component": "5.15.2",
     "@loadable/server": "5.15.2",
     "@react-spring/web": "9.4.5-beta.1",
--- a/src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
+++ b/src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
@@ -43,7 +43,7 @@ export const GeneTrackConfig = () => {
     updateTrackName,
     updateTrackLabel,
     updateShowSeveralTranscripts,
-    updateShowTranscriptIds
+    updateShowTranscriptLabels
   } = useBrowserTrackConfig();
 
   if (!selectedTrackConfig) {
@@ -54,7 +54,7 @@ export const GeneTrackConfig = () => {
   const shouldShowTrackLabel = selectedTrackConfig.showFeatureLabel;
   const shouldShowSeveralTranscripts =
     selectedTrackConfig.showSeveralTranscripts;
-  const shouldShowTranscriptIDs = selectedTrackConfig.showTranscriptIds;
+  const shouldShowTranscriptLabels = selectedTrackConfig.showTranscriptLabels;
 
   return (
     <div className={styles.section}>
@@ -73,8 +73,10 @@ export const GeneTrackConfig = () => {
         <div className={styles.toggleWrapper}>
           <label>Transcript IDs</label>
           <SlideToggle
-            isOn={shouldShowTranscriptIDs}
-            onChange={() => updateShowTranscriptIds(!shouldShowTranscriptIDs)}
+            isOn={shouldShowTranscriptLabels}
+            onChange={() =>
+              updateShowTranscriptLabels(!shouldShowTranscriptLabels)
+            }
             className={styles.slideToggle}
           />
         </div>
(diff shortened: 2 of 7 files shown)


src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -185,35 +185,35 @@ const useBrowserTrackConfig = () => {
     });
   };
 
-  const updateShowTranscriptIds = (isTranscriptIdsShown: boolean) => {
+  const updateShowTranscriptLabels = (isTranscripLabelsShown: boolean) => {
```
azangru, line 188: `is` -> `are` in the function parameter? or perhaps just `shouldShow`, without the `transcriptLabels` part?

jyothishnt: I have used `shouldShowTranscriptLabels`, Is that OK?

azangru: Verbose, but ok 🤷‍♂️

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
```diff
@@ -222,7 +222,8 @@ const useBrowserTrackConfig = () => {
     analyticsTracking.trackEvent({
       category: 'track_settings',
       label: selectedCog,
-      action: 'several_transcripts_' + (isTranscriptIdsShown ? 'on' : 'off')
+      action:
+        'several_transcripts_' + (shouldShowTranscriptLabels ? 'on' : 'off')
```
azangru, line 226: Why `several_transcripts`?

package.json
```diff
@@ -30,7 +30,7 @@
     "coverage": "jest --coverage"
   },
   "dependencies": {
-    "@ensembl/ensembl-genome-browser": "0.3.0",
+    "@ensembl/ensembl-genome-browser": "0.3.1-test",
```
azangru, line 33: remember...

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
```diff
@@ -73,8 +73,10 @@ export const GeneTrackConfig = () => {
         <div className={styles.toggleWrapper}>
           <label>Transcript IDs</label>
           <SlideToggle
-            isOn={shouldShowTranscriptIDs}
-            onChange={() => updateShowTranscriptIds(!shouldShowTranscriptIDs)}
+            isOn={shouldShowTranscriptLabels}
+            onChange={() =>
+              updateShowTranscriptLabels(!shouldShowTranscriptLabels)
+            }
```
imransl, line 79: I am a bit conflicted with this syntax that seems to be used more frequently in the code. Why do you prefer this (inline function) over using a function expression?

src/content/app/genome-browser/components/browser-track-config/useBrowserTrackConfig.ts
jyothishnt: oops again!

src/content/app/genome-browser/components/browser-track-config/track-config-views/GeneTrackConfig.tsx
jyothishnt: haha that is true and you get so annoyed 😒   I will change it

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -286,7 +305,7 @@ const useGenomeBrowser = () => {
     });
   };
 
-  const toggleTrackLabel = (params: {
+  const toggleFeatureLabel = (params: {
```
azangru, line 308: - why `feature` rather than specifically `transcript`? the message we are sending to the genome browser inside this function is specific to transcripts
- why `label` (singular) rather than `labels` (plural), or `ids`?

jyothishnt: this is a generic method to turn on label, not just for transcript. It can be variant, GC, contig etc.

azangru: label still not renamed to labels

azangru (approved):

imransl (approved):

PR #763, opened 2022-06-14
Blast error icon

ens-ma7:
## Description
Adding alert icon (shared component created in this [PR](https://github.com/Ensembl/ensembl-client/pull/751)) to blast. When a sequence input box becomes invalid (invalid character or not matching the type), an alert icon is shown next to that input sequence box and also at the top header (where the sequence count is - see XD in jira).
A context had to be added to update the top header since we had to keep track of the validity of multiple boxes.

There are also minor code update (use AppSelector instead of use selector)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1559

## Deployment URL(s)
http://blast-error-icon.review.ensembl.org

## Views affected
http://blast-error-icon.review.ensembl.org/blast

**Tip to review PR:**
The alert icon for each sequence box is an easy one and changes made to `BlastInputSequence.tsx`
The other part is adding the react context to keep track of all sequences boxes. This was added in `BlastForm.tsx` and the context was called/updated in `BlastInputSequence.tsx` each time a sequence input box is updated.
And the final check to see if one of the sequence is invalid then show the alert icon is in `BlastInputSequenceHeader.tsx`
As a result of the context being added, the test `BlastInputSequence.test.tsx` (where the context is being used) had to be updated.


src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
```diff
@@ -117,17 +157,22 @@ describe('<BlastInputSequence />', () => {
...
 
-      rerender(<BlastInputSequence {...commonProps} sequence={sequence} />);
+      const reRenderComponent = (
+        <Provider store={mockStore(mockState)}>
+          <TestContext.Provider value={testContext}>
+            <BlastInputSequence {...commonProps} sequence={sequence} />
+          </TestContext.Provider>
+        </Provider>
+      );
+      rerender(reRenderComponent);
```
imransl, line 167: For these kind of rerenders what we have usually done in our tests is to render the component all over again like this:

```container = renderComponent({sequence}).container;```

File: src/content/app/tools/blast/views/blast-form/BlastForm.tsx [hunk 2]
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
```
imransl, line 18: This might be slightly out of scope, but since it should be a minor change can we switch to `useAppSelector` and `useAppDispatch`?

ens-ma7: Thats done.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
```diff
@@ -46,10 +49,21 @@ const BlastInputSequencesHeader = (props: Props) => {
...
 
-  const dispatch = useDispatch();
+  const sequenceValidityContext = useContext(BlastFormContext);
+  let shouldShowAlert = false;
+
+  if (sequenceValidityContext) {
+    shouldShowAlert = Object.values(
+      sequenceValidityContext.sequenceValidityStatus
+    ).includes(false);
+  }
```
imransl, line 66: Another, more concise way to write this would be:

```  const shouldShowAlert = sequenceValidityContext && Object.values(
    sequenceValidityContext.sequenceValidityStatus
  ).includes(false);```

File: src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.scss [hunk 4]
```diff
@@ -86,3 +86,10 @@
   font-weight: $light;
   margin-right: 14px;
 }
+
+.alertButton {
+  display: inline-block;
```
imransl, line 91: Don't think we need `inline-block` here. Doesn't make any difference when with or without it.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
```diff
@@ -14,30 +14,74 @@
...
+  blast: {
+    blastForm: initialBlastFormState
+  }
+};
+
+const emptyTestContext = {
+  sequenceValidityStatus: undefined,
+  updateSequenceValidity: jest.fn()
+};
+let testContext: typeof emptyTestContext;
```
imransl, line 58: We don't need to reassign `emptyTestContext` to another variable. Instead use it directly?

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
```diff
@@ -60,12 +60,23 @@ const BlastForm = () => {
 
 const Main = () => {
   const isSmallViewport = useMediaQuery('(max-width: 1900px)');
+  const [sequenceValidityStatus, setSequenceValidityStatus] = useState({});
```
azangru, line 63: "status" (singular) suggests that there is only one thing; whereas in your code it is a key-value collection of things, where the key is an index, and the value is a boolean representing whether this sequence is valid or not.

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
```diff
@@ -60,12 +60,23 @@ const BlastForm = () => {
...
+
+  const updateSequenceValidity = (index: number, status: boolean) => {
+    setSequenceValidityStatus({ ...sequenceValidityStatus, [index]: status });
```
azangru, line 66: so what do we do with keys for indexes that no longer exist?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -64,6 +70,9 @@ const BlastInputSequence = (props: Props) => {
   const deleteButtonRef = useRef<HTMLButtonElement | null>(null);
   const textareaRef = useRef<HTMLTextAreaElement | null>(null);
 
+  const errorTooltipDescription =
+    'Please check that all your sequences are nucleotide or protein, and that they do not contain any invalid characters';
```
azangru, line 74: let's move this out of the component function

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -132,6 +141,17 @@ const BlastInputSequence = (props: Props) => {
     ? checkSequenceValidity(input, sequenceType)
     : true; // treat empty input as valid
 
+  const { sequences } = useBlastForm();
```
azangru, line 144: I'm confused. Why does a single sequence box need access to the full list of sequences? Why doesn't it just validate its own input?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -132,6 +141,17 @@ const BlastInputSequence = (props: Props) => {
...
+  const sequenceValidityKey = index === null ? sequences.length : index;
+  const sequenceValidityContext = useContext(BlastFormContext);
+
+  useEffect(() => {
```
azangru, line 148: Prefer to keep all hooks close to the beginning of the component function. The reason being that you cannot place an if-statement returning null above a hook.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: well it only needs to know how many sequences we have to deal with the empty input sequence. remember we will know that there is an empty input sequence which is valid by looking at the sequences.length

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: I have moved it out but wasnt sure where else to put it so i have made it constant just outside the component function. let me know if there is a better place for it.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: Ok i manage to move it further up in the code. Please have a look and make sure i didnt break any flow (specially the function to check the sequence validity).

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -88,6 +88,21 @@ const BlastInputSequence = (props: Props) => {
...
+    sequenceValidityContext?.updateSequenceValidity(
+      sequenceValidityKey,
+      isInputValid
+    );
+  }, [isInputValid]);
```
imransl, line 104: These hook calls can be moved even further up. `useBlastForm` can definitely go just below the last `useRef` at line 71. ` useContext` right below it? And then you can move the other parts of this highlighted section to be right below them.

ens-ma7: whats the benefit of moving it even further? To my understanding its pretty much at the top of the code functionality where it wont be affected by any if statement unless i have missed something.

ens-ma7: I have moved the `useContext` at the top and the inputvalid check also but i left the useeffect where it was since it was straight after another useeffect which kind of make sense to keep them together. The code looks much neater this way, hope thats ok now.

imransl: It is usually good to move the hooks to the top. Like Andrey mentioned in another comment, we need them at the top where it is more visible so that we don't unknowingly place any if-statements.

ens-ma7: @imransl i have moved it to the top where the other `useEffect` is, could you check and let me know if we need to move it higher up. If thats the case then we need to move the other `useEffect` that was already there.

imransl: It looks better. However, the `useEffect` right below the `isInputValid` variable.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
```diff
@@ -14,30 +14,72 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { createContext } from 'react';
+import { Provider } from 'react-redux';
```
azangru, line 18: So, to clarify, `BlastInputSequence` component no longer needs redux? Then all redux code should be removed from the test.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
```diff
@@ -14,30 +14,72 @@
...
+  blast: {
+    blastForm: initialBlastFormState
+  }
+};
+
+const emptyTestContext = {
+  sequencesValidity: undefined,
+  updateSequenceValidity: jest.fn()
+};
+const TestContext = createContext(emptyTestContext);
```
azangru, line 58: Woah! Very confused. Is `TestContext` interchangeable with `BlastFormContext`? Doesn't a React component care where it's getting its values from?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
ens-ma7: Good spot, i forgot about this after i changed the place of the context. I dont need to make any changes to the test file then and i have removed all my changes.

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
```diff
@@ -60,12 +62,35 @@ const BlastForm = () => {
...
-  return isSmallViewport ? <MainSmall /> : <MainLarge />;
+  return (
+    <BlastFormContext.Provider
+      value={{
+        updateSequenceValidity,
+        removeSequenceValidity,
+        sequencesValidity
+      }}
+    >
+      {isSmallViewport ? <MainSmall /> : <MainLarge />}
```
azangru, line 91: By doing so, `MainSmall` and `MainLarge` will probably re-render every time the state of the `Main` container will update, which is probably going to be very frequently. We can avoid that by either wrapping those components in React.memo, or moving the logic related to the context into its own component, which would do the following:

```
type Props = {
  children: ReactNode
}

const BlastFormContextContainer = (props: Props) => {
   // create state
   // have function for updating the state 

  return (
    <BlastFormContext.Provider
      value={...}
    >{props.children}</BlastFormContext.Provider>
  );
}
```
and will be used in Main as:

```
<BlastFormContextContainer>{ isSmallViewport ? <MainSmall /> : <MainLarge /> }</BlastFormContextContainer>
```

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
```diff
@@ -57,6 +60,7 @@ const BlastInputSequences = () => {
     } else if (typeof index === 'number') {
       const newSequences = [...sequences].filter((_, i) => i !== index);
       updateSequences(newSequences);
+      sequenceValidityContext?.removeSequenceValidity(index);
```
azangru, line 63: Really not sure that this is even necessary.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: This is not relevant anymore since the code has been moved out of BlastInputSequence

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
ens-ma7: Done

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.test.tsx
ens-ma7: Done

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: Done

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: Done

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: This should now be moved...

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -40,7 +40,7 @@ import DeleteButton from 'src/shared/components/delete-button/DeleteButton';
 import AlertButton from 'src/shared/components/alert-button/AlertButton';
 
 import type { ParsedInputSequence } from 'src/content/app/tools/blast/types/parsedInputSequence';
-import { SequenceType } from 'src/content/app/tools/blast/types/blastSettings';
+import { type SequenceType } from 'src/content/app/tools/blast/types/blastSettings';
```
azangru, line 43: import type { SequenceType }

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -18,12 +18,15 @@ import React, {
...
   useEffect,
   useRef,
+  useContext,
   type FormEvent,
   type ClipboardEvent,
   type FocusEvent
 } from 'react';
 import classNames from 'classnames';
 
+import { BlastFormContext } from 'src/content/app/tools/blast/views/blast-form/BlastFormContextContainer';
```
azangru, line 28: Let's now separate BlastFormContext into its own file. It's odd that it has to be imported from a container; it should be the other way around: the container imports the context.

src/content/app/tools/blast/views/blast-form/BlastFormContextContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { ReactNode, useState } from 'react';
+
+import useBlastForm from 'src/content/app/tools/blast/hooks/useBlastForm';
+
+type SequenceValidity = {
```
azangru, line 21: `type SequenceValidity` -> `type BlastFormContextType`

src/content/app/tools/blast/views/blast-form/BlastFormContextContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  updateSequenceValidity: (index: number | null, status: boolean) => void;
+  removeSequenceValidity: (index: number) => void;
+  sequencesValidity: boolean[];
+};
+
+export const BlastFormContext = React.createContext<
+  SequenceValidity | undefined
+>(undefined);
```
azangru, line 29: Is the `undefined` argument required? I thought if you don't pass any arguments, it will be undefined.

src/content/app/tools/blast/views/blast-form/BlastForm.tsx
azangru: I am still not seeing in the code anything that would change the size of the array based on the length of the sequences plus the presence/absence of the empty input box.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
ens-ma7: I agree its feels odd for the context to be imported from the container but at the same time it feels odd to have a separate file just for 2 lines of code for the context:
```export const BlastFormContext = React.createContext<
  SequenceValidity | undefined
>(undefined);```

What about i renamed `BlastForContextContainer` to `BlastFormContext` and it will also contains the container as a separate export. Does that sound better?

Comment (azangru) [reply src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx:28]: > What about i renamed BlastForContextContainer to BlastFormContext and it will also contains the container as a separate export. Does that sound better?

Yes.

File: src/content/app/tools/blast/views/blast-form/BlastFormContextContainer.tsx (hunk 19 above)
Comment (ens-ma7) [reply src/content/app/tools/blast/views/blast-form/BlastFormContextContainer.tsx:29]: If i removed the `undefined` in the argument, the function complains saying it needs at least one argument passed. This is the error:
image

Comment (azangru) [reply src/content/app/tools/blast/views/blast-form/BlastFormContextContainer.tsx]: Ok, I thought the argument for `createContext` was optional; but apparently not, according both to the docs and type definitions.

File: src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx [hunk 20]
```diff
@@ -0,0 +1,69 @@
...
+    index: number | null,
+    status: boolean
+  ) => {
+    const array = [...sequenceValidityFlags];
+    index = index === null ? sequences.length : index;
+    array[index] = status;
+    setSequenceValidityFlags(array);
+  };
+
+  const removeSequenceValidityFlags = (index: number) => {
```
azangru, line 51: `flags` -> `flag` (we are only removing one)

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+  BlastFormContextType | undefined
+>(undefined);
+
+type Props = {
+  children: ReactNode;
+};
+
+export const BlastFormContextContainer = (props: Props) => {
+  const [sequenceValidityFlags, setSequenceValidityFlags] = useState<boolean[]>(
+    []
```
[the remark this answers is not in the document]
ens-ma7: Are you saying that the first sequence box (empty) is valid and hence set the array with [True]? Does that not complicate things when deleting box we have to make sure the array isn't empty if we delete everything.
Whats the benefit are having a filled array?

azangru:
The problem we saw yesterday is caused by the "Clear all":


https://user-images.githubusercontent.com/6834224/176373414-a6f04ae8-bcd7-4cd8-93c5-07e48138e4f8.mov

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -34,13 +37,14 @@ import {
...
+import AlertButton from 'src/shared/components/alert-button/AlertButton';
 
 import type { ParsedInputSequence } from 'src/content/app/tools/blast/types/parsedInputSequence';
-import { SequenceType } from 'src/content/app/tools/blast/types/blastSettings';
+import type { SequenceType } from 'src/content/app/tools/blast/types/blastSettings';
 
 import styles from './BlastInputSequence.scss';
 
-type Props = {
+export type Props = {
```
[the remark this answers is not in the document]
ens-ma7: Not anymore(i was using it for the test but since then the code has changed). Good spot, i have removed it. thanks

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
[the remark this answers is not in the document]
ens-ma7: Agreed in hte PR review to leave this for now.

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+  );
+  const { sequences } = useBlastForm();
+
+  const updateSequenceValidityFlags = (
+    index: number | null,
+    status: boolean
+  ) => {
+    const array = [...sequenceValidityFlags];
+    index = index === null ? sequences.length : index;
```
imransl, line 47: You could probably rewrite this as: `index = index ?? sequences.length;`

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
```diff
@@ -69,6 +81,11 @@ const BlastInputSequencesHeader = (props: Props) => {
     lastInputBox?.scrollIntoView({ block: 'end', behavior: 'smooth' }); // Safari doesn't properly support this (see https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView); but other options scroll just as jerkily on Safari
   };
 
+  const clearAllClick = () => {
+    clearAllSequences();
+    blastFormContext && blastFormContext.clearSequenceValidityFlags();
```
imransl, line 86: Doesn't `blastFormContext?.clearSequenceValidityFlags();` work here?

azangru: `clearAllClick` -> `onClearAll`

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+    array[index] = status;
+    setSequenceValidityFlags(array);
+  };
+
+  const removeSequenceValidityFlag = (index: number) => {
+    setSequenceValidityFlags(
+      sequenceValidityFlags.filter((value, i) => i !== index)
```
azangru, line 54: So what happens when we are deleting the empty box that doesn't have an index?

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
ens-ma7: Done

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequencesHeader.tsx
ens-ma7: Done both

src/content/app/tools/blast/views/blast-form/BlastFormContext.tsx
azangru: Yeah, that's what I thought — it's another way of breaking the UI:

https://user-images.githubusercontent.com/6834224/176670113-e465dc14-c9ae-456a-84b9-a00c4e25e375.mov

ens-ma7: Nothing happen :) And it wasn't that bad since it was getting overwritten the next time a new box is created but i agree its better to clean the array when this happen. I have updated the code to handle the deletion of empty box. But i spotted one more thing which is you can also delete a box by removing all the sequences in that box  and then click outside the box and the box will be deleted (this doesnt apply to the first box though). That behavior did surprise me but i am just trying to figure out the code where its deleting that box when its out of focus.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequence.tsx
```diff
@@ -50,6 +54,9 @@ type Props = {
   onRemoveSequence?: (index: number | null) => void;
 };
 
+const errorTooltipDescription =
+  'Please check that all your sequences are nucleotide or protein, and that they do not contain any invalid characters';
```
azangru, line 58: Why does this error message, which appears in a single input box, say "all your sequences"?

ens-ma7: Its per the XD in the Jira, not sure if Andrea was trying to keep it consistent but it makes sense to what you are saying, i will just double check with her and update the text. Should be easy to update it for just one box.

azangru (approved):

PR #769, opened 2022-06-27
Replace gene-focus to focus

jyothishnt:
## Description
Currently, our code first assigns a "gene-focus" id to the focus track and then renames it to "focus" before sending the data to the genome browser. This indirection is clearly unnecessary.
The task is to just use the string "focus" as the identifier of the focus track.

I see a problem here and would like to know your suggestions.
On this [line](https://github.com/Ensembl/ensembl-client/pull/769/files#diff-285d5a75f69c1ff6ef15d53d6aaea4e49ed4ccd2c3ab789adf58f4f12b39188dR133) I have added a condition to consider focus track as a GENE type config. When we have variant or other tracks as focus in the future, this is not going to work properly

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1626

## Deployment URL(s)
http://use-focus.review.ensembl.org

## Views affected
Genome browser

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
@@ -174,10 +174,8 @@ const useGenomeBrowser = () => {
     const trackStateForLabels = cloneDeep(emptyOnOffLists);
 
     trackConfigs &&
-      Object.keys(trackConfigs).forEach((key) => {
-        const trackId = key;
-
-        const config = trackConfigs[key];
+      Object.keys(trackConfigs).forEach((trackId) => {
+        const config = trackConfigs[trackId];
 
         config.showTrackName
           ? trackStateForNames.on.push(trackId)
--- a/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
@@ -175,10 +175,7 @@ const useGenomeBrowser = () => {
 
     trackConfigs &&
       Object.keys(trackConfigs).forEach((key) => {
-        let trackId = key;
-        if (trackId.match('focus')) {
-          trackId = 'focus';
-        }
+        const trackId = key;
 
         const config = trackConfigs[key];
 
--- a/src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
+++ b/src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
@@ -134,7 +134,7 @@ export const getTrackType = (trackId: string) => {
     return null;
   }
 
-  if (trackId.startsWith('gene')) {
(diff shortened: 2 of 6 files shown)


src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
```diff
@@ -134,7 +134,7 @@ export const getTrackType = (trackId: string) => {
     return null;
   }
 
-  if (trackId.startsWith('gene')) {
+  if (trackId.startsWith('gene') || trackId.startsWith('focus')) {
```
ens-ma7, line 137: should it be `.startsWith` or `.match`?

azangru: startsWith is good for the first part of the if; but the second part of the if should be just an equality comparison (`trackId === 'focus'`)

ens-ma7: sorry i meant the changes that was made (the second part) and yeah equality is better. Also whats the reason we are keeping the first condition `trackId.startsWith('gene')` since we have replaced all occurence of `gene-*`

azangru (approved):

src/content/app/genome-browser/state/track-config/trackConfigSlice.ts
azangru: > Also whats the reason we are keeping the first condition trackId.startsWith('gene') since we have replaced all occurence of gene-*

We only have full control over the name of the focus track. The if-condition above is checking for both the focus track and all the other gene-and-transcripts tracks that are coming from the track api:

image

We talked about changing the way of how we identify the type of a track, because track ids will stop being meaningful and human-readable in the foreseeable future. But that's for the future and will need proper coordination with the BE team.

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -175,10 +175,7 @@ const useGenomeBrowser = () => {
 
     trackConfigs &&
       Object.keys(trackConfigs).forEach((key) => {
-        let trackId = key;
-        if (trackId.match('focus')) {
-          trackId = 'focus';
-        }
+        const trackId = key;
```
ens-ma7, line 178: just a minor thing, we dont need to assign `const trackId= key` if we are removing the `if` statement and the `.match`. we could probably write it this way:
`Object.keys(trackConfigs).forEach((trackId) => {` 
or if we want to go with `key` because on line 180 we are using key but then further down we are using trackId. I think to keep the code clean its better to keep it consistent.

jyothishnt: good catch. changed now. Thanks

ens-ma7 (approved):

PR #770, opened 2022-06-27
set min height for sequence view

ridwanamo:
## Description
See description in Jira, we need to set a min height for the sequence view since there is a huge empty space for smaller sequence. 

Note regarding code change:
It looks like the `minmax(170px, 400px)` in the parent wasn't working as it was defaulting to 400px and thats because the parent of the parent is a flex container. So i have set the min-height and max-height in the sequence div instead.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1604

## Deployment URL(s)
http://drawersequenceview-minheight.review.ensembl.org

## Views affected
Sequence view in Genome browser  drawer

### DIFF
--- a/src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
+++ b/src/content/app/genome-browser/components/drawer/components/sequence-view/DrawerSequenceView.scss
@@ -86,7 +86,6 @@
   grid-area: main-bottom;
   display: flex;
   justify-content: center;
-  padding-top: 110px;
 }
 
 .loadFailureContainer {


azangru:
Looks like a wrong positioning of the "Failed to get data" element in the container with the updated styles:

image

azangru:
Something to check with Andrea: how do we position the radio buttons given that a sequence can be short (as short as a single line)?

image

andrea-ebi:
> Looks like a wrong positioning of the "Failed to get data" element in the container with the updated styles
Think this just needs to move up:
.loadFailureContainer__DrawerSequenceView__BjOuL
padding-top: 0;

andrea-ebi:
- Something to check with Andrea: how do we position the radio buttons given that a sequence can be short (as short as a single line)?
Fine as is 
if people don't like the white space they can collapse the Sequence section

azangru (approved):

PR #771, opened 2022-06-27
Update transcript track eye behaviour

imransl:
## Description
Multiple transcripts will be available in the genome browser from `ensembl-genome-browser` 0.4.0 version onwards. So we need to update the behaviour of the transcript tracks to make sure it works after this. This PR updates the code to do this and also fixes the bug that allowed us to disable only one transcript at a time.

## Related JIRA Issue(s)
[ENSWBSITES-1594](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1594)

## Deployment URL(s)
http://update-transcript-eye-behaviour.review.ensembl.org

## Views affected
Genome browser

### DIFF
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
@@ -106,7 +106,7 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
     dispatch(
       updateObjectTrackStates({
         status: transcriptIds.length ? Status.SELECTED : Status.UNSELECTED,
-        transcripts: transcriptIds
+        transcriptIds
       })
     );
   };


src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -68,18 +67,55 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
+          ) {
+            return true;
+          } else {
+            return false;
+          }
+        });
+
+      updateTranscriptTracks(visibleTranscriptIds);
+    }
+  }, [sortedTranscripts]);
```
imransl, line 102: This gets called whenever an individual transcript track is toggled. Is there a better place to notify the genome browser on which transcripts need to be visible on load?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -68,18 +67,55 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
+        .map((transcript) => transcript.stable_id)
+        .filter((transcriptId) => {
+          if (
+            transcriptTrackStatuses[transcriptId] === Status.SELECTED ||
+            transcriptTrackStatuses[transcriptId] === undefined
+          ) {
+            return true;
+          } else {
+            return false;
+          }
```
azangru, line 97: Nitpick; but this function can just return what is being checked in the if-condition

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -68,18 +67,55 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
       trackId: GENE_TRACK_ID
     })
   );
-  const { toggleTrack, genomeBrowser } = useGenomeBrowser();
-
+  const browserGenomeTrackStates = useAppSelector(
+    getBrowserActiveGenomeTrackStates
+  );
+  const { toggleTrack, updateTranscriptTracks, genomeBrowser } =
```
azangru, line 73: > `updateTranscriptTracks`

Wondering about the name here. On Dan's side, there are no transcript tracks. Are we keeping the metaphor on our side, or do we want to change the names (e.g. to focusGeneTranscripts) to keep consistency with Dan's view of the world?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
azangru: You are re-declaring `sortedTranscripts` as a new array on line 79. I would expect this to mean that this useEffect will re-run every render cycle.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelTranscript.tsx
```diff
@@ -14,58 +14,55 @@
...
 const TrackPanelTranscript = (props: Props) => {
-  const { genomeId, transcript, focusObjectId, trackId } = props;
-  const activeGenomeId = useSelector(getBrowserActiveGenomeId);
-  const trackVisibilityStatus = useSelector((state: RootState) =>
-    getBrowserTrackState(state, {
+  const { genomeId, transcript, focusObjectId } = props;
+  const browserGenomeTrackStates = useAppSelector(
+    getBrowserActiveGenomeTrackStates
+  );
+  const trackVisibilityStatus = useAppSelector((state: RootState) =>
```
azangru, line 52: `useAppSelector` knows the type of the redux state; no need to specify it here

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelTranscript.tsx
```diff
@@ -96,7 +120,7 @@ const TrackPanelTranscript = (props: Props) => {
   return (
     <SimpleTrackPanelItemLayout
       visibilityStatus={trackVisibilityStatus}
-      onChangeVisibility={() => props.onChangeVisibility(trackVisibilityStatus)}
+      onChangeVisibility={() => onChangeVisibility()}
```
azangru, line 123: `() => onChangeVisibility()` is identical to just `onChangeVisibility`

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelTranscript.tsx
```diff
@@ -83,6 +80,33 @@ const TrackPanelTranscript = (props: Props) => {
...
+        ? Status.UNSELECTED
+        : Status.SELECTED;
+
+    const updatedTranscriptTracks = {
+      ...transcriptTrackStatuses,
+      [transcript.stable_id]: newStatus
+    };
+    const visibleTranscriptIds = Object.keys(updatedTranscriptTracks).filter(
+      Boolean
+    );
```
azangru, line 95: I am confused — why should `filter(Boolean)` work in this case? `Status.UNSELECTED` is a string.

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
```diff
@@ -54,7 +54,12 @@ export type TrackStates = {
 
 export type GenomeTrackStates = {
   commonTracks?: TrackStates;
-  objectTracks?: { [objctId: string]: TrackStates };
+  objectTracks?: { [objectId: string]: TrackStates };
+  transcriptTracks?: {
+    [objectId: string]: {
+      [transcriptId: string]: TrackActivityStatus;
+    };
+  };
```
azangru, line 62: This is something to think about — whether it's the appropriate way of representing the states of the transcripts of a focus gene. I wonder what alternative ways of representing them we can think of.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
imransl: Ah that's what I missed and was wondering why is it rerunning. Thanks for spotting this!

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -405,6 +406,26 @@ const useGenomeBrowser = () => {
...
+
+    if (trackIdToSend === 'focus' && !isTurnedOn) {
+      genomeBrowser?.send({
+        type: OutgoingActionType.SET_VISIBLE_TRANSCRIPTS,
+        payload: {
+          track_id: 'focus',
+          transcript_ids: []
+        }
+      });
+    }
```
azangru, line 418: I don't like this. `trackIdToSend === 'focus'` does not have to refer to focus gene; there will be other focus objects in the focus track as well. Plus, we could just fire the `updateTranscriptTracks` action from the track panel gene component.

azangru:
One thing to think about is whether we should disable all transcript eyes when the focus gene track is disabled; and enable all transcript eyes when the focus gene track is re-enabled; and whether this should be part of this PR. Something to confirm with Andrea and @ens-st3

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -63,67 +66,76 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
     })
   );
-  const { toggleTrack, genomeBrowser } = useGenomeBrowser();
-
+  const visibleTranscriptIds = useAppSelector(getVisibleTranscriptIds);
+  const { toggleTrack, updateTranscriptTracks, genomeBrowser } =
+    useGenomeBrowser();
   const dispatch = useAppDispatch();
 
+  let allTranscriptsInGene: TrackPanelTranscriptType[] | undefined; // eslint-disable-line
```
azangru, line 77: why is eslint disabled for this line?

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
```diff
@@ -52,9 +55,19 @@ export type TrackStates = {
   };
 };
 
+// we might need to add other types for variants and regulation?
+export type GeneTrackStates = {
+  [objectId: string]: {
+    gene: {
```
azangru, line 61: The `gene` field inside a focus object doesn't make sense. The focus object is a gene in this case.

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
```diff
@@ -52,9 +55,19 @@ export type TrackStates = {
...
+      status?: TrackActivityStatus;
+      transcripts?: string[];
+    };
+  };
+};
+
 export type GenomeTrackStates = {
   commonTracks?: TrackStates;
-  objectTracks?: { [objctId: string]: TrackStates };
+  objectTracks?: GeneTrackStates;
```
azangru, line 70: We are probably looking for something like this instead:

```ts
type FocusTrack = FocusGeneTrack; // will be a union with FocusRegionTrack, FocusVariantTrack, etc.

type FocusGeneTrack = {
  type: 'gene';
  status: TrackActivityStatus;
  transcripts: string[] | null
};

type GenomeTrackStates = {
  objectTracks?: {
    [objectId: string]: FocusTrack;
  }
};
```

tests/setup-jest.js
```diff
@@ -4,3 +4,8 @@ jest.mock('src/services/google-analytics');
 process.on('unhandledRejection', (reason, promise) => {
   console.log('unhandledRejection', reason, promise);
 });
+
+const { TextEncoder, TextDecoder } = require('util'); // eslint-disable-line @typescript-eslint/no-var-requires
```
azangru, line 8: Since this is an import, could you please move it to the top of the file?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
imransl: Because it gets reassigned at line number 138 and eslint keeps complaining of not reassigning it.

azangru: Why don't we assign an empty array to both `allTranscriptsInGene` and `sortedTranscripts`? Then you wouldn't need to run checks on `allTranscriptsInGene` to make sure that it exists.

imransl: This was something we did previously. However, that caused these variables to get reset all the time, which resulted in `useEffect` getting executed multiple times: https://github.com/Ensembl/ensembl-client/pull/771#discussion_r909306155

azangru: `const allTranscriptsInGene = currentData?.gene.transcripts ?? []`?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelTranscript.tsx
imransl: ✅

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -33,13 +36,15 @@ import TrackPanelItemsCount from './TrackPanelItemsCount';
...
-  TrackId
-} from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
+import { TrackId } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
+import type { TrackPanelTranscript as TrackPanelTranscriptType } from 'src/content/app/genome-browser/state/types/track-panel-gene';
 import type { RootState } from 'src/store';
 
 import styles from './TrackPanelItem.scss';
+import {
+  IncomingActionType,
+  ReportVisibleTranscriptsAction
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
imransl: That did the trick! ✅

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
imransl: ✅

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
azangru: Why does it need to be reassigned? I can also see the reassignment on line 136, which says ` allTranscriptsInGene = gene.transcripts;`; but on line 77 you are already assigning this value to the `allTranscriptsInGene` variable.

src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
```diff
@@ -265,8 +257,53 @@ const browserGeneralSlice = createSlice({
...
+
+      if (transcripts) {
+        state.trackStates = set(
+          `${activeGenomeId}.objectTracks.${activeFocusObjectId}.transcripts`,
+          transcripts,
+          state.trackStates
+        );
+      }
+
+      browserStorageService.saveTrackStates(state.trackStates);
```
azangru, line 287: Here and below — redux team urges people not to run side effects in reducers.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
imransl: Yes that's true. Will remove line 136.

imransl:
@azangru @msmanoj Do you think it is safe to remove the production flag from `TrackPanelGene.tsx`?
https://github.com/Ensembl/ensembl-client/pull/771/files/c892765dd19677b2dafb9254ba736f761f422068#diff-37e74b832bf16f992fdb57930b90f9eb7f59ce085377d1d393c2947251a5416cR140

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
imransl: ✅

tests/setup-jest.js
imransl: ✅

src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
imransl: ✅

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -63,67 +66,74 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
+      (action: ReportVisibleTranscriptsAction) => {
+        setVisibleTranscriptIds(action.payload.transcript_ids);
+      }
+    );
 
-  const toggleExpand = () => {
-    setIsCollapsed(!isCollapsed);
-  };
+    return () => subscription?.unsubscribe();
+  }, [genomeBrowser, genomeId, focusObjectId]);
```
azangru, line 91: Only the `toggleTrack` in this useEffect depends on `genomeId` and `focusObjectId` (AND, on the trackStatus, I would have thought). The subscription to the genome browser, on the other hand, depends only on the presence of the genome browser. These should be two different useEffects.

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
```diff
@@ -24,7 +24,10 @@ export enum TrackItemColour {
 
 export type TrackItemColourKey = keyof typeof TrackItemColour;
 
-export type TrackActivityStatus = Status.SELECTED | Status.UNSELECTED;
+export type TrackActivityStatus =
+  | Status.SELECTED
+  | Status.UNSELECTED
+  | Status.PARTIALLY_SELECTED;
```
azangru, line 30: This is no longer correct, right? We aren't storing `Status.PARTIALLY_SELECTED` in redux.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
imransl: ✅

src/content/app/genome-browser/components/track-panel/trackPanelConfig.ts
imransl: ✅

src/content/app/genome-browser/state/browser-general/browserGeneralSelectors.ts
```diff
@@ -54,31 +54,45 @@ export const getBrowserTrackState = (
...
-        ];
+      ? allBrowserTrackStates?.[genomeId]?.[tracksGroup]?.[params.objectId]
+          ?.status
+      : allBrowserTrackStates?.[genomeId]?.[tracksGroup]?.[
+          params.categoryName
+        ]?.[params.trackId];
   return savedTrackStatus ?? Status.SELECTED;
 };
 
+export const getVisibleTranscriptIds = (state: RootState) => {
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -63,67 +66,77 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
+
+    return () => subscription?.unsubscribe();
   }, [genomeBrowser]);
 
-  if (!currentData) {
-    return null;
-  }
+  useEffect(() => {
+    updateObjectTrackStatus(trackStatus);
+  }, [genomeId, focusObjectId]);
```
azangru, line 96: `trackStatus` also a dependency?

azangru (approved):

src/content/app/genome-browser/state/browser-general/browserGeneralSelectors.ts
```diff
@@ -54,28 +54,27 @@ export const getBrowserTrackState = (
...
-          categoryName
-        ]?.[trackId]
-      : allBrowserTrackStates?.[genomeId]?.[tracksGroup]?.[categoryName]?.[
-          trackId
-        ];
+      ? allBrowserTrackStates?.[genomeId]?.[tracksGroup]?.[params.objectId]
+          ?.status
+      : allBrowserTrackStates?.[genomeId]?.[tracksGroup]?.[
+          params.categoryName
+        ]?.[params.trackId];
```
msmanoj, line 77: Should we consider replacing this with `if else` to make it more readable?

imransl: ✅

src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
```diff
@@ -152,6 +145,55 @@ export const updateBrowserActiveFocusObjectIdsAndSave = (
   };
 };
 
+export const updateObjectTrackStates: ActionCreator<
+  ThunkAction<void, any, void, Action<string>>
+> = (payload: { status: Status; transcripts?: string[] }) => {
```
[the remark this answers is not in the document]
imransl: ✅

msmanoj (approved):

PR #772, opened 2022-07-01
Migrate GA to GA4

msmanoj:
## TODO:
- Ask Kamal to update the Google Analytics key in the deployment env to `G-MB6YFL24VC` before merging

## Description
This PR is to migrate from universal analytics to GA4 by doing the following:

- Updated the [loader script](https://github.com/Ensembl/ensembl-client/blob/dev/src/services/google-analytics/loadGoogleAnalytics.ts) to load gtag.js instead of analytics.js
- Updated the api for sending custom events to google

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1638

## Deployment URL(s)
http://ga4.review.ensembl.org

### DIFF
--- a/src/content/app/species-selector/hooks/useSpeciesSelectorAnalytics.ts
+++ b/src/content/app/species-selector/hooks/useSpeciesSelectorAnalytics.ts
@@ -25,8 +25,8 @@ import type {
   PopularSpecies,
   SearchMatch
 } from 'src/content/app/species-selector/types/species-search';
-import { type CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
-import { type AnalyticsOptions } from 'src/analyticsHelper';
+import type { CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import type { AnalyticsOptions } from 'src/analyticsHelper';
 
 const useSpeciesSelectorAnalytics = () => {
   const committedSpecies = useSelector(getCommittedSpecies);
--- a/src/content/app/species-selector/hooks/useSpeciesSelectorAnalytics.ts
+++ b/src/content/app/species-selector/hooks/useSpeciesSelectorAnalytics.ts
@@ -18,23 +18,31 @@ import { useSelector } from 'react-redux';
 import analyticsTracking from 'src/services/analytics-service';
 
 import { getSpeciesAnalyticsName } from 'src/content/app/species-selector/speciesSelectorHelper';
-
 import { getCommittedSpecies } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
 
-import {
+import { AppName } from 'src/global/globalConfig';
+import type {
   PopularSpecies,
   SearchMatch
 } from 'src/content/app/species-selector/types/species-search';
-import { CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import { type CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import { type AnalyticsOptions } from 'src/analyticsHelper';
 
 const useSpeciesSelectorAnalytics = () => {
   const committedSpecies = useSelector(getCommittedSpecies);
+
+  const trackEvent = (ga: AnalyticsOptions) => {
+    analyticsTracking.trackEvent({
+      ...ga,
+      app: AppName.SPECIES_SELECTOR
+    });
+  };
 
   /*  Species Selector Page Events */
   const trackAutocompleteSpeciesSelect = (species: SearchMatch) => {
     const speciesNameForAnalytics = getSpeciesAnalyticsName(species);
 
-    analyticsTracking.trackEvent({
+    trackEvent({
       category: 'species_search',
       action: 'preselect',
       label: speciesNameForAnalytics,
@@ -48,7 +56,7 @@ const useSpeciesSelectorAnalytics = () => {
   ) => {
     const speciesNameForAnalytics = getSpeciesAnalyticsName(species);
 
-    analyticsTracking.trackEvent({
+    trackEvent({
       category: 'popular_species',
       action: action,
       label: speciesNameForAnalytics,
@@ -59,7 +67,7 @@ const useSpeciesSelectorAnalytics = () => {
   const trackCommittedSpecies = (species: CurrentItem) => {
     const speciesNameForAnalytics = getSpeciesAnalyticsName(species);
 
-    analyticsTracking.trackEvent({
+    trackEvent({
       category: 'species_selector',
       label: speciesNameForAnalytics,
       action: 'add',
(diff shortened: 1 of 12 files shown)


src/server/routes/unsupportedBrowserRouter.tsx
```diff
@@ -67,14 +67,15 @@ const unsupportedBrowserRouter = (_: Request, res: Response) => {
...
   process.env.GOOGLE_ANALYTICS_KEY
-    ? `<script>
-    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-    ga('create', '${process.env.GOOGLE_ANALYTICS_KEY}', 'auto');
-    ga('send', 'pageview');
-  </script>`
+    ? `<script async src="https://www.googletagmanager.com/gtag/js?id=${process.env.GOOGLE_ANALYTICS_KEY}"></>
```
azangru, line 70: `</>` -> `</script>`

src/services/analytics-service.ts
```diff
@@ -39,50 +39,61 @@ class AnalyticsTracking {
...
+      page_path: pagePath
+    });
   }
 
   // Track an event
-  public trackEvent(ga: AnalyticsOptions) {
-    ga.species && this.setSpeciesDimension(ga.species);
-    ga.app && this.setAppDimension(ga.app);
-    ga.feature && this.setFeatureDimension(ga.feature);
+  public trackEvent(gtag: AnalyticsOptions) {
```
azangru, line 56: both the `gtag` and the previously used `ga` are a strange name for the parameter of this function. I suggest we call it either `params`, or `options`, or `event`, or something similar.

src/services/google-analytics/index.ts
```diff
@@ -16,46 +16,79 @@
...
   static initialize(googleAnalyticsKey: string) {
     loadGoogleAnalytics(googleAnalyticsKey);
+    this.googleAnalyticsKey = googleAnalyticsKey;
   }
 
-  static ga(...args: any[]) {
-    window.ga(...args);
+  static gtag(...args: any[]) {
+    window.gtag(...args);
   }
```
azangru, line 53: I don't know if we are using this method at all; but as we learned, gtag doesn't accept an array of arguments; it wants the actual `Arguments` object

src/services/google-analytics/index.ts
```diff
@@ -16,46 +16,79 @@
...
 
-  // referencee: example from Google's docs for analytics.js
-  // https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#hitType
-  static pageview(pagePath: string) {
-    window.ga('send', {
-      hitType: 'pageview',
-      page: pagePath
+  // referencee: example from Google's docs for gtag.js
```
azangru, line 55: referencee -> reference ? (remnants from my broken keyboard on the MacbookAir)?

src/services/google-analytics/index.ts
```diff
@@ -16,46 +16,75 @@
...
-      hitType: 'pageview',
-      page: pagePath
+  // reference: example from Google's docs for gtag.js
+  // https://developers.google.com/analytics/devguides/collection/gtagjs/pages
+  static pageview(params: TrackPageView) {
+    window.gtag('event', 'page_view', {
+      send_to: this.googleAnalyticsKey,
+      ...params
     });
   }
```
azangru, line 58: There are slightly different instructions for single-page applications [here](https://developers.google.com/analytics/devguides/collection/gtagjs/single-page-applications#measure_virtual_pageviews).

Also, is the `send_to` option really necessary? We are always using the same ga measurement id

azangru: `...params` no longer makes any sense; could as well be just `params`

src/services/google-analytics/index.ts
```diff
@@ -16,46 +16,75 @@
...
+      options['value'] = params.event_label;
+    }
+
+    window.gtag('event', params.event_action, options);
+  }
+
+  static setDimension(params: SetDimension) {
+    window.gtag('event', params.event_name, {
+      [params.dimension_key]: params.dimension_value
+    });
```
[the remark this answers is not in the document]
msmanoj: I have removed the dedicated methods to set and clear the dimensions and they are now passed along with the events. However, this won't be the case for genome browser events which I think it can be updated when we work on updating the GB analytics.

src/services/google-analytics/index.ts
msmanoj: That's done

src/content/app/species-selector/hooks/useSpeciesSelectorAnalytics.ts
```diff
@@ -18,23 +18,31 @@ import { useSelector } from 'react-redux';
...
 
-import {
+import { AppName } from 'src/global/globalConfig';
+import type {
   PopularSpecies,
   SearchMatch
 } from 'src/content/app/species-selector/types/species-search';
-import { CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import { type CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import { type AnalyticsOptions } from 'src/analyticsHelper';
```
azangru, line 29: here and below `import { type  }` -> `import type`?

`import { type  }` makes sense if you are also importing values from the same module; but if you are only importing types, then `import type` is more idiomatic

src/services/google-analytics/loadGoogleAnalytics.ts
```diff
@@ -49,6 +49,12 @@ const loadGoogleAnalytics = (trackerId: string) => {
 // until the real google analytics object is ready to replace it
 const createGAShim = (trackerId: string) => {
   window.gtag = function () {
+    // Cleanup the dataLayer array after every 100 events
+    if (window.dataLayer.length > 100) {
+      window.dataLayer.push(function (this: any) {
+        this.reset();
+      });
+    }
```
azangru, line 57: 1) The function should not take `this` as an argument
2) Is this any better than linking the cleanup of the data layer to the navigation from one app to another?

azangru: I found the part of the gtag script that prevents the infinite growth of the queue:

image

Gtag allows the queue to grow up to a thousand elements, and then for every new push, removes the oldest element(s) from the queue so that its size never exceeds 1000 elements. I don't know why they allow the queue to get to 1000; but in any case, I now have proof that the queue will not grow indefinitely.

Let's delete our cleanup code then. Sorry I brought this up.

azangru (approved):

src/services/google-analytics/loadGoogleAnalytics.ts
msmanoj: Thanks for sharing that. It's good to get these checked anyways.

PR #773, opened 2022-07-01
Spacing adjustments

jyothishnt:
## Description
Andrea has created a spec for the recurring distances on the site:

https://www.ebi.ac.uk/seqdb/confluence/display/ENSWEB/Global+site+structure+specification

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1602

## Deployment URL(s)
http://global-spacing.review.ensembl.org

## Views affected
Global

### DIFF
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
@@ -75,7 +75,7 @@ export const TrackPanelList = () => {
       {activeFocusObject?.type === 'gene' &&
       activeGenomeId &&
       activeFocusObject.stable_id ? (
-        <section className="mainTrackItem">
+        <section className={`${styles.mainTrackItem}`}>
           <TrackPanelGene
             focusObjectId={activeFocusObject.object_id}
             genomeId={activeGenomeId}
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
@@ -75,7 +75,7 @@ export const TrackPanelList = () => {
       {activeFocusObject?.type === 'gene' &&
       activeGenomeId &&
       activeFocusObject.stable_id ? (
-        <section className={`${styles.mainTrackItem} mainTrackItem`}>
+        <section className="mainTrackItem">
           <TrackPanelGene
             focusObjectId={activeFocusObject.object_id}
             genomeId={activeGenomeId}
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
@@ -75,7 +75,7 @@ export const TrackPanelList = () => {
       {activeFocusObject?.type === 'gene' &&
       activeGenomeId &&
       activeFocusObject.stable_id ? (
-        <section className="mainTrackItem">
+        <section className={`${styles.mainTrackItem} mainTrackItem`}>
           <TrackPanelGene
             focusObjectId={activeFocusObject.object_id}
             genomeId={activeGenomeId}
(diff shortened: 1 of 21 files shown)


src/header/launchbar/Launchbar.scss
```diff
@@ -5,7 +5,11 @@
   justify-content: space-between;
   border-bottom: 1px solid $grey;
   border-top: 1px solid $grey;
-  padding: 11px 6px;
+  // padding: 11px 6px;
```
azangru, line 8: ❓

src/shared/components/app-bar/AppBar.scss
```diff
@@ -8,7 +8,9 @@
     'top top'
     'main right';
   align-items: center;
-  padding: 8px 20px;
+  padding: 8px 0px;
```
azangru, line 11: this should be changed either to `padding-top`, or to `padding: 8px $page-right-padding 0 $page-left-padding`;

src/header/launchbar/Launchbar.scss
```diff
@@ -45,6 +49,10 @@
   width: 32px;
   margin: 0 12px;
 
+  &:first-child {
+    margin-left: 0px;
+  }
+
```
azangru, line 55: Something in the launchbar styles have pushed the buttons to the left of center in between the separators. Compare internal 2020:

image

with the review deployment:

image

azangru:
image

azangru:
Could you update the description of the PR, please? It looks copied from the Jira ticket; but now that you've actually written the code, the last paragraph, which discusses possible approaches, looks out of place:

image

jyothishnt:
The last fix was a minor change in right padding for the RH panel.
Andrea asked to make the below change to padding. 
sidebar__StandardAppLayout__Y3IBc {
    padding: 21px 12px;
}

src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
```diff
@@ -75,7 +75,7 @@ export const TrackPanelList = () => {
       {activeFocusObject?.type === 'gene' &&
       activeGenomeId &&
       activeFocusObject.stable_id ? (
-        <section className="mainTrackItem">
+        <section className={`${styles.mainTrackItem} mainTrackItem`}>
```
azangru, line 78: What's happening here?

If the `mainTrackItem` class that is in the deleted line was used in a test, then you no longer need it (`styles.mainTrackItem` will evaluate to `mainTrackItem` in the test). If it was used for some other purpose, then what was it?

jyothishnt: It was just for the test. I didnt know that it would evaluate the same in test. Will update.

azangru:
Something broken here

image

src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.tsx
azangru: The change you made should have been the opposite. Not this:

image

but this:

` <section className={`${styles.mainTrackItem}`}>`

azangru:
Also, after speaking with Andrea — she doesn't like the horizontal shift of the content in the right-hand panel of the genome browser that occurs when the scrollbar appears or hides, such as in this example:

https://user-images.githubusercontent.com/6834224/178374351-bafab97b-42c6-4b9b-ac4b-0af7f042784e.mov

To avoid the horizontal shift, we will probably need to introduce a special wrapper (I built a small [proof-of-concept](https://jsfiddle.net/a648w7u3) in jsfiddle)

azangru:
Here's something that I'd like to discuss with you and Andrea, and maybe Anne:

- Space 6 in [Andrea's diagram](https://www.ebi.ac.uk/seqdb/confluence/display/ENSWEB/Global+site+structure+specification) has turned into two variables: `$page-left-padding`, and `$page-right-padding`.
  - Is there any point in having them as two different variables when they have the same value
  - Is the variable appropriately named? Both you and Andrea are using these spaces not just as a page padding
- Space 5 in [Andrea's diagram](https://www.ebi.ac.uk/seqdb/confluence/display/ENSWEB/Global+site+structure+specification) has not been implemented anywhere.

jyothishnt:
We can discuss about space. 6.
Space 5 is implemented in EV.

PR #775, opened 2022-07-06
Zmenu bugfixes

ens-ma7:
## Description
Two bug fix related to ZMenu
1639 - Move dispatch into useEffect but had to rearrange code so that useEffect is before any return
1640 - Close Zmenu when dragging genome browser...Use a different approach as suggested in the jira as after speaking to dan there is a possibility that the genome browser can be moved by keyboard or some other actions and the best way to capture this is to check if the location is changing (meaning the genome browser is moving) then close the zmenu.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1639
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1640

## Deployment URL(s)
http://zmenu-bugfix.review.ensembl.org

## Views affected
Genome browser

### DIFF
--- a/src/content/app/genome-browser/components/zmenu/Zmenu.test.tsx
+++ b/src/content/app/genome-browser/components/zmenu/Zmenu.test.tsx
@@ -110,13 +110,14 @@ const defaultProps: ZmenuProps = {
 
 describe('<Zmenu />', () => {
   describe('rendering', () => {
-    test('renders zmenu content', () => {
+    it('renders zmenu content', () => {
       const { queryByTestId } = renderComponent();
       expect(queryByTestId('zmenuContent')).toBeTruthy();
     });
   });
-  describe('destroying ZMenu', () => {
-    it('closes zmenu when location change', () => {
+
+  describe('behaviour', () => {
+    it('sends a signal to be closed when genome browser’s location changes', () => {
       const { store } = renderComponent();
       expect(mockSetZmenus).not.toHaveBeenCalled();
       act(() => {
@@ -128,7 +129,7 @@ describe('<Zmenu />', () => {
           ])
         );
       });
-      expect(mockSetZmenus).toHaveBeenCalledWith({});
+      expect(mockSetZmenus).toHaveBeenCalledWith({}); // the new zmenu object will not have the id of the current zmenu
     });
   });
 });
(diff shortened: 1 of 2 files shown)


src/content/app/genome-browser/components/zmenu/Zmenu.tsx
```diff
@@ -54,8 +56,19 @@ export type ZmenuProps = {
...
   const anchorRef = useRefWithRerender<HTMLDivElement>(null);
   const { zmenus, setZmenus } = useGenomeBrowser();
+  const actualChrLocation = useAppSelector(getActualChrLocation);
+  const [renderCount, setRenderCount] = useState(0);
+
   const dispatch = useAppDispatch();
 
+  useEffect(() => {
+    setRenderCount(renderCount + 1);
+    // Do not try to close the Zmenu on first render
```
azangru, line 66: omg, this is so ugly! let's see if the `usePrevious` hook helps:

```ts
// no need to call it actual here
const chromosomeLocation = useAppSelector(getActualChrLocation);
const previousChromosomeLocation = usePrevious(chromosomeLocation)

useEffect(() => {
  if (chromosomeLocation.toString() !== previousChromosomeLocation.toString()) {
    destroyZmenu();
  }
}, [chromosomeLocation])
```

src/content/app/genome-browser/components/zmenu/Zmenu.test.tsx
```diff
@@ -50,6 +53,21 @@ jest.mock('./ZmenuInstantDownload', () => () => (
...
+
+const mockState = {
+  browser: {
+    browserGeneral: {
+      actualChrLocations: {
+        human: [chrName, startPosition, endPosition] as ChrLocation
+      }
+    }
+  }
+};
```
azangru, line 69: Well, now you can write a proper test case for the logic that you added 😄 

- replace the mock redux store with the real redux store
- dispatch an event that changes the chromosome location of human
- confirm that the zmenu was destroyed by observing the consequences that you can control (`setZmenus` would be called on `useGenomeBrowser`)

src/content/app/genome-browser/components/zmenu/Zmenu.tsx
msmanoj: That was my idea 😃 
`usePrevious` should work as well.

azangru (approved):

PR #784, opened 2022-07-25
Fix server-side status codes for EntityViewer

azangru:
## Description
Discovered the following bugs:
- Requests sent to Entity Viewer urls with an invalid genome id, but without an entity id return a response with a 200:OK instead of a 404 status. Example: https://staging-2020.ensembl.org/entity-viewer/grc
- Requests sent to Entity Viewer urls with an improperly formatted entity id responds with a 500 status code rather than a 404. Example: https://staging-2020.ensembl.org/entity-viewer/grch38/foo

## Deployment URL(s)
http://fix-entity-viewer-ssr.review.ensembl.org

## Views affected
Enitty Viewer

### DIFF
--- a/src/content/app/entity-viewer/EntityViewerPage.tsx
+++ b/src/content/app/entity-viewer/EntityViewerPage.tsx
@@ -71,9 +71,10 @@ export const serverFetch: ServerFetch = async (params) => {
   const dispatch: AppDispatch = store.dispatch;
   const { genomeId: genomeIdFromUrl, entityId } = getPathParameters<
     'genomeId' | 'entityId'
-  >('/entity-viewer/:genomeId/:entityId', path);
+  >(['/entity-viewer/:genomeId', '/entity-viewer/:genomeId/:entityId'], path);
 
-  if (!(genomeIdFromUrl && entityId)) {
+  // If the url is just /entity-viewer, there is nothing more to do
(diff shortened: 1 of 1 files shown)


src/content/app/entity-viewer/EntityViewerPage.tsx
```diff
@@ -83,21 +84,30 @@ export const serverFetch: ServerFetch = async (params) => {
...
-    return; // this shouldn't happen
+  // If the url is /entity-viewer/:genomeId, there is nothing more to do
+  if (!entityId) {
+    return;
   }
 
-  // FIXME: gene symbol parsing can explode!
-  const geneStableId = parseFocusIdFromUrl(entityId).objectId;
+  // NOTE: we will have to be smarter here when entities are no longer just genes
+  let geneStableId;
```
imransl, line 101: Can't this be named as `objectStableId` or just `stableId` instead?

azangru: It won't help much — down below, we are passing this id to the `fetchGenePageMeta.initiate` function, which fetches a gene from Thoas. When we get more entities than just gene, we will have to be able to choose which function to call for different entity types.

imransl (approved):

msmanoj (approved):

PR #785, opened 2022-07-25
Remove the 'track:' prefix from track ids

azangru:
## Description
After the track api has updated track ids to remove the prefix `track:` from them (see https://github.com/Ensembl/ensembl-track-api/pull/9), in this PR we are deleting the code that was responsible for the removal of this prefix before sending the track id to the genome browser.

Also, updated types to remove the outdated `FocusObjectTrack` type.

**NOTE:** release to be coordinated with Andres

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1623

## Deployment URL(s)
http://updated-track-ids.review.ensembl.org

## Views affected
Genome browser

### DIFF
--- a/src/server/middleware/proxyMiddleware.ts
+++ b/src/server/middleware/proxyMiddleware.ts
@@ -56,31 +56,13 @@ const staticAssetsMiddleware = createProxyMiddleware('/static', {
   target: 'http://localhost:8081'
 });
 
-// const apiProxyMiddleware = createProxyMiddleware('/api', {
-//   target: 'https://staging-2020.ensembl.org',
-//   changeOrigin: true,
-//   secure: false
-// });
-
-const apiProxyMiddleware = createProxyMiddleware(
-  ['/api/**', '!/api/tracks/**'],
-  {
-    target: 'https://staging-2020.ensembl.org',
-    changeOrigin: true,
-    secure: false
-  }
-);
-
-const tracksProxyMiddleware = createProxyMiddleware('/api/tracks/**', {
-  target: 'http://change-track-ids.review.ensembl.org',
-  // pathRewrite: {
-  //   '^/api/docs': '/api', // rewrite path
-  // },
+const apiProxyMiddleware = createProxyMiddleware('/api', {
+  target: 'https://staging-2020.ensembl.org',
   changeOrigin: true,
   secure: false
 });
 
-let proxyMiddleware = [apiProxyMiddleware, tracksProxyMiddleware];
+let proxyMiddleware = [apiProxyMiddleware];
 
 if (process.env.NODE_ENV === 'development') {
   proxyMiddleware = proxyMiddleware.concat([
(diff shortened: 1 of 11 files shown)


src/server/middleware/proxyMiddleware.ts
```diff
@@ -56,13 +56,31 @@ const staticAssetsMiddleware = createProxyMiddleware('/static', {
   target: 'http://localhost:8081'
 });
 
-const apiProxyMiddleware = createProxyMiddleware('/api', {
-  target: 'https://staging-2020.ensembl.org',
+// const apiProxyMiddleware = createProxyMiddleware('/api', {
+//   target: 'https://staging-2020.ensembl.org',
+//   changeOrigin: true,
+//   secure: false
+// });
```
imransl, line 63: Can this be removed?

src/server/middleware/proxyMiddleware.ts
```diff
@@ -56,13 +56,31 @@ const staticAssetsMiddleware = createProxyMiddleware('/static', {
...
+    secure: false
+  }
+);
+
+const tracksProxyMiddleware = createProxyMiddleware('/api/tracks/**', {
+  target: 'http://change-track-ids.review.ensembl.org',
+  // pathRewrite: {
+  //   '^/api/docs': '/api', // rewrite path
+  // },
   changeOrigin: true,
```
imransl, line 79: Does this need to be left here?

src/server/middleware/proxyMiddleware.ts
```diff
@@ -56,13 +56,31 @@ const staticAssetsMiddleware = createProxyMiddleware('/static', {
...
+    changeOrigin: true,
+    secure: false
+  }
+);
+
+const tracksProxyMiddleware = createProxyMiddleware('/api/tracks/**', {
+  target: 'http://change-track-ids.review.ensembl.org',
+  // pathRewrite: {
+  //   '^/api/docs': '/api', // rewrite path
+  // },
```
imransl, line 78: Does this need to be left here?

src/server/middleware/proxyMiddleware.ts
azangru: Ouch, forgot the dev code. I've reverted the changes to the `proxyMiddleware` file.

msmanoj:
PR looks good. I am just wondering what will happen to the existing values store in the local storage?

azangru:
> PR looks good. I am just wondering what will happen to the existing values store in the local storage?

🤷‍♂️ 

I think the overall feeling in the team is that we shouldn't yet worry about this much. But if you can think about how we would be dealing with all that if we had a production-grade service, I'll be interested to hear that 😃

msmanoj:
> > PR looks good. I am just wondering what will happen to the existing values store in the local storage?
> 
> 🤷‍♂️
> 
> I think the overall feeling in the team is that we shouldn't yet worry about this much. But if you can think about how we would be dealing with all that if we had a production-grade service, I'll be interested to hear that 😃

You are right that we shouldn't worry about it much right now. I don't think there is any simple solution for this at the moment.

msmanoj (approved):

imransl (approved):

jyothishnt (approved):
Look ok to me. Not worried about the localstorage entries.

PR #786, opened 2022-07-27
blast error fix - reset sequence validity flags on sequences change

ens-ma7:
## Description
See jira ticket on how to reproduce bug. The issue was that the sequenceValidityFlags array wasn't getting clear properly when sequence boxes was deleted. 2 ways we can delete a sequence box:
1. click on the delete icon
2. delete all sequences in a box and then click outside box

The previous approach was using `useEffect` on unmounting the component but it was getting confused with the index of which box was getting removed and there was no way to find out the index of the sequence box being removed.

The solution that i choose which might not be the perfect one (but still better than rewriting the whole thing) is to have the sequences object in redux (the committed sequences) within a useEffect and everytime the object change, we revalidate all the sequences.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1654

## Deployment URL(s)
http://blast-error-fix.review.ensembl.org

## Views affected
blast - http://blast-error-fix.review.ensembl.org/blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
@@ -73,7 +73,6 @@ const BlastInputSequences = () => {
     } else if (typeof index === 'number') {
       const newSequences = [...sequences].filter((_, i) => i !== index);
       updateSequences(newSequences);
-      blastFormContext?.removeSequenceValidityFlag(index);
     }
   };
 
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
@@ -17,8 +17,6 @@
 import React, { useEffect, useContext } from 'react';
 import { useAppSelector } from 'src/store';
 
-import usePrevious from 'src/shared/hooks/usePrevious';
-
 import BlastFormContext from 'src/content/app/tools/blast/views/blast-form/BlastFormContext';
 
 import { getEmptyInputVisibility } from 'src/content/app/tools/blast/state/blast-form/blastFormSelectors';
@@ -46,15 +44,11 @@ const BlastInputSequences = () => {
 
   const blastFormContext = useContext(BlastFormContext);
 
-  const previousSequences = usePrevious(sequences);
-
   useEffect(() => {
-    if (previousSequences?.length !== sequences.length) {
-      const sequenceFlags = sequences.map((sequence) => {
-        return checkSequenceValidity(sequence.value, sequenceType);
-      });
-      blastFormContext?.setSequenceValidityFlags(sequenceFlags);
-    }
+    const sequenceFlags = sequences.map((sequence) => {
+      return checkSequenceValidity(sequence.value, sequenceType);
+    });
+    blastFormContext?.setSequenceValidityFlags(sequenceFlags);
   }, [sequences]);
 
   const onSequenceAdded = (input: string, index: number | null) => {
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
@@ -50,8 +50,6 @@ const BlastInputSequences = () => {
 
   useEffect(() => {
     if (previousSequences?.length !== sequences.length) {
-      blastFormContext?.clearSequenceValidityFlags();
-
       const sequenceFlags = sequences.map((sequence) => {
         return checkSequenceValidity(sequence.value, sequenceType);
       });
--- a/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
+++ b/src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
@@ -17,6 +17,8 @@
 import React, { useEffect, useContext } from 'react';
 import { useAppSelector } from 'src/store';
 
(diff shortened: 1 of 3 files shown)


jyothishnt (approved):

ens-ma7:
I have added an `if` statement in the useeffect which will hopefully make the state update a bit less. But the fact that the blast input form is complicated already (there are many ways you can add/remove sequence box), its making the code for this check a bit ugly. I still think this is better than rewriting the code. We can revise this code again when it stop working or breaks. 

The other possibility that a sequence box can be created is having sequences separated by break line for example:
```AAA

AAA
```

which means the check needs to be for all sequences that get committed.

The way to look at this is we have 2 ways to update the states one which is when typing in the each individual sequence box which is only updating the state for that specific box and the other one is when adding or deleting a box which because there are different ways that this can happen, it makes sense to revalidate all boxes everytime the sequences box change.

azangru:
The big question about this PR for me is how upset we are that the same sequence now gets run through the `checkSequenceValidity` checker. You will notice that:

1) `checkSequenceValidity` is executed in `BlastInputSequences.tsx` every time sequences change
2) It is then executed again in `BlastInputSequence.tsx` every time BlastInputSequence receives a new sequence from BlastInputSequences.

On the one hand, checking a string must be a very fast operation that should take next to no time (we can actually profile this 😃 ). On the other hand, I wonder whether this signals that we should do a big refactor and move all the code responsible for sequences in the form to redux. I didn't want to do this initially to avoid unnecessary rerenders; but maybe I was wrong and keeping all sequences, even the ones that the user is editing, in redux won't be too bad.

I'm very uncertain on this.

ens-ma7:
> The big question about this PR for me is how upset we are that the same sequence now gets run through the `checkSequenceValidity` checker. You will notice that:
> 
> 1. `checkSequenceValidity` is executed in `BlastInputSequences.tsx` every time sequences change
> 2. It is then executed again in `BlastInputSequence.tsx` every time BlastInputSequence receives a new sequence from BlastInputSequences.

To answer your concern above, we do need to check the sequence again because if you look at my example above your comment. Take this example sequence:

```
AAA

AAA
```

This is a valid sequence for one input box meaning `checkSequenceValidity` returns true for `BlastInputSequence.tsx`. And on committing this sequence (clicking outside the sequence box), it gets split in 2 different sequence box with each sequence box now being invalid and this is executed by `checkSequenceValidity` in `BlastInputSequences.tsx`

By now going through this and explaining all of it, i am now wondering if even moving it to redux will be the solution for it since the issue is we have too many ways of creating/deleting sequence boxes.

azangru:
>  And on committing this sequence (clicking outside the sequence box), it gets split in 2 different sequence box with each sequence box now being invalid and this is executed by checkSequenceValidity in BlastInputSequences.tsx

I am confused. Isn't each of these `AAA` sequences then passed into its own BlastInputSequence component, and validated there as well?

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
```diff
@@ -35,8 +41,24 @@ const BlastInputSequences = () => {
...
+
   const shouldAppendEmptyInput = useAppSelector(getEmptyInputVisibility);
 
+  const blastFormContext = useContext(BlastFormContext);
+
+  const previousSequences = usePrevious(sequences);
+
+  useEffect(() => {
+    if (previousSequences?.length !== sequences.length) {
+      blastFormContext?.clearSequenceValidityFlags();
```
msmanoj, line 53: We don't need to clear it here. We will be anyways resetting it on L58 below.

msmanoj:
If we are considering to do a big refactor or to asses if we need to, I think it would be better have a separate ticket for that. 
If the changes in this PR fixes the bug and doesn't break anything else, I think we should review and merge it.

ens-ma7:
> > And on committing this sequence (clicking outside the sequence box), it gets split in 2 different sequence box with each sequence box now being invalid and this is executed by checkSequenceValidity in BlastInputSequences.tsx
> 
> I am confused. Isn't each of these `AAA` sequences then passed into its own BlastInputSequence component, and validated there as well?

Yes it does but the validitycheck wasn't being updated for the first sequence box. It was fine for the second one.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
ens-ma7: Good point... i have removed this line now.

src/content/app/tools/blast/components/blast-input-sequences/BlastInputSequences.tsx
```diff
@@ -35,8 +41,22 @@ const BlastInputSequences = () => {
...
+  const previousSequences = usePrevious(sequences);
+
+  useEffect(() => {
+    if (previousSequences?.length !== sequences.length) {
+      const sequenceFlags = sequences.map((sequence) => {
+        return checkSequenceValidity(sequence.value, sequenceType);
+      });
+      blastFormContext?.setSequenceValidityFlags(sequenceFlags);
+    }
+  }, [sequences]);
```
azangru, line 58: I think, in this case, we can just rely on the array of `[sequences]` to indicate the change when the useEffect should run. There shouldn't be any need for the additional `if (previousSequences?.length !== sequences.length)` check that we are seeing on line 52.

azangru (approved):

ens-ma7:
Created this ticket (https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1694) for the rewriting of the state in redux

PR #787, opened 2022-07-28
Update show hide

msmanoj:
## Description
- Updated the ShowHide component to accept a single `className` prop

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1664

## Deployment URL(s)
http://update-show-hide.review.ensembl.org

## Views affected

- External references
- GB drawer sequence download
- EV transcripts download and more information
- Help page `Show me what to do`

### DIFF
--- a/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -146,8 +146,6 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
     dispatch(setFilterPanel(!isFilterPanelOpen));
   };
 
-  const filterLabelClass = isFilterPanelOpen ? styles.openFilterLabel : '';
-
   useEffect(() => {
     if (!genomeId || !props.gene) {
       return;
@@ -186,9 +184,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
           {props.gene.transcripts.length > 5 && (
             <div className={styles.filterLabelWrapper}>
               <ShowHide
-                classNames={{
-                  label: filterLabelClass
-                }}
+                className={isFilterPanelOpen ? styles.showHide : undefined}
                 onClick={toggleFilterPanel}
                 isExpanded={isFilterPanelOpen}
                 label={filterLabel}
--- a/src/shared/components/show-hide/ShowHide.scss
+++ b/src/shared/components/show-hide/ShowHide.scss
@@ -1,6 +1,9 @@
 @import 'src/styles/common';
 
 .showHide {
+  cursor: pointer;
+  display: inline-block;
+
   // increase specificity to be sure to override chevron's own styles
   .chevron {
     margin-left: 10px;
@@ -9,7 +12,6 @@
 
   .label {
     color: var(--show-hide-label-color, $blue);
-    cursor: pointer;
     white-space: nowrap;
   }
 }
--- a/src/shared/components/show-hide/ShowHide.scss
+++ b/src/shared/components/show-hide/ShowHide.scss
@@ -1,15 +1,15 @@
 @import 'src/styles/common';
 
 .showHide {
-  // increase specificity to be sure to override chevron's own styles 
+  // increase specificity to be sure to override chevron's own styles
   .chevron {
     margin-left: 10px;
-    height: 6px;
+    --chevron-height: 6px;
   }
-}
 
-.label {
-  color: $blue;
-  cursor: pointer;
-  white-space: nowrap;
+  .label {
+    color: var(--show-hide-label-color, $blue);
+    cursor: var(--show-hide-label-cursor, pointer);
+    white-space: nowrap;
+  }
 }
--- a/src/shared/components/show-hide/ShowHide.scss
+++ b/src/shared/components/show-hide/ShowHide.scss
@@ -4,12 +4,12 @@
   // increase specificity to be sure to override chevron's own styles
   .chevron {
     margin-left: 10px;
-    --chevron-height: 6px;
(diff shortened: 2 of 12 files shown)


src/shared/components/show-hide/ShowHide.scss
```diff
@@ -1,15 +1,15 @@
 @import 'src/styles/common';
 
 .showHide {
-  // increase specificity to be sure to override chevron's own styles 
+  // increase specificity to be sure to override chevron's own styles
   .chevron {
     margin-left: 10px;
-    height: 6px;
+    --chevron-height: 6px;
```
azangru, line 7: We want this property to be overridable from the parent as well:

```
--chevron-height: var(--show-hide-chevron-height, 6px);
```

src/shared/components/show-hide/ShowHide.scss
```diff
@@ -1,15 +1,15 @@
...
   }
-}
 
-.label {
-  color: $blue;
-  cursor: pointer;
-  white-space: nowrap;
+  .label {
+    color: var(--show-hide-label-color, $blue);
+    cursor: var(--show-hide-label-cursor, pointer);
```
azangru, line 12: Is `cursor` something that needs configuring? In `ShowHide.tsx`, doesn't a click on the label (if the label exist) always toggle the show/hide?

src/shared/components/show-hide/ShowHide.tsx
```diff
@@ -30,33 +30,23 @@ import styles from './ShowHide.scss';
...
-    svg: classNames(styles.chevron, props.classNames?.chevron)
-  };
+  const wrapperClasses = classNames(styles.showHide, props.className);
 
   return (
     <div className={wrapperClasses}>
-      <span onClick={props.onClick} className={labelClasses}>
-        {props.label}
+      <span onClick={props.onClick} className={styles.label}>
+        {props.label && props.label}
```
azangru, line 45: This looks odd. Is the `&& props.label` even required?

azangru: BTW, don't we want to change the structure of the ShowHide component to something like this:


```ts
<div onClick={props.onClick} className={wrapperClasses}>
  {props.label && (
    <span className={styles.label}>
      {props.label}
    </span>  
  )}
  <Chevron
    direction={props.isExpanded ? 'up' : 'down'}
    animate={true}
    classNames={{ wrapper: styles.chevron }}
  />
</div>
```

i.e. move the onClick onto the outer element, and wrap the label in its own span with its own class?

src/shared/components/show-hide/ShowHide.tsx
```diff
@@ -30,33 +30,23 @@ import styles from './ShowHide.scss';
...
         <Chevron
           direction={props.isExpanded ? 'up' : 'down'}
           animate={true}
-          classNames={chevronClasses}
+          classNames={{ wrapper: styles.chevron }}
```
azangru, line 49: OMG! The chevron also has a `classNames` property? Can we transform it into a single `className` as well?

src/shared/components/show-hide/ShowHide.scss
```diff
@@ -1,15 +1,15 @@
 @import 'src/styles/common';
 
 .showHide {
-  // increase specificity to be sure to override chevron's own styles 
+  // increase specificity to be sure to override chevron's own styles
   .chevron {
     margin-left: 10px;
-    height: 6px;
```
ens-ma7, line 7: `cursor: pointer` is missing, if you hoover over chevron for showHide (Entityviewer Filter or Entity Viewer more information), the mouse isn't a pointer.

azangru: Well spotted 👍 I've pushed an update.

ens-ma7: Its not properly fixed, now the whole area (even the empty space after the chevron) is clickable (mouse pointer)

azangru: > Its not properly fixed, now the whole area (even the empty space after the chevron) is clickable (mouse pointer)

It's due to the context where the show/hide is in. If you check out the show/hide itself, it behaves correctly as an inline-block element:

https://user-images.githubusercontent.com/6834224/182147655-aa235ec0-4a9b-4d5d-8e0f-0e0d5aac2f2b.mov

The problem that you are seeing is probably on the entity viewer, where the show/hide expands to fill the css grid column. I'll update the styles there.

azangru: I pushed an update that should prevent the growing of the show/hide container inside the EntityViewer transcript info panels.

jyothishnt (approved):
Looks good

src/content/app/entity-viewer/gene-view/GeneView.tsx
```diff
@@ -186,9 +184,7 @@ const GeneViewWithData = (props: GeneViewWithDataProps) => {
           {props.gene.transcripts.length > 5 && (
             <div className={styles.filterLabelWrapper}>
               <ShowHide
-                classNames={{
-                  label: filterLabelClass
-                }}
+                className={isFilterPanelOpen ? styles.showHide : undefined}
```
jyothishnt, line 187: Is it better to write undefined or ''? In my PR I used the latter

azangru: `className={undefined}` will result in no `class` attribute on the DOM element:

```
<span>Hello</span>
```

whereas `className=""` will most likely result in an empty `class` attribute on the DOM element (unless React has changed its behaviour):

```
<span class>Hello</span>
```

I think the first result is cleaner.

ens-ma7 (approved):

PR #789, opened 2022-08-02
Species home page UI improvement

ens-ma7:
## Description
Updating the UI for removing species and also going back to species selector. (See Jira ticket description and XD in jira)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1630

## Deployment URL(s)
http://speciespage-improvement.review.ensembl.org

## Views affected
species home page

### DIFF
--- a/src/content/app/species/SpeciesPage.scss
+++ b/src/content/app/species/SpeciesPage.scss
@@ -11,7 +11,8 @@
   grid-template-columns: 1fr auto;
   align-items: center;
   line-height: 1;
-  padding-left: 11px;
+  padding-left: 10px;
+  padding-right: 20px;
 }
 
 .topbarLeft {
@@ -19,10 +20,6 @@
   align-items: center;
   cursor: pointer;
   justify-self: start;
-}
-
-.backChevron {
-  margin-left: 5px;
 }
 
 .pageTitle {
--- a/src/content/app/species/SpeciesPage.scss
+++ b/src/content/app/species/SpeciesPage.scss
@@ -11,7 +11,7 @@
   grid-template-columns: 1fr auto;
   align-items: center;
   line-height: 1;
-  padding-left: 10px;
+  padding-left: 30px;
   padding-right: 20px;
 }
 
@@ -25,7 +25,7 @@
 .pageTitle {
   color: $blue;
   font-weight: $bold;
-  margin-left: 13px;
+  margin-left: 16px;
 }
 
 .dataForSpecies {
--- a/src/content/app/species/SpeciesPage.scss
+++ b/src/content/app/species/SpeciesPage.scss
@@ -25,7 +25,7 @@
 .pageTitle {
   color: $blue;
   font-weight: $bold;
-  margin-left: 16px;
+  margin-left: 17px;
 }
 
 .dataForSpecies {
--- a/src/content/app/species/SpeciesPage.scss
+++ b/src/content/app/species/SpeciesPage.scss
@@ -11,19 +11,24 @@
   grid-template-columns: 1fr auto;
   align-items: center;
   line-height: 1;
+  padding-left: 11px;
 }
 
 .topbarLeft {
   display: flex;
   align-items: center;
+  cursor: pointer;
+  justify-self: start;
 }
 
-.close {
-  margin-top: 2px;
+.backChevron {
+  margin-left: 5px;
 }
 
 .pageTitle {
-  margin-right: 1rem;
+  color: $blue;
+  font-weight: $bold;
+  margin-left: 13px;
 }
 
 .dataForSpecies {
--- a/src/content/app/species/SpeciesPageContent.tsx
+++ b/src/content/app/species/SpeciesPageContent.tsx
@@ -35,7 +35,7 @@ import SpeciesAppBar from './components/species-app-bar/SpeciesAppBar';
 import SpeciesSidebar from './components/species-sidebar/SpeciesSidebar';
 import { StandardAppLayout } from 'src/shared/components/layout';
 import SpeciesMainView from 'src/content/app/species/components/species-main-view/SpeciesMainView';
-import CloseButton from 'src/shared/components/close-button/CloseButton';
+import Chevron from 'src/shared/components/chevron/Chevron';
 import MissingGenomeError from 'src/shared/components/error-screen/url-errors/MissingGenomeError';
 
 import { BreakpointWidth } from 'src/global/globalConfig';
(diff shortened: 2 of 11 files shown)


src/content/app/species/SpeciesPageContent.tsx
```diff
@@ -119,12 +119,13 @@ const TopBar = () => {
...
   return (
     <div className={styles.topbar}>
-      <div className={styles.topbarLeft}>
-        <span className={styles.pageTitle}>Species Home page</span>
-        <CloseButton
-          className={styles.close}
-          onClick={returnToSpeciesSelector}
+      <div className={styles.topbarLeft} onClick={returnToSpeciesSelector}>
+        <Chevron
+          direction={'left'}
```
imransl, line 124: This can passed as just a string: `direction="left"`

src/content/app/species/components/species-title-area/species-remove/SpeciesRemove.tsx
```diff
@@ -27,7 +27,10 @@ import { getCommittedSpeciesById } from 'src/content/app/species-selector/state/
 
 import { deleteSpeciesAndSave } from 'src/content/app/species-selector/state/speciesSelectorSlice';
 
-import { PrimaryButton } from 'src/shared/components/button/Button';
+import {
+  PrimaryButton,
+  SecondaryButton
+} from 'src/shared/components/button/Button';
 
 import { RootState } from 'src/store';
```
imransl, line 35: Not related to this PR but noticed this should be a type import

azangru:
Doesn't have to be addressed in this PR (although might as well); but we will need to fix this space:

Deployment:

image


XD:

image

src/content/app/species/SpeciesPage.scss
```diff
@@ -11,19 +11,24 @@
...
   align-items: center;
+  cursor: pointer;
+  justify-self: start;
 }
 
-.close {
-  margin-top: 2px;
+.backChevron {
+  margin-left: 5px;
 }
```
azangru, line 26: What distance are you trying to achieve here, and why make it a left margin of the chevron rather than the left padding of one of its parents?

image

src/content/app/species/components/species-title-area/SpeciesTitleArea.scss
```diff
@@ -32,12 +32,12 @@ $container-padding-right: 20px;
...
 .speciesTitleArea {
   display: grid;
-  grid-template-areas:
-    'icon species-name usage-toggle remove';
+  grid-template-areas: 'icon species-name usage-toggle remove';
   grid-template-columns: 60px fit-content(60%) 200px minmax(280px, 1fr);
   grid-column-gap: 20px;
   align-items: center;
   height: 90px;
+  max-width: 1321px;
```
azangru, line 40: This is a peculiarly precise max-width, which I don't think we've ever seen before. Why is it like this?

azangru: Should we wrap these two in a container then, if we want to keep their right border the same?
 
image

src/content/app/species/components/species-title-area/species-remove/SpeciesRemove.scss
```diff
@@ -40,3 +40,8 @@ $fake-left-margin: 60px;
   color: $blue;
   cursor: pointer;
 }
+
+.buttonContainer {
+  display: flex;
+  justify-content: flex-end;
+}
```
azangru, line 47: If you are also modifying the species removal, notice how the XD says "Remove species" instead of "Remove"?

image

and how it also keeps the "do not remove" flush right?

image

src/content/app/species/SpeciesPageContent.tsx
ens-ma7: Done

ens-ma7:
> Doesn't have to be addressed in this PR (although might as well); but we will need to fix this space:
> 
> Deployment:
> 
> image
> 
> XD:
> 
> image

We don't know yet what's going to be in here and that could decide how we align the text, probably best to leave it for now and once we know what is going here then we can addressed it. 
I am happy to change it now to align it but my concern is it will change again later.

src/content/app/species/SpeciesPage.scss
ens-ma7: > What distance are you trying to achieve here, and why make it a left margin of the chevron rather than the left padding of one of its parents?
> 
> image
> 
> Should the topbar's left padding be adjusted instead? Why is it like it is?
> 
> image

I dont need this margin-left, i forgot to remove it after adding padding to the container. Good spot. I have now removed it.

src/content/app/species/components/species-title-area/species-remove/SpeciesRemove.scss
ens-ma7: Good spot. Updated the text

src/content/app/species/components/species-title-area/SpeciesTitleArea.scss
ens-ma7: Agreed. i was trying to align both the titlebar and the stats box. I have now moved the max-width to the container.

ens-ma7:
> Doesn't have to be addressed in this PR (although might as well); but we will need to fix this space:

Done

imransl (approved):

azangru:
The left space in the grey bar is very odd. It is comprised of two paddings, none of which our global space sizes:

image

image

We'll need to:
- find out if the grey top bar should have a consistent left padding across all apps
- if not (a more likely answer), remove the left padding from the top bar of the StandardAppLayout component, and configure the left space individually in individual apps
- if yes, then set that padding in the StandardAppLayout component so that it is set only once

Doesn't have to be done in this PR; but will need a dedicated ticket if it's moved out of this PR

src/content/app/species/components/species-title-area/species-remove/SpeciesRemove.scss
azangru: You aren't aligning it to the right corner?


https://user-images.githubusercontent.com/6834224/182611039-298b05a3-ea77-4641-919a-cbbf157b8c7b.mov

ens-ma7:
> The left space in the grey bar is very odd. It is comprised of two paddings, none of which our global space sizes:
To add more info to this, we do have the topbar with a padding-left: 18px on all the app and we are adding more padding left in each app topbar (GB, EV, blast) and it makes sense to keep all consistent and feels like padding-left: 30px is the right size matching the bars above (launchbar and appbar). I can update it in this PR if we decide to make it consistent (which will be a better solution) but will need to check with Andrea when she is back

azangru:
> I can update it in this PR if we decide to make it consistent

I have a vague recollection that Andrea didn't want the target icon that we use to refocus on a focus object to move to the right in the genome browser. If this is the case, we won't be able to have a consistent left space in the top bar; but then we should remove the 18px left padding, entirely, and just set the required padding in individual apps.

src/content/app/species/components/species-title-area/species-remove/SpeciesRemove.scss
ens-ma7: sorry miss that, have now fix it.

ens-ma7:
Regarding the left space on the top bar, clarification from Andrea is that there is a consistent 60px space on the left which can either be empty or used for icon (focus icon, chevron) which will have their own alignment within that space. 

Todo:
Remove the 18px padding-left on the topBar and leave it to each app to customise what goes on the left (for EV it will have a padding left of 60px for now)

ens-ma7:
Also agreed to set the max-width for the stats container to 1380px and to align the left border of the stats table with the species name

azangru (approved):
Looks good

PR #791, opened 2022-08-02
Remove cogs position from redux

msmanoj:
## Description
- Removed the cogs position from redux and moved it to `useBrowserCogList`
-  Moved `selectedCog` one level up in redux
- Updated the tests

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1616

## Deployment URL(s)
http://cogs-out-of-redux.review.ensembl.org

## Views affected
Cogs

### DIFF
--- a/src/content/app/genome-browser/components/browser-cog/useBrowserCogList.ts
+++ b/src/content/app/genome-browser/components/browser-cog/useBrowserCogList.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useContext, useEffect } from 'react';
 
 import { useAppDispatch, useAppSelector } from 'src/store';
 import { useGenomeTracksQuery } from 'src/content/app/genome-browser/state/api/genomeBrowserApiSlice';
@@ -26,6 +26,8 @@ import {
   getBrowserActiveFocusObject
 } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSelectors';
 
+import { GenomeBrowserContext } from '../../Browser';
+
 import {
   setInitialTrackSettingsForGenome,
   getDefaultGeneTrackSettings,
@@ -33,8 +35,7 @@ import {
   getTrackType,
   TrackType,
   type TrackSettings,
-  type TrackSettingsForGenome,
-  type CogList
+  type TrackSettingsForGenome
 } from 'src/content/app/genome-browser/state/track-settings/trackSettingsSlice';
 import { TrackId } from 'src/content/app/genome-browser/components/track-panel/trackPanelConfig';
 
@@ -46,7 +47,15 @@ const useBrowserCogList = () => {
   const { data: trackCategories } = useGenomeTracksQuery(genomeId);
   const focusObject = useAppSelector(getBrowserActiveFocusObject); // should we think about what to do if there is no focus object
 
-  const [cogList, setCogList] = useState<CogList | null>(null);
+  const genomeBrowserContext = useContext(GenomeBrowserContext);
+
+  if (!genomeBrowserContext) {
+    throw new Error(
+      'useGenomeBrowser must be used with GenomeBrowserContext Provider'
+    );
+  }
+
+  const { cogList, setCogList } = genomeBrowserContext;
 
   const dispatch = useAppDispatch();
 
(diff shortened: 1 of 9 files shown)


src/content/app/genome-browser/components/browser-cog/useBrowserCogList.ts
```diff
@@ -33,7 +33,8 @@ import {
   getTrackType,
   TrackType,
   type TrackSettings,
-  type TrackSettingsForGenome
+  type TrackSettingsForGenome,
+  CogList
```
azangru, line 37: `type`?

msmanoj: That's done

azangru:
Could you merge the latest dev, and check whether it's something in your PR that broke 1/5 transcripts, or it was something in the old dev.

https://user-images.githubusercontent.com/6834224/182581683-5f40c58c-13d6-4b22-86a4-daf828b44a10.mov

msmanoj:
> Could you merge the latest dev, and check whether it's something in your PR that broke 1/5 transcripts, or it was something in the old dev.
> 
It's something I broke in this PR. I'll take a look

msmanoj:
> > Could you merge the latest dev, and check whether it's something in your PR that broke 1/5 transcripts, or it was something in the old dev.
> 
> It's something I broke in this PR. I'll take a look

This is now fixed. The problem was that the state of the `cogList` was not shared between trackSettingsPanel and cogs.

azangru:
> The problem was that the state of the `cogList` was not shared between trackSettingsPanel and cogs.

image

Confused: why is it needed on the browser context?

msmanoj:
> > The problem was that the state of the `cogList` was not shared between trackSettingsPanel and cogs.
> 
> image
> 
> Confused: why is it needed on the browser context?

It needs to be in some context so that the state of the cogList is shared between different places where the `useBrowserCogList` is used. I thought it would be better to add to the existing browser context rather than creating another one.

msmanoj:
Another solution will be is to make the `useBrowserCogList` hook directly listen to the cog updates payload which I am not sure if it'll be better.

azangru:
> It needs to be in some context so that the state of the cogList is shared between different places where the `useBrowserCogList` is used. I thought it would be better to add to the existing browser context rather than creating another one.

I'm trying to decide whether passing the cog list through the genome browser context is significantly better than just passing it through props  (BrowserCogList -> BrowserCog -> TrackPanelSettings). Normally, I would go for props. Is there a stronger argument in favour of the context?

msmanoj:
> I'm trying to decide whether passing the cog list through the genome browser context is significantly better than just passing it through props (BrowserCogList -> BrowserCog -> TrackPanelSettings). Normally, I would go for props. Is there a stronger argument in favour of the context?

Not an issue now but passing it as props will restrict the hierarchy of these three components which will be an issue if we decide to separate out `TrackPanelSettings` from `BrowserCog`. Also, we'll need something like a context if we need to access `cogList` outside of `BrowserCogList`. Again, it's not an issue now. 
I think using context is fine here as it doesn't create any clutter but I don't have any preference.

msmanoj:
> Not an issue now but passing it as props will restrict the hierarchy of these three components which will be an issue if we decide to separate out `TrackPanelSettings` from `BrowserCog`. Also, we'll need something like a context if we need to access `cogList` outside of `BrowserCogList`. Again, it's not an issue now. I think using context is fine here as it doesn't create any clutter but I don't have any preference.

I also noticed that it won't be easy for `useBrowserTrackSettings` to access the cogList if we decide to pass it as props.

azangru:
> I also noticed that it won't be easy for `useBrowserTrackSettings` to access the cogList if we decide to pass it as props.

If we did that, we would pass the cogs list as another argument to the functions called from TrackPanelSettings

msmanoj:
> > I also noticed that it won't be easy for `useBrowserTrackSettings` to access the cogList if we decide to pass it as props.
> 
> If we did that, we would pass the cogs list as another argument to the functions called from TrackPanelSettings

This would require passing the props further down to `RegularTrackSettings` & `GeneTrackSettings`.

azangru (approved):

PR #792, opened 2022-08-03
Don't set zmenu if there is no transcript data

msmanoj:
## Description
- Fixed the issue with zmenu not opening on the first click
- Currently it validates the data that comes from the GB to check if there is any data for transcript. If not, it simply doesn't try draw the zmenu.

**To reproduce**:
Click on the "-----" lines first and then click on a transcript.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1678

## Deployment URL(s)
http://fix-zmenu-ref.review.ensembl.org

## Views affected
Genome browser zmenu

### DIFF
--- a/src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
+++ b/src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
@@ -19,9 +19,13 @@ import React, { useEffect, memo } from 'react';
 import useGenomeBrowser from 'src/content/app/genome-browser/hooks/useGenomeBrowser';
 
 import {
-  ZmenuCreateAction,
+  type ZmenuCreateAction,
+  type ZmenuCreatePayload,
+  type ZmenuContentGene,
+  type ZmenuContentTranscript,
+  type ZmenuPayloadVariety,
   IncomingActionType,
-  ZmenuCreatePayload
+  ZmenuPayloadVarietyType
 } from '@ensembl/ensembl-genome-browser';
 
 import Zmenu from './Zmenu';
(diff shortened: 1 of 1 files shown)


azangru:
What [ENSWBSITES-1678](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1678) has revealed is that the `useRefWithRerender` hook has been written incorrectly. My fault. Although why this only becomes apparent after the Zmenu component first returns `null`, I do not know.

If you look at our [useRefWithRerender](https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/hooks/useRefWithRerender.ts), you will notice that we were expecting useEffect to rerun when ref.current changes. This would only happen, however, if the component that was calling `useRefWithRerender` was itself rerendering; because useEffect would only run during a rerender.

```
const useRefWithRerender = <T>(
  initialValue: T | null
): MutableRefObject<T | null> => {
  const [currentRef, setCurrentRef] = useState<T | null>(null);
  const ref = useRef<T>(initialValue);

  useEffect(() => {
    if (currentRef !== ref.current) {
      setCurrentRef(ref.current);
    }
  }, [ref.current]);

  return ref;
};
```

A working implementation of this hook (I tested locally) would look something like this:

```ts
const useRefWithRerender = <T>(
  initialValue: T | null
): MutableRefObject<T | null> => {
  const [currentRef, setCurrentRef] = useState<T | null>(null);
  const ref = useRef<T | null>(initialValue);

  const callbackRef = (node: T) => {
    if (currentRef !== node) {
      ref.current = node;
      setCurrentRef(node);
    }
  }

  return {
    ref,
    callbackRef
  };
};
```
In this case, `callbackRef` returned from this hook will be passed to the `ref` property of a React element; and this function will be called when the DOM element created from the React element gets mounted.

I'll push a PR to demonstrate; but will have to fix the two components that are using this hook, as well as tests.

msmanoj:
Superseded by https://github.com/Ensembl/ensembl-client/pull/795

PR #793, opened 2022-08-03
Use real genome id instead of the id in the url for external references in Entity Viewer

azangru:
## Description
External References component in Entity Viewer is querying Thoas using the genome id it gets from the url instead of the "real" genome id:

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1680

## Deployment URL(s)
http://fix-ev-external-references.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
@@ -15,11 +15,11 @@
  */
 
 import React, { useState } from 'react';
-import { useParams } from 'react-router';
 import sortBy from 'lodash/sortBy';
 
-import { parseFocusObjectIdFromUrl } from 'src/shared/helpers/focusObjectHelpers';
 import { defaultSort } from 'src/content/app/entity-viewer/shared/helpers/transcripts-sorter';
+
+import useGeneViewIds from 'src/content/app/entity-viewer/gene-view/hooks/useGeneViewIds';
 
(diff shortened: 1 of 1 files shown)


imransl (approved):

src/content/app/entity-viewer/gene-view/components/gene-view-sidebar/external-references/GeneExternalReferences.tsx
```diff
@@ -41,8 +41,7 @@ type ExternalReferencesGroupType = {
 };
 
 const GeneExternalReferences = () => {
-  const { entityId, genomeId } = useParams<'genomeId' | 'entityId'>();
-  const geneId = entityId ? parseFocusObjectIdFromUrl(entityId).objectId : null;
+  const { genomeId, geneId } = useGeneViewIds();
```
ens-ma7, line 44: can geneId be null or undefined or '' ? My understanding of the hook `useGeneViewIds` it will return undefined if there is no geneId. is that right?

ens-ma7: so what happen if there is no entityId from this line?
https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/entity-viewer/gene-view/hooks/useGeneViewIds.ts#L28

I am just confused with the if statement in the hook above. If entityId cannot be undefined or null then we dont need that if statement

azangru: The way that the hook is written, it will return `undefined`.

But `GeneExternalReferences` is only mounted on the route that has an entity id in the url; so when called inside of `GeneExternalReferences`, it should return a string.

ens-ma7 (approved):

PR #797, opened 2022-08-09
[FOR DISCUSSION]: another approach to table

azangru:
## Description
This PR presents an approach to building a table that is diametrically opposite to the one used in https://github.com/Ensembl/ensembl-client/pull/776 In this PR, the table, as well as its parts, is a dumb presentational component that does not have any logic of its own, and Instead delegates all logic to its parent component.

A clear disadvantage of this approach is that the logic present in the parent would have to be replicated across different instances of the table.

The advantages, on the other hand, are:
- flexibility
- keeping table state in one place (i.e. in the parent)

**Questions:**
- is anything else required for the validation of the approach to the building of the table?
- which approach do we prefer?
- is there any benefit in having both tables?
- if https://github.com/Ensembl/ensembl-client/pull/776 is preferred, can the smart table in https://github.com/Ensembl/ensembl-client/pull/776 be built utilising the components of the dumb table presented here?

### DIFF
--- a/src/shared/components/table/Table.scss
+++ b/src/shared/components/table/Table.scss
@@ -1,0 +1,43 @@
+@import 'src/styles/common';
+
+.table {
+  border-collapse: collapse;
+}
+
+.tableHeadCell {
+  font-size: 12px;
+  font-weight: $light;
+  border-bottom: 1px solid $blue;
+}
+
+.tableHeadCell,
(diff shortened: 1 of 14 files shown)


azangru:
Closed in favour of https://github.com/Ensembl/ensembl-client/pull/803

PR #798, opened 2022-08-10
Blast sequence alignment

azangru:
## Description
This PR adds a simple string representation of an alignment between a query sequence and a match found by the BLAST tool.

**NOTE:** lots of lines in the diff are just sample data in the storybook. We can delete the BlastSequenceAlignment story when we are done with reviewing this PR.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1558

## Deployment URL(s)
~View Storybook locally.~ deleted the storybook story after the review


src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
```diff
@@ -0,0 +1,165 @@
...
+  startIndex: number;
+  endIndex: number;
+  startPosition: number;
+  strandDirectionMultiplier: number;
+}) => {
+  let position = startPosition;
+
+  for (let i = startIndex; i < endIndex; i++) {
+    if (sequence[i] !== '-') {
+      position += 1 * strandDirectionMultiplier;
```
msmanoj, line 160:
```suggestion
      position += strandDirectionMultiplier;
```

src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
```diff
@@ -0,0 +1,165 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const ALIGNMENT_LINE_LENGTH = 60;
+
+type Params = {
```
msmanoj, line 19: This is same as `BlastSequenceAlignmentInput`. Can we import and use that instead?

src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
azangru: I think I explicitly added the `1` for clarity, i.e. "either add or subtract one based on whether you are moving in the same or in the opposite direction as the query sequence". I realise that `1 * strandDirectionMultiplier` looks silly, because it's the same as just `strandDirectionMultiplier`; but then, will it not be confusing that the code would read as if you are adding a multiplier.

src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
azangru: Yeah, I was planning to do just that, and then forgot 🙂

src/content/app/tools/blast/types/blastJob.ts
```diff
@@ -0,0 +1,54 @@
...
+  hits: BlastHit[];
+};
+
+type BlastHit = {
+  hit_acc: string;
+  hit_def: string;
+  hit_hsps: HSP[];
+};
+
+type HSP = {
```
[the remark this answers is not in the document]
azangru: Yes, I git checked-out these files from that branch into this one 🙂  I needed these changes in this branch anyway.

src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
msmanoj: Fair enough. Happy to leave it as it is.

src/content/app/tools/blast/types/blastJob.ts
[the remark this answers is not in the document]
azangru: I git checked-out this file from the branch with the hit diagram, where it was renamed to blastJob:

image

Didn't realise the blastApiResponse file would remain unchanged in this branch 😞

src/content/app/tools/blast/components/blast-sequence-alignment/blastSequenceAlignmentHelper.ts
azangru: ✅

src/content/app/tools/blast/types/blastJob.ts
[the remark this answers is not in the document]
azangru: ✅

src/content/app/tools/blast/components/blast-sequence-alignment/formatters/simpleStringFormatter.ts
```diff
@@ -0,0 +1,95 @@
...
+  queryLinePrefix = queryLinePrefix.padEnd(prefixLength, ' ');
+  hitLinePrefix = hitLinePrefix.padEnd(prefixLength, ' ');
+
+  const alignmentPrefix = ''.padStart(prefixLength, ' ');
+
+  return [
+    `${queryLinePrefix} ${queryLine} ${queryLineSuffix}`,
+    `${alignmentPrefix} ${middleLine}`,
+    `${hitLinePrefix} ${hitLine} ${hitLineSuffix}`
+  ].join('\n');
```
msmanoj, line 84: Wouldn't it be simpler to automatically align these using a table?

azangru: This is specifically a formatter for generating a raw string output if we need one (at least for testing purposes; but even more importantly, if we ever need to generate download files ourselves). We can create other formatters if we need them, or just generate a table out of the output of the `createBlastSequenceAlignment` function if that is necessary.

msmanoj (approved):
I am not sure what our plans are regarding the middle line if it needs to count or not.
Other than that, the PR looks good.

Please remember to remove the sample data.

azangru:
> I am not sure what our plans are regarding the middle line if it needs to count or not.

I updated the formatter code; it now adds the position in the beginning and the end of the middle line. Plus I changed the labels in the beginning of the line to align right. The alignment looks like this now:

image

PR #802, opened 2022-08-12
Update genome browser to version 0.5.0

azangru:
## Description
Update genome browser to version ~0.4.4~ 0.5.0.

See the [changelog](https://github.com/Ensembl/ensembl-dauphin-style-compiler/blob/staging/CHANGELOG.md#044) for the list of features.

## Deployment URL(s)
http://update-genome-browser-044.review.ensembl.org


imransl:
@azangru The gene and transcript tracks are all shown as being turned off by default in the track panel:

Screenshot 2022-08-15 at 17 47 36

Also noticed that when I switched genes using the zmenu and then clicked on one of the previously viewed gene in the track panel, the genome browser focussed without showing the gene (as what happened before). Unfortunately I couldn't reproduce it a second time. Will post a screenshot if I get to though.

imransl (approved):
Don't see the above issues any more. Looks good now.

ens-st3 (approved):

PR #805, opened 2022-08-16
Make sequence box in BLAST results view collapsible

ens-ma7:
## Description
Adding chevron to result view to show/hide blast alignments. Also fix some padding based on dev spec provided in jira ticket.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1650

## Deployment URL(s)
http://blast-result-chevron.review.ensembl.org

## Views affected
[Blast result](http://blast-result-chevron.review.ensembl.org/blast)

### DIFF
--- a/src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
+++ b/src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
@@ -8,7 +8,7 @@ $double_standard_gutter: calc($standard-gutter * 2);
 }
 
 .sequenceBoxWrapper {
-  padding: 24px 26px; // TODO: the 26px padding looks like a constant asking to be extracted
+  padding: 24px 20px 0px 20px;
   border: 1px solid $grey;
 }
 
@@ -24,10 +24,9 @@ $double_standard_gutter: calc($standard-gutter * 2);
 
 .resultsSummaryRow {
   display: grid;
-  grid-template-columns: [sequence-label] 11ch [hits-diagram] minmax(
-      48%,
-      1050px
-    ) [species] minmax(20%, 1fr) 10px;
+  grid-template-columns:
+    [sequence-label] 11ch [hits-diagram] minmax(48%, 1050px)
+    [species] minmax(20%, 1fr) 10px;
   column-gap: 20px;
   padding-bottom: 20px;
 }
@@ -75,4 +74,12 @@ $double_standard_gutter: calc($standard-gutter * 2);
   justify-content: flex-end;
   column-gap: 40px;
   width: 268px;
+}
+
+.showHideWrapper {
+  --show-hide-chevron-height: var(--chevron-height, 8px);
+}
+
+.rulerContainerClass {
+  padding-bottom: 10px;
 }
--- a/src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
+++ b/src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
@@ -77,9 +77,9 @@ $double_standard_gutter: calc($standard-gutter * 2);
 }
 
 .showHideWrapper {
-  --show-hide-chevron-height: var(--chevron-height, 8px);
+  --show-hide-chevron-height: 8px;
 }
 
-.rulerContainerClass {
+.rulerWrapper {
   padding-bottom: 10px;
 }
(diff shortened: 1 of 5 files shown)


src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.tsx
```diff
@@ -117,28 +124,37 @@ const SequenceBox = (props: SequenceBoxProps) => {
...
-            species={speciesInfo[0]}
-            jobId={result.jobId}
-            diagramWidth={plotwidth}
-          />
-        );
-      })}
-      <div className={styles.resultsSummaryRow}>
+      {isExpanded &&
+        blastResults.map((result) => {
+          const speciesInfo = species.filter(
```
imransl, line 137: Can't `find` be used here instead?

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
```diff
@@ -76,3 +75,11 @@ $double_standard_gutter: calc($standard-gutter * 2);
   column-gap: 40px;
   width: 268px;
 }
+
+.showHideWrapper {
+  --show-hide-chevron-height: var(--chevron-height, 8px);
```
azangru, line 80: No need for the `var(--chevron-height,` part

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
```diff
@@ -76,3 +75,11 @@ $double_standard_gutter: calc($standard-gutter * 2);
...
+}
+
+.rulerContainerClass {
```
azangru, line 83: No need for the `Class` suffix?

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.tsx
```diff
@@ -107,6 +109,11 @@ const SequenceBox = (props: SequenceBoxProps) => {
     parsedBlastSequence;
   const rulerContainer = useRef<HTMLDivElement | null>(null);
   const { width: plotwidth } = useResizeObserver({ ref: rulerContainer });
+  const [isExpanded, setExpanded] = useState(false);
```
azangru, line 112: > Default view is panels expanded.

-- from the ticket description

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
ens-ma7: I agreed no need for class but when i removed the Class the name `rulerContainer` clashes with the ref variable name `rulerContainer` so i have renamed it to `rulerWrapper`

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.tsx
ens-ma7: I didnt write this line of code but just added the isExpanded only. Not sure if we want to change this now. I am curious about the difference between filter and find though

imransl: `find` will give you one item that matches the condition. This is what we need to use if there is only one unique item that we are searching for.

ens-ma7: agreed with Imran to leave it with `filter` since the code is working fine as it is and there wont be much difference in using find instead of filter.

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.scss
azangru: Sorry, I meant that the word "Class" is unnecessary in a CSS file, because anything starting with a `.` is a class name.

In tsx, when you are writing:

image

— the word `Class`, or, better, `ClassName`, is actually very helpful, to the point of being necessary:

```ts
// rulerWrapperClassName, or rulerWrapperClass, or rulerWrapperClasses, or rulerWrapperClassNames
const rulerWrapperClassName = classNames(
  styles.resultsSummaryRow,
  styles.rulerWrapper // here, though, it would still be styles.rulerWrapper; not styles.rulerWrapperClass
);
```

azangru (approved):

src/content/app/tools/blast/views/blast-submission-results/BlastSubmissionResults.tsx
```diff
@@ -25,6 +26,7 @@ import { useFetchBlastSubmissionQuery } from 'src/content/app/tools/blast/state/
 
 import { parseBlastInput } from '../../utils/blastInputParser';
 import { pluralise } from 'src/shared/helpers/formatters/pluralisationFormatter';
+import ShowHide from 'src/shared/components/show-hide/ShowHide';
```
imransl, line 29: Just noticed this import is in the wrong place

imransl (approved):

PR #810, opened 2022-08-24
Display all tracks returned from the track api by default

azangru:
## Description
[TODO]

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1698

## Deployment URL(s)
http://show-all-genomic-tracks.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomicTracks.ts
+++ b/src/content/app/genome-browser/hooks/useGenomicTracks.ts
@@ -1,0 +1,107 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect } from 'react';
+import { useAppSelector } from 'src/store';
+
+import useGenomeBrowser from 'src/content/app/genome-browser/hooks/useGenomeBrowser';
+import useGenomeBrowserIds from './useGenomeBrowserIds';
+import { useGenomeTracksQuery } from 'src/content/app/genome-browser/state/api/genomeBrowserApiSlice';
+
+import { getBrowserTrackStates } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSelectors';
+
+import { Status } from 'src/shared/types/status';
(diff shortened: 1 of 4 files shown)


src/content/app/genome-browser/hooks/useGenomicTracks.ts
```diff
@@ -0,0 +1,173 @@
...
+  const savedGenomicTrackStates =
+    allBrowserTrackStates[activeGenomeId ?? '']?.commonTracks ?? {};
+
+  useEffect(() => {
+    if (!genomeBrowser) {
+      return;
+    }
+
+    const trackIdsList = combineWithSavedData(
+      prepareTrackIdsList(genomeTrackCategories ?? []),
```
msmanoj, line 69: Can `prepareTrackIdsList` be updated to accept `savedGenomicTrackStates` and do the combining as well?

src/content/app/genome-browser/hooks/useGenomicTracks.ts
```diff
@@ -0,0 +1,173 @@
...
+const combineWithSavedData = (
+  trackIdsList: TrackIdsList,
+  savedTrackStates: TrackStates
+): TrackIdsList => {
+  const savedTrackVisibilityMap = new Map<string, TrackActivityStatus>();
+  Object.values(savedTrackStates)
+    .flatMap((trackCategory) => Object.entries(trackCategory))
+    .forEach(([trackId, trackStatus]) => {
+      savedTrackVisibilityMap.set(trackId, trackStatus);
+    }, {});
```
msmanoj, line 116: Why do we need `, {}`?

azangru: It's leftover of me writing a reduce first 😄 Surprised that it's valid for the `forEach`; at least typescript swallowed it.

msmanoj:
Toggling the track settings panel seems to be send messages to toggle tracks. I am not sure what is causing it.

azangru:
> Toggling the track settings panel seems to be send messages to toggle tracks. I am not sure what is causing it.

Could you describe this a bit more?

The way the code is written, togging track settings for a single track will cause the useGenomicTracks hook to send unnecessary messages about track settings, because the track settings object in redux will be updated. I don't see an easy way of avoiding it. But you seem to be describing something different?

msmanoj:
>The way the code is written, togging track settings for a single track will cause the useGenomicTracks hook to send unnecessary messages about track settings, because the track settings object in redux will be updated. I don't see an easy way of avoiding it.
 
Yes, I did notice that.

>But you seem to be describing something different? 

Yes, what I was mentioning is that the messages are getting sent when I just click on the cogs to open the settings panel.

src/content/app/genome-browser/hooks/useGenomicTracks.ts
azangru: ✅

src/content/app/genome-browser/hooks/useGenomicTracks.ts
azangru: ✅

imransl (approved):

msmanoj (approved):

PR #812, opened 2022-08-25
Update tooltip text for the eye next to the gene in right-hand panel

msmanoj:
## Description
- optionally passing `getVisibilityIconHelpText` as a prop to customise the help text. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1679

## Deployment URL(s)
http://eye-tooltip-update.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
@@ -144,12 +144,21 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
     );
   };
 
+  const getVisibilityIconHelpText = (status: Status) => {
+    if (status === Status.SELECTED) {
+      return 'Hide all transcripts';
+    }
+
+    return 'Show all transcripts';
+  };
+
   return (
     <>
       <GroupTrackPanelItemLayout
         isCollapsed={isCollapsed}
         visibilityStatus={geneVisibilityStatus}
         onChangeVisibility={onGeneVisibilityChange}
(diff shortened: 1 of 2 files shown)


imransl (approved):

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -144,12 +144,21 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
...
+    return 'Show all transcripts';
+  };
+
   return (
     <>
       <GroupTrackPanelItemLayout
         isCollapsed={isCollapsed}
         visibilityStatus={geneVisibilityStatus}
         onChangeVisibility={onGeneVisibilityChange}
+        getVisibilityIconHelpText={getVisibilityIconHelpText}
```
azangru, line 161: I'm confused: doesn't `TrackPanelGene` at this point know what its visibility status is? Why wouldn't it just tell `GroupTrackPanelItemLayout` as a string what the help text ought to be? Why pass the function?

msmanoj: I've updated it to pass the help text as  a prop instead of a function.

azangru:
image

We should probably modify this help text as well? "Show this transcript" / "Hide this transcript"?

msmanoj:
> image
> 
> We should probably modify this help text as well? "Show this transcript" / "Hide this transcript"? The transcripts are all shown in the same track...

Yes, I've updated it.

azangru (approved):

PR #819, opened 2022-09-06
Add confirmation dialog for deletion of a BLAST job

ens-ma7:
## Description
Blast job deletion confirmation message. As per the jira ticket, on clicking the thrash icon, we show a confirmation to delete or not to delete. 
And the extra logic (mentioned in the jira), if it is the last job on either the unviewed job view or the job list we redirect to new job form and also when deleting a job on a result view (one submission), we redirect to new job form.

Note:
I am checking which view we are currently on by using `useLocation` which is looking at the URL. currently the view is passed as props to `BlastAppBar.tsx`, maybe for another ticket/PR we can have the view in redux state so that its easily retrieved.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1686

## Deployment URL(s)
http://blast-deletion.review.ensembl.org

## Views affected
blast (try submitting a few jobs and have some in unviewed and job lists)

### DIFF
--- a/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
+++ b/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
@@ -28,9 +28,7 @@ import blastFormReducer, {
 import speciesSelectorReducer, {
   SpeciesSelectorState
 } from 'src/content/app/species-selector/state/speciesSelectorSlice';
-import blastGeneralReducer, {
-  initialState as initialBlastGeneralState
-} from 'src/content/app/tools/blast/state/general/blastGeneralSlice';
+import blastGeneralReducer from 'src/content/app/tools/blast/state/general/blastGeneralSlice';
 
 import BlastAppBar from './BlastAppBar';
 
@@ -52,7 +50,6 @@ const mockCommittedItems = [
 
 const initialState = {
   blast: { blastForm: initialBlastFormState },
-  general: { view: initialBlastGeneralState },
   speciesSelector: {
     committedItems: mockCommittedItems
   } as SpeciesSelectorState
--- a/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
+++ b/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
@@ -28,6 +28,9 @@ import blastFormReducer, {
 import speciesSelectorReducer, {
   SpeciesSelectorState
 } from 'src/content/app/species-selector/state/speciesSelectorSlice';
+import blastGeneralReducer, {
+  initialState as initialBlastGeneralState
+} from 'src/content/app/tools/blast/state/general/blastGeneralSlice';
 
 import BlastAppBar from './BlastAppBar';
 
@@ -49,6 +52,7 @@ const mockCommittedItems = [
 
 const initialState = {
   blast: { blastForm: initialBlastFormState },
+  general: { view: initialBlastGeneralState },
   speciesSelector: {
     committedItems: mockCommittedItems
   } as SpeciesSelectorState
@@ -56,7 +60,8 @@ const initialState = {
 
 const rootReducer = combineReducers({
   blast: combineReducers({
-    blastForm: blastFormReducer
+    blastForm: blastFormReducer,
+    blastGeneral: blastGeneralReducer
   }),
   speciesSelector: speciesSelectorReducer
 });
@@ -70,7 +75,7 @@ const renderComponent = () => {
   const renderResult = render(
     <Provider store={store}>
       <MemoryRouter>
-        <BlastAppBar view="blast-form" />
+        <BlastAppBar />
       </MemoryRouter>
     </Provider>
   );
--- a/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
+++ b/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
@@ -4,7 +4,7 @@
   display: grid;
   grid-template-areas:
     'sequence-number submission-name submission-details control-buttons'
-    '. . . delete-message';
+    '. . delete-message delete-message';
 
   grid-template-columns: auto auto 1fr 20%;
   column-gap: 20px;
--- a/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
+++ b/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
@@ -7,27 +7,12 @@
     '. . . delete-message';
 
   grid-template-columns: auto auto 1fr 20%;
-  grid-template-rows: auto auto;
   column-gap: 20px;
   align-items: center;
   padding-left: 26px; // TODO: seems to be a constant shared with sequence boxes (see ListedBlastSubmission.scss)
   padding-top: 20px;
   margin-bottom: 14px;
   white-space: nowrap;
-}
-
-@media screen and (max-width: 1720px) {
-  .grid {
-    grid-template-columns: auto auto 1fr auto;
-    grid-template-rows: auto auto;
-  }
-  .controlButtons {
-    padding-left: 20px;
-  }
-  .deleteMessageContainer {
-    grid-area: delete-message;
-    padding-left: 20px;
-  }
 }
 
 .submissionName {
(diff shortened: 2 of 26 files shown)


src/content/app/tools/blast/state/general/blastGeneralSlice.ts
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
+
+type BlastView =
+  | 'blast-form' //new job form
+  | 'unviewed-submissions' //unviewed jobs
+  | 'submissions-list' // jobs list (list of viewed)
+  | 'submission-results'; // one submission result
```
azangru, line 23: Regarding the comments:

`new job form` — the form for submitting a new batch of BLAST jobs
`unviewed jobs` — unviewed BLAST submissions
`jobs list (list of viewed)` -> list of viewed BLAST submissions
`one submission result` -> results of BLAST jobs in a single submission

There are two concepts here: a BLAST job is a single 1 sequence vs 1 species BLAST run performed by EBI BLAST. Each of the jobs will receive an id from the EBI BLAST tool.

We can submit one or more jobs at once via the form (which makes the whole thing very confusing). This group of jobs submitted in a single form is referred to in our code as a submission.

What you then see in the lists are the submissions (potentially many sequences against many species at once). And when you press on the "Results" button, you are seeing a page that shows results from multiple BLAST jobs grouped in a single submission.

ens-ma7: Thanks. I will update the comments....its indeed confusing and i think i will add one more comment of top of this to explain what a job and a submission is.

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -81,6 +93,16 @@ export const BlastSubmissionHeader = (props: Props) => {
 
   const handleDeletion = () => {
     dispatch(deleteBlastSubmission(submissionId));
+
+    //redirect to new job form/view if deleting on one submission/job result or it is the last submissions/job in either unviewed jobs and job lists
```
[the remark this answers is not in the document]
ens-ma7: Done. Updated both comments (the one above and this one).

src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
```diff
@@ -49,14 +52,16 @@ const mockCommittedItems = [
 
 const initialState = {
   blast: { blastForm: initialBlastFormState },
+  general: { view: initialBlastGeneralState },
```
azangru, line 55: This isn't in the right place

ens-ma7: good point. Not sure why i added this here but i remember i had to add the blastGeneralReducer and looks like this is enough, it doesnt require the state. I have updated the code.

azangru: I mean if the test doesn't require it (is doesn't seem to; it wasn't complaining when the state was set up incorrectly), then don't include it. Otherwise it makes an impression that this piece of state is required for the test to run.

azangru:
Something happened to our right-hand white space... I thought we had it sorted. Doesn't seem to be something introduced in this PR; but can be fixed here. Or separately.

image

azangru:
Small screen broken


https://user-images.githubusercontent.com/6834224/188904808-b8446149-6746-4b2e-9894-904dde0851cb.mov

src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.test.tsx
ens-ma7: But if i removed the blastGeneralReducer then the test fails. So i left the reducer in there. And i thought if you are adding the reducer, you also need to add the state but thats not the case.

azangru: Oh, ok then.

ens-ma7:
> Something happened to our right-hand white space... I thought we had it sorted. Doesn't seem to be something introduced in this PR; but can be fixed here. Or separately.


Happy to fix this as part of this PR, i think its just one padding-right that needs adjusting.

ens-ma7:
> Small screen broken

it looks fine to me.  What other solution do we have for smaller screen when you click on the trash icon, should it wrap the message? we can check with Andrea?

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
```diff
@@ -100,3 +85,18 @@
   color: $blue;
   cursor: pointer;
 }
+
+// Leave media queries at the bottom of the file to increase css specifity
```
azangru, line 89: `to increase css specifity` -> `to make sure they override CSS rules defined above`

The specificity of the CSS selectors inside of media queries remains the same; that's the whole point of having them at the bottom, meaning that they will override rules from selectors of the same specificity defined above.

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
```diff
@@ -73,3 +69,35 @@
...
+  color: $red;
+}
+
+.clickable {
+  color: $blue;
+  cursor: pointer;
+}
+
+// Leave media queries at the bottom of the file to make sure they override CSS rules defined above
+@media screen and (max-width: 1720px) {
```
jyothishnt, line 91: @ens-ma7 Are we going ahead with this view for screen size 1720px? 
Screenshot 2022-09-13 at 10 53 11

ens-ma7: @jyothishnt i will probably revert this change if we are going with the show/hide option for showing the deletion message

ens-ma7:
waiting for PR: https://github.com/Ensembl/ensembl-client/pull/828 to be merged and then need to update this branch (rebase dev) and add the shared component

msmanoj:
Is there any use in greying out the `Results` button that is currently selected?

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
```diff
@@ -73,3 +58,32 @@
 .controlButtons > * {
   flex-shrink: 0;
 }
+
+.deleteMessageContainer {
+  margin: 20px 0px;
```
azangru, line 63: `0` doesn't need units. `0px` is the same as `0em` or `0rem` or `0ch`, or `0%`. It's just `0`.

azangru (approved):
Pushed a couple of minor commits. Happy with the PR otherwise.

ens-ma7:
Thanks @azangru for making the changes needed. I am also happy with the changes. Lets see if others want to review it and then we can merge. thanks

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -81,48 +93,71 @@ export const BlastSubmissionHeader = (props: Props) => {
...
+          <span className={styles.timeStamp}>
+            <span>{submissionTime}</span>
+            <span className={styles.timeZone}>GMT</span>
+          </span>
+          {sequenceCount && sequenceCount > 1 && (
+            <ShowHide
+              className={styles.showHide}
+              isExpanded={props.isExpanded || false}
+              onClick={() =>
+                props.toggleExpanded && props.toggleExpanded(!props.isExpanded)
```
imransl, line 130: This also can be written as `props.toggleExpanded?.(!props.isExpanded)`

jyothishnt (approved):
Looks good to me, though I found a different bug while testing for which I will create a JIRA

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
ens-ma7: Thanks for reviewing @imransl. But this is not something that i have changed in this PR (it came up in the diff since i added some html to the same container and shifted the tabbing of the code). 

I can make that change but not sure whats the impact and are we sure this is correct (the `.` after the `props.toggleExpanded` doesn't feel right, its not an object here but a function

azangru: >  are we sure this is correct (the . after the props.toggleExpanded doesn't feel right

This is called [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax), a recent (two years or thereabouts) addition to the javascript syntax.

ens-ma7: I made the suggested change @imransl. could you please review and approve if its all good. Thanks

imransl (approved):

msmanoj (approved):

PR #822, opened 2022-09-07
Remove "data for this species" text from Species page

ens-ma7:
## Description
As per jira ticket but leaving it comment out since there will be some text to be shown here.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1689

## Deployment URL(s)
http://specieshomepage-cleanup.review.ensembl.org

## Views affected
species homepage. Any species top bar

### DIFF
--- a/src/content/app/species/SpeciesPageContent.tsx
+++ b/src/content/app/species/SpeciesPageContent.tsx
@@ -123,7 +123,9 @@ const TopBar = () => {
         <Chevron direction="left" animate={false} />
         <span className={styles.pageTitle}>Find a Species</span>
       </div>
-      <div className={styles.dataForSpecies}>Data for this species</div>
+      <div className={styles.dataForSpecies}>
+        {/* placeholder for future species data */}
+      </div>
     </div>
   );
 };


src/content/app/species/SpeciesPageContent.tsx
```diff
@@ -123,7 +123,9 @@ const TopBar = () => {
         <Chevron direction="left" animate={false} />
         <span className={styles.pageTitle}>Find a Species</span>
       </div>
-      <div className={styles.dataForSpecies}>Data for this species</div>
+      <div className={styles.dataForSpecies}>
+        {/* placeholder for future species data */}
+      </div>
```
azangru, line 128: I'm curious if there is any value in keeping this here. Surely it will be very easy to recreate if we have to put some info here later on? Did Andrea say that she wants to put something in this space?

ens-ma7: Yes she said that there will be some text for this space. I notice we do this often in the code, in fact my comment is exactly the same as to what i saw in the code.

imransl (approved):

azangru (approved):
Ok then. I wonder if we'll still be having this placeholder sitting here in a year's time 😄

PR #818, opened 2022-09-04
Update dependencies

azangru:
## Description
Routine update of project dependencies.

### Notable updates
- typescript updated from 4.7 to 4.8
- jest updated from v. 28 to v. 29 (changed defaults for snapshot formatter, which required updating the snapshots)

## Deployment URL(s)
http://update-dependencies.review.ensembl.org

## Checks
- `npm test` — ✅ 
- `npm run check-types` — ✅ 
- development build — ✅ 
- production build — ✅ 
- `npm run storybook` — ✅

PR #825, opened 2022-09-08
Remove the focusable attribute from the svg

azangru:
## Description
Just a one-line change — addressing a lint error about an invalid `focusable` attribute for svg that @imransl noticed in https://github.com/Ensembl/ensembl-client/pull/818.

The linter is correct; the `focusable` attribute has been removed from the current svg spec. The current recommendation is to use tabindex if you want to make sure the svg element isn't focusable; but I played with the Select in the Storybook, and I don't think it is focusable even without these attributes.

## Views affected
Might check the Select element in Storybook if you wish.


imransl (approved):
Looks good locally

PR #828, opened 2022-09-15
create delete confirmation component and added to storybook

ens-ma7:
## Description
we will have a common component for deleting or removing confirmation (example will be blast job deletion or removing species in species selector).

The component should show a message in red and have 1 button to confirm deletion and 1 link to cancel the deletion.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1732

## Deployment URL(s)
Will need to deploy storybook locally to visualize it

## Views affected
storybook (shared component/DeletionConfirmation)

### DIFF
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -3,15 +3,15 @@
 .deleteConfirmationContainer {
   display: flex;
   align-items: center;
-  justify-content: var(--container-justify-content, flex-end);
+  justify-content: flex-end;
   column-gap: 30px;
 }
 
 .warningText {
-  color: var(--warning-text-color, #{$red});
+  color: var(--deletion-confirmation-warning-color, #{$red});
 }
 
 .cancel {
-  color: var(--clickable-color, #{$blue});
+  color: var(--deletion-confirmation-cancel-color, #{$blue});
   cursor: pointer;
 }
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -3,16 +3,16 @@
 .deleteConfirmationContainer {
   display: flex;
   align-items: center;
-  justify-content: flex-end;
-  column-gap: 30px;
-  margin: 20px 0px;
+  justify-content: var(--container-justify-content, flex-end);
+  column-gap: var(--container-column-gap, 30px);
+  margin: var(--container-margin, 20px 0px);
 }
 
 .warningText {
-  color: $red;
+  color: var(--warning-text-color, #{$red});
 }
 
 .clickable {
-  color: $blue;
+  color: var(--clickable-color, #{$blue});
   cursor: pointer;
 }
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -4,8 +4,12 @@
   display: flex;
   flex-wrap: wrap;
   align-items: center;
+  column-gap: 30px;
+  row-gap: 10px;
+}
+
+.contentAlignRight {
   justify-content: flex-end;
-  gap: 30px;
 }
 
 .warningText {
@@ -15,7 +19,6 @@
 .controlsContainer {
   display: flex;
   align-items: center;
-  justify-content: flex-end;
   column-gap: 30px;
 }
 
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -1,6 +1,6 @@
 @import 'src/styles/common';
 
-.deleteMessageContainer {
+.deleteConfirmationContainer {
   display: flex;
   align-items: center;
   justify-content: flex-end;
@@ -8,7 +8,7 @@
   margin: 20px 0px;
 }
 
-.deleteMessage {
+.warningText {
   color: $red;
 }
 
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -1,0 +1,18 @@
+@import 'src/styles/common';
+
+.deleteMessageContainer {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  column-gap: 30px;
+  margin: 20px 0px;
+}
+
+.deleteMessage {
+  color: $red;
+}
+
+.clickable {
+  color: $blue;
+  cursor: pointer;
+}
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -2,13 +2,21 @@
 
 .deleteConfirmationContainer {
   display: flex;
+  flex-wrap: wrap;
   align-items: center;
   justify-content: flex-end;
-  column-gap: 30px;
+  gap: 30px;
 }
 
 .warningText {
   color: var(--deletion-confirmation-warning-color, #{$red});
+}
+
+.controlsContainer {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  column-gap: 30px;
 }
 
 .cancel {
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.scss
@@ -4,15 +4,14 @@
   display: flex;
   align-items: center;
   justify-content: var(--container-justify-content, flex-end);
-  column-gap: var(--container-column-gap, 30px);
-  margin: var(--container-margin, 20px 0px);
+  column-gap: 30px;
 }
 
 .warningText {
   color: var(--warning-text-color, #{$red});
 }
 
-.clickable {
+.cancel {
   color: var(--clickable-color, #{$blue});
   cursor: pointer;
 }
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
@@ -52,7 +52,7 @@ describe('DeletionConfirmation', () => {
     const deleteButton = container.querySelector('button') as HTMLButtonElement;
     expect(deleteButton?.textContent).toBe('Remove');
 
-    const cancelLink = container.querySelector('.clickable') as HTMLElement;
+    const cancelLink = container.querySelector('.cancel') as HTMLElement;
     expect(cancelLink?.textContent).toBe('Do not remove');
 
     const warningText = container.querySelector('.warningText') as HTMLElement;
@@ -65,9 +65,10 @@ describe('DeletionConfirmation', () => {
     );
 
     const deletionConfirmation = container.firstChild as HTMLElement;
-    expect(
-      deletionConfirmation.classList.contains('.componentClass')
-    ).toBeTruthy();
+
+    expect(deletionConfirmation.classList.contains('componentClass')).toBe(
+      true
+    );
   });
 
   it('it calls correct callback on confirmation', async () => {
@@ -81,7 +82,7 @@ describe('DeletionConfirmation', () => {
 
   it('it calls correct callback on cancellation', async () => {
     const { container } = render(<DeletionConfirmation {...props} />);
-    const cancelLabel = container.querySelector('.clickable') as HTMLElement;
+    const cancelLabel = container.querySelector('.cancel') as HTMLElement;
 
     await userEvent.click(cancelLabel);
 
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
@@ -43,8 +43,8 @@ describe('DeletionConfirmation', () => {
     const { container } = render(
       <DeletionConfirmation
         {...props}
-        confirmationText="Remove"
-        cancellationText="Do not remove"
+        confirmText="Remove"
+        cancelText="Do not remove"
         warningText="Do you want to remove?"
       />
     );
@@ -64,7 +64,10 @@ describe('DeletionConfirmation', () => {
       <DeletionConfirmation {...props} className="componentClass" />
     );
 
-    expect(container.querySelector('.componentClass')).toBeTruthy();
+    const deletionConfirmation = container.firstChild as HTMLElement;
+    expect(
+      deletionConfirmation.classList.contains('.componentClass')
+    ).toBeTruthy();
   });
 
   it('it calls correct callback on confirmation', async () => {
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
@@ -39,13 +39,13 @@ describe('DeletionConfirmation', () => {
     const { container } = render(<DeletionConfirmation {...props} />);
 
     const deleteButton = container.querySelector('button') as HTMLButtonElement;
-    expect(deleteButton?.textContent).toBe('Remove');
+    expect(deleteButton?.textContent).toBe(props.confirmText);
 
     const cancelLink = container.querySelector('.cancel') as HTMLElement;
-    expect(cancelLink?.textContent).toBe('Do not remove');
+    expect(cancelLink?.textContent).toBe(props.cancelText);
 
     const warningText = container.querySelector('.warningText') as HTMLElement;
-    expect(warningText?.textContent).toBe('Do you want to remove?');
+    expect(warningText?.textContent).toBe(props.warningText);
   });
 
   it('applies custom class name passed from the parent', () => {
@@ -62,7 +62,7 @@ describe('DeletionConfirmation', () => {
     );
   });
 
-  it('applies alignment class set from the parent', () => {
+  it('applies alignment set from the parent', () => {
     const { container, rerender } = render(<DeletionConfirmation {...props} />);
 
     const deletionConfirmation = container.firstChild as HTMLElement;
--- a/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
+++ b/src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
@@ -62,6 +62,20 @@ describe('DeletionConfirmation', () => {
     );
   });
 
+  it('applies alignment class set from the parent', () => {
+    const { container, rerender } = render(<DeletionConfirmation {...props} />);
+
+    const deletionConfirmation = container.firstChild as HTMLElement;
+    expect(
+      deletionConfirmation.classList.contains('deleteConfirmationContainer')
+    ).toBe(true);
+
+    rerender(<DeletionConfirmation {...props} alignContent="right" />);
+    expect(deletionConfirmation.classList.contains('contentAlignRight')).toBe(
+      true
+    );
+  });
+
   it('it calls correct callback on confirmation', async () => {
     const { container } = render(<DeletionConfirmation {...props} />);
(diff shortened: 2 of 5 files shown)


src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  it('calls cancelDeletion onClick when clicked', async () => {
+    const { container } = render(<DeletionConfirmation {...props} />);
+    const cancelLabel = container.querySelector('.clickable') as HTMLElement;
+
+    await userEvent.click(cancelLabel);
+
+    expect(props.onCancel).toHaveBeenCalled();
+  });
+});
```
azangru, line 61: here's the behaviour that you have in your component:

```
- it renders with default messages
- it applies custom messages provided by the parent
  - here, you would check, in a single test, the text of the warning, of the button, and of the cancel element
- it applies custom class name passed from the parent
- it calls correct callback on confirmation
- it calls correct callback on cancellation
```

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import React from 'react';
+import classNames from 'classnames';
+
+import { PrimaryButton } from 'src/shared/components/button/Button';
+
+import styles from './DeletionConfirmation.scss';
+
+export type DeletionConfirmationProps = {
+  message?: string;
```
azangru, line 25: warningText?

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,58 @@
...
+  confirmButtonLabel?: string;
```
azangru, line 26: confirmationText?

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,58 @@
...
+  cancelLabel?: string;
```
azangru, line 27: cancellationText

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,58 @@
...
+  onConfirm: () => void;
+  onCancel: () => void;
+  className?: string;
+};
+
+const DeletionConfirmation = (props: DeletionConfirmationProps) => {
+  const containerClass = classNames(
+    styles.deleteMessageContainer,
```
azangru, line 35: `deleteMessageContainer` -> `deleteConfirmation` (name of the whole element)

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
+@import 'src/styles/common';
```
azangru, line 1: Think about which CSS properties in this component need to be customizable, and express them as CSS variables.

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: its already been used in the file with the class `.clickable`

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: ah sorry i understood what you meant now, to make use of the --var. will update the code

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,58 @@
...
+import classNames from 'classnames';
+
+import { PrimaryButton } from 'src/shared/components/button/Button';
+
+import styles from './DeletionConfirmation.scss';
+
+export type DeletionConfirmationProps = {
+  warningText?: string;
+  confirmationText?: string;
+  cancellationText?: string;
```
jyothishnt, line 27: Minor, but do you to change it to `confirmText`/`cancelText`?

jyothishnt (approved):

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: i prefer `confirmText` / `cancelText` but lets leave it as it is since it is another variable debate which wouldn't make much difference here.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
azangru: I preferred confirmText and cancelText as well, but convinced myself to call them confirmation and cancellation, as nouns. Will be happy if you change them to confirm and cancel texts.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
...
+
+.deleteConfirmationContainer {
+  display: flex;
+  align-items: center;
+  justify-content: var(--container-justify-content, flex-end);
```
azangru, line 6: You shouldn't need this CSS rule.

ens-ma7: But what if we want the text to be at the start, do we not need to overwrite this?

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
...
+  column-gap: var(--container-column-gap, 30px);
+  margin: var(--container-margin, 20px 0px);
```
azangru, line 8: No; you should not set a margin on the container element here. You are building a new component, which should be adaptable to its environment as well as it can be. Margin comes from a specific context in which it is used. It is not a general property of this component.

If you want to add a margin, you will pass it set in the class name passed from the parent.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
+@import 'src/styles/common';
+
+.deleteConfirmationContainer {
+  display: flex;
+  align-items: center;
+  justify-content: var(--container-justify-content, flex-end);
+  column-gap: var(--container-column-gap, 30px);
```
azangru, line 7: Here and below: `--container-column-gap` is a poor name for a CSS variable, because it isn't precise enough. Which "container"? CSS variables are global; the overrides — if they come — will come from the parent. Would you expect the parent to say `--container-column-gap` and mean "the distance between the delete button and what's to the left and to the right of it"? I wouldn't.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
...
+  margin: var(--container-margin, 20px 0px);
+}
+
+.warningText {
+  color: var(--warning-text-color, #{$red});
```
azangru, line 12: `--deletion-confirmation-warning-color`

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.clickable {
+  color: var(--clickable-color, #{$blue});
```
azangru, line 16: `--deletion-confirmation-cancel-color`

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+describe('DeletionConfirmation', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('renders with default message', () => {
+    const { container } = render(<DeletionConfirmation {...props} />);
+
+    const deleteButton = container.querySelector('button') as HTMLButtonElement;
+    expect(deleteButton?.textContent).toBe('Delete');
```
azangru, line 39: What's the warning message? What's the cancel message? Why are we checking specifically the button message, but not the other two?

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const warningText = container.querySelector('.warningText') as HTMLElement;
+    expect(warningText?.textContent).toBe('Do you want to remove?');
+  });
+
+  it('applies custom class name passed from the parent', () => {
+    const { container } = render(
+      <DeletionConfirmation {...props} className="componentClass" />
+    );
+
+    expect(container.querySelector('.componentClass')).toBeTruthy();
```
azangru, line 67: 1) Find the top-level component element

`const deletionConfirmation = container.firstChild`

2) Make sure it has the class name that you passed in:

`expect(deletionConfirmation.classList.has('componentClass')).toBe(true)`.

Otherwise, it isn't obvious from the test what element the provided class gets applied to

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: but the css will be inside the component css and will have the component name attached to the class. I am not sure why you will need the component name with the css variable. the way we used our css variable is closely coupled with the react component so dont see where the confusion will be

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: i am making the gap fixed (not customisable)

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
ens-ma7: The warning message or the cancel message can be anything, why are we checking for the text? Should we not be checking if the component renders. 
I am a bit confused as to why we need to test these default text. i thought we dont check the exact text in a component?

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: same as above?

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
azangru: >  why are we checking for the text

I don't know. Why are you checking the text of the delete button? It can be anything as well.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
azangru: > but the css will be inside the component css and will have the component name attached to the class

If you use CSS variables to modify the behaviour of your component, it means that the component that will actually set this variable will be one of the parents of your component.

image

ens-ma7: Agreed from slack discussion that css variable should be unique so that we can identify them further up in the tree and also we can more than 1 variable doing the same thing (changing color) but in different component

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: Agreed this is not needed as a variable.

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -0,0 +1,18 @@
...
+  justify-content: var(--container-justify-content, flex-end);
+  column-gap: var(--container-column-gap, 30px);
+  margin: var(--container-margin, 20px 0px);
+}
+
+.warningText {
+  color: var(--warning-text-color, #{$red});
+}
+
+.clickable {
```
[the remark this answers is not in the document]
ens-ma7: ✔

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
```diff
@@ -3,15 +3,15 @@
 .deleteConfirmationContainer {
   display: flex;
   align-items: center;
-  justify-content: var(--container-justify-content, flex-end);
+  justify-content: flex-end;
```
azangru, line 6: I still don't get it — what if you removed this CSS rule at all?

If the problem is that the component behaves like this:

image

and you are trying to make it to behave like this:

image

then why not make it behave like this instead:

image

by changing the `display` property from `flex` to `inline-flex`?

jyothishnt (dismissed):
Things have changed since I approved. We need to finalise the position of the component.
Component should not worry about the position, parent should take care of it.

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { PrimaryButton } from 'src/shared/components/button/Button';
+
+import styles from './DeletionConfirmation.scss';
+
+export type DeletionConfirmationProps = {
+  warningText: string;
+  confirmText: string;
+  cancelText: string;
+  contentAlignRight?: boolean;
```
azangru, line 28: `contentAlignRight?: boolean;` -> `alignContent: 'left' | 'right'`

...

```
const containerClass = classNames(
    styles.deleteConfirmationContainer,
   {[styles.contentAlignRight]: props.alignContent === 'right' },
    props.className
);
```

ens-ma7: so my thinking to this is by default the content will be aligned left and hence no need to specify this. And hence making the alignRight optional, its whether you want to align it right or not. Less props to pass through when you need a simple component.

azangru: Yes, I know what you mean.

For my part, I though the name `contentAlignRight` slightly odd; and wondered what if we thought of this property in the same way we think about CSS properties, for which there is a default value (in this case, "left"); and other values (in this case, "right"). Passing the value "left" won't do anything; whereas passing the value "right" would change the arrangement of the element.

But I understand the reluctance to having a value that doesn't do anything. Let's see what the rest of the team thinks at scrum.

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import DeletionConfirmation, {
+  DeletionConfirmationProps
+} from './DeletionConfirmation';
+
+const props: DeletionConfirmationProps = {
+  confirmText: 'Remove',
```
jyothishnt, line 26: Component and props talk about Delete and why the text says Remove?

ens-ma7: Its just a test and we are checking if we can change the text to anything else (can be delete or it can be any other text) as this is part of the component props.
The text and the props dont mention delete.

jyothishnt: Ah yes, good point!

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
msmanoj: I think `alignContent` would be a better option.

msmanoj: Looking at the css [here](https://github.com/Ensembl/ensembl-client/pull/828/files#diff-7fce8969ad25904d78fb32760364f592a85b2b3328655b938dc54e989528e940R12), I wonder if this should be called `justifyContent` instead of `alignContent `?

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+describe('DeletionConfirmation', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('applies custom messages provided by parent', () => {
+    const { container } = render(<DeletionConfirmation {...props} />);
+
+    const deleteButton = container.querySelector('button') as HTMLButtonElement;
+    expect(deleteButton?.textContent).toBe('Remove');
```
msmanoj, line 42: Here and below, `'Remove'` -> props.confirmText?

src/shared/components/deletion-confirmation/DeletionConfirmation.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      deletionConfirmation.classList.contains('deleteConfirmationContainer')
+    ).toBe(true);
+
+    rerender(<DeletionConfirmation {...props} className="componentClass" />);
+    expect(deletionConfirmation.classList.contains('componentClass')).toBe(
+      true
+    );
+  });
+
+  it('applies alignment class set from the parent', () => {
```
msmanoj, line 65: `applies alignment class set from the parent` - > `uses the alignment set by the parent`?

azangru (approved):

src/shared/components/deletion-confirmation/DeletionConfirmation.scss
ens-ma7: Agreed to leave some of the alignment/display to the parent but also make it as an optional props whether to align the content right or not

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: Again i am not sure to me `alignContent` is clearer than `justifyContent`

msmanoj (approved):

src/shared/components/deletion-confirmation/DeletionConfirmation.tsx
ens-ma7: Agreed to go with `alignContent` since it describes what we are trying to achieve. `justifyContent` is specific to css and also with display flex.

PR #831, opened 2022-09-19
Download blast submission

azangru:
## Description
Download all files for a BLAST submission when the user presses the "Download" button in the header

TODO:
- [x] Update name of the zip file (called "test" at the moment)
- [x] Update directory names for sequences
- [x] Reuse same function for fetching all raw output files
- [x] Change the behaviour of `downloadAsFile` function

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1734

## Deployment URL(s)
http://download-blast-submission.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
+++ b/src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
@@ -45,7 +45,11 @@ export const createCSVForGenomicBlast = (blastJobResult: BlastJobResult) => {
         queryStart,
         queryEnd
       } = getCommonFields({ blastHsp: hsp });
-      const genomicLocation = `${hit.hit_acc}:${hsp.hsp_hit_from}-${hsp.hsp_hit_to}`;
+      const genomicStart =
+        hitOrientation === 'Forward' ? hsp.hsp_hit_from : hsp.hsp_hit_to;
+      const genomicEnd =
+        hitOrientation === 'Forward' ? hsp.hsp_hit_to : hsp.hsp_hit_from;
+      const genomicLocation = `${hit.hit_acc}:${genomicStart}-${genomicEnd}`;
       const newRow = [
         evalue,
         hitLength,
@@ -172,7 +176,7 @@ const getCommonFields = ({
   hitLength: blastHsp.hsp_align_len,
   percentId: blastHsp.hsp_identity,
   score: blastHsp.hsp_bit_score,
-  hitOrientation: blastHsp.hsp_hit_frame,
+  hitOrientation: blastHsp.hsp_hit_frame === '1' ? 'Forward' : 'Reverse',
   hitStart: blastHsp.hsp_hit_from,
   hitEnd: blastHsp.hsp_hit_to,
   queryStart: blastHsp.hsp_query_from,
--- a/src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
+++ b/src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
@@ -1,0 +1,186 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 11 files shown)


src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
```diff
@@ -0,0 +1,184 @@
...
+        hitLength,
+        percentId,
+        score,
+        hitOrientation,
+        hitStart,
+        hitEnd,
+        queryStart,
+        queryEnd
+      } = getCommonFields({ blastHsp: hsp });
+      const genomicLocation = `${hit.hit_acc}:${hsp.hsp_hit_from}-${hsp.hsp_hit_to}`;
```
msmanoj, line 48: Should the position of `from` and `to` values be based on whichever is higher?

src/content/app/tools/blast/blast-download/submissionDownload.ts
```diff
@@ -0,0 +1,185 @@
...
+    speciesFolderNamesMap.set(species.genome_id, folderName);
+  }
+
+  const zip = new JSZip();
+  const rootFolder = zip.folder(getNameForZipRoot(submission));
+
+  for (const blastResult of results) {
+    const { genomeId, sequenceId, csv, raw } = blastResult;
+    const csvFileName = 'table'; // NOTE: this will probably change
+    const rawFileName = 'output'; // NOTE: this will probably change
```
msmanoj, line 151: Should this be `alignments` as per L49?

azangru: You mean the name of the file? Probably; but it contains more than just alignments... It's the raw output of the tool. We can call it alignments for now since this is what Lucy wanted, but it's more than just that.

msmanoj: `output` is okay here.  It's just the comments above mention that this file will be called `alignments.txt`.

azangru: The comment above is a very rough approximation of the contents of the files.

But I can rename the file to `alignments` here if it helps.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -130,18 +119,62 @@ const blastApiSlice = restApiSlice.injectEndpoints({
         }
       }
     }),
-    fetchBlastSubmission: builder.query<BlastJobResultResponse, string>({
-      query: (jobId) => ({
-        url: `${config.toolsApiBaseUrl}/blast/jobs/result/${jobId}/json`
-      })
+    fetchAllBlastRawResults: builder.query<{ result: string }[], string[]>({
+      queryFn: async (jobIds, _queryApi, __extraOptions, baseQuery) => {
```
msmanoj, line 123: Out of curiosity, what is the difference between using `_` &  `__`?

src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
```diff
@@ -0,0 +1,184 @@
...
+  const rows = [columnNames] as Array<string | number>[];
+
+  for (const hit of blastJobResult.hits) {
+    for (const hsp of hit.hit_hsps) {
+      const {
+        evalue,
+        hitLength,
+        percentId,
+        score,
+        hitOrientation,
```
msmanoj, line 42: Appears as `1`/`-1` in the CSV. Should it be `forward` / `reverse` instead?

src/content/app/tools/blast/blast-download/submissionDownload.ts
msmanoj: No, don't worry. I just brought this up just to check if the file name is correct.

msmanoj:
For a job with ~5000 hits, it took ~21 seconds for the download to begin after clicking the button. It looks like we need some sort of an indication that the file is getting prepared. 

Also, I think we need to temporarily disable the button while it is generating the zip file to prevent spam clicking.

azangru:
> For a job with ~5000 hits, it took ~21 seconds for the download to begin after clicking the button. It looks like we need some sort of an indication that the file is getting prepared.
> 
> Also, I think we need to temporarily disable the button while it is generating the zip file to prevent spam clicking.
> 
> Not sure if these should be addressed in this PR.

Both valid concerns; but none are addressed in this PR. I am waiting for Andrea to come up with a UI. Depending on its complexity, it may be added either in this PR or in a subsequent one.

src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
azangru: ✅

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
azangru: ✅

src/content/app/tools/blast/blast-download/createBlastCSVTable.ts
azangru: ✅

msmanoj (approved):

PR #833, opened 2022-09-21
Skip request to track categories if genomeId is null

msmanoj:
## Description
- this PR prevents the call to track categories endpoint when genomeId is null

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1729

## Deployment URL(s)
http://fix-null-track-categories.review.ensembl.org

## Views affected
- Genome browser interstitial

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
@@ -72,7 +72,10 @@ const useGenomeBrowserTracks = () => {
     // skip the effect if track categories have not yet been fetched
     // or if they have already been set
     if (
-      !(trackCategories && focusObject && trackSettingsForGenome && genomeId)
+      !trackCategories ||
+      !focusObject ||
+      trackSettingsForGenome ||
+      !genomeId
     ) {
       return;
     }
--- a/src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
@@ -54,7 +54,9 @@ import type { FocusObject } from 'src/shared/types/focus-object/focusObjectTypes
 
 const useGenomeBrowserTracks = () => {
   const genomeId = useAppSelector(getBrowserActiveGenomeId) as string;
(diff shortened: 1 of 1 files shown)


src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
```diff
@@ -54,7 +54,9 @@ import type { FocusObject } from 'src/shared/types/focus-object/focusObjectTypes
 
 const useGenomeBrowserTracks = () => {
   const genomeId = useAppSelector(getBrowserActiveGenomeId) as string;
```
azangru, line 56: `as string` is lying to us here; try removing it

msmanoj: That's done

src/content/app/genome-browser/hooks/useGenomeBrowserTracks.ts
```diff
@@ -66,7 +71,9 @@ const useGenomeBrowserTracks = () => {
   useEffect(() => {
     // skip the effect if track categories have not yet been fetched
     // or if they have already been set
-    if (!trackCategories || !focusObject || trackSettingsForGenome) {
+    if (
+      !(trackCategories && !focusObject && trackSettingsForGenome && genomeId)
```
imransl, line 75: `focusObject` here is getting negated twice here. I wonder if this makes the check a bit more confusing to read.

azangru, line 75: image

msmanoj: I've remove the extra `!`

azangru: I'm sorry; but the change is just not the same:

image

The original comment says:

```
// skip the effect if track categories have not yet been fetched
// or if they have already been set
```
And the original condition was saying to return if `trackSettingsForGenome` were truthy, not falsey

msmanoj: oops! Sorry, I think I should have avoided this change. I've reverted it and just included a check for `genomeId`

azangru (approved):

PR #834, opened 2022-09-21
Fix GB in-app search bug selecting current feature

msmanoj:
## Description
This PR fixes the issue with the disappearing content in the horizontal bar above the genome browser when the user tries to navigate to a feature that is currently active (by searching for it and clicking on a search result). This fix however, does not prevent the location information in the URL from getting lost but this does not cause any problem.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1721

## Deployment URL(s)
http://fix-gb-in-app-search-bug.review.ensembl.org

## Views affected
- Genome browser bar

### DIFF
--- a/src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
+++ b/src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
@@ -324,6 +324,8 @@ const browserGeneralSlice = createSlice({
 
       if (chrLocation) {
         state.chrLocations[activeGenomeId] = chrLocation;
+      } else {
+        delete state.chrLocations[activeGenomeId];
       }
     },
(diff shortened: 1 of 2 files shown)


src/content/app/genome-browser/state/browser-general/browserGeneralSlice.ts
```diff
@@ -324,8 +324,6 @@ const browserGeneralSlice = createSlice({
 
       if (chrLocation) {
         state.chrLocations[activeGenomeId] = chrLocation;
-      } else {
-        delete state.chrLocations[activeGenomeId];
```
azangru, line 328: Umm, this doesn't look right 😞

azangru (approved):

PR #835, opened 2022-09-23
disable species lozenge on blast small screens

jyothishnt:
## Description
For blast, on small screen, we need to disable all lozenge (when the Add sequence/ Select species buttons appear)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1685

## Deployment URL(s)
http://disable-species-blast.review.ensembl.org/

## Views affected
Blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.scss
+++ b/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.scss
@@ -1,0 +1,7 @@
+@import 'src/styles/common';
+
+.speciesLozenge {
+  @media (max-width: 1900px) {
+    --species-lozenge-bg-color: #{$grey};
+  }
+}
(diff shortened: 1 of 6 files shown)


azangru (approved):

imransl (approved):

msmanoj (approved):

ens-ma7:
sorry to add to this late but the behavior is wrong. According to jira and also to my understanding, the species lozenge should be reenable when moving to the species selection.

image

azangru:
> According to jira and also to my understanding, the species lozenge should be reenable when moving to the species selection.

Oh. I didn't remember that. It makes sense. Shame; the CSS solution was so simple and elegant.

azangru:
@jyothishnt @ens-ma7 I've pushed a commit to address Ridwan's point above.

- Moved back to a JS option
- Disabling the lozenges in a single-column view for the sequence step and enabling them for the species step
- Extracted the breakpoint into a shared js variable that gets reused in two different components, to keep consistency
- Updated/fixed some CSS related to the right page padding

Right padding before:

https://user-images.githubusercontent.com/6834224/192834251-b5f4abb4-ede3-4461-bd68-5737f19a3ec6.mov

Right padding after:

https://user-images.githubusercontent.com/6834224/192834292-6db7ebf1-9485-450b-b77e-91f5b24f4fa9.mov

jyothishnt:
Looks good to me

ens-ma7 (approved):

PR #836, opened 2022-09-24
Update the types and code of the ViewInApp component

azangru:
## Description
This is a refactoring PR that does not add any extra functionality or fix any bugs. It updates the api of the `ViewInApp` component, such that:
- the buttons shown in the `ViewInApp` component are determined only by the keys of the `links` property
- a member of the `links` property can be either an object describing a link, or a function that will perform the navigation task

### Rationale
The fields of ViewInAppProps have the following semantics:
- the buttons rendered by the ViewInApp component,  as well as the default behaviour upon a click on a given button, are determined by the keys of the `links` object passed as a mandatory property
- the optional `onAppClick` property describes additional behaviour upon a click on a given button
- the optional `onAnyAppClick` property describes additional behaviour upon a click on any button.

**QUESTION**: why not make the `links` property only ever contain url strings, and pass functions describing more complex on-click behaviour only via the `onAppClick` property?

**ANSWER**: because the `links` property is mandatory. Consider an edge case in which the on-click behaviour on any button has to be described in a function. That would require the consumer of the component to pass an empty `links` object into the component, which makes for an awkward api.

## Deployment URL(s)
http://update-view-in-app.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/components/zmenu/ZmenuAppLinks.tsx
+++ b/src/content/app/genome-browser/components/zmenu/ZmenuAppLinks.tsx
@@ -23,7 +23,9 @@ import useGenomeBrowserIds from 'src/content/app/genome-browser/hooks/useGenomeB
 import useGenomeBrowser from '../../hooks/useGenomeBrowser';
 
 import { ToggleButton as ToolboxToggleButton } from 'src/shared/components/toolbox';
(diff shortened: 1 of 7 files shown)


msmanoj (approved):
Looks good 👍

PR #837, opened 2022-09-26
Add tests for DataTable

msmanoj:
## Description
- Added some simple tests to the DataTable components

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1728

## Deployment URL(s)
http://data-table-tests.review.ensembl.org


src/shared/components/data-table/DataTable.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+): { data: TableData; columns: DataTableColumns } => {
+  return {
+    data: times(rows, (row) =>
+      times(columns, (column) => `Cell ${row},${column}`)
+    ),
+    columns: times(columns, (column) => ({
+      columnId: `${column}`,
+      title: `Column ${column}`,
+      helpText: `Column ${column} help text`,
+      isSortable: true
```
azangru, line 40: Which of these fields are optional? Since you have a test for rendering without errors, I would use it to render the table with a just a minimum required amount of props.

src/shared/components/data-table/DataTable.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    container = renderDataTable({
+      state: {
+        ...defaultProps.state,
+        rowsPerPage: Infinity
+      }
+    }).container;
+
+    expect(container.querySelectorAll('tbody tr').length).toBe(
+      defaultProps.state.data.length
+    );
```
azangru, line 100: No difference from the check above?

src/shared/components/data-table/DataTable.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    expect(container.querySelectorAll('tbody tr').length).toBe(5);
+  });
+
+  it('respects onStateChange', async () => {
+    const onStateChange = jest.fn();
+    container = renderDataTable({
+      onStateChange
+    }).container;
+
+    expect(onStateChange).toBeCalledWith(defaultProps.state);
```
azangru, line 119: I am not even seeing a state being changed here 🤔

src/shared/components/data-table/DataTable.tsx
```diff
@@ -33,12 +33,12 @@ import {
...
-type TableContextType = DataTableState & {
+export type TableContextType = DataTableState & {
   dispatch: React.Dispatch<AllTableActions>;
   columns: DataTableColumns;
-  theme: TableTheme;
-  selectableColumnIndex: number;
-  expandedContent: { [rowId: string]: ReactNode };
+  theme?: TableTheme;
+  selectableColumnIndex?: number;
+  expandedContent?: { [rowId: string]: ReactNode };
```
azangru, line 41: Ah, this reminds me. The `ReactNode` type is an output type of a React component; it suggests that React had to prerender the expanded content for _all_ rows before the user even expands a row. I think expanded content should be rendered lazily and on demand rather than all at once beforehand.

src/shared/components/data-table/components/main/components/table-row/TableRow.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  it('renders without error', () => {
+    container = renderTableRow().container;
+    expect(() => container).not.toThrow();
+  });
+
+  it('respects hiddenRowIds and renders nothing', () => {
+    container = renderTableRow({
+      hiddenRowIds: { [defaultProps.rows[0].rowId]: true }
```
azangru, line 70: Didn't have this question earlier; but looking at the test now — why is this shaped as a `Record<string, boolean>`? The `false` value wouldn't make sense here, would it? This looks something like a list of ids. In the old days, we would have used an array for this; but these days, a Set would be a better option.

src/shared/components/data-table/components/main/components/table-row/TableRow.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }).container;
+
+    expect(container.querySelector('tr')).toBeFalsy();
+
+    container = renderTableRow({
+      searchText: '1'
+    }).container;
+
+    expect(container.querySelector('tr')).toBeTruthy();
+  });
```
azangru, line 136: I'm uneasy about this. Why is a row responsible for search — why isn't it the table's responsibility?

src/shared/components/data-table/components/table-controls/components/table-actions/components/find-in-table/FindInTable.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      payload: '1'
+    });
+
+    expect(defaultProps.dispatch).toHaveBeenNthCalledWith(3, {
+      type: 'set_search_text',
+      payload: '3'
+    });
+  });
+
+  it('click on cancel link dispatched the correct action', async () => {
```
azangru, line 78: this doesn't read like a sentence ("it click on cancel link..." — ?)

src/shared/components/data-table/DataTable.test.tsx
msmanoj: Here we check if the correct number of rows are displayed after setting `rowsPerPage` to `Infinity` (it was `20` in the previous check)

src/shared/components/data-table/DataTable.test.tsx
msmanoj: Here it just checks if `onStateChange` is called with the currentState when the table loads initially. 
I am not sure what is the best way to trigger a change and then test it.

src/shared/components/data-table/DataTable.tsx
msmanoj: Currently, `expandedContent` gets updated only on demand.

src/shared/components/data-table/components/main/components/table-row/TableRow.test.tsx
msmanoj: I agree that Set would fit better here. I'll update it

src/shared/components/data-table/components/main/components/table-row/TableRow.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    columns[indexOfColumnWithRenderer].renderer = renderer;
+
+    container = renderTableRow({
+      columns
+    }).container;
+
+    expect(renderer).toBeCalledWith({
+      rowData: defaultProps.rows[0].cells,
+      rowId: String(defaultProps.rows[0].rowId),
+      cellData: defaultProps.rows[0].cells[indexOfColumnWithRenderer]
```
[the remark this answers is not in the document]
msmanoj: I pass the rowData because the data displayed in a particular column could be derived by combining multiple columns.
I think it can be updated to be the entire table data when necessary.

src/shared/components/data-table/DataTable.test.tsx
azangru: > Here it just checks if onStateChange is called with the currentState when the table loads initially.

That's my point: there was no change, but the onStateChange callback was triggered.

src/shared/components/data-table/DataTable.test.tsx
azangru: I know 😄 

- We start with 10, right? Do you want to skip `rowsPerPage` in one of the checks to see the default behaviour?
- The behaviour in the Infinity check is no different from the behaviour in the 20 rows per page check. Perhaps generate the number of rows that is above the maximum number offered by the UI (over 100), and make sure all rows are present?

src/shared/components/data-table/components/main/components/table-row/TableRow.test.tsx
msmanoj: Moved the search handler to `useDataTable` hook.

src/shared/components/data-table/components/table-controls/components/table-actions/components/find-in-table/FindInTable.test.tsx
msmanoj: Updated it

src/shared/components/data-table/DataTable.test.tsx
msmanoj: I've updated it to cover the following scenarios:

Total rows: 15, rowsPerPage: default, Displayed rows should be: 10
Total rows: 15, rowsPerPage: Infinity, Displayed rows should be: 15
Total rows: 15, rowsPerPage: default, currentPageNumber:2, Displayed rows should be: 10

src/shared/components/data-table/DataTable.test.tsx
azangru: I think so. Something native like an input element or a textarea element don't fire a change event when they are rendered with a provided value.

I wonder what should happen if the parent is in control of the state and decides to change it. The table probably shouldn't tell the parent back that the state has changed either; because the parent already knows it.

msmanoj: I have pushed an update to call `onStateChange` only on state change

src/shared/components/data-table/DataTable.tsx
```diff
@@ -63,13 +63,17 @@ const DataTable = (props: TableProps) => {
...
     initialDataTableState
   );
 
   useEffect(() => {
-    props.onStateChange?.(tableState);
+    if (!firstRenderRef.current) {
+      props.onStateChange?.(tableState);
+    }
+    firstRenderRef.current = false;
   }, [tableState]);
```
azangru, line 77: Not liking this idea, with useEffect and the firstRenderRef.

Could we pull off something like this: create a wrapper around the reducer function, which behaves like the reducer, but also notifies the parent of the updated state:

```ts
// use this in the useReducer instead of the tableReducer
const reducerWithCallback: typeof tableReducer = (state, action) => {
  const newState = tableReducer(state, action);

  // if there are some actions when we want not to notify the parent
  // (e.g. if the state has updated because the parent said so),
  // then skip the callback. Plus, don't notify parent if the state hasn't changed.
  if (newState !== state) {
    props?.onStateChange(newState);  
  }

  return newState;
}
```

src/shared/components/data-table/DataTable.tsx
```diff
@@ -79,12 +79,23 @@ const DataTable = (props: TableProps) => {
         payload: initialState
       });
     }
+
+    /* 
+      Here we reset shouldResetStateRef to false so that the
```
azangru, line 84: to true 😄

azangru (approved):
I've checked the latest DataTable code, and am satisfied that it works. The internals of the DataTable component have now become confusing, in order to support updates of the table state both within the table component, and from the parent component — the code now uses a couple of flags to signal when to update the table state from the parent, and when to ignore the parent. It looks like, theoretically, it is should possible to break the table code if the parent ignores any given `onStateChange` callback and tries to update the table state at some later time, when the flag has not yet been removed.

The discussion on Slack that has prompted this change is:

image

We may want to revisit the DataTable component if we think of something better. Meanwhile, however, the component is working well enough to be merged.

azangru:
There are still problems with the data table; but at they also exist in the dev branch; so they aren't regressions.

I'm going to merge this PR.

PR #839, opened 2022-09-27
Add loading states to the DownloadButton

azangru:
## Description
- DownloadButton will cycle through the following download states:
  - not_requested (regular button appearance)
  - loading (spinner)
  - success (green checkmark) or error (red cross)
- While the button is in the loading, success, or error state, it will not register additional clicks and thus won't start unnecessary downloads

**Note:** visual appearance of the button in different states is somewhat different from XD; but was discussed with and approved by Andrea.

### Questions to consider for subsequent PRs
- `DownloadButton` now is very similar to the `LoadingButton`; do they need to be merged together?
- `DownloadButton` can't avoid showing the spinner for near-instantaneous downloads; not does it know how to show the spinner for a minimal duration to avoid unpleasant flickering (can be addressed in a dedicated PR) 
- What should happen if the user starts the download, switches to a different screen before it finishes, and then comes back?

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1749

## Deployment URL(s)
http://submission-download-ui.review.ensembl.org

## Views affected
- BLAST submissions list view
- BLAST submission details view

### DIFF
--- a/src/shared/components/download-button/DownloadButton.scss
+++ b/src/shared/components/download-button/DownloadButton.scss
@@ -1,15 +1,15 @@
 @import 'src/styles/common';
 
 .downloadButton {
+  --_download-button-size: var(--download-button-size, 18px);
   background-color: var(--download-button-background, #{$blue});
   border: none;
   outline: none;
-  height: var(--download-button-size, 18px);
-  width: var(--download-button-size, 18px);
+  height: var(--_download-button-size);
+  width: var(--_download-button-size);
 
   svg {
-    height: 10px;
-    width: 10px;
+    width: 56%; // about 10px for a 18px-wide button
     fill: $white;
   }
 }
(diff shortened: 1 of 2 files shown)


src/shared/components/download-button/DownloadButton.tsx
```diff
@@ -15,34 +15,114 @@
...
+    }
+  };
+
+  // const { className, ...otherProps } = props;
+
+  // const elementClasses = classNames(
+  //   styles.downloadButton,
+  //   { [styles.downloadButtonDisabled]: props.disabled },
+  //   className
+  // );
```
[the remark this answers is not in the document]
azangru: ✅

imransl (approved):

src/shared/components/download-button/DownloadButton.scss
```diff
@@ -18,3 +18,45 @@
...
+.downloadButtonError {
+  background-color: var(--download-button-success-background, transparent);
+  border: none;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+
+  svg {
+    height: auto;
+  }
```
msmanoj, line 44: Are all these styles needed?

src/shared/components/download-button/DownloadButton.tsx
```diff
@@ -15,34 +15,106 @@
...
+    default:
+      return <DownloadIcon />;
+  }
+};
+
+const ControlledDownloadButtonWithForwardedRef = forwardRef(
+  ControlledDownloadButton
+);
+
+export { ControlledDownloadButtonWithForwardedRef as ControlledDownloadButton };
```
msmanoj, line 119: Why do we need to support/use `ref` in this component?

azangru: `DownloadButton` is a shared component that can be used in various places; so I am trying to treat it as if it were just a regular HTML button. Which means that, since I can access a regular `button` element in the DOM by passing a `ref` to a React component representing it, I want to also be able to access the `button` element via a ref passed to the `DownloadButton` component. It is a useful thing to be able to do when you need it; and it's easy to set up.

src/shared/components/download-button/DownloadButton.scss
azangru: Checked again. `border:none` not required. Display inline-flex with the centering of the child isn't really necessary for the success button, but is needed for the error button, because the cross in the error state is smaller than the size of the button. I will leave the inline-flex for both success and error to avoid confusion.

✅

msmanoj (approved):

ens-st3 (approved):

PR #841, opened 2022-09-27
EXPERIMENTAL: Add a utility for running async visual tasks

azangru:
## Description
This PR adds a utility for wrapping potentially long-running functions such that:
- if the function completes quickly, nothing additional happens
- if the function takes a while to complete, the wrapper utility sends a signal for showing a loading indicator
- the wrapper utility will take some (configurable) time before sending a new signal to remove the loading indicator, thus avoiding the undesirable flickering of the loading indicator caused by too brief a time of showing it

This utility function can then be used to create a hook that can be used in individual components that display loading spinners (e.g. in the `LoadingButton` component); or to send signals to the redux store.

### DIFF
--- a/src/shared/hooks/tests/useAsyncTask.test.ts
+++ b/src/shared/hooks/tests/useAsyncTask.test.ts
@@ -1,0 +1,159 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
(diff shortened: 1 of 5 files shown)


imransl:
@azangru Went through this and like this approach. Can we use this for loading the genome browser as well?

PR #842, opened 2022-09-28
Remove duplicate import from useEntityViewerAnalytics

msmanoj:
## Description
- Remove duplicate import from `useEntityViewerAnalytics.ts`

### DIFF
--- a/src/content/app/entity-viewer/hooks/useEntityViewerAnalytics.ts
+++ b/src/content/app/entity-viewer/hooks/useEntityViewerAnalytics.ts
@@ -27,8 +27,8 @@ import {
 } from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSelectors';
 import { getCommittedSpeciesById } from 'src/content/app/species-selector/state/speciesSelectorSelectors';
 
-import {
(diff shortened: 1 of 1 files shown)


azangru:
While you are at it, can you annotate type imports (that are type aliases rather than enums). I can see at least RootState that could benefit from `import type`

msmanoj:
> While you are at it, can you annotate type imports (that are type aliases rather than enums)? I can see at least RootState that could benefit from `import type`. `AnalyticsOptions`, `Filters`, `SortingRule` seem to be type aliases as well.

That's done

azangru:
Aren't both of these type aliases?

image

If yes, then it's `import type {Filters, SortingRule}`

azangru (approved):
I just pushed a commit for my latest comment.

msmanoj:
Sorry, I missed to get back early. `SortingRule` is an enum.

azangru:
> `SortingRule` is an enum.

It's not used as a value here though. Not as `SortingRule.SOMETHING`.

Typescript was quite happy with the build after my latest commit.

PR #848, opened 2022-10-03
Download table data

msmanoj:
## Description
- Implemented "Download this table" option on DataTable and enable it on Blast result table

The DataTable can either handle the download by itself by generating a csv or it can give the control over to the parent component using the `downloadHandler` prop.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1736

## Deployment URL(s)
http://datatable-export.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/blast-download/submissionDownload.ts
+++ b/src/content/app/tools/blast/blast-download/submissionDownload.ts
@@ -168,7 +168,7 @@ const createZipArchive = async (submission: EnrichedBlastSubmission) => {
   return zip;
 };
 
-const getNameForZipRoot = (submission: BlastSubmission) => {
+export const getNameForZipRoot = (submission: BlastSubmission) => {
   const {
     id,
     submittedData: {
--- a/src/content/app/tools/blast/blast-download/submissionDownload.ts
+++ b/src/content/app/tools/blast/blast-download/submissionDownload.ts
@@ -100,7 +100,7 @@ const downloadBlastSubmission = async (
       csv = createCSVForGenomicBlast(job.data);
     } else if (blastedAgainst === 'cdna') {
       csv = createCSVForTranscriptBlast(job.data);
-    } else if (blastedAgainst === 'protein') {
+    } else if (blastedAgainst === 'pep') {
       csv = createCSVForProteinBlast(job.data);
     }
 
--- a/src/content/app/tools/blast/blast-download/submissionDownload.ts
+++ b/src/content/app/tools/blast/blast-download/submissionDownload.ts
@@ -168,7 +168,7 @@ const createZipArchive = async (submission: EnrichedBlastSubmission) => {
   return zip;
 };
 
-export const getNameForZipRoot = (submission: BlastSubmission) => {
+const getNameForZipRoot = (submission: BlastSubmission) => {
   const {
     id,
     submittedData: {
--- a/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
+++ b/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
@@ -25,8 +25,7 @@ import BlastHitsDiagram from 'src/content/app/tools/blast/components/blast-hits-
 import BlastSequenceAlignment from 'src/content/app/tools/blast/components/blast-sequence-alignment/BlastSequenceAlignment';
 
 import { createCSVForGenomicBlast } from 'src/content/app/tools/blast/blast-download/createBlastCSVTable';
-import { getNameForZipRoot } from 'src/content/app/tools/blast/blast-download/submissionDownload';
-import { downloadBlobAsFile } from 'src/shared/helpers/downloadAsFile';
+import { downloadTextAsFile } from 'src/shared/helpers/downloadAsFile';
 
 import type {
   BlastHit,
@@ -193,11 +192,7 @@ const SingleBlastJobResult = (props: SingleBlastJobResultProps) => {
       </div>
 
       {isExpanded && (
-        <HitsTable
-          jobResult={jobResult}
-          submission={submission}
-          species={speciesInfo}
-        />
+        <HitsTable jobResult={jobResult} submission={submission} />
       )}
     </div>
   );
@@ -206,15 +201,12 @@ const SingleBlastJobResult = (props: SingleBlastJobResultProps) => {
 type HitsTableProps = {
   jobResult: SingleBlastJobResultProps['jobResult'];
   submission: BlastSubmission;
-  species: Species;
 };
 const HitsTable = (props: HitsTableProps) => {
-  const { jobResult, submission, species } = props;
+  const { jobResult, submission } = props;
 
   const blastDatabase = submission.submittedData.parameters
     .database as DatabaseType;
-
-  const zipFileName = getNameForZipRoot(submission) + '.zip';
 
   const [tableState, setTableState] = useState<Partial<DataTableState>>({
     rowsPerPage: 100,
@@ -347,26 +339,9 @@ const HitsTable = (props: HitsTableProps) => {
   };
 
   const downloadHandler = async () => {
-    const JSZip = await import('jszip').then((module) => module.default); // use a dynamic import to split this library off in a separate chunk
     const csv = createCSVForGenomicBlast(jobResult.data);
 
-    const zip = new JSZip();
-    const rootFolder = zip.folder(getNameForZipRoot(submission));
-
-    const { sequenceId } = jobResult;
-
-    const speciesFolderName = `${species.scientific_name}-${species.genome_id}`;
-    const sequenceFolderName = `Query sequence ${sequenceId}`; // NOTE: this name will probably change as well
-    const csvFileName = 'table';
-
-    rootFolder?.file(
-      `${speciesFolderName}/${sequenceFolderName}/${csvFileName}.csv`,
-      csv
-    );
-
-    const blob = await zip.generateAsync({ type: 'blob' });
-
-    await downloadBlobAsFile(blob, zipFileName);
+    await downloadTextAsFile(csv, 'table.csv');
   };
 
   return (
@@ -384,7 +359,6 @@ const HitsTable = (props: HitsTableProps) => {
           TableAction.DOWNLOAD_SHOWN_DATA
         ]}
         downloadHandler={downloadHandler}
-        downloadFileName={zipFileName}
       />
     </div>
   );
--- a/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
+++ b/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
@@ -241,7 +241,7 @@ const HitsTable = (props: HitsTableProps) => {
           hitHsp.hsp_align_len,
           '', // view_alignment
           hitHsp.hsp_identity,
-          hitHsp.hsp_score,
+          hitHsp.hsp_bit_score,
           getDynamicColumnContent({
             hit,
             blastDatabase,
--- a/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
+++ b/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
@@ -24,7 +24,11 @@ import ShowHide from 'src/shared/components/show-hide/ShowHide';
 import BlastHitsDiagram from 'src/content/app/tools/blast/components/blast-hits-diagram/BlastHitsDiagram';
 import BlastSequenceAlignment from 'src/content/app/tools/blast/components/blast-sequence-alignment/BlastSequenceAlignment';
 
-import { createCSVForGenomicBlast } from 'src/content/app/tools/blast/blast-download/createBlastCSVTable';
+import {
+  createCSVForGenomicBlast,
+  createCSVForProteinBlast,
+  createCSVForTranscriptBlast
+} from 'src/content/app/tools/blast/blast-download/createBlastCSVTable';
 import { downloadTextAsFile } from 'src/shared/helpers/downloadAsFile';
 
 import type {
@@ -339,7 +343,14 @@ const HitsTable = (props: HitsTableProps) => {
   };
 
   const downloadHandler = async () => {
-    const csv = createCSVForGenomicBlast(jobResult.data);
+    let csv = '';
+    if (blastDatabase === 'dna') {
+      csv = createCSVForGenomicBlast(jobResult.data);
+    } else if (blastDatabase === 'cdna') {
+      csv = createCSVForTranscriptBlast(jobResult.data);
+    } else if (blastDatabase === 'pep') {
+      csv = createCSVForProteinBlast(jobResult.data);
+    }
 
     await downloadTextAsFile(csv, 'table.csv');
   };
@@ -397,6 +408,7 @@ const getDynamicColumnContent = (props: DynamicColumnContentProps) => {
 
   return (
     <span className={styles.nowrap}>
+      <span>hello</span>
       {`${hit.hit_acc}:${[hitHsp.hsp_hit_from, hitHsp.hsp_hit_to]
         .sort()
         .join('-')}`}
--- a/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
+++ b/src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
@@ -408,7 +408,6 @@ const getDynamicColumnContent = (props: DynamicColumnContentProps) => {
 
   return (
     <span className={styles.nowrap}>
-      <span>hello</span>
       {`${hit.hit_acc}:${[hitHsp.hsp_hit_from, hitHsp.hsp_hit_to]
         .sort()
         .join('-')}`}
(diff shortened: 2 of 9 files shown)


ens-st3:
Download works as expected.
What do we expect to happen after I've clicked the 'Download' button ? Screen to stay as it is, ie still showing the 'Download', or reset and 'Actions' in the drop down
Screenshot 2022-10-04 at 16 59 01
Screenshot 2022-10-04 at 16 56 08

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { TableAction } from 'src/shared/components/data-table/dataTableTypes';
+
+import styles from './DownloadData.scss';
+
+const getReactNodeText = (node: ReactNode): string => {
+  if (['string', 'number'].includes(typeof node)) {
+    return node?.toString() || '';
+  }
+
+  if (typeof node === 'object' && node && 'props' in node) {
```
imransl, line 32: Checking if `node` is truthy is redundant here

src/content/app/tools/blast/blast-download/submissionDownload.ts
```diff
@@ -168,7 +168,7 @@ const createZipArchive = async (submission: EnrichedBlastSubmission) => {
   return zip;
 };
 
-const getNameForZipRoot = (submission: BlastSubmission) => {
+export const getNameForZipRoot = (submission: BlastSubmission) => {
```
azangru, line 171: Since you reverted the commit that was using this function outside this module, it probably shouldn't be exported.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,113 @@
...
+import useDataTable from 'src/shared/components/data-table/hooks/useDataTable';
+import { downloadTextAsFile } from 'src/shared/helpers/downloadAsFile';
+
+import { PrimaryButton } from 'src/shared/components/button/Button';
+
+import { TableAction } from 'src/shared/components/data-table/dataTableTypes';
+
+import styles from './DownloadData.scss';
+
+const getReactNodeText = (node: ReactNode): string => {
```
azangru, line 27: There's a `renderToString` function exported from `'react-dom/server'`; I wonder if it won't be more suitable here.

(You can even pass the resulting string to a DOM element as `innerHTML` and request `textContent` back, I suppose; in order to strip all html tags that can be there).

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That is because `typeof(null)` is also `object`

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
azangru: As a follow-up (continuing the discussion we had during scrum), I am still confused about whether this function will return the right thing for something like:

```
<div>
  <span>One</span>
  <span>Two</span>
</div>
```

From what I am seeing in the console (playing with someone else's code sandbox that uses React 16-something; but I doubt it will be different in React 18), `children` in such a case will be an array of react nodes, which will fail the first check (not string or number), fail the second check (no 'props' in array), and will return an empty string? I'm just running this in my head, so might be wrong.

When using the inbuilt browser methods, we won't need to think about all those edge cases.

src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
```diff
@@ -24,12 +24,18 @@ import ShowHide from 'src/shared/components/show-hide/ShowHide';
 import BlastHitsDiagram from 'src/content/app/tools/blast/components/blast-hits-diagram/BlastHitsDiagram';
 import BlastSequenceAlignment from 'src/content/app/tools/blast/components/blast-sequence-alignment/BlastSequenceAlignment';
 
+import { createCSVForGenomicBlast } from 'src/content/app/tools/blast/blast-download/createBlastCSVTable';
```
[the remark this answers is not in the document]
msmanoj: Updated it to handle other types

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: No, it'll throw `TypeError`

src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
```diff
@@ -382,6 +408,7 @@ const DynamicColumnContent = (props: DynamicColumnContentProps) => {
 
   return (
     <span className={styles.nowrap}>
+      <span>hello</span>
```
azangru, line 411: Hi, who are you?

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React, { ReactNode } from 'react';
+import ReactDOM from 'react-dom/client';
+import { memoize } from 'lodash';
```
azangru, line 18: this will import the whole lodash; we are importing individual lodash modules.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    const csv = formatCSV(dataForExport);
+
+    downloadTextAsFile(csv, downloadFileName ?? 'Table export.csv');
+  };
+
+  const onSuccess = () => {
+    // show the green tick momentarily before we close it
+    setTimeout(onCancel, 1000);
```
azangru, line 106: `onCancel` looks out of place here; perhaps the function should be called `restoreDefaults`?

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,126 @@
...
+      });
+    });
+
+    const csv = formatCSV(dataForExport);
+
+    downloadTextAsFile(csv, downloadFileName ?? 'Table export.csv');
+  };
+
+  const onSuccess = () => {
+    // show the green tick momentarily before we close it
```
azangru, line 105: No similar courtesy to onError?

src/shared/components/data-table/DataTable.tsx
```diff
@@ -48,14 +50,16 @@ export const TableContext = React.createContext(
...
   columns: DataTableColumns;
-  state?: Partial<DataTableState>;
   theme: TableTheme;
   selectableColumnIndex: number;
-  className?: string;
   expandedContent: { [rowId: string]: ReactNode };
   disabledActions?: TableAction[];
+  className?: string;
+  downloadFileName?: string;
+  downloadHandler?: () => void;
```
azangru, line 61: Promise<void>?

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const onCancel = () => {
+    dispatch({
+      type: 'set_selected_action',
+      payload: TableAction.DEFAULT
+    });
+  };
+
+  const handleDownload = async () => {
+    if (downloadHandler) {
+      downloadHandler();
```
azangru, line 65: this is probably an asynchronous function, and you probably want to await it

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+          dataForExport[rowIndex + 1].push(getReactNodeText(cellExportData));
+        }
+      });
+    });
+
+    const csv = formatCSV(dataForExport);
+
+    downloadTextAsFile(csv, downloadFileName ?? 'Table export.csv');
+  };
```
azangru, line 102: would you like to move the setTimeout with the restore mechanism inside this function? It should cover both success and error paths, I think.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
imransl: Okay that makes sense. I just tried with `node && 'props' in node` (without `typeof node === 'object'`) and there was no error thrown even when it was of a different type i.e. array.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That's done

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That's done

src/shared/components/data-table/DataTable.tsx
msmanoj: That's done

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: renamed it

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That's done

src/content/app/tools/blast/views/blast-submission-results/components/single-blast-job-result/SingleBlastJobResult.tsx
msmanoj: oops! removed it

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,122 @@
...
+          dataForExport[rowIndex + 1].push(getReactNodeText(cellExportData));
+        }
+      });
+    });
+
+    const csv = formatCSV(dataForExport);
+
+    downloadTextAsFile(csv, downloadFileName ?? 'Table export.csv');
+
+    setTimeout(restoreDefaults, 1000);
```
azangru, line 104: Regarding the timeout (both here and in the if-block above). Theoretically, the component can get unmounted at any time (especially in the async if-block above), and the code will try to change the state of the unmounted component. Nothing will break; but in dev mode, React complains if something attempts to update the state of unmounted components.

Should we do something like:

```
const allowComponentResetRef = useRef(true);


useEffect(() => {
  return () => allowComponentResetRef.current = false;
}, []);

// and then...

const restoreDefaults = () => {
  setTimeout(() => {
    if (allowComponentResetRef.current) {
      dispatch({
        type: 'set_selected_action',
        payload: TableAction.DEFAULT
      });
    }
  }, 1000);
};
```

imransl:
Reviewed the PR again and haven't got much to add

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: I've made this change. Thanks

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,134 @@
...
+        if (isExportable !== false) {
+          const cellExportData = renderer
+            ? renderer({
+                rowData: row.cells,
+                rowId: String(row.rowId),
+                cellData: cell
+              })
+            : cell;
+
+          dataForExport[rowIndex + 1].push(getReactNodeText(cellExportData));
```
azangru, line 107: Don't you want to check here if `cellExportData` is a primitive (string / number), in which case you could push it into `dataForExport` directly?

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const restoreDefaults = () => {
+    setTimeout(() => {
+      if (allowComponentResetRef.current) {
+        dispatch({
+          type: 'set_selected_action',
+          payload: TableAction.DEFAULT
+        });
+      }
+    }, 1000);
+  };
```
azangru, line 73: Sorry, we messed up. There is now a second's delay when you click on "Cancel", which shouldn't be there.

https://user-images.githubusercontent.com/6834224/195531427-f7f9d940-a525-4903-86d3-7ff7e3e0d025.mov

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    downloadTextAsFile(csv, downloadFileName ?? 'Table export.csv');
+
+    restoreDefaults();
+  };
+
+  return (
+    <div className={styles.downloadData}>
+      <span>{downloadFileName ?? 'table.csv'}</span>
+      <LoadingButton onClick={handleDownload}>Download</LoadingButton>
```
azangru, line 122: This is super minor, but since all data is already on the client, the download completes before the spinner stops:

https://user-images.githubusercontent.com/6834224/195536179-bd8beb17-a4f6-48d5-af6d-76a993db3b6e.mov

It can be addressed by changing from LoadingButton to ControlledLoadingButton, and driving the state of that button from the `DownloadData` component. Or we can ignore this for now, and see how https://github.com/Ensembl/ensembl-client/pull/841 pans out.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That's done. Previously it was getting checked inside `getReactNodeText` but not since we updated it.

azangru:
The downloaded table isn't default-sorted (by e-value). Is this correct? @ens-st3 ?

azangru:
After sorting the downloaded data locally, I am seeing a discrepancy in the Score column between what's downloaded and what's shown in the table site. Must be an error somewhere — could easily be in my download code, or could be in the table.

image

image

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: I have addressed this issue in the latest commit.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: This is now fixed.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const handleDownload = async () => {
+    setDownloadState(LoadingState.LOADING);
+    if (downloadHandler) {
+      try {
+        await downloadHandler();
+        setDownloadState(LoadingState.SUCCESS);
+        setTimeout(restoreDefaults, 1000);
+      } catch {
+        setDownloadState(LoadingState.ERROR);
+        setTimeout(() => setDownloadState(LoadingState.NOT_REQUESTED), 2000);
```
azangru, line 87: You should probably also check for `allowComponentResetRef.current` in the callback to this setTimeout function. You are changing component's state here.

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -0,0 +1,154 @@
...
+      }
+
+      return;
```
azangru, line 90: this is unnecessary?

msmanoj: I'm not returning within the try block so this return here is required

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
msmanoj: That's done

src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
azangru: Oh, sorry, didn't see where the function ends 🙂

azangru (approved):

PR #849, opened 2022-10-04
Analytics - filling the gaps

msmanoj:
## Description
- Created useGlobalAnalytics to track custom events dispatched from shared components

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1743

## Deployment URL(s)
http://analytics-filling-the-gaps.review.ensembl.org

### DIFF
--- a/src/global/hooks/useGlobalAnalytics.ts
+++ b/src/global/hooks/useGlobalAnalytics.ts
@@ -15,16 +15,15 @@
  */
 
 import { useEffect } from 'react';
-import { useLocation } from 'react-router-dom';
 
 import analyticsTracking from 'src/services/analytics-service';
 
 import { type AnalyticsOptions } from 'src/analyticsHelper';
+import { useAppSelector } from 'src/store';
+import { getCurrentApp } from '../globalSelectors';
 
 const useGlobalAnalytics = () => {
-  const location = useLocation();
-  const appName: string =
-    location.pathname.split('/').filter(Boolean)[0] || 'homepage';
+  const appName = useAppSelector(getCurrentApp) || 'homepage';
 
   const sendTrackEvent = (ga: AnalyticsOptions) => {
     analyticsTracking.trackEvent({
--- a/src/global/hooks/useGlobalAnalytics.ts
+++ b/src/global/hooks/useGlobalAnalytics.ts
@@ -1,0 +1,55 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+
+import analyticsTracking from 'src/services/analytics-service';
+
+import { type AnalyticsOptions } from 'src/analyticsHelper';
+
+const useGlobalAnalytics = () => {
+  const location = useLocation();
+  const appName: string =
+    location.pathname.split('/').filter(Boolean)[0] || 'homepage';
+
+  const handleAnalyticsCustomEvent = (event: CustomEvent<AnalyticsOptions>) => {
+    const { category, action, species, feature, app } = event.detail;
+
+    analyticsTracking.trackEvent({
+      category,
+      action,
+      species,
+      feature,
+      app: app ?? appName
+    });
+  };
+
+  useEffect(() => {
+    window.addEventListener(
+      'analytics',
+      handleAnalyticsCustomEvent as EventListener
+    );
+    return () => {
+      window.removeEventListener(
+        'analytics',
+        handleAnalyticsCustomEvent as EventListener
+      );
+    };
+  }, []);
+};
+
+export default useGlobalAnalytics;
--- a/src/global/hooks/useGlobalAnalytics.ts
+++ b/src/global/hooks/useGlobalAnalytics.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 
 import analyticsTracking from 'src/services/analytics-service';
 
@@ -24,6 +24,7 @@ import { getCurrentApp } from '../globalSelectors';
 
 const useGlobalAnalytics = () => {
   const appName = useAppSelector(getCurrentApp) || 'homepage';
+  const globalAnalyticsRef = useRef<HTMLDivElement | null>(null);
 
   const sendTrackEvent = (ga: AnalyticsOptions) => {
     analyticsTracking.trackEvent({
@@ -39,17 +40,21 @@ const useGlobalAnalytics = () => {
   };
 
   useEffect(() => {
-    window.addEventListener(
+    globalAnalyticsRef.current?.addEventListener(
       'analytics',
       handleAnalyticsCustomEvent as EventListener
     );
     return () => {
-      window.removeEventListener(
+      globalAnalyticsRef.current?.removeEventListener(
         'analytics',
         handleAnalyticsCustomEvent as EventListener
       );
     };
-  }, []);
+  }, [globalAnalyticsRef, appName]);
+
+  return {
+    globalAnalyticsRef
+  };
 };
 
 export default useGlobalAnalytics;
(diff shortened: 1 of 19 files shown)


src/global/hooks/useGlobalAnalytics.ts
```diff
@@ -0,0 +1,55 @@
...
+      category,
+      action,
+      species,
+      feature,
+      app: app ?? appName
+    });
+  };
+
+  useEffect(() => {
+    window.addEventListener(
```
azangru, line 42: Let's make the hook return a ref `(useRef<HTMLElement|null>)`, which will be assigned to a sufficiently high-level DOM element by the parent. No need to do everything on window.

src/shared/components/help-popup/HelpPopupButton.tsx
```diff
@@ -34,6 +34,15 @@ const HelpPopupButton = (props: Props) => {
...
 
   const openModal = () => {
+    const trackHelpPopupOpen = new CustomEvent('analytics', {
+      detail: {
+        category: 'contextual-help',
+        action: 'opened'
+      }
+    });
+
+    window.dispatchEvent(trackHelpPopupOpen);
```
azangru, line 44: Here are a couple of things I don't like about this:

- it's dispatched against the window rather than against the DOM node of this element
- it makes HelpPopupButton aware of:
  - analytics in general (yuck!)
  - Google analytics in particular (because of the payload) — double yuck!

Imagine HelpPopupButton were a component of a design system separate from the `ensembl-client` codebase.

We could, potentially, produce an event called something like `ensembl-component-info` (or something better, but sufficiently non-analyticsy), with a payload that looks something like:

```
{
  componentName: 'help-popup-button', // or 'help-popup'
  action: 'help-shown',
  resourceType: 'article',
  resourceName: '<title of the article or of the video>'
}
```
The downside, obviously, is that there will need to be code that transforms this into Google analytics events. The upside is that this code won't be in the HelpPopupButton.

msmanoj: Why do we need to hide the actual purpose of events like this from these components?

Analytics is the only purpose these events are dispatched.

When someone reads this code in future, this event being `non-analyticsy` will only increase the time to understand what is going on here.

azangru: > When someone reads this code in future, this event being non-analyticsy will only increase the time to understand what is going on here.

On the other hand, if we extracted these components into a components library, they would have these weird analytics events associated with them and tailored to a very specific purpose — reporting to Google analytics.

I suppose one could look at such components as being more chatty than others in terms of what they want to tell the world about themselves.

msmanoj: Even if these components are moved to a components library, are these events going to be used for something other than analytics?

I think these events should be dispatched only if they are used for something.

src/global/hooks/useGlobalAnalytics.ts
msmanoj: Will this `ref` be used to dispatch the events?

azangru: Here, it would be listening to the events.

src/global/hooks/useGlobalAnalytics.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useLocation } from 'react-router-dom';
+
+import analyticsTracking from 'src/services/analytics-service';
+
+import { type AnalyticsOptions } from 'src/analyticsHelper';
+
+const useGlobalAnalytics = () => {
+  const location = useLocation();
+  const appName: string =
+    location.pathname.split('/').filter(Boolean)[0] || 'homepage';
```
azangru, line 27: Don't we save our app names in redux? Why not read it back from there?

src/shared/components/help-popup/HelpPopupBody.tsx
```diff
@@ -42,6 +42,18 @@ const HelpPopupBody = (props: Props) => {
   const { article, loadingState } = useHelpArticle(currentReference);
   const historyRef = useRef<HelpPopupHistory | null>(null);
 
+  useEffect(() => {
+    const trackHelpArticleViewed = new CustomEvent('analytics', {
+      detail: {
+        category: 'contextual-help',
+        action: article?.type + '_article_opened',
```
azangru, line 49: - As written, this code will dispatch some events even when there is no article. What would this mean?
- Why are we tracking this independently from "contextual-help/opened" in `HelpPopupButton`? Why is it not a single event? If contextual help opens, what else can it do other than display an article?

src/global/hooks/useGlobalAnalytics.ts
msmanoj: Here is the error I get when I use redux:

```
No errors found.
TypeError: Cannot read properties of undefined (reading 'currentApp')
    at getCurrentApp (webpack://ensembl-new/./src/global/globalSelectors.ts?:24:45)
    at memoizedSelector (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js:79:30)
    at getServerSnapshotWithSelector (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js:138:14)
    at Object.useSyncExternalStore (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5221:10)
    at useSyncExternalStore (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/react/cjs/react.development.js:1676:21)
    at useSyncExternalStoreWithSelector (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js:145:15)
    at useSelector (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/react-redux/lib/hooks/useSelector.js:54:27)
    at useGlobalAnalytics (webpack://ensembl-new/./src/global/hooks/useGlobalAnalytics.ts?:33:76)
    at eval (webpack://ensembl-new/./src/content/app/App.tsx?:68:82)
    at renderWithHooks (/Users/mapasa/Documents/work/ebi/repos/ensembl-client/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5662:16)
```

src/shared/components/help-popup/HelpPopupBody.tsx
msmanoj: >As written, this code will dispatch some events even when there is no article. What would this mean?

I've added an `if` wrapper to check if an article exists

src/global/hooks/useGlobalAnalytics.ts
msmanoj: Would it still work if the events are dispatched to `window`?

src/shared/components/communication-framework/ConversationIcon.tsx
```diff
@@ -31,24 +31,30 @@ type Props = {
 
 const ConversationIcon = (props: Props) => {
   const dispatch = useDispatch();
+  const analyticsRef = useRef<HTMLDivElement | null>(null);
```
azangru, line 34: `analyticsRef` -> `elementRef` (it's a reference to an html element)

src/shared/components/communication-framework/contact-us/ContactUs.tsx
```diff
@@ -29,6 +29,7 @@ import styles from './ContactUs.scss';
 
 const ContactUs = () => {
   const [shouldShowForm, setShouldShowForm] = useState(false);
+  const analyticsRef = useRef<HTMLDivElement | null>(null);
```
azangru, line 32: Same as above

src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
```diff
@@ -223,10 +224,11 @@ const ContactUsInitialForm = () => {
       detail: {
         category: 'contact_us',
         action: 'contact_form_submited'
-      }
+      },
+      bubbles: true
     });
 
-    window.dispatchEvent(trackContactUsSubmission);
+    analyticsRef.current?.dispatchEvent(trackContactUsSubmission);
```
azangru, line 231: We could probably reuse `dropAreaRef` here?

msmanoj: I tried but `dropAreaRef` is a function.

azangru: I pushed a commit to address this.

src/shared/components/communication-framework/ConversationIcon.tsx
msmanoj: That's done

src/shared/components/communication-framework/contact-us/contact-us-form/initial/ContactUsInitialForm.tsx
msmanoj: Nice. Thanks.

azangru (approved):

PR #852, opened 2022-10-12
Display submission name and adapt to small screen

ridwanamo:
## Description
Show submission name in the result view (job list and unviewed jobs). Setting it to a min width and truncating the text. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1737

## Deployment URL(s)
http://blast-name.review.ensembl.org

## Views affected
blast (you need to submit a job with name - a long name)

### DIFF
--- a/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
+++ b/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
@@ -47,7 +47,7 @@ const selectedMouse = {
 };
 
 const selectedSpecies = [selectedHuman, selectedMouse];
-const jobName = faker.lorem.words();
+const submissionName = faker.lorem.words();
 const database = faker.lorem.word();
 const blastParameters = {
   alignments: '50',
@@ -68,7 +68,7 @@ const mockState = merge({}, initialState, {
   sequences,
   selectedSpecies,
   settings: {
-    jobName,
+    submissionName,
     sequenceType: 'dna',
     program: 'blastn',
     preset: 'normal',
@@ -86,8 +86,8 @@ const expectedPayload = {
     value: toFasta(seq)
   })),
   preset: 'normal',
+  submissionName,
   parameters: {
-    jobName,
     database,
     program: 'blastn',
     stype: 'dna',
--- a/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
+++ b/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
@@ -87,7 +87,7 @@ const expectedPayload = {
   })),
   preset: 'normal',
   parameters: {
-    jobName: jobName,
+    jobName,
     database,
     program: 'blastn',
     stype: 'dna',
--- a/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
+++ b/src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
@@ -87,7 +87,7 @@ const expectedPayload = {
   })),
   preset: 'normal',
   parameters: {
-    title: jobName,
+    jobName: jobName,
     database,
     program: 'blastn',
     stype: 'dna',
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
@@ -33,7 +33,7 @@ import {
   getSelectedBlastProgram,
   getSelectedSearchSensitivity,
   getBlastSearchParameters,
-  getBlastJobName
+  getBlastSubmissionName
 } from 'src/content/app/tools/blast/state/blast-form/blastFormSelectors';
 
 import type {
@@ -177,7 +177,7 @@ const BlastSettings = ({ config }: Props) => {
             />
           </div>
           <div className={styles.blastJobBlock}>
-            <BlastJobName />
+            <BlastSubmissionName />
             <BlastJobSubmit />
           </div>
         </div>
@@ -296,19 +296,23 @@ const BlastSettings = ({ config }: Props) => {
   );
 };
 
-const BlastJobName = () => {
-  const jobName = useAppSelector(getBlastJobName);
-  const { setBlastJobName } = useBlastForm();
+const BlastSubmissionName = () => {
+  const submissionName = useAppSelector(getBlastSubmissionName);
+  const { setBlastSubmissionName } = useBlastForm();
 
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const name = e.currentTarget.value;
-    setBlastJobName(name);
+    setBlastSubmissionName(name);
   };
 
   return (
-    <div className={styles.blastJobName}>
+    <div className={styles.blastSubmissionName}>
       <span>Submission name</span>
-      <ShadedInput value={jobName} onChange={onChange} placeholder="optional" />
+      <ShadedInput
+        value={submissionName}
+        onChange={onChange}
+        placeholder="optional"
+      />
     </div>
   );
 };
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
@@ -307,7 +307,7 @@ const BlastJobName = () => {
 
   return (
     <div className={styles.blastJobName}>
-      <span>Name this job</span>
+      <span>Submission name</span>
       <ShadedInput value={jobName} onChange={onChange} placeholder="optional" />
     </div>
   );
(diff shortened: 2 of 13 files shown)


src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -115,9 +116,12 @@ export const BlastSubmissionHeader = (props: Props) => {
     <>
       <div className={styles.grid}>
         <div>{blastProgram}</div>
-        <div className={styles.submissionName}>
-          {/* placeholder for submission name */}
-        </div>
+        {jobName && (
+          <div className={styles.submissionName}>
+            <span className={styles.submissionIdLabel}>Submission name</span>
```
imransl, line 121: This class name is a bit misleading since there is another `submissionIdLabel` below meant for the actual `submissionId`. How about `submissionNameLabel` instead?

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -70,19 +70,20 @@ export const BlastSubmissionHeader = (props: Props) => {
...
     );
-    const { title, program, stype, ...otherParameters } = parameters;
+    const { jobName, program, stype, ...otherParameters } = parameters;
 
     const payload = {
       sequences: parsedSequences,
       selectedSpecies: species,
       settings: {
-        jobName: title,
+        jobName: jobName,
```
imransl, line 86: Can just be `jobName` rather than writing this as a key-value pair

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
```diff
@@ -87,7 +87,7 @@ const expectedPayload = {
   })),
   preset: 'normal',
   parameters: {
-    title: jobName,
+    jobName: jobName,
```
imransl, line 90: Same as below, can just be `jobName`.

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
ridwanamo: :heavy_check_mark:

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
ridwanamo: :heavy_check_mark:

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.test.tsx
ridwanamo: :heavy_check_mark:

jyothishnt (approved):
Looks good to me. Tested working.

azangru:
@ridwanamo In this PR, the width of the window can't go down to 980px, which is our minimal width. Is this intentional?

ridwanamo:
> @ridwanamo In this PR, the width of the window can't go down to 980px, which is our minimal width. Is this intentional?

No it wasn't intentional but i guess its a side effect of showing the submission name. If there is no submission name is behaves fine.

azangru:
> No it wasn't intentional but i guess its a side effect of showing the submission name. If there is no submission name is behaves fine.

I'd treat it as a regression...

In earlier Andrea's designs, she was hiding the submission name field for very small screens. I wonder if we can resurrect that behaviour (the media query you removed in `BlastSubmissionHeader.scss`

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -70,19 +70,20 @@ export const BlastSubmissionHeader = (props: Props) => {
...
   const submissionTime = getFormattedDateTime(new Date(submission.submittedAt));
+  const { jobName } = submission.submittedData.parameters;
 
   const editSubmission = () => {
     const { sequences, species, preset, parameters } = submission.submittedData;
     const parsedSequences = sequences.flatMap((sequence) =>
       parseBlastInput(sequence.value)
     );
-    const { title, program, stype, ...otherParameters } = parameters;
+    const { jobName, program, stype, ...otherParameters } = parameters;
```
imransl, line 80: There is a `jobName` at line number 73 too which is the same. Won't it be better to set these variables at the top itself rather than within the function?

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
```diff
@@ -52,7 +52,7 @@ export type OptionalSubmissionParameters = Partial<
 >;
 export type BlastSubmissionParameters = MandatorySubmissionParameters &
   OptionalSubmissionParameters & {
-    title: string;
+    jobName: string;
```
azangru, line 55: Woah, woah. Wait a second. The name of the parameter that gets submitted to EBI BLAST is `title`. Since EBI BLAST only has a concept of job (and not of our submission), the `title` field will be applied to every job.

- The `parameters` field of the payload (see blastApiSlice) should only contain valid BLAST parameters. `jobName` isn't one.
- We can either save submission name at the top level of a submission, or in the `title` field of submission parameters. Not sure which to prefer, but slightly leaning toward the `title` parameter. In either case, it shouldn't be called _job_ name

ridwanamo: i found title confusing and in the frontend code we were calling it jobName which is better even though it is submission name. i will look into this maybe next week now.

azangru: For reference: here's where `title` comes from:

https://www.ebi.ac.uk/seqdb/confluence/pages/viewpage.action?pageId=94147939#NCBIBLAST+HelpandDocumentation-RESTAPI

image

But again, for EBI, it's the title of a job, rather than of a batch of jobs, like it is for us. However, we can still make use of this field for our purposes.

ridwanamo:
> > No it wasn't intentional but i guess its a side effect of showing the submission name. If there is no submission name is behaves fine.
> 
> I'd treat it as a regression...
> 
> In earlier Andrea's designs, she was hiding the submission name field on very small screens. I wonder if we can resurrect that behaviour (the media query you removed in `BlastSubmissionHeader.scss`

Agreed to hide the submission name on smaller screen

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
ridwanamo: was that agreed with Backend (toolsApi) that the submission name will be the same title thats required from EBI. Also are we happy with the name `title` in our code. should it not be something like `submissionTitle` | `submissionName` or keep it just `title` to make it match whats required from the backend?

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
ridwanamo: This shouldn't be valid anymore. Changed the code so that submissionName (jobName rename) is now separated from parameters

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
ridwanamo: The code has been refactored to rename jobName to submissionName and moved it outside parameters and as agreed we are not going to use `title` thats provided by the EBI backend since `title` is specific for how EBI tools work which is associating a title to each submitted job whereas submissionName is very specific to our FrontEnd UI where it is the name of each submission which can include multiple jobs.

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -296,19 +296,23 @@ const BlastSettings = ({ config }: Props) => {
...
-    setBlastJobName(name);
+    setBlastSubmissionName(name);
   };
 
   return (
-    <div className={styles.blastJobName}>
-      <span>Name this job</span>
-      <ShadedInput value={jobName} onChange={onChange} placeholder="optional" />
+    <div className={styles.blastSubmissionName}>
+      <span>Submission name</span>
```
azangru, line 310: Oh, interesting! I didn't know Andrea has changed this label text; but now I can see in an [XD](https://xd.adobe.com/view/879a4244-8420-48c1-b3c9-c998e77651fd-d2c9/screen/571dfdb2-6465-4a1d-9d0f-300cf632aa27?fullscreen) that she has. Hoping I am looking at the up-to-date XD 😄 

However, you will notice that the style of this label in the XD is different; and that the placeholder text in the input starts with a capital "O".

ridwanamo: Done.

imransl (approved):

azangru (approved):

PR #853, opened 2022-10-12
More tracking events

msmanoj:
## Description
- Added an event to track species change link in entity viewer app bar.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1743

## Deployment URL(s)
http://more-analytics-events.review.ensembl.org


azangru (approved):
It's a bit strange that we called this action `change_link_clicked` rather than `species_change_link_clicked`; and it's a bit unfortunate that we define this event multiple times in different apps; but as long as it works, I don't mind.

PR #855, opened 2022-10-13
Make the label of FeatureLengthRuler more customizable

azangru:
## Description
- Add customisability for FeatureLengthRuler label
  - Enable colouring it differently from the labels of the ticks
  - Make the distance between the ruler label and the first tick label tweakable 
- Changed the `unitsLabel` property to `rulerLabel`
- Add the label `Length` to the ruler on the BLAST submission details page

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1738

## Deployment URL(s)
http://ruler-label.review.ensembl.org

### DIFF
--- a/src/shared/components/feature-length-ruler/FeatureLengthRuler.tsx
+++ b/src/shared/components/feature-length-ruler/FeatureLengthRuler.tsx
@@ -50,7 +50,8 @@ export type TicksAndScale = Ticks & {
 type Props = {
   length: number; // number of biological building blocks (e.g. nucleotides) in the feature
   width: number; // number of pixels allotted to the axis on the screen
-  unitsLabel?: string; // optional label showing what the ruler is measuring; will be displayed at the start of the ruler if present
+  rulerLabel?: string; // optional label showing what the ruler is measuring; will be displayed at the start of the ruler if present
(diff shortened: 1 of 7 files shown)


src/shared/components/feature-length-ruler/FeatureLengthRuler.tsx
```diff
@@ -79,7 +80,15 @@ const FeatureLengthRuler = (props: Props) => {
...
-          {props.unitsLabel ? `${props.unitsLabel} 1` : 1}
+          {props.rulerLabel && (
+            <tspan
+              className={styles.rulerName}
+              x={`-${props.rulerLabelOffset ?? 2}ch`}
+            >
+              {props.rulerLabel}
+            </tspan>
+          )}
+          <tspan x={0}>1</tspan>
```
msmanoj, line 91: Does this need the `label` className?

azangru: Hmm. At the moment, the parent `text` element has the `label` classname, which should apply to this `tspan`. I am not sure whether it makes more sense to move the `label` from the parent to this tspan. It works as it is though.

PR #857, opened 2022-10-22
Add visualisation for BLAST against genomic targets

azangru:
## Description
This PR introduces a new visualisation specifically for results of BLAST jobs against top-level genomic sequences (the DNA database). See the XD with the design [here](https://xd.adobe.com/view/3b27cc95-2a60-4407-855d-5a310e783e9e-6e74/?fullscreen)

The genomic visualisation makes an api call to fetch the species karyotype, which it then uses as a reference when drawing genomic regions. This allows us both to display top-level regions in the same order as they are typically represented in a karyotype, and to filter out the regions that are not part of karyotype (such as chromosome patches).

The rendering of the visualisation of BLAST results is now conditional on which database the input sequences were BLASTed against. For the DNA database, we will be showing the new visualisation; while for the transcripts and the proteins databases, we will be showing the one that we developed previously.

**Also:**
Removed extra horizontal padding from row elements in a BLAST results box:

image

notice the unnecessary inner padding (green):

image

which resulted in this misalignment in the layout:

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1735

## Deployment URL(s)
http://genomic-blast-viz.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
+++ b/src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
@@ -20,6 +20,7 @@ import sortBy from 'lodash/sortBy';
 import { useGenomeKaryotypeQuery } from 'src/shared/state/genome/genomeApiSlice';
 
 import BlastGenomicRegionHeatmap from './BlastGenomicRegionHeatmap';
+import RegionWithoutMatches from './BlastGenomicRegionWithoutMatches';
 import PillButton from 'src/shared/components/pill-button/PillButton';
 import CloseButton from 'src/shared/components/close-button/CloseButton';
 
@@ -66,6 +67,10 @@ const BlastGenomicHitsDiagram = (props: Props) => {
   const regionNamesForDisplay = isExpanded
     ? sortedRegionNamesForExpandedView
     : sortedRegionNamesForCompactView;
+
+  if (!regionNamesForDisplay.length) {
+    return <RegionWithoutMatches width={width} />;
+  }
 
   return (
     <div>
--- a/src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
+++ b/src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
@@ -14,14 +14,18 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import sortBy from 'lodash/sortBy';
 
 import { useGenomeKaryotypeQuery } from 'src/shared/state/genome/genomeApiSlice';
 
 import BlastGenomicRegionHeatmap from './BlastGenomicRegionHeatmap';
+import PillButton from 'src/shared/components/pill-button/PillButton';
+import CloseButton from 'src/shared/components/close-button/CloseButton';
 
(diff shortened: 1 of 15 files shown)


src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const hitHspsWithRegionName = job.hits.flatMap((hit) => {
+    return hit.hit_hsps.map((hit_hsp) => ({
+      start: Math.min(hit_hsp.hsp_hit_from, hit_hsp.hsp_hit_to),
+      end: Math.max(hit_hsp.hsp_hit_from, hit_hsp.hsp_hit_to),
+      eValue: hit_hsp.hsp_expect,
+      regionName: hit.hit_acc
+    }));
+  });
+
+  const sortedHsps = sortBy(hitHspsWithRegionName, 'eValue');
```
jyothishnt, line 128: The order of top hits in svg and table doesn't match.

jyothishnt:
Missing line for 0 hits. (as in the xd)

Screenshot 2022-10-24 at 11 35 52

jyothishnt:
Also when there are no hits, we do not need to render the table at all and also remove the arrow at the end of species name.

azangru:
> Also when there are no hits, we do not need to render the table at all and also remove the arrow at the end of species name.

This should certainly be done in a dedicated PR.

azangru:
> Missing line for 0 hits. (as in the xd)

Updated.

src/content/app/tools/blast/components/blast-genomic-hits-diagram/BlastGenomicHitsDiagram.tsx
ens-st3: The Patches are excluded from the image, but not the table, yes ?

ens-st3:
Having fewer than five visible circles is going to be unavoidable. Any concerns (for a separate  PR) Andrea ?
Screenshot 2022-10-24 at 18 52 05

src/content/app/tools/blast/views/blast-submission-results/components/blast-results-per-sequence/BlastResultsPerSequence.tsx
```diff
@@ -61,18 +62,24 @@ const BlastResultsPerSequence = (props: BlastResultsPerSequenceProps) => {
   const { width: plotwidth } = useResizeObserver({ ref: rulerContainer });
   const [shouldShowJobResult, showJobResult] = useState(true);
   const [shouldShowParamaters, showParamaters] = useState(false);
+  const usesGenomicHitsDiagram = parameters.database === 'dna'; // NOTE: works for now; but likely to expand in the future
```
imransl, line 65: There is `needsGenomicHitsDiagram` used in `SingleBlastJobResult`. Can it be used here too instead of `usesGenomicHitsDiagram`?

azangru: I didn't particularly like that name; but I suppose `uses...` may me too similar with names of hooks? 🤷‍♂️  I renamed the variable to `needs...`

jyothishnt (approved):

imransl (approved):

PR #859, opened 2022-10-24
hide chevron when there is only one transcript

jyothishnt:
## Description
Hide chevron in track panel list when there is only one transcript.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1745

## Deployment URL(s)
http://remove-chevron.review.ensembl.org

## Views affected
GB

### DIFF
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
@@ -37,6 +37,7 @@ import GroupTrackPanelItemLayout from './track-panel-item-layout/GroupTrackPanel
 import SimpleTrackPanelItemLayout from './track-panel-item-layout/SimpleTrackPanelItemLayout';
 
 import { Status } from 'src/shared/types/status';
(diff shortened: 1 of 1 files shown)


imransl (approved):

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelGene.tsx
```diff
@@ -171,6 +172,7 @@ const TrackPanelGene = (props: TrackPanelGeneProps) => {
         visibilityIconHelpText={getVisibilityIconHelpText(geneVisibilityStatus)}
         onShowMore={onShowMore}
         toggleExpand={toggleExpand}
+        hideToggle={hideToggle}
       >
```
azangru, line 176: Mmm. A `GroupTrackPanelItemLayout` component is like a `SimpleTrackPanelItemLayout` component, but with two extra parameters: `isCollapsed`, and `toggleExpand`. So if we don't want to show the chevron for genes with a single transcript, why don't we use the `SimpleTrackPanelItemLayout` component for such cases in `TrackPanelGene`? It would keep the purpose of these two components clearly separated instead of blurring the distinction between them.

azangru (approved):

PR #863, opened 2022-10-26
Move grch38 to top of blast species list

jyothishnt:
## Description
GRCh38 needs to given more prominence and it should be in the top of the list of species to be selected in the right hand pane of the BLAST form. GRCh37 needs to be moved just below it.
This is a temporary fix on the client side. Species list should eventually come from the backend.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1746

## Deployment URL(s)
http://reorder-blast-species-list.review.ensembl.org

## Views affected
Blast

### DIFF
--- a/src/content/app/tools/blast/components/blast-species-selector/speciesList.ts
+++ b/src/content/app/tools/blast/components/blast-species-selector/speciesList.ts
@@ -18,12 +18,6 @@
 // Check if we need the sorting and which field we need to submit a job
 const speciesList = [
   {
-    assembly_name: 'GRCh38.p13',
-    common_name: 'Human',
-    genome_id: 'a7335667-93e7-11ec-a39d-005056b38ce3',
-    scientific_name: 'Homo sapiens'
-  },
-  {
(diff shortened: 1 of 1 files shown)


azangru (approved):
My immediate question was why pass that array through the sorting function; why not just hard-code the order of the species in the array. I am not convinced the implementation of the sorting function is even correct. But then, it's a strange file that will sooner or later be deleted...

imransl (approved):

azangru (changes requested):
Since we decided to use the popular species endpoint from the species selector to get the species list (see [ENSWBSITES-1773](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1773)), there is no point in merging this pull request. Either use the same PR for ENSWBSITES-1773, or create a new PR (and we will close this one when the other one is accepted).

azangru:
This PR is superseded by https://github.com/Ensembl/ensembl-client/pull/871

PR #867, opened 2022-10-27
Update ImageButton component

azangru:
## Description
The purpose of this PR is to refactor the ImageButton component such that:
- The statusClasses property accepting multiple CSS class names is removed. Instead, the component uses CSS variables for properties that can be styled from the parent
- ImageButton is now using properties of the native button component, thus getting closer to the native button element.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1618

## Deployment URL(s)
http://image-button-refactoring.review.ensembl.org

### DIFF
--- a/src/shared/components/image-button/ImageButton.scss
+++ b/src/shared/components/image-button/ImageButton.scss
@@ -1,40 +1,39 @@
 @import 'src/styles/common';
 
-/* stylelint-disable no-descending-specificity */
-// stylelint doesn't like that the 'imageButton button > svg' selector is more specific than ".default svg" and others;
-// but moving it below the rest of the selectors feels weird
+.wrapper {
+  font-size: 0;
+  line-height: 0;
+  position: relative;
+  display: inline-block; // buttons are normally inline-block elements
+  vertical-align: middle;
+}
 
 .imageButton {
-  position: relative;
-  font-size: 0;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  width: var(--image-button-width, auto);
+  height: var(--image-button-height, auto);
 
-  svg {
+  & > svg {
+    width: var(--image-button-svg-width, 100%);
+    height: var(--image-button-svg-height, 100%);
     transition: opacity 0.4s;
-  }
-
-  button {
-    width: 100%;
-    height: 100%;
-  }
-
-  // if a button's immediate child is an svg, it should be the same size as the button
-  button > svg {
-    width: 100%;
-    height: 100%;
   }
 }
 
 .default {
+  background-color: var(--image-button-default-bg-color, initial);
   svg {
-    fill: $blue;
+    fill: var(--image-button-default-svg-color, #{$blue});
   }
 }
 
 .selected {
-  background-color: $blue;
+  background-color: var(--image-button-selected-bg-color, #{$blue});
 
   svg {
-    fill: $white;
+    fill: var(--image-button-selected-svg-color, #{$white});
   }
   &svg:hover {
     opacity: 0.8;
@@ -42,8 +41,10 @@
 }
 
 .unselected {
+  background-color: var(--image-button-unselected-bg-color, none);
+
   svg {
-    fill: $blue;
+    fill: var(--image-button-unselected-svg-color, #{$blue});
   }
   &svg:hover {
     opacity: 0.8;
@@ -51,11 +52,10 @@
 }
 
 .disabled {
+  background-color: var(--image-button-disabled-bg-color, none);
+  cursor: default;
   svg {
-    fill: $grey;
+    fill: var(--image-button-disabled-svg-color, #{$grey});
     opacity: 1;
-  }
-  button {
-    cursor: default;
   }
 }
--- a/src/shared/components/image-button/ImageButton.scss
+++ b/src/shared/components/image-button/ImageButton.scss
@@ -41,7 +41,7 @@
 }
 
 .unselected {
-  background-color: var(--image-button-unselected-bg-color, none);
+  background-color: var(--image-button-unselected-bg-color);
 
   svg {
     fill: var(--image-button-unselected-svg-color, #{$blue});
@@ -52,7 +52,7 @@
 }
 
 .disabled {
-  background-color: var(--image-button-disabled-bg-color, none);
+  background-color: var(--image-button-disabled-bg-color);
   cursor: default;
   svg {
     fill: var(--image-button-disabled-svg-color, #{$grey});
--- a/src/shared/components/image-button/ImageButton.scss
+++ b/src/shared/components/image-button/ImageButton.scss
@@ -23,7 +23,7 @@
 }
 
 .default {
-  background-color: var(--image-button-default-bg-color, initial);
+  background-color: var(--image-button-default-bg-color);
   svg {
     fill: var(--image-button-default-svg-color, #{$blue});
   }
--- a/src/shared/components/image-button/ImageButton.scss
+++ b/src/shared/components/image-button/ImageButton.scss
@@ -8,6 +8,9 @@
   vertical-align: middle;
 }
 
+/* stylelint-disable no-descending-specificity */
+/* "imageButton button svg" has a higher specificity than ".default svg", and linter doesn't like it */
+
 .imageButton button {
   display: inline-flex;
   width: 100%;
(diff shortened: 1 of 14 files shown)


src/shared/components/image-button/ImageButton.scss
```diff
@@ -1,61 +1,61 @@
...
-  svg {
+.imageButton {
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  width: var(--image-button-width, auto);
+  height: var(--image-button-height, auto);
+
+  & > svg {
+    width: var(--image-button-svg-width, 100%);
```
ridwanamo, line 19: why are we using percentage here instead of an actual value?

src/shared/components/image-button/ImageButton.tsx
```diff
@@ -24,52 +28,57 @@ import Tooltip from 'src/shared/components/tooltip/Tooltip';
...
   image: FunctionComponent | string;
-  className?: string;
-  statusClasses?: { [key in ImageButtonStatus]?: string };
-  onClick?: (event?: React.MouseEvent<HTMLDivElement>) => void;
 };
 
 export const ImageButton = (props: Props) => {
+  const {
+    status = Status.DEFAULT,
+    description,
```
ridwanamo, line 53: should description be `description = ''` for the default value because what will happen if no description is passed. will it not be undef instead of empty string?

src/shared/components/image-button/ImageButton.scss
```diff
@@ -1,61 +1,61 @@
...
-
-  // if a button's immediate child is an svg, it should be the same size as the button
-  button > svg {
-    width: 100%;
-    height: 100%;
-  }
 }
 
 .default {
+  background-color: var(--image-button-default-bg-color, initial);
```
ridwanamo, line 26: Do we need to set this? if we dont specify this will it not take the browser's default

src/shared/components/image-button/ImageButton.tsx
```diff
@@ -24,52 +28,57 @@ import Tooltip from 'src/shared/components/tooltip/Tooltip';
...
 export type ImageButtonStatus =
   | Status.DEFAULT
   | Status.SELECTED
   | Status.UNSELECTED
   | Status.DISABLED;
 
-export type Props = {
-  status: ImageButtonStatus;
-  description: string;
+type ChildlessButtonProps = Omit<
```
ridwanamo, line 39: why do we need this?

src/shared/components/image-button/ImageButton.tsx
```diff
@@ -24,52 +28,57 @@ import Tooltip from 'src/shared/components/tooltip/Tooltip';
...
-    <div ref={hoverRef} className={imageButtonClasses} {...buttonProps}>
-      <button>
+    <div ref={hoverRef} className={styles.wrapper}>
+      <button
+        className={imageButtonClasses}
+        {...otherProps}
+        disabled={status === Status.DISABLED}
+      >
         {typeof props.image === 'string' ? (
           <img src={props.image} alt={props.description} />
```
ridwanamo, line 79: should this be:
``

src/shared/components/image-button/ImageButton.scss
```diff
@@ -1,61 +1,61 @@
...
-    fill: $white;
+    fill: var(--image-button-selected-svg-color, #{$white});
   }
   &svg:hover {
     opacity: 0.8;
   }
 }
 
 .unselected {
+  background-color: var(--image-button-unselected-bg-color, none);
```
azangru, line 44: should delete this

src/shared/components/image-button/ImageButton.tsx
ridwanamo: we are not passing children through

src/shared/components/image-button/ImageButton.tsx
ridwanamo: Both undef and '' empty string will not show the alt even though accessibility reason there should always be an alt but thats not applicable to us for now.

src/shared/components/image-button/ImageButton.scss
ridwanamo: The percentage is relative to the container.

stories/shared-components/image-button/ImageButton.stories.tsx
```diff
@@ -27,97 +26,47 @@ import { Status } from 'src/shared/types/status';
...
+        description="I am a disabled button; but I will still show a tooltip when moused over"
+        image={DownloadIcon}
+        onClick={toggleImage}
+      />
+      <p>
+        I am a disabled button; but I will still show a tooltip when moused over
+      </p>
+      <ImageButton
+        status={Status.DEFAULT}
+        description="I am a disabled button; but I will still show a tooltip when moused over"
```
ridwanamo, line 63: description is wrong here, its not a disabled button.

src/shared/components/image-button/ImageButton.tsx
azangru: Yes; I also removed the `disabled` property; so that the only way to influence whether the ImageButton is disabled is to pass appropriate props to it.

src/shared/components/image-button/ImageButton.scss
azangru: Browser is perfectly happy without the fallback. I'm removing it...

image

src/shared/components/image-button/ImageButton.tsx
```diff
@@ -24,52 +28,60 @@ import Tooltip from 'src/shared/components/tooltip/Tooltip';
...
+    <div ref={hoverRef} className={styles.wrapper}>
+      <button
+        className={imageButtonClasses}
+        type={buttonType}
+        {...otherProps}
+        disabled={status === Status.DISABLED}
+      >
         {typeof props.image === 'string' ? (
-          <img src={props.image} alt={props.description} />
+          <img src={props.image} alt={description} />
```
ridwanamo, line 82: There is one more place to replace `props.description` (line 89 inside <Tooltip>)

azangru: ✅

stories/shared-components/image-button/ImageButton.stories.tsx
azangru: ✅

src/shared/components/image-button/ImageButton.tsx
azangru: `src={Image}`, with capital-case `Image` feels wrong. Let it be as it is.

azangru:
- Moved the parent-customizable `className` property to the top level of the component.
- Removed the `--image-button-width` and `--image-button-height` variables. The parent would need to set the width and height on the ImageButton component directly, by passing it a css class.

src/shared/components/visibility-icon/VisibilityIcon.tsx
```diff
@@ -39,29 +40,24 @@ type VisibilityIconProps = {
 };
 
 export const VisibilityIcon = (props: VisibilityIconProps) => {
-  const status = props.status;
-  let imageButtonStatus: ImageButtonStatus;
-  let eyeIcon = Eye;
-  let className = styles.default;
+  const { status } = props;
+  const isPartiallySelected = props.status === Status.PARTIALLY_SELECTED;
```
imransl, line 44: Can't just `status` be used here?

src/shared/components/visibility-icon/VisibilityIcon.tsx
```diff
@@ -39,29 +40,24 @@ type VisibilityIconProps = {
...
-  if (status === Status.PARTIALLY_SELECTED) {
-    imageButtonStatus = Status.DEFAULT;
-    eyeIcon = EyePartial;
-    className = styles.partiallySelected;
-  } else {
-    imageButtonStatus = status;
-  }
+  const elementClasses = classNames(styles.visibilityIcon, {
+    [styles.visibilityIconPartiallySelected]:
+      status === Status.PARTIALLY_SELECTED
```
imransl, line 52: `isPartiallySelected` here?

src/shared/components/visibility-icon/VisibilityIcon.tsx
azangru: ✅

src/shared/components/visibility-icon/VisibilityIcon.tsx
azangru: ✅

ridwanamo (approved):

imransl (approved):

PR #868, opened 2022-10-31
[BLOCKED] Use the new Node global fetch api, and remove the dependency on cross-fetch

azangru:
## Description
The new long-term support version of Node (18.12.0 and later) includes the global `fetch` function, which has the same api as the browser fetch (see the [docs](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch)). Switching to it will allow us to remove the dependency on the cross-fetch package.

The annoying bit is that the Node team still considers the fetch api "experimental", and we will see the following warning in the console:

image

**BLOCKER:**
Tests that rely on the mock service worker (`msw`) package have failed, because `msw` doesn't currently support Node's native `fetch`. See PR `https://github.com/mswjs/msw/issues/1388` in the `msw` repository.

## Deployment URL(s)
http://remove-server-side-fetch.review.ensembl.org


imransl:
@azangru Going through that issue there doesn't seem any interest in looking into it the issue. Are we going to give it a bit more time before deciding what to do with this PR?

azangru:
> there doesn't seem be any interest by the msw devs to look into it the issue

Why do you think so?

> Are we going to give it a bit more time before deciding what to do with this PR?

Sure. As long as it takes.

azangru:
Superseded by https://github.com/Ensembl/ensembl-client/pull/1049

PR #870, opened 2022-11-01
Old blast submissions

azangru:
## Description
The results of EBI BLAST jobs are stored only for 7 days; thus making it possible for the app to have BLAST submissions whose results are no longer accessible. This PR addresses this problem.

- Added a notice about how long BLAST submission results are available (and how long a submission is stored in the browser) to the grey bar in the top of BLAST pages
- After 7 days, will show a notice in the header of a BLAST submission informing user that results for this submission are no longer available
- Added error screens to the BLAST submission details page (for when user navigates to this page directly using the url)
- Added a check at every app startup that will delete submissions older than 28 days
- Added tests for the blast storage service method that deletes expired blast submissions

## Also
- Switched the the version of the `idb` library that uses async iterators (good for ergonomics)
- Fixed the bug when updated BLAST jobs status did not get written to IndexedDB after the polling

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1716
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1744

## Deployment URL(s)
http://old-blast-submissions.review.ensembl.org

## How to test
Ok, so you submitted your BLAST jobs; but you probably don't want to wait for 7 days until the jobs have properly expired. To artificially age your BLAST submission, you would need to modify the record in IndexedDB manually, by following the steps below:
- In browser's dev tools, find IndexedDB, then find the `blast-submissions` storage in it, and read an id of a BLAST submission that you want to make older
- Then, in your console, do something along the following lines:

```js
blastSubmissionId = the_id_that_you_see_in_indexed_db

dbRequest = window.indexedDB.open("ensembl-website");
dbRequest.onsuccess = (event) => {
  const db = event.target.result;
  const transaction = db.transaction("blast-submissions", 'readwrite')
  const objectStore = transaction.objectStore("blast-submissions")
  objectStore.get(blastSubmissionId).onsuccess = (event) => {
    updateSubmitTime(event.target.result, objectStore);
  };
};

updateSubmitTime = (data, objectStore) => {
  data.submittedAt = Date.now() - 1000 * 60 * 60 * 24 * 7 // you can set however many days you want here 
  const requestUpdate = objectStore.put(data, blastSubmissionId);
  requestUpdate.onsuccess = (event) => {
    console.log('blast submission date updated')
  };
}
```

### DIFF
--- a/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
+++ b/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
@@ -52,7 +52,7 @@
 }
 
 .controlsSection {
-  grid-column: control-buttons;
+  grid-column: controls-section;
   display: flex;
   align-items: center;
   justify-content: flex-end;
@@ -86,7 +86,7 @@
 // Leave media queries at the bottom of the file to make sure they override CSS rules defined above
 @media screen and (max-width: 1439px) {
   .grid {
-    grid-template-columns: auto [submission-name] auto [submission-details] 1fr [control-buttons] auto;
+    grid-template-columns: auto [submission-name] auto [submission-details] 1fr [controls-section] auto;
   }
   .controlButtons {
     padding-left: 20px;
@@ -94,6 +94,10 @@
 }
 
 @media screen and (max-width: 1279px) {
+  .grid {
+    grid-template-columns: auto [submission-details] 1fr [controls-section] auto;
+  }
+
   .submissionName {
     display: none;
   }
--- a/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
+++ b/src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
@@ -2,7 +2,7 @@
 
 .grid {
   display: grid;
-  grid-template-columns: 11ch [submission-name] auto [submission-details] 1fr [control-buttons] 20%;
+  grid-template-columns: 11ch [submission-name] auto [submission-details] 1fr [controls-section] 20%;
   column-gap: 20px;
   align-items: center;
   padding-left: 26px; // TODO: seems to be a constant shared with sequence boxes (see ListedBlastSubmission.scss)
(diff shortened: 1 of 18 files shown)


src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -172,4 +167,54 @@ export const BlastSubmissionHeader = (props: Props) => {
...
+      <DeleteButton onClick={onDelete} disabled={isInDeleteMode} />
+      <div className={styles.unavailableResultsMessage}>
+        <span>{UNAVAILABLE_RESULTS_WARNING}</span>
+        <QuestionButton helpText={<UnavailableResultsHelpText />} />
+      </div>
+    </div>
+  );
+};
+
+const UnavailableResultsHelpText = () => (
```
ridwanamo, line 206: `UnavailableResultsHelpText` to `ResultsExpiryHelpText`

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
```diff
@@ -2,7 +2,7 @@
 
 .grid {
   display: grid;
-  grid-template-columns: 11ch [submission-name] auto [submission-details] 1fr [control-buttons] 20%;
+  grid-template-columns: 11ch [submission-name] auto [submission-details] 1fr [controls-section] 20%;
```
imransl, line 5: Just curious why this is `controls-section` now?

ridwanamo (approved):

imransl (approved):

src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.scss
azangru: 🤷‍♂️  It contains more than buttons.

ridwanamo (approved):

PR #871, opened 2022-11-02
Blast species list

jyothishnt:
## Description
Blast should use species list from species selector so that hardcoded list currently in use can be removed. 
Test also updated.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1773

## Deployment URL(s)
http://blast-species-list.review.ensembl.org

## Views affected
Blast


src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.test.tsx
```diff
@@ -65,12 +91,13 @@ const renderComponent = (
...
 describe('SpeciesSelector', () => {
   it('updates the selectedSpecies state', async () => {
-    const { container, store } = renderComponent();
+    const { container, store, debug } = renderComponent();
 
     const speciesCheckbox = container.querySelector(
       'tbody tr [data-test-id="checkbox"]'
     ) as HTMLElement;
 
+    debug(container);
```
ridwanamo, line 100: you left this debug here, it should be removed and same for above?

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
```diff
@@ -14,30 +14,36 @@
...
 const BlastSpeciesSelector = () => {
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
+  const popularSpeciesList = useSelector(getPopularSpecies);
   const selectedSpeciesList = useSelector(getSelectedSpeciesList);
   const selectedGenomeIds = selectedSpeciesList.map(
     ({ genome_id }) => genome_id
   );
 
+  useEffect(() => {
```
ridwanamo, line 43: why do we need to dispatch this?

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
```diff
@@ -14,30 +14,36 @@
  * limitations under the License.
  */
 
-import React from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import React, { useEffect } from 'react';
+import { useSelector } from 'react-redux';
```
ridwanamo, line 18: we should use `useAppSelector` from src/store

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
jyothishnt: If a user come to blast view first, we need to fetch the popular species

ridwanamo: Then should the useEffect not have `popularSpeciesList` as a check. My understanding of useEffect and the way its written here is everytime the component is loaded it will fetch the species. And we only want to fetch it when there is no species.

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
```diff
@@ -46,6 +53,10 @@ const BlastSpeciesSelector = () => {
     }
   };
 
+  const availableSpeciesList = popularSpeciesList.filter(
+    (species) => species.scientific_name
```
azangru, line 57: This is the field to use:

image

ridwanamo (approved):

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
jyothishnt: Oh cool. Thanks!

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.test.tsx
```diff
@@ -27,7 +27,33 @@ import blastFormReducer, {
...
+    genome_id: 'a73357ab-93e7-11ec-a39d-005056b38ce3',
+    scientific_name: 'Triticum aestivum'
+  },
+  {
+    assembly_name: 'GRCh37.p13',
+    common_name: 'Human',
+    genome_id: '3704ceb1-948d-11ec-a39d-005056b38ce3',
+    scientific_name: 'Homo sapiens'
+  }
+];
```
azangru, line 49: We have a function for generating popular species objects: https://github.com/Ensembl/ensembl-client/blob/dev/tests/fixtures/popular-species.ts

jyothishnt: I tried to use this, but now I see the out-of-scope variable referencing issue you suspected yesterday. :-(

imransl (approved):

PR #872, opened 2022-11-04
Add Hotjar tracking script

azangru:
## Description
Adding hotjar tracking script.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1713

## Deployment URL(s)
Nothing to see here; the injection of the script depends on the presence of an environment variable which is only available in production

### DIFF
--- a/src/server/routes/viewsRouter.tsx
+++ b/src/server/routes/viewsRouter.tsx
@@ -36,6 +36,8 @@ import Root from 'src/root/Root';
 const paths = getPaths();
 const statsFile = path.resolve(paths.buildStaticPath, 'loadable-stats.json');
 
+const configForClient = getConfigForClient();
+
 const viewRouter = async (req: Request, res: Response) => {
   // maintainers of the loadable component say that ChunkExtractor is stateful
   // and needs to be initialized at every request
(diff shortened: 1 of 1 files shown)


imransl (approved):

jyothishnt (approved):

src/server/routes/viewsRouter.tsx
```diff
@@ -126,4 +127,25 @@ const viewRouter = async (req: Request, res: Response) => {
   res.status(status).send(responseString);
 };
 
+// TODO: remove as soon as it is no longer needed
+const hotjarId = 2555715; // if we discover that this needs to be updated, we will extract it into an environment variable
+const hotjarTrackingScript = configForClient.environment.shouldReportAnalytics
```
ridwanamo, line 132: Just making sure this is the right environment check we want. I dont understand how enabling GA analytics reporting is link to hotjar.

ridwanamo: sorry ignore my question above, you are checking if it is a production deployment (i was expecting an expilcit flag for checking production environment). Was just making sure its the right flag to use.

azangru: Yes; I am checking here that the code is executed in the production deployment. Hotjar is similar to Google Analytics in that both are horrible third-party scripts collecting user data; so both will be enabled in an environment where we need this data collection.

The alternative would have been to create a dedicated environment variable for Hotjar; which I don't think deserves it at this point.

ridwanamo (approved):
code looks ok but we can only test this once we push the changes to production.

PR #873, opened 2022-11-04
update icons in GB sidebar toolstrip

ridwanamo:
## Description
Remove the following icons:
* TrackManager
* Personal Data

Add one icon: Navigate

Confirm with Andrea size of navigate icon should be 26px x 26px and tooltip on icon is "Navigate browser image"

Had to update one test.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1771

## Deployment URL(s)
http://toolstrip-update.review.ensembl.org

## Views affected
Genome browser only


src/content/app/genome-browser/components/browser-sidebar-toolstrip/BrowserSidebarToolstrip.tsx
```diff
@@ -133,6 +119,16 @@ const BrowserSidebarToolstrip = () => {
         onClick={() => toggleModalView(BrowserSidebarModalView.DOWNLOADS)}
         image={DownloadIcon}
       />
+      <ImageButton
+        className={classNames(
+          [layoutStyles.sidebarIcon],
+          [layoutStyles.navigateIcon]
```
azangru, line 125: why so many arrays?

src/shared/components/layout/StandardAppLayout.scss
```diff
@@ -130,6 +130,11 @@ $drawerWindowWidth: 45px;
   margin-bottom: 30px;
 }
 
+.navigateIcon {
+  width: 26px;
+  height: 26px;
+}
```
azangru, line 136: Wait, you still had all this padding (green) to shave off before starting tweaking width and height.

Screenshot 2022-11-04 at 15 12 57

BTW: might be worth waiting until the updated ImageButton component is merged; and check how it works there.

src/content/app/genome-browser/components/browser-sidebar-toolstrip/BrowserSidebarToolstrip.tsx
ridwanamo: sorry my bad, i was trying something else before and forgot to remove the square bracket. Done now.

src/shared/components/layout/StandardAppLayout.scss
ridwanamo: So the padding is set to 4px on all the icons....whats the better solution, set the padding to 2px or set the height and width? Feels like we might want to keep the padding for the icons the same isn't it? I

azangru: The updated `ImageButton` component has `--image-button-svg-width` and `--image-button-svg-height` properties. You can use either them or the padding to tweak the icon size; whichever is more convenient/appropriate.

ridwanamo: I went with the padding instead of readjusting the width and height because the padding was affecting the size of the image so that sounds like a better fix.

azangru: > so that sounds like a better fix

I'm confused by why you would consider this a better fix; and drew a diagram to explain what I meant:

image

I'm not saying that this is better or worse than the padding; I'm just saying that this allows you to control the dimensions of the svg directly, rather than calculating them from image button width(height) - padding *2.

azangru (approved):

src/shared/components/layout/StandardAppLayout.scss
ridwanamo: I am also confused. I understand your point of being able to override the image width and height using the css variable. So which one should we go with:
* set padding to 2px or
* set width and height to 24px

azangru: 24x24px is the dimensions of the ImageButton, which is this rectangle:



whereas the `--image-button-svg-width` and `--image-button-svg-height` properties control the size of the svg image inside the rectangle:



Both the padding and the CSS variable are valid options for setting the size of the svg image inside the button. The option with the variable is more straightforward because it is easier to reason about than the padding option (e.g. `--image-button-svg-width: 20px` and `--image-button-svg-height: 20px` gives you an svg with a size of 20x20px; whereas `padding: 4px` means that the size of the svg is the size of the ImageButton (24x24px) minus twice the size of the padding (-4px left, -4px right, -4px top, -4px bottom)).

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/NavigateModal.tsx
```diff
@@ -16,13 +16,13 @@
 
 import React from 'react';
 
-const TracksManagerModal = () => {
+const NavigateModal = () => {
   return (
-    <section className="tracksManagerModal">
-      <p>Save multiple browser configurations</p>
+    <section className="NavigateModal">
```
imransl, line 21: What is the purpose of this CSS class?

src/styles/main.scss
```diff
@@ -27,7 +27,7 @@ dl {
 button {
   cursor: pointer;
 
-  &.disabled {
+  &[disabled] {
```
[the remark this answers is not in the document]
ridwanamo: Removed it and it still works fine.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/NavigateModal.tsx
ridwanamo: I have removed the css class. Its funny though all the other modal has the same thing.

azangru (approved):

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/NavigateModal.tsx
imransl: We need to remove them too if we aren't using them. But let's do that in another PR :)

imransl (approved):

PR #877, opened 2022-11-11
store submitted blast sequences as header and value

ridwanamo:
## Description
we currently stored the blast input sequences as combined header and value in the local storage but there is more use of having it separate that is a field for header and a field for value (see JIRA for more detail).

This PR address storing the sequence as header and value separately in the local storage. And then combined the header and sequence together just before submitting it to the back end tool api.

Update the code to remove using `parseBlastInput` to separate the header and sequence.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1761

## Deployment URL(s)
http://blast-parsed-sequences.review.ensembl.org

## Views affected
blast


src/content/app/tools/blast/components/blast-submission-header/BlastSubmissionHeader.tsx
```diff
@@ -78,13 +77,10 @@ export const BlastSubmissionHeader = (props: Props) => {
...
   const editSubmission = () => {
     const { sequences, species, preset, parameters } = submission.submittedData;
-    const parsedSequences = sequences.flatMap((sequence) =>
-      parseBlastInput(sequence.value)
-    );
     const { program, stype, ...otherParameters } = parameters;
 
     const payload = {
-      sequences: parsedSequences,
+      sequences: sequences,
```
imransl, line 83: SInce both key and value are the same just `sequences` should do

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
```diff
@@ -72,7 +72,7 @@ export type BlastSubmission = {
   id: string;
   submittedData: {
     species: Species[];
-    sequences: { id: number; value: string }[]; // TODO: consider whether to have strings or parsed sequences
+    sequences: { id: number; header?: string; value: string }[]; // TODO: consider whether to have strings or parsed sequences
     preset: string;
     submissionName: string;
     parameters: BlastSubmissionParameters;
```
imransl, line 78: Shouldn't this be `BlastSubmissionPayload` type?

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -19,14 +19,16 @@ import type { BaseQueryFn } from '@reduxjs/toolkit/dist/query/baseQueryTypes';
...
+
 import type { BlastSettingsConfig } from 'src/content/app/tools/blast/types/blastSettings';
 import type { Species } from 'src/content/app/tools/blast/state/blast-form/blastFormSlice';
 import type { BlastSubmission } from '../blast-results/blastResultsSlice';
 import type { BlastJobResultResponse } from 'src/content/app/tools/blast/types/blastJob';
 
 export type BlastSubmissionPayload = {
   species: Species[];
-  sequences: { id: number; value: string }[];
+  sequences: { id: number; header?: string; value: string }[];
```
imransl, line 31: `sequences` is defined inside multiple types. Can't it be its own type and reused instead?

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
ridwanamo: we already have an export type `BlastSubmissionPayload` in blastApiSlice.ts  and in this one there is the extra part for the result which is related to a submission in blast hence just `BlastSubmission`. Anyway i didnt choose this name in this PR and it feels like it is the right name for it. Happy to discuss this.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
ridwanamo: we do have the following:

`export type ParsedInputSequence = {
  header?: string;
  value: string;
};`

Note the extra `id` field in the above sequences. This is needed once the sequences are submitted. I can still use the `ParsedInputSequence` type above but How do i include it with the sequences above?

imransl: We can create a new type using `ParsedInputSequence`:

```
type BlastInputSequence = ParsedInputSequence & { id: number; };
```

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -62,9 +64,15 @@ const blastApiSlice = restApiSlice.injectEndpoints({
       BlastSubmissionPayload
     >({
       query(payload) {
+        // merge the sequence and header using tofasta
+        const submittedSequences = payload.sequences.map((item) => ({
```
azangru, line 68: Technically, we haven't yet submitted these sequences. We are about to. Maybe `querySequences` instead of `submittedSequences`?

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -62,9 +64,15 @@ const blastApiSlice = restApiSlice.injectEndpoints({
       BlastSubmissionPayload
     >({
       query(payload) {
+        // merge the sequence and header using tofasta
```
azangru, line 67: This comment describes _what_ the code below is doing (which can be read from the code itself), rather than _why_ this is being done. Maybe replace with something like "The backend accepts sequences as FASTA strings".

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
```diff
@@ -72,7 +72,7 @@ export type BlastSubmission = {
   id: string;
   submittedData: {
     species: Species[];
-    sequences: { id: number; value: string }[]; // TODO: consider whether to have strings or parsed sequences
+    sequences: { id: number; header?: string; value: string }[]; // TODO: consider whether to have strings or parsed sequences
```
azangru, line 75: Re. the comment: this has been considered and todone in this pull request?

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
azangru: Regarding the name of the proposed new type. As the code is written (see `BlastJobSubmit`, line 93), sequences only acquire the `id` field at the moment of submission. Maybe call the type `StoredInputSequence` or `SubmittedInputSequence`?

src/content/app/tools/blast/state/blast-results/blastResultsSlice.ts
azangru: > Shouldn't this be BlastSubmissionPayload type?

A submission payload included, apart from BLAST parameters, a list of selected species, and a list of query sequences. The `BlastSubmissionParameters` type is only about the parameters (the ones that you select using the dropdowns in the BLAST submission form).

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
ridwanamo: what about just `SubmittedSequence`

src/content/app/tools/blast/types/parsedInputSequence.ts
```diff
@@ -18,3 +18,5 @@ export type ParsedInputSequence = {
   header?: string;
   value: string;
 };
+
+export type submittedSequence = ParsedInputSequence & { id: number };
```
azangru, line 22: Please start type names with a capital letter.

azangru: Also, I wonder if the declaration of this type more naturally belongs in the `blastResultsSlice.ts` file. Submitted sequences only exist around and after a submission. But I don't have a strong opinion about this.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -64,15 +65,15 @@ const blastApiSlice = restApiSlice.injectEndpoints({
       BlastSubmissionPayload
     >({
       query(payload) {
-        // merge the sequence and header using tofasta
-        const submittedSequences = payload.sequences.map((item) => ({
+        // backend toolsAPI accepts sequences as FASTA strings
```
azangru, line 68: toolsAPI -> tools API?

azangru:
Here's what I am seeing in IndexedDB. I wonder whether it won't make sense to only add the header field (and store it in the db) when it is really present. 

Screenshot 2022-11-16 at 15 31 35

azangru:
Also, did you lose the headers in the list view?

Screenshot 2022-11-16 at 15 39 06

compared to your `blast-result-layout` branch:

Screenshot 2022-11-16 at 15 38 48

ridwanamo:
> Here's what I am seeing in IndexedDB. I wonder whether it won't make sense to only add the header field (and store it in the db) when it is really present.
> 
> Screenshot 2022-11-16 at 15 31 35

@azangru push my changes to remove the header in the submitted sequence if there is no header.

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
```diff
@@ -89,13 +89,17 @@ const BlastJobSubmit = () => {
...
+  // labelling sequences with completely artificial identifiers
   // so that job ids in the response can be matched to individual combinations of sequences and genome ids
   const sequences = blastFormData.sequences.map((sequence, index) => ({
     id: index + 1,
     header: sequence.header,
     value: sequence.value
   }));
+  sequences.forEach((item) => {
+    if (!item.header) delete item.header;
+  });
```
azangru, line 101: Yuck! Let's do this inside the map instead.

If `blastFormData.sequences` sometimes have the header field and sometimes don't (please check), then it's as easy as:

```
const sequences = blastFormData.sequences.map((sequence, index) => ({
  ...sequence,
  id: index + 1,
}));
``` 

Otherwise, we'll need to do something like this:
```
const sequences = blastFormData.sequences.map((sequence, index) => {
 const sequenceForSubmission: SubmittedSequence = {
  id: index + 1,
  value: sequence.value
 };

 if (sequence.header) {
  sequenceForSubmission.header = sequence.header;
 }

 return sequenceForSubmission;
});
```

src/content/app/tools/blast/views/blast-submission-results/components/blast-results-per-sequence/BlastResultsPerSequence.tsx
```diff
@@ -37,6 +35,7 @@ import styles from './BlastResultsPerSequence.scss';
 type BlastResultsPerSequenceProps = {
   sequence: {
     id: number;
+    header?: string;
     value: string;
```
azangru, line 39: You now have declared a `SubmittedSequence` type; please use it throughout the PR.

ridwanamo: i thought i did replace everywhere but looks like i have missed a few places. should all be there now.

src/content/app/tools/blast/components/blast-job-submit/BlastJobSubmit.tsx
ridwanamo: Turn out the first option works since `blastFormData.sequences` dont have the header field when its empty. But if it did i am curious to why my solution wasn't good. 
i have made the suggested change but just curious why not the delete option.

azangru: >  But if it did i am curious to why my solution wasn't good. whats the difference between the delete and the map. Is it because we dont want to mutate the object?

It's probably a matter of style; but between (1) creating a header field and then deleting the header field if empty, and (2) not creating the header field if empty in the first place, the second option feels much more satisfying.

azangru (approved):

imransl (approved):

PR #878, opened 2022-11-13
Use React 18's own apis for server-side rendering

azangru:
## Description
Use React 18's own apis for server-side rendering, and give the whole `document` to React to control.

As a result of this PR we will be able to:
- Remove dependency on the `loadable-component` package, which hasn't been updated for React 18. (Version 5.15.2 that we use still has `"react": ">=16.3.0"` as its peer dependency.
- Remove dependency on `react-helmet-async`, which is a somewhat hacky way to inject page metadata into html head.

**Details**
- Updated viewRouter to use the new React `renderToPipeableStream` api
- Created the `Html` component, which gives React full control over the whole HTML document (based on Dan Abramov's [example sandbox](https://codesandbox.io/s/kind-sammet-j56ro))
- Added webpack-manifest-plugin to be able to read webpack manifest on the server, and inject it into the html
- Updated eslint-plugin-react to make it aware of the `noModule` attribute on the script tag
- Created a new redux slice responsible for page metadata (for data to be used in html head)
-  Removed all code that was relying on the `loadable-component` package; using `React.lazy` instead
- Removed all `React Helmet` code
- Only relevant fields are included in the assets manifest before transferring it to client (i.e. `client.css`)
- Remove packages:
  - @loadable/component
  - @loadable/server
  - react-helmet-async
  - @loadable/babel-plugin
  - @loadable/webpack-plugin
  - @types/loadable__component
  - @types/loadable__server
  - @types/loadable__webpack-plugin

### Note
- This PR disables hot module replacement. It was already semi-broken after we added server-side rendering; and the further complication is that the old implementation would rerun the `hydrateRoot` function on the same `document` node when hot module replacement occurs, which causes errors. A promising direction is to research [react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin); but this is a slow and labor-intensive task, which is better done in a subsequent PR.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1677

## Deployment URL(s)
http://react-full-html.review.ensembl.org


src/content/app/species/SpeciesPage.tsx
```diff
@@ -35,32 +35,39 @@ import { getPathParameters, useUrlParams } from 'src/shared/hooks/useUrlParams';
 import type { ServerFetch } from 'src/routes/routesConfig';
 import type { AppDispatch } from 'src/store';
 
-const LoadableSpeciesPageContent = loadable(
+const LazylyLoadedSpeciesPageContent = React.lazy(
   () => import('./SpeciesPageContent')
 );
 
+const defaultTitle = 'Species page — Ensembl';
+const defaultDescription = 'Species home page';
```
jyobhai, line 43: all other places you have pageTitle and pageDescription. Can we keep it the same?

webpack/client/production.ts
```diff
@@ -37,7 +37,7 @@ export default (): Configuration => {
           test: /.*\.(gif|png|jpe?g)$/i,
           type: 'asset/resource',
           generator: {
-            filename: 'images/[name].[hash][ext]'
+            filename: 'images/[name].[contenthash][ext]'
```
ridwanamo, line 40: why this change? whats the difference betwee `hash` key and the `contenthash` key

azangru: Here's the [full documentation](https://webpack.js.org/configuration/output/#template-strings) of webpack's options. `hash` is a deprecated option, replaced by `fullhash`. In [another place](https://webpack.js.org/concepts/under-the-hood/#output) in the docs, they say that the most common option is `contenthash`. Contenthash should refer to the hash from the content of the given generated file; so in theory, should be the most appropriate for caching purposes.

src/server/routes/viewsRouter.tsx
```diff
@@ -63,89 +55,59 @@ const viewRouter = async (req: Request, res: Response) => {
...
+};
 
-  res.status(status).send(responseString);
+const getBootstrapScripts = (assetsManifest: Record<string, string>) => {
+  // In development environment, the only entry point is the client.js file (it's huge and contains the runtime, third-party libs, and css)
+  // In production, webpack will code-split, and extract vendors.js and runtime-client.js chunks
+  return [
+    assetsManifest['client.js'],
+    assetsManifest['vendors.js'],
+    assetsManifest['runtime~client.js']
```
imransl, line 102: I see the tilde used here instead of the dash. Is this intentional?

src/routes/routesConfig.tsx
```diff
@@ -16,18 +16,30 @@
...
+  serverFetch as genomeBrowserPageServerFetch
+} from 'src/content/app/genome-browser/BrowserPage';
 import EntityViewerPage, {
   serverFetch as entityViewerServerFetch
 } from 'src/content/app/entity-viewer/EntityViewerPage';
-import BlastPage from 'src/content/app/tools/blast/BlastPage';
-import AboutPage from 'src/content/app/about/AboutPage';
-import HelpPage from 'src/content/app/help/HelpPage';
+import BlastPage, {
+  serverFetch as blastServerFetch
```
imransl, line 35: Do we need to keep the server fetch names consistent? Entity viewer and Blast don't have `page` in their imported names.

src/server/routes/viewsRouter.tsx
azangru: I didn't even notice it before you mentioned it; but yes; this is the name of the key that webpack generates in its manifest, and which I copied out of the manifest file. I don't know why it replaces hyphens with tildes.

image

azangru: @imransl there seems to be a problem with the screenshot (i didn't know this is even possible). The manifest file actually says this: 
```
{
  "client.css": "/static/client.8430ec0c8e0a9769a447.css",
  "client.js": "/static/client.30cdeb509016a05f4881.js",
  "unsupportedBrowser.css": "/static/unsupportedBrowser.7a874b03bf636ac9519c.css",
  "unsupportedBrowser.js": "/static/unsupportedBrowser.3423f30df83b48f77fd6.js",
  "runtime~client.js": "/static/runtime~client.51c005dbd8799cc97d18.js",
  "runtime~unsupportedBrowser.js": "/static/runtime~unsupportedBrowser.f3b0f2b750aee9b85d82.js",
  "771.css": "/static/771.9f148a96bff8837bd91e.css",
  "771.js": "/static/771.948ae1eccfce94966307.js"
  ...
}
```

azangru: Also; including `runtime~client.js` is a requirement for the app to bootstrap (tested this in a build that did not include this file). This file is not imported independently from any other file.

imransl: Wow that is some nice graphic processing by MacOS to change the characters like that in case you didn't accidentally change it 😄 

On a different note, that's a very interesting file naming convention.

imransl (approved):

ridwanamo (approved):

PR #881, opened 2022-11-16
show - when description is null for protein family

ridwanamo:
## Description
As per jira description to show "-" when there is no description (or its null) for protein domain. It turns out to be a more complicated fix than a simple check for null. Thoas is returning the correct value as null but we are using the description field as a key in an object which is causing the null to appeared since the key is treated as a string. 

Its not a good idea to have description as a key to an object since it can be empty or null. so had to refactor the code to take name as a key and having description in the field.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1776

## Deployment URL(s)
http://empty-protein-description.review.ensembl.org

## Views affected
Gene function under entity viewer (click on the first protein)

Example of gene with null description:
http://empty-protein-description.review.ensembl.org/entity-viewer/grch37/gene:ENSG00000139618?view=protein

### DIFF
--- a/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
+++ b/src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
@@ -18,7 +18,7 @@ import React from 'react';
 import classNames from 'classnames';
 import { scaleLinear, ScaleLinear } from 'd3';
 
-import { FamilyMatchInProduct } from 'src/content/app/entity-viewer/state/api/queries/proteinDomainsQuery';
+import { type FamilyMatchInProduct } from 'src/content/app/entity-viewer/state/api/queries/proteinDomainsQuery';
 
 import styles from './ProteinDomainImage.scss';
 
@@ -35,12 +35,17 @@ export type ProteinDomainImageProps = {
   };
 };
 
+type location = {
+  start: number;
+  end: number;
+};
+
 type ProteinDomainImageData = {
-  [resource_type: string]: {
-    [resource_description: string]: {
-      start: number;
-      end: number;
-    }[];
+  [resource_name: string]: {
+    [domain_name: string]: {
+      description: string;
+      locations: location[];
+    };
   };
 };
 
@@ -51,7 +56,7 @@ export const getDomainsByResourceGroups = (
 
   proteinDomains.forEach((domain) => {
     const {
-      sequence_family: { description: domainName },
+      sequence_family: { name: domainName, description: description },
       sequence_family: {
         source: { name: resource_name }
       },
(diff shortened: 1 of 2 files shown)


src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -18,7 +18,7 @@ import React from 'react';
 import classNames from 'classnames';
 import { scaleLinear, ScaleLinear } from 'd3';
 
-import { FamilyMatchInProduct } from 'src/content/app/entity-viewer/state/api/queries/proteinDomainsQuery';
+import { type FamilyMatchInProduct } from 'src/content/app/entity-viewer/state/api/queries/proteinDomainsQuery';
```
azangru, line 21: `import { type...` -> `import type {...` (you are only importing a type.

src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -35,12 +35,17 @@ export type ProteinDomainImageProps = {
   };
 };
 
+type location = {
```
azangru, line 38: - we begin type names with a capital letter
- however, we already have a `Location` type elsewhere

ridwanamo: I cant see where we have `Location` type, we do have a `location.ts` and a type `LocationWithinRegion` but not just `Location`. i know we used it elsewhere but i dont think its exported anywhere else.

src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -51,7 +56,7 @@ export const getDomainsByResourceGroups = (
 
   proteinDomains.forEach((domain) => {
     const {
-      sequence_family: { description: domainName },
+      sequence_family: { name: domainName, description: description },
```
imransl, line 59: Just `description` should do here and below

azangru (approved):

src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
imransl: @ridwanamo This hasn't been changed yet. It isn't a very important one, but do you intend to keep it this way?

ridwanamo: @imransl sorry i missed that. I have now pushed this change.

src/content/app/entity-viewer/gene-view/components/protein-domain-image/ProteinDomainImage.tsx
```diff
@@ -61,12 +66,14 @@ export const getDomainsByResourceGroups = (
     if (!groupedDomains[resource_name]) {
       groupedDomains[resource_name] = {};
     }
-
     if (!groupedDomains[resource_name][domainName]) {
-      groupedDomains[resource_name][domainName] = [];
+      groupedDomains[resource_name][domainName] = {
+        description: description,
```
imransl, line 71: You probably missed this one @ridwanamo 😄

imransl (approved):

PR #882, opened 2022-11-18
Update genome browser to 0.5.2

azangru:
## Description
[Related PR in the ensembl-genome-browser repo](https://github.com/Ensembl/ensembl-genome-browser)

Changes associated with this update

### Vertical scrolling

**Previous model**

Notice that:
- The canvas could be higher than the container element passed from chrome, into which the canvas was rendering. The height of the canvas dictated whether the scrollbar would appear.
- We used to render Zmenu and BrowserCogs as **siblings** of the element that we passed to the genome browser.

image

**New model**

Notice that:
- The canvas is always as high as the container element passed to the genome browser from chrome
- Genome browser also renders an empty div which is as high as the full vertical content of genome browser. This empty div is what is responsible for the appearance of the scrollbar
- Zmenu and BrowserCogs are now rendered as **children** of the container element passed to the genome browser 

image

### Error messaging
Genome browser now sends an error message when the client's version is out of date with the backend; and we respond to it by showing a screen that asks the user to refresh their browser.

Unfortunately, there was no time to properly refactor error messages at the Rust level; which is why it's a bit messy for this release, both in ensembl-genome-browser repo and here. Error messages are expected to be improved and made more standard in a future release of the genome browser.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1778

## Deployment URL(s)
http://gb-vertical-scrolling.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/components/browser-image/BrowserImage.tsx
+++ b/src/content/app/genome-browser/components/browser-image/BrowserImage.tsx
@@ -14,9 +14,15 @@
  * limitations under the License.
  */
 
-import React, { useRef, useEffect, memo } from 'react';
+import React, { useState, useRef, useEffect, memo } from 'react';
 import { useSelector } from 'react-redux';
 import classNames from 'classnames';
+import {
+  IncomingActionType,
+  GenomeBrowserErrorType,
+  type GenomeBrowserErrorAction,
+  type GenomeBrowserError as GenomeBrowserErrorObj
+} from '@ensembl/ensembl-genome-browser';
 
 import useGenomeBrowser from 'src/content/app/genome-browser/hooks/useGenomeBrowser';
 import useGenomeBrowserPosition from 'src/content/app/genome-browser/hooks/useGenomeBrowserPosition';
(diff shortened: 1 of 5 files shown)


src/content/app/genome-browser/components/browser-image/BrowserImage.tsx
```diff
@@ -14,9 +14,15 @@
...
 
-import React, { useRef, useEffect, memo } from 'react';
+import React, { useState, useRef, useEffect, memo } from 'react';
 import { useSelector } from 'react-redux';
 import classNames from 'classnames';
+import {
+  IncomingActionType,
+  GenomeBrowserErrorType,
+  type GenomeBrowserErrorAction,
+  type GenomeBrowserError as GenomeBrowserErrorObj
```
imransl, line 24: Won't `GenomeBrowserErrorType` be better here?

src/content/app/genome-browser/components/genome-browser-error/GenomeBrowserError.scss
```diff
@@ -0,0 +1,23 @@
...
+  padding-top: 56px;
+  row-gap: 34px;
+}
+
+.container p {
+  margin: 0;
+  text-align: center;
+}
+
+.container p + p {
```
imransl, line 17: Curious about why the sibling selector is used in this way with the same element

imransl (approved):

ens-st3:
I've not been able to consistently break anything

src/content/app/genome-browser/components/genome-browser-error/GenomeBrowserError.scss
imransl: Ah gotcha. So it is an alternative to `:not(:first-child)`.

azangru: True. I suppose `:not(:first-child)` would have been more expressive here.

PR #883, opened 2022-11-23
Global padding for blast

ridwanamo:
## Description
We have a repeating padding (0 30px 90px 60px) in all the blast view and creating a css variable for the padding to be use in all the blast view.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1788

## Deployment URL(s)
http://blast-margin.review.ensembl.org

## Views affected
blast (blast form - new job, blast submission view - unviewed jobs and job list, job result view)

### DIFF
--- a/src/content/app/tools/blast/BlastPage.scss
+++ b/src/content/app/tools/blast/BlastPage.scss
@@ -1,0 +1,6 @@
+@import 'src/styles/common';
+
+.blastPageContainer {
+  --blast-page-standard-padding: 0 #{$standard-gutter} #{$global-padding-bottom}
+    #{$double-standard-gutter};
+}
--- a/src/content/app/tools/blast/BlastPage.scss
+++ b/src/content/app/tools/blast/BlastPage.scss
@@ -1,7 +1,6 @@
 @import 'src/styles/common';
 
 .blastPage {
-  --blast-page-standard-padding: 0 #{$standard-gutter} #{$global-padding-bottom}
-    #{$double-standard-gutter};
+  --blast-page-standard-padding: 0 20px #{$global-padding-bottom} #{$double-standard-gutter};
   height: 100%;
 }
--- a/src/content/app/tools/blast/BlastPage.scss
+++ b/src/content/app/tools/blast/BlastPage.scss
@@ -1,6 +1,6 @@
 @import 'src/styles/common';
 
-.blastPageContainer {
+.blastPage {
   --blast-page-standard-padding: 0 #{$standard-gutter} #{$global-padding-bottom}
     #{$double-standard-gutter};
 }
--- a/src/content/app/tools/blast/BlastPage.scss
+++ b/src/content/app/tools/blast/BlastPage.scss
@@ -1,6 +1,8 @@
 @import 'src/styles/common';
 
 .blastPage {
-  --blast-page-standard-padding: 0 20px #{$global-padding-bottom} #{$double-standard-gutter};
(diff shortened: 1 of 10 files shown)


src/content/app/tools/blast/BlastPage.scss
```diff
@@ -0,0 +1,6 @@
+@import 'src/styles/common';
+
+.blastPageContainer {
```
azangru, line 3: Just `.blastPage` would be sufficient

src/content/app/tools/blast/BlastPage.scss
```diff
@@ -0,0 +1,6 @@
...
+  --blast-page-standard-padding: 0 #{$standard-gutter} #{$global-padding-bottom}
+    #{$double-standard-gutter};
+}
```
azangru, line 6: So here's what worries me; and what I tried to express yesterday.

The element wrapped inside the newly added div has a `height: 100%` rule:

image

But the parent element has no notion of a height:

image

which means that the child's `height: 100%` becomes meaningless. If we wanted to maintain the height appropriately, we would need to set it on this parent element as well.

azangru (approved):

src/content/app/tools/blast/views/blast-form/BlastForm.scss
```diff
@@ -9,20 +9,19 @@
...
-  padding-bottom: $global-padding-bottom;
+  padding: var(--blast-page-standard-padding);
 }
 
 .grid {
   display: grid;
   grid-template-columns: repeat(2, min-content);
   column-gap: 45px;
-  padding-bottom: $global-padding-bottom;
+  padding: var(--blast-page-standard-padding);
```
azangru, line 24: So here's the problem with the scrollbar that we observed. It suggests that our approach with the global BLAST-wide padding may be an incorrect one; because we may want to hand-adjust paddings for scrollable and not-scrollable containers.

_(or maybe that we need a `--blast-page-scrollable-padding` variable, besides  the `--blast-page-standard-padding` one)_


Uploading Screen Recording 2022-11-24 at 11.47.32.mov…

ridwanamo: @azangru this should be fixed with my last commit.

src/root/Root.tsx
```diff
@@ -74,4 +75,24 @@ export const Root = () => {
...
+
+  outer.appendChild(inner);
+  document.body.appendChild(outer);
+  const scrollbarWidth = outer.offsetWidth - outer.scrollWidth;
+
+  const rootNode = document.querySelector(':root') as HTMLElement;
+  rootNode?.style.setProperty('--scrollbar-width', `${scrollbarWidth}px`);
+
+  outer.parentNode?.removeChild(outer);
+};
```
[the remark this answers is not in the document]
ridwanamo: Done all 3 items and push the changes.

azangru:
Pushed several changes:

1. Changed top: 0px to top: -100px

**Why?** To move the element off the screen. With top: 0; it would still occupy space on the screen, as can be seen on this screenshot:

image

2. Removed the element's child

**Why?** The child wasn't being used for anything. For example, we weren't setting the child's height to be greater than the outer element's height.

3. Set initial value for the `scrollbar-width` custom property

**Why?** Just to feel safer.

imransl (approved):

azangru (approved):

PR #885, opened 2022-11-24
Fix tracks disappearing when drawer opened

imransl:
## Description
The tracks apart from the the focus track disappears when the drawer has been opened for a while. This however happens only when you leave the genome browser app and return to it.

Steps to reproduce:

1. Open the drawer by clicking on the ellipsis (three dots) button of any track
2. Click on another app or home
3. Wait for a minute and then click on genome browser
4. You will see the drawer is still opened. Close it and notice only the focus track being displayed in the genome browser.

## Related JIRA Issue(s)
[ENSWBSITES-1782](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1782)

## Deployment URL(s)
http://fix-disappearing-tracks.review.ensembl.org

## Views affected
Genome browser

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomicTracks.ts
+++ b/src/content/app/genome-browser/hooks/useGenomicTracks.ts
@@ -55,6 +55,7 @@ const useGenomicTracks = () => {
     if (
       !genomeBrowser ||
       !trackSettingsForGenome ||
+      !genomeTrackCategories ||
       genomeIdInitialisedRef.current === activeGenomeId
     ) {
       return;


imransl:
I couldn't get around why this was happening even when `useEffect` is supposed to fire when `genomicTrackCategories` changes: https://github.com/Ensembl/ensembl-client/blob/fix-disappearing-tracks/src/content/app/genome-browser/hooks/useGenomicTracks.ts#L72

azangru:
There's a bug that is almost certainly very closely related to what you were doing in this PR.

Steps to reproduce:
- select more than one species
- visit genome browser and go past the interstitial screen for one of the species
- now toggle between species such that genome browser gets mounted and unmounted (replaced by the interstitial screen)
**Bug:** you will again see that the tracks aren't being shown.

Don't know if it's something to be tackled in this PR or in another one


https://user-images.githubusercontent.com/6834224/203937784-ff07748c-fdfc-4b8c-b6c8-98f35ee92660.mov

imransl:
@azangru I've fixed that issue now. Turned out that we need to reset the `genomeIdInitialisedRef` to rerun the `useEffect`. Also noticed that this was enough to fix the tracks disappearing for the first issue (returning from another app) too. However, I've left the `genomeTrackCategories` check as it would be useless to run the `useEffect` when it is `undefined`.

src/content/app/genome-browser/hooks/useGenomicTracks.ts
```diff
@@ -66,6 +67,11 @@ const useGenomicTracks = () => {
     );
     trackIdsList.forEach(genomeBrowserMethods.toggleTrack);
     genomeIdInitialisedRef.current = activeGenomeId as string;
+
+    // reset this ref on unmount to rerun this useEffect on mount for the last selected genome id
+    return () => {
+      genomeIdInitialisedRef.current = '';
+    };
```
azangru, line 74: I am awfully confused by this. The cleanup function from a useEffect runs when a component is unmounted; but when it is unmounted, the ref should disappear as well. Refs don't survive after a component is removed.

imransl: It doesn't seem to be resetting by itself once I remove that function.

Screenshot 2022-11-25 at 17 18 36

src/content/app/genome-browser/hooks/useGenomicTracks.ts
```diff
@@ -66,6 +67,11 @@ const useGenomicTracks = () => {
     );
     trackIdsList.forEach(genomeBrowserMethods.toggleTrack);
     genomeIdInitialisedRef.current = activeGenomeId as string;
+
+    // reset this ref on unmount to rerun this useEffect on mount for the last selected genome id
```
azangru, line 71: We should probably update — or maybe even delete — this comment. `useEffect`'s cleanup function doesn't run only when the component unmounts, but also when any of the members of the dependences array changes. Since this hook is run in the `Browser` component, we don't expect any unmounting when you switch between species, even if one of the species shows the interstitial. The comment doesn't really add anything to what's already written on line 73.

imransl: That's done now ✅

azangru (approved):

PR #887, opened 2022-11-27
Update launchbar button for Entity Viewer to avoid the flickering interstitial

azangru:
## Description
We were generating rather imprecise links to EntityViewer:
- In the launchbar, the link associated with the EntityViewer button was just `/entity-viewer`
- Within EntityViewer, the links generated by clicks on species lozenges were `/entity-viewer/:genomeId`

We were relying on EntityViewer to figure our on its own what it wants to display. Which it eventually did, but in the process, it could cycle through the interstitial screen, while figuring out whether it had an active genome id and gene id.

This PR updates links generated by the launchbar button and species lozenges, such that they open the exact EntityViewer screen based on which genome is active, which entity is active, and which view the user was in last.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1588

## Deployment URL(s)
http://update-link-to-entity-viewer.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/shared/components/entity-viewer-app-bar/EntityViewerAppBar.tsx
+++ b/src/content/app/entity-viewer/shared/components/entity-viewer-app-bar/EntityViewerAppBar.tsx
@@ -15,15 +15,23 @@
  */
 
 import React, { memo, useMemo } from 'react';
-import { useSelector } from 'react-redux';
 import { Link, useNavigate } from 'react-router-dom';
 import isEqual from 'lodash/isEqual';
 
+import { useAppSelector } from 'src/store';
 import * as urlFor from 'src/shared/helpers/urlHelper';
 import { AppName } from 'src/global/globalConfig';
+import {
+  parseFocusObjectId,
+  buildFocusIdForUrl
+} from 'src/shared/helpers/focusObjectHelpers';
 
-import { getEntityViewerActiveGenomeId } from 'src/content/app/entity-viewer/state/general/entityViewerGeneralSelectors';
+import {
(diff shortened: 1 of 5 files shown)


imransl (approved):

jyothishnt:
I was thinking of another solution and wanted to discuss this. I think we could tackle it where you create the Entityviewer routes to show interstitial here https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/entity-viewer/EntityViewer.tsx#L85

A validation before creating the routes like below is enough to solve this problem I think. I tested it on my sandbox and it works. Do you think any issues with this approach @azangru @imransl @ridwanamo ?

`  const renderEntityViewerRoutes = () => (
    <Routes>
      {!activeGenomeId && !activeEntityId && <Route index element={<EntityViewerInterstitial />} /> }
      {!activeEntityId && <Route path="/:genomeId" element={<EntityViewerInterstitial />} /> }
      <Route path="/:genomeId/:entityId" element={<EntityViewerForGene />} />
    </Routes>
  );
`

azangru:
@jyothishnt could you push your branch so that we can inspect the review deployment? 

Though personally, I'd much rather we didn't tinker with the router. I would much prefer the routes to be predictably set at the boot-up of the app rather than getting constantly changed underneath.

jyothishnt:
Review deployment here. I have only added this in EV, so any route from other apps to EV should not see any flickering.

http://update-ev-link-approach2.review.ensembl.org/

jyothishnt (approved):

PR #888, opened 2022-11-29
Update database value for softmasked sequences

jyothishnt:
## Description
Changes to blast config to use softmaksed sequences

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1784

## Deployment URL(s)
http://new-blast-db.review.ensembl.org

## Views affected
BLAST

### DIFF
--- a/tests/fixtures/blast/blastSettingsConfig.json
+++ b/tests/fixtures/blast/blastSettingsConfig.json
@@ -594,15 +594,26 @@
     }
   },
   "database_sequence_types": {
+    "dna_sm": "dna",
     "dna": "dna",
     "cdna": "dna",
     "pep": "protein"
   },
   "programs_configurator": [
     {
+      "sequence_type": "dna_sm",
+      "database_type": "dna",
+      "programs": ["blastn", "tblastx"]
+    },
+    {
       "sequence_type": "dna",
       "database_type": "dna",
       "programs": ["blastn", "tblastx"]
+    },
+    {
+      "sequence_type": "dna_sm",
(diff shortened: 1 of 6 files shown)


tests/fixtures/blast/blastSettingsConfig.json
```diff
@@ -590,16 +594,27 @@
...
   },
   "database_sequence_types": {
+    "dna_sm": "dna",
     "dna": "dna",
     "cdna": "dna",
     "pep": "protein"
   },
   "programs_configurator": [
+    {
+      "sequence_type": "dna_sm",
```
azangru, line 604: This (and same in another entry below) is probably wrong. Look at the configurator for protein below — it says `sequence_type: "protein"`, rather that `sequence_type: "pep"`.

I actually suspect that these two entries that you added to the configurator are unnecessary. Need to check and update Andres's config.

jyothishnt: Doh! right, this is sequence type and not database type!

azangru (approved):
Generally looks good.

Due to this change, however, we can't deploy the client before the server. We'll need to tell Andres to deploy the updated tools api on staging first – and merge this PR right after that. Same with deploying to production.

Screenshot 2022-12-08 at 14 09 50

jyothishnt:
Tested on my sandbox and it works fine with the current blastconfig in staging. Means we can merge this without breaking. @azangru

imransl (approved):

PR #889, opened 2022-12-01
Focus location

azangru:
## Description
Enable locations as focus objects for the genome browser, even before the genome browser update to a version that properly supports this.

Related PR: https://github.com/Ensembl/ensembl-genome-browser/pull/18

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1789

## Deployment URL(s)
http://focus-location.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/hooks/useFocusTrack.ts
+++ b/src/content/app/genome-browser/hooks/useFocusTrack.ts
@@ -34,10 +34,7 @@ import type {
   FocusGeneTrack,
   FocusGeneTrackSettings
 } from 'src/content/app/genome-browser/state/track-settings/trackSettingsSlice';
-import type {
-  FocusGene,
-  FocusObjectIdConstituents
-} from 'src/shared/types/focus-object/focusObjectTypes';
+import type { FocusGene } from 'src/shared/types/focus-object/focusObjectTypes';
 
 /**
  * The purposes of this hook are:
@@ -59,8 +56,7 @@ const useFocusTrack = () => {
     genomeBrowserMethods,
     focusGene:
       parsedFocusObjectId?.type === 'gene' ? (focusObject as FocusGene) : null,
-    focusObjectId,
-    parsedFocusObjectId
+    focusObjectId
   });
 
   /**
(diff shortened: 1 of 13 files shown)


ridwanamo:
Not sure if this is part of this PR, we need some validation for the focus location (same like we have for gene). So if you go to the following URL:
http://focus-location.review.ensembl.org/genome-browser/grch38?focus=location:foo (rubbish in location parameters) - its broken

And this one is a very interesting one:
http://focus-location.review.ensembl.org/genome-browser/grch38?focus=location:13:32,547,042-32,717,406&location=13:32272495-32442859

Notice the comma in the focus location. The interesting part of this URL is that it works and also create a bookmark(previously viewed) of this location but when you click on the previously viewed for this object, it breaks.

azangru:
> Not sure if this is part of this PR, we need some validation for the focus location

Yeah, validations for urls are definitely a different ticket.

imransl:
I came across another issue while playing with different locations. Once a location object breaks, the other valid locations don't work properly too.


https://user-images.githubusercontent.com/32512909/205713385-91548814-d940-46ce-ac5e-d3eaa27a3586.mov

azangru:
I don't know how we are going to handle invalid urls, but I would argue that it's outside of scope of this PR.

ridwanamo:
Just notice this ticket in the next sprint which is about URL validation:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-314

ridwanamo (approved):

src/content/app/genome-browser/hooks/useFocusTrack.ts
```diff
@@ -103,8 +93,8 @@ const useFocusGene = (params: Params) => {
 
   useEffect(() => {
     focusObjectIdRef.current = focusObjectId;
-    geneIdRef.current = parsedFocusObjectId?.objectId;
-  }, [focusObjectId, parsedFocusObjectId?.objectId]);
+    geneIdRef.current = focusGene?.stable_id;
```
imransl, line 96: Why not use `geneStableId` here, defined above, instead?

azangru: Good point. Updated.

imransl (approved):

PR #893, opened 2022-12-07
Update protein domains image to show domain information in a tooltip

azangru:
## Description
Upon a click on a protein domain rectangle in the protein diagram in Entity Viewer, the user will see a tooltip containing details about this domain.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1674

## Deployment URL(s)
http://protein-domains-image.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.scss
+++ b/src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction.scss
@@ -10,7 +10,7 @@
   min-height: 300px;
   margin-right: 30px;
   padding-bottom: 100px;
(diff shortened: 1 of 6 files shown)


ridwanamo:
One thing that i have noticed and you mentionned before is the first protein is not open by default when you go to the view first time. I tried in incognito and i also noticed that even the transcript view is not showing the first one as open.

azangru:
> One thing that i have noticed and you mentionned before is the first protein is not open by default when you go to the view first time.

Yes, this only happens locally when you run the site in dev mode. I suspect it may have something to do with the double execution of `useEffect`s within react StrictMode wrapper in development. It's fine in production builds.

ridwanamo (approved):

imransl (approved):

PR #894, opened 2022-12-07
Add new navigation panel to Genome browser

imransl:
## Description
A new navigation panel has been added as a sidebar modal on the Genome browser page. There are two versions of the modal: one for navigating within a region, and the other for switching between regions.

The obsolete browser navigation components and redundant tests were removed in: #897.

## Related JIRA Issue(s)
[ENSWBSITES-1783](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1783)
[ENSWBSITES-1791](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1791)
[ENSWBSITES-1801](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1801)

## Deployment URL(s)
http://new-navigation-panel.review.ensembl.org


src/content/app/genome-browser/components/browser-nav-button/BrowserNavButton.tsx
```diff
@@ -15,7 +15,8 @@
  */
 
 import React, { memo } from 'react';
-import type { OutgoingAction } from '@ensembl/ensembl-genome-browser';
+import { OutgoingAction } from '@ensembl/ensembl-genome-browser';
```
azangru, line 18: Why the change here?

src/content/app/genome-browser/components/browser-nav-button/browserNavConfig.ts
```diff
@@ -34,20 +32,20 @@ export type BrowserNavItem = {
 
 export const browserNavConfig: BrowserNavItem[] = [
```
azangru, line 33: I've been wondering why do we even need this file. Why not write most of this inline in jsx in the component responsible for rendering the nav buttons?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,80 @@
...
+  height: 22px;
+  width: 22px;
+}
+
+.zoomInButton,
+.zoomOutButton,
+.browserReset {
+  height: 16px;
+  width: 16px;
+}
```
azangru, line 60: By doing so, aren't you now competing with the selectors of the same specificity on these buttons? If so, then the styles will be relying on the order in which CSS rules are declared, which will eventually rely on the order in which Webpack puts together the CSS files.

Is this perhaps an argument for the use of CSS variables?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+  };
+
+  const getKaryotypeOptions = () =>
+    genomeKaryotype.map(({ name }) => ({
+      value: name,
+      label: name,
+      isSelected: stickInput === name
+    }));
+
+  const updateStickInput = (value: string) => {
```
azangru, line 131: `stick` is a term used inside the genome browser. In browser chrome, I think we call them regions.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,80 @@
+@import 'src/styles/common';
```
azangru, line 1: Weren't you planning to add a grey background to the navigation modal in this PR?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+import { getNumberWithoutCommas } from 'src/shared/helpers/formatters/numberFormatter';
+import {
+  validateRegion,
+  type RegionValidationErrors
+} from 'src/content/app/genome-browser/helpers/browserHelper';
+
+import Input from 'src/shared/components/input/Input';
+import { PrimaryButton } from 'src/shared/components/button/Button';
+import Tooltip from 'src/shared/components/tooltip/Tooltip';
+import Select from 'src/shared/components/select/Select';
```
azangru, line 35: We are using a different select component now. The one you are importing looks like this:

image

whereas what the XD shows is this:

image

[Here](https://ensembl.github.io/ensembl-client/storybook/index.html?path=/story/components-shared-components-simpleselect--simple-select-default-story) is what you are after.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
azangru: Also, check distances from the left (i.e. from the toolstrip).

Here's what I am seeing in the review deployment:

image

whereas here is what I am seeing in the XD:

image

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,80 @@
...
+.zoomInButton,
+.zoomOutButton,
+.browserReset {
+  height: 16px;
+  width: 16px;
+}
+
+.formButtons {
+  margin-top: 30px;
+  width: 272px; // TODO: find a better way to position the buttons rather than relying on the width
```
azangru, line 64: Perhaps these two should be wrapped in a single element with a fixed width?

image

Or (possibly even better) perhaps the whole section can be set to have a certain fixed width?

image

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -30,7 +30,8 @@ export enum BrowserSidebarModalView {
   BOOKMARKS = 'Previously viewed',
   SHARE = 'Share',
   DOWNLOADS = 'Downloads',
-  NAVIGATE = 'Navigate'
+  NAVIGATE_REGION = 'Navigate region',
+  NAVIGATE_LOCATION = 'Navigate location'
```
azangru, line 34: I can't be 100% certain; but I think the behaviour currently seen on the review deployment is incorrect. Specifically, pressing the button in the toolstrip when the modal is on should probably close the modal regardless of which of the two modals the user had open, whereas on the review deployment, pressing the button when you are on the second modal will take you back to the first one:

https://user-images.githubusercontent.com/6834224/208416013-c8bdd4fc-1dd5-43e5-8255-12c242411351.mov

src/content/app/genome-browser/components/browser-nav-button/browserNavConfig.ts
imransl: ✅

src/content/app/genome-browser/components/browser-nav-button/BrowserNavButton.tsx
imransl: Probably missed this one during the merge

✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: renamed this to `chrInput` as there is already a `regionInput`. Couldn't think of any better name. ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
imransl: Fixed the background colour and created a new ticket to check the paddings: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1821

✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+        <div className={styles.inputGroup} ref={regionRef}>
+          <label>
+            <span>Go to</span>
+            <Input
+              type="text"
+              onFocus={onRegionInputFocus}
+              onChange={onRegionChange}
+              disabled={coordInputsActive}
+              value={regionInput}
+              placeholder="Location to co-ordinates..."
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
imransl: Applied fixed width to the whole section ✅

src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
azangru: > as there is already a regionInput

Shouldn't that `regionInput` be renamed to `locationInput`?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
azangru: Check XD for the behaviour of inputs:

https://user-images.githubusercontent.com/6834224/208934407-f71be796-223c-4015-ab9b-9b4f6a05b3e4.mp4

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
azangru: Do we start with 1? Or should we start with an unselectable empty option "Select"?


https://user-images.githubusercontent.com/6834224/208935285-9b4439af-5939-4649-bbaf-b9a9707c7577.mp4

ens-st3:
Two comments.
1. The coordinate input box doesn't work
Screenshot 2022-12-22 at 14 30 47

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
imransl: Fixed that

✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: That's done now ✅ 

The option selected is not reset for now as that seems to be causing other issues at the moment. Will look into this later.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,317 @@
...
+          </label>
+        </div>
+      </div>
+      <div className={styles.formButtons}>
+        {(coordInputsActive || regionInputActive) && (
+          <span
+            className={classNames(styles.cancel, styles.clickableText)}
+            onClick={resetForm}
+          >
+            cancel
```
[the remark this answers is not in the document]
imransl: ✅

imransl:
> Two comments.
> 
> 1. The coordinate input box doesn't work (sorry the popup hasn't appeared on my screenshot)

I have fixed issues related to the navigation.

> 2. Going to a new focus location doesn't set it as a focus location. It therefore doesn't appear in the 'previously viewed'

I was not sure if we need to change the new location to a focus location. I thought that this is similar to the navigation bar fields that only changed the location rather than the region (the new focus location).

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
azangru: image

I am seeing blank space now.

src/content/app/genome-browser/components/browser-sidebar-modal/BrowserSidebarModal.tsx
```diff
@@ -27,6 +27,8 @@ import { closeDrawer } from 'src/content/app/genome-browser/state/drawer/drawerS
 
 import SidebarModal from 'src/shared/components/layout/sidebar-modal/SidebarModal';
 
+import navigateModalStyles from './modal-views/navigate-modal/NavigateModal.scss';
```
azangru, line 30: Mmm, the grey background feels like it belongs in BrowserSidebarModal's own scss file. What if any modal other than the navigation one wants to be grey?

imransl: Yes that's a good point. However, isn't there a possibility grey background can be used in other apps as well? So should it go in SidebarModal.scss?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: Argh I forgot to add the text `Select` as the label.

src/content/app/genome-browser/components/browser-sidebar-modal/BrowserSidebarModal.tsx
azangru: If it goes there, then the way to invoke the grey background would be to add something like a `theme` prop to SidebarModal component. I am trying to think if there are any downsides to doing so, but can't think of any. Unless Andrea gets super inventive with the modals.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,102 @@
...
+  .moveRightButton {
+    height: 22px;
+    width: 22px;
+  }
+
+  .zoomInButton,
+  .zoomOutButton {
+    height: 16px;
+    width: 16px;
+  }
```
azangru, line 32: Shouldn't we do for the move and zoom buttons the same as you did for the browser reset button? Your BrowserNavButton defines its own width and height.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,102 @@
...
+  .zoomOutButton {
+    display: inline-block;
+    margin-right: 35px;
+  }
+
+  .moveLeftButton,
+  .moveRightButton {
+    height: 22px;
+    width: 22px;
+  }
```
azangru, line 26: Aren't these the same as the default?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,102 @@
...
+}
+
+.browserNavBarControls {
+  margin: 15px 0 0 40px;
+
+  .moveLeftButton,
+  .moveRightButton,
+  .zoomInButton,
+  .zoomOutButton {
+    display: inline-block;
```
azangru, line 18: Shouldn't `display` be BrowserNavButton's property?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,345 @@
...
+    const {
+      startError = null,
+      endError = null,
+      regionError = null
+    } = errorMessages;
+    updateErrorMessages(startError, endError, regionError);
+  };
+
+  const resetForm = () => {
+    // setRegionNameInput('');
```
azangru, line 184: :sciss

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,102 @@
...
+  label {
+    & > span {
+      display: inline-block;
+      margin-right: 10px;
+      text-align: right;
+      width: 32px;
+    }
+  }
+
+  input {
```
azangru, line 54: Our Input component accepts class names. Let's use that rather than the tag name.

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -0,0 +1,102 @@
...
+    border: 1px solid $light-grey;
+    height: 30px;
+    padding: 2px 6px;
+
+    &:disabled {
+      border: 1px solid $grey;
+    }
+  }
```
azangru, line 62: image

The styles that we are seeing here should probably be defined on the Input component, and be adjustable through variables. Probably not for this PR; but we should probably have a ticket for this.

azangru:
To confirm with Andrea:
- Should the input boxes for start and end coordinates (only in the `NavigateRegionModal` though) start as empty or as showing the current location?
- Should the only way to navigate to a region be by pressing the green "Go" button; or should the "Enter" key also be supported (we would ignore Enter key presses if the form is only patrly completed)?
- After the user completes the navigation, should the inputs clear, or should they keep showing the user's input (until the user closes the modal)?

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/BrowserSidebarModal.tsx
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateLocationModal.tsx
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    changeBrowserLocation({
+      genomeId: activeGenomeId as string,
+      chrLocation: newChrLocation,
+      focus: {
+        type: 'location',
+        id: `location:${newChrLocationStr}`
+      }
+    });
```
imransl, line 210: I tried using this to change the focus object. However, this doesn't work. Should it be done different?

azangru: - create a new genome browser url like [here](https://github.com/Ensembl/ensembl-client/pull/906/files#diff-a31338b9080b764e86eb24c99ee592ef35fa20e46932d3f212e9ea5691abd214R489-R492)
- navigate to that url

imransl: ✅

azangru:
General comment: when testing the UI, I am seeing that sometimes things work and other times they don't. For example, here's me entering region coordinates and pressing enter:

https://user-images.githubusercontent.com/6834224/210514450-f18471fa-b365-48e6-a6b6-04e225fa9f79.mov

Also, the Esc key does not work (does not act as Cancel).

azangru (approved):

PR #897, opened 2022-12-09
Remove browser navigation bar components and tests

imransl:
## Description
Created a new PR to remove the browser navigation bar components rather than doing it in #894 to reduce the number of diff lines. Some other parts of the code (redux and higher level browser components) will also have deletions.

## Related JIRA Issue(s)
[ENSWBSITES-1811](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1811)

## Deployment URL(s)
http://remove-browser-nav-bar.review.ensembl.org

## Views affected
Genome browser


src/content/app/genome-browser/components/browser-nav-icon/BrowserNavIcon.tsx
```diff
@@ -23,7 +23,7 @@ import ImageButton from 'src/shared/components/image-button/ImageButton';
 
 import { browserNavIconActionMap } from 'src/content/app/genome-browser/state/browser-nav/browserNavSlice';
 
-import { BrowserNavItem } from 'src/content/app/genome-browser/components/browser-nav/browserNavConfig';
+import { BrowserNavItem } from 'src/content/app/genome-browser/components/browser-nav-icon/browserNavConfig';
```
azangru, line 26: What does the component called `browser-nav-icon` do? Is it now responsible for the five buttons that move the genome browser left/right, zoom in/out and refocus on the focus object? If yes, then should we rename the directory (`browser-nav-buttons`, or something)?

azangru (approved):

src/content/app/genome-browser/components/browser-nav-icon/BrowserNavIcon.tsx
imransl: Agree with this. `BrowserNavIcon` uses `ImageButton` which itself suggests that it is more than a simple icon. I have renamed the directory and the component to be `BrowserNavButton`.

imransl:
I pushed a couple of commits just now renaming `BrowserNavIcon` to `BrowserNavButton`. Appreciate if you can have another look at it 🙏

azangru (approved):
Go ahead with the merge; we'll pick up the discussion when your main PR is ready.

PR #898, opened 2022-12-09
update StandardAppLayout

imransl:
## Description
The sidebar modal for Genome browser needs to be customisable because of the new navigation panel. However, this was not possible as there was a padding specified for the parent sidebar element. So this PR does the following to fix the issue (as suggested in the JIRA ticket):

- Add a `className` prop to `SidebarModal` to customise the styling
- Add a new `Sidebar` component

## Related JIRA Issue(s)
[ENSWBSITES-1810](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1810)

## Deployment URL(s)
http://update-standard-app-layout.review.ensembl.org

## Views affected
Sidebars in Genome browser, Species page and Entity viewer

### DIFF
--- a/src/shared/components/layout/StandardAppLayout.scss
+++ b/src/shared/components/layout/StandardAppLayout.scss
@@ -86,14 +86,13 @@ $drawerWindowWidth: 45px;
   height: 100%;
   width: $sidebarContentWidth;
   border-left: 1px solid $grey;
-  padding: 21px 0 0 12px;
   background-color: white;
   overflow: auto;
   scrollbar-gutter: stable;
 
   & > div {
     padding-bottom: $global-padding-bottom;
-    max-width: calc($sidebarContentWidth - 25px);
+    max-width: calc($sidebarContentWidth - 13px);
   }
 }
 
--- a/src/shared/components/layout/sidebar-modal/SidebarModal.scss
+++ b/src/shared/components/layout/sidebar-modal/SidebarModal.scss
@@ -8,6 +8,7 @@
   grid-template-columns: 1fr 20px;
   grid-template-rows: 38px 1fr;
   height: 100%;
+  padding: 21px 0 0 12px;
 }
 
 .title {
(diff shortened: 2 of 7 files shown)


src/shared/components/layout/StandardAppLayout.scss
```diff
@@ -86,14 +86,13 @@ $drawerWindowWidth: 45px;
...
-  padding: 21px 0 0 12px;
   background-color: white;
   overflow: auto;
   scrollbar-gutter: stable;
 
   & > div {
     padding-bottom: $global-padding-bottom;
-    max-width: calc($sidebarContentWidth - 25px);
+    max-width: calc($sidebarContentWidth - 13px);
   }
```
azangru, line 96: Does this selector even make sense here anymore? Shouldn't padding-bottom be defined in the Sidebar component now? 

As for max-width, we should probably do `$sidebarContentWidth - var(--scrollbar-width)` (and also on the Sidebar component).

src/shared/components/layout/sidebar/Sidebar.scss
```diff
@@ -0,0 +1,5 @@
+@import 'src/styles/common';
+
+.wrapper {
```
azangru, line 3: `.wrapper` -> `.sidebar`? Same as the component name?

src/shared/components/layout/sidebar-modal/SidebarModal.scss
```diff
@@ -8,6 +8,7 @@
   grid-template-columns: 1fr 20px;
   grid-template-rows: 38px 1fr;
   height: 100%;
+  padding: 21px 0 0 12px;
```
azangru, line 11: I see these are the same values as inside the `Sidebar` component. Does `SidebarModal` want to be inside the `Sidebar`, by any chance?

src/shared/components/layout/sidebar/Sidebar.scss
imransl: ✅

src/shared/components/layout/StandardAppLayout.scss
imransl: ✅

PR #904, opened 2022-12-19
Use a pill button to toggle the list of transcripts in the track panel of genome browser

azangru:
## Description
In the Track Panel of the genome browser, replace the chevron next to gene name:

image

with a pill button under gene's transcripts:

image

**QUESTION:** should we delete the `GroupTrackPanelItemLayout` component? With the chevron gone from the gene, it is no longer used anywhere...

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1809

## Deployment URL(s)
http://gb-track-panel-showhide.review.ensembl.org


imransl:
> QUESTION: should we delete the GroupTrackPanelItemLayout component? With the chevron gone from the gene, it is no longer used anywhere...

Yes, let's delete this.

imransl (approved):

PR #906, opened 2022-12-21
Add links from the BLAST table to genome browser

azangru:
## Description
For BLAST submissions against assembly databases (DNA, DNA softmasked), add links from genomic locations in the BLAST results table to the genome browser. Clicking on such link would treat the BLAST hit location it as a focus location for the genome browser.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1787

## Deployment URL(s)
http://blast-to-gb-link.review.ensembl.org

### DIFF
--- a/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.tsx
+++ b/src/content/app/tools/blast/components/blast-app-bar/BlastAppBar.tsx
@@ -70,7 +70,8 @@ const BlastAppBar = () => {
           genome_id: species.genome_id,
           common_name: species.common_name,
           scientific_name: species.scientific_name,
-          assembly_name: species.assembly_name
+          assembly_name: species.assembly_name,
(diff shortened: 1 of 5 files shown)


ens-st3 (approved):
Looks great, immediately feels more useful. A question for me is what should happen when I click on a link to the GB for a species that I've not previously selected. Currently that is not added to the ones I've selected but I suspect it might be. A decision for another day though

azangru:
> Currently that is not added to the ones I've selected but I suspect it might be.

Not sure what you mean. Here's a screen recording where it is. It would be a real challenge not to have it added.

https://user-images.githubusercontent.com/6834224/209157653-77c000a7-0a57-4d00-add6-063f04606b49.mov

imransl (approved):

PR #907, opened 2022-12-21
Add example location links to species and genome browser pages

imransl:
## Description
Previously region links were shown in the interstitial page of the genome browser and the species page. However, they were removed after some time. Recently, the genome browser started supporting location focus objects, which replaces the regions. So we can now display location example  links in species and genome browser pages.

The data sent by the endpoint still mentions `region` rather than `location` as the type. So this might need to be changed in the future when it sends `location` as the type.

To test this in the species page, you just need to go the species page of a particular genome. However, for the genome browser you will need to go to the interstitial page of a genome.

## Related JIRA Issue(s)
(ENSWBSITES-1780)[https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1780]

## Deployment URL(s)
http://restore-links.review.ensembl.org

## Views affected
Species page
Genome browser

### DIFF
--- a/src/content/app/genome-browser/components/interstitial/BrowserInterstitial.tsx
+++ b/src/content/app/genome-browser/components/interstitial/BrowserInterstitial.tsx
@@ -23,7 +23,10 @@ import useGenomeBrowserAnalytics from 'src/content/app/genome-browser/hooks/useG
 
 import { fetchExampleFocusObjects } from 'src/content/app/genome-browser/state/focus-object/focusObjectSlice';
 
-import { getExampleGenes } from 'src/content/app/genome-browser/state/focus-object/focusObjectSelectors';
+import {
+  getExampleGenes,
+  getExampleLocations
+} from 'src/content/app/genome-browser/state/focus-object/focusObjectSelectors';
 
 import {
   parseFocusObjectId,
@@ -72,8 +75,9 @@ const BrowserInterstitial = () => {
 const ExampleLinks = () => {
   const { genomeIdForUrl } = useGenomeBrowserIds();
   const focusObjects = useAppSelector(getExampleGenes);
+  const focusLocations = useAppSelector(getExampleLocations);
 
-  const links = focusObjects.map((exampleObject) => {
+  const geneLinks = focusObjects.map((exampleObject) => {
     const parsedFocusObjectId = parseFocusObjectId(exampleObject.object_id);
     const focusId = buildFocusIdForUrl(parsedFocusObjectId);
     const path = urlFor.browser({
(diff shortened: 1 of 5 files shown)


src/content/app/genome-browser/components/interstitial/BrowserInterstitial.tsx
```diff
@@ -72,8 +75,26 @@ const BrowserInterstitial = () => {
...
+      <div key={exampleObject.object_id}>
+        <Link to={path} replace>
+          Example {exampleObject.type}
+        </Link>
+      </div>
+    );
+  });
 
-  const links = focusObjects.map((exampleObject) => {
+  const locationLinks = focusLocations.map((exampleObject) => {
```
azangru, line 97: What is different in how location links are processed as opposed to gene links? Why is this a separate function?

src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -678,31 +677,30 @@ const getExampleLinks = (props: {
         }
       };
     }
+  } else if (section === SpeciesStatsSection.ASSEMBLY) {
+    // TODO: Change the type check to location when data type returned switches to location
+    const regionExample = exampleFocusObjects.find(
+      (object) => object.type === 'region'
```
azangru, line 683: **Step 1:** unless I am mistaken, the only place that `region`s are currently coming into our app is from an api response that we are handling [here](https://github.com/Ensembl/ensembl-client/blob/85e253b136dba717d0acd1dd644287dff98c1307/src/shared/state/genome/genomeApiSlice.ts#L41). Let's change "region" to "location" in that piece of code.

**Step 2:** Line 683 above would then change to `object.type === 'location'`

**Step 3:** We probably won't have `region`s in `focusObjectSlice`.

src/content/app/genome-browser/state/focus-object/focusObjectSlice.ts
```diff
@@ -164,8 +164,12 @@ export const fetchFocusObject = createAsyncThunk(
       return;
     }
 
-    if (payload.type === 'location') {
-      const focusLocationObject = buildFocusLocationObject(payload);
+    // TODO: Remove checking for region and overriding the payload type when location is properly supported
+    if (payload.type === 'location' || payload.type === 'region') {
```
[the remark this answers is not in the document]
imransl: It comes from the backend. Didn't think about `transformResponse` that we use in slice to change this. Thanks for pointing that out. Will make the change there.

src/content/app/species/state/general/speciesGeneralHelper.ts
```diff
@@ -678,31 +677,30 @@ const getExampleLinks = (props: {
         }
       };
     }
+  } else if (section === SpeciesStatsSection.ASSEMBLY) {
+    // TODO: Change the type check to location when data type returned switches to location
+    const regionExample = exampleFocusObjects.find(
```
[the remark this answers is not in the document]
imransl: ✅

src/content/app/species/state/general/speciesGeneralHelper.ts
imransl: ✅

src/content/app/genome-browser/state/focus-object/focusObjectSlice.ts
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/components/interstitial/BrowserInterstitial.tsx
imransl: They are almost the same. Removed the code for `locationLinks`.

✅

azangru (approved):

PR #908, opened 2022-12-25
Enable programmatic reset for the SimpleSelect component

azangru:
## Description
This PR adds a mechanism for resetting the `SimpleSelect` component to its initial state (either to show the placeholder text, which is the most likely use case, or to show a preselected default option).

This may come handy in https://github.com/Ensembl/ensembl-client/pull/894 if inputs in the navigation modal need to be cleared after the user has entered and confirmed the coordinates; but is a useful behaviour for the `SimpleSelect` component regardless of whether it is required for https://github.com/Ensembl/ensembl-client/pull/894.

Run the Storybook locally to check the added behaviour.

### DIFF
--- a/src/shared/components/simple-select/SimpleSelect.test.tsx
+++ b/src/shared/components/simple-select/SimpleSelect.test.tsx
@@ -1,0 +1,101 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useRef, useEffect } from 'react';
+import { render } from '@testing-library/react';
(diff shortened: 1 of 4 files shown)


src/shared/components/simple-select/SimpleSelect.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+];
+
+describe('SimpleSelect', () => {
+  it('renders a list of options', () => {
+    const { container } = render(<SimpleSelect options={options} />);
+    const simpleSelectElement = container.firstElementChild as HTMLElement;
+
+    // SimpleSelect renders a div element (for styling purposes)
+    expect(simpleSelectElement.tagName.toLowerCase()).toBe('div');
```
imransl, line 44: Wouldn't checking if the `select` class exists be more useful here?

azangru: Added checks for the `select` class and for a CSS class passed from the parent. I think there's still value in checking that the top-level element rendered by this component is not a `select`.

imransl (approved):

azangru:
I'm also changing the behaviour of the placeholder in this PR. Before, we hid the placeholder from the dropdown with the options, which made the first option appeared selected (notice the tick next to it), although in fact it was not (the user had to click on the first option to select it, although it looks as if it were selected):

https://user-images.githubusercontent.com/6834224/210380693-65c03a07-47b7-486d-920f-bb920d6ff208.mp4

Other design systems that use the native browser select show the placeholder in the list of options, but make it unselectable (disabled). I am doing the same in this PR:

https://user-images.githubusercontent.com/6834224/210381127-c4467a39-339d-474f-a00b-288e5624026e.mp4

Asked Andrea; she approved.

PR #900, opened 2022-12-14
Add validation of focus location

azangru:
## Description
- Checks the validity of the focus location before showing the genome browser image
- Checks the validity of the `location` url parameter before showing the genome browser image

**Examples to try while testing:**
- Use a chromosome name that doesn't exist for this species
- Make the `start` parameter greater than the `end` parameter
- Omit a colon or a dash in the location

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-314

## Deployment URL(s)
http://validate-url-location.review.ensembl.org

PR #913, opened 2023-01-10
Validate genome browser page url (alternative PR)

azangru:
An alternative to PR https://github.com/Ensembl/ensembl-client/pull/900

## Description
This PR checks the validity of the focus object parameter, and of the location parameter in the genome browser page url before showing the genome browser image.

In contrast to https://github.com/Ensembl/ensembl-client/pull/900, this PR achieves this functionality by expanding `GenomeBrowserIdsContext` and encapsulating the validation logic inside of custom hooks.

## What to check while reviewing
Stress-test all three variable parts of the genome browser url. Remember that so far we can have focus genes and focus locations. Remember, too, that we aren't allowing commas in location coordinates in the url

image

image

**Examples to try while testing:**
- Use a chromosome name that doesn't exist for this species
- Make the start parameter greater than the end parameter
- Format the location incorrectly (e.g. omit the colon or the dash)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-314

## Deployment URL(s)
http://url-validation-context.review.ensembl.org


jyothishnt:
If we add commas in the location it gets trimmed off and loads the genome browser. But the value that gets into the previously viewed is the raw input from the user. Check this one below
Screenshot 2023-01-12 at 11 29 48

jyothishnt:
Also I see the new hooks inside the context directory. Wonder if it should go inside `hooks` dir?

azangru:
> Also I see the new hooks inside the context directory. Wonder if it should go inside hooks dir?

No, these two are only going to be used by by the `GenomeBrowserIdsContext`. I'll move `GenomeBrowserIdsContext` inside the folder that now contains the two new hooks.

jyothishnt:
Thanks. Still something is not right. I can show you tomorrow.

jyothishnt (approved):

PR #914, opened 2023-01-12
Enable genome browser navigation buttons

azangru:
## Description
Enable genome browser navigation buttons in the region navigation panel; and use the information available to the browser chrome to tell the genome browser where to go.

Notice that the proposed component keeps its own internal state of the genome browser position in order to enable instantaneous feedback to button clicks (no need to rely on genome browser messages, which may come at a slower rate than user clicks). This internal state gets re-synced with messages coming from the genome browser after the user stops clicking the buttons.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1792

## Deployment URL(s)
http://wire-up-gb-nav-buttons.review.ensembl.org


src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/GenomeBrowserNavigationButtons.tsx
```diff
@@ -0,0 +1,199 @@
...
+      browserLocationRef.current = {
+        regionLength: regionData?.region.length ?? Infinity,
+        regionName,
+        start,
+        end
+      };
+    }
+  }, [activeGenomeId, browserLocation, regionData?.region.length]);
+
+  // TODO: add query for the region so as to never exceed the region length
```
imransl, line 88: Is this for the future?

azangru: No, that's done. Odd; I was sure I had deleted those comments. At least I removed them now ✅

imransl (approved):

jyothishnt:
I am getting lots of errors from GB when zoom in/out and panning. 
Screenshot 2023-02-09 at 11 32 55

azangru:
> I am getting lots of errors from GB when zoom in/out and panning.

That's something to do with the genome browser backend not being up to date. Not related to navigation. You can try in on your local machine.

PR #916, opened 2023-01-13
update track settings with new design

jyothishnt:
## Description
Updates to track setting panel with new design

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1798

## Deployment URL(s)
http://new-track-settings-panel.review.ensembl.org

## Views affected
GB

## Possible Issues
In the code we have 2 types of tracks - Gene/FocusGene and Regular
Feature labels are not available for any gene tracks at the moment, so I haven't changed the current logic in the code. But at some point we will have to update the logic to include other tracks.

We could use the `selectedTrackSettings.settings` to see if there is setting available for a track. But somehow the logic is tied behind the tracktype Gene and Regular. Hard to explain in words, easy to show and tell. I will do that in case needed.


src/content/app/genome-browser/components/track-settings-panel/TrackSettingsPanel.scss
```diff
@@ -2,7 +2,7 @@
 
 .trackSettingsPanel {
   display: grid;
-  grid-template-columns: 140px 250px 1fr;
+  grid-template-columns: max-content max-content 1fr;
```
azangru, line 5: `grid-template-columns: repeat(3, max-content);`, I think. Although that won't work with Andrea's 4-column variation panel design.

Do we even need a grid here? Can we do flex (if we need vertical alignment) or just plain old display: block containing display: inline-block (if we don't)?

src/content/app/genome-browser/components/track-settings-panel/TrackSettingsPanel.scss
```diff
@@ -22,12 +22,17 @@
...
+      padding-right: 10px;
     }
 
     &:not(:last-child) {
       border-right: 1px solid $grey;
     }
+
+    &:last-child {
+      padding: 0 25px;
+      min-height: 62px;
```
azangru, line 34: 1) what's the purpose of the `min-height`?
2) should the padding be the property of the `moveTracksIcon` class?

src/content/app/genome-browser/components/track-settings-panel/TrackSettingsPanel.scss
```diff
@@ -52,3 +57,11 @@
     vertical-align: middle;
   }
 }
+
+.moveTracksIcon {
```
azangru, line 61: `.moveTracksIcon` -> `reorderTrack`?

- we are dealing here with a single track, not multiple tracks
- the purpose of this element will be to change track order

src/content/app/genome-browser/components/track-settings-panel/TrackSettingsPanel.scss
```diff
@@ -52,3 +57,11 @@
...
+  display: flex;
+  justify-content: center;
+  width: 30px;
+  height: auto;
```
azangru, line 65: ❓

src/content/app/genome-browser/components/track-settings-panel/track-settings-views/GeneTrackSettings.tsx
```diff
@@ -20,14 +20,12 @@ import { useAppSelector } from 'src/store';
 import useTrackSettings from '../useTrackSettings';
 
 import SlideToggle from 'src/shared/components/slide-toggle/SlideToggle';
-import GlobalTrackSwitch from './components/global-track-switch/GlobalTrackSwitch';
 
-import {
-  getApplyToAllSettings,
-  getTrackSettingsForTrackId
-} from 'src/content/app/genome-browser/state/track-settings/trackSettingsSelectors';
```
azangru, line 28: If we are getting rid of "this track/all tracks" distinction, we should hunt down the code responsible for that and delete it.

Example:

image

src/content/app/genome-browser/components/track-settings-panel/TrackSettingsPanel.scss
jyothishnt: repeat sounds good. Can't we just use repeat(4, max-content)

azangru: I think flex for this case will be superior. You aren't creating any grid here.

jyothishnt: Yes I can replace it with flex.
Although with repeat max-content, we haven't added any gutter space so it is still fine. But yes we are not creating a grid layout here.

azangru (approved):

PR #918, opened 2023-01-17
Disable proxying to api servers in Ensembl deployments

azangru:
## Description
Kamal has been asking us not to enable in kubernetes deployments request handlers that we would normally use when developing locally, to proxy requests to appropriate backends. Although ingress rules set in kubernetes mean that requests intended for other services will never reach our rendering server, it is safer to know that if something goes wrong and such a request does get to the rendering server, the response will be a 404 error rather than something that our staging server would respond with. Failing early helps with debugging.

This PR makes use of the fact that all our kubernetes deployments will have an environment variable called "ENVIRONMENT" set (we use it to distinguish between production, staging, internal, and review deployments).
Knowing this, we can assume that if this this environment variable is not set, then the code is running in locally, and will need to proxy api requests and requests for static assets to relevant backends.

### Result

**Before,** the log from a kubernetes pod with the rendering service showed that we instantiated proxy handlers for `/api` and `/static` routes:

image

**Now,** the log from the pod shows that the rendering service does not initialize any proxies:

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1517

## Deployment URL(s)
http://no-proxy-in-deployment.review.ensembl.org

## Tested
- local development mode (`npm start`) — ✅ 
- local production build (`npm run build`, then `node dist/server/server.js`) — ✅

### DIFF
--- a/src/server/helpers/getConfigForServer.ts
+++ b/src/server/helpers/getConfigForServer.ts
@@ -1,0 +1,30 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// To distinguish between Ensembl deployments (production, staging, internal, development),
+// we use an environment variable called "ENVIRONMENT". When this variable is not set,
+// we can assume that the code is running in local development mode, where the locally running server
+// will need to proxy api requests and requests for static assets to relevant backends.
+const isEnsemblDeployment = 'ENVIRONMENT' in process.env;
+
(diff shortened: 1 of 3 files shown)


kamaldodiya:
Looks good. Other way to get away with proxy middleware, setting up environment variable and getting as close to production as it can be is to use telepresence :). 
With telepresence 
1) Create branch review app
2) Intercept the FE service to route traffic
3) Do development as you would

The only thing this would need is telepresence installed on laptop, knowing few commands. Would be easy to integrate complex workflows as well. Most importantly, everyone needs to adopt it and involves bit learning curve. 
Here is the link to my presentation [Improving-Dev-Workflow.pptx](https://www.ebi.ac.uk/seqdb/confluence/download/attachments/97124916/Improving-Dev-Workflow.pptx?version=1&modificationDate=1669370036939&api=v2)

As it stands looks good to me

kamaldodiya (approved):
Looks good. Other way to get away with proxy middleware, setting up environment variable and getting as close to production as it can be is to use telepresence :). 
With telepresence 
1) Create branch review app
2) Intercept the FE service to route traffic
3) Do development as you would

The only thing this would need is telepresence installed on laptop, knowing few commands. Would be easy to integrate complex workflows as well. Most importantly, everyone needs to adopt it and involves bit learning curve. 
Here is the link to my presentation [Improving-Dev-Workflow.pptx](https://www.ebi.ac.uk/seqdb/confluence/download/attachments/97124916/Improving-Dev-Workflow.pptx?version=1&modificationDate=1669370036939&api=v2) If you would like to explore it further.

As it stands looking at the k8s log looks good to me.

PR #919, opened 2023-01-19
Add sorting of BLAST results table by genomic location

azangru:
## Description
This PR includes several modifications of `DataTable` in order to support customised sorting of rows and rendering of cells. Specifically:
- The type of `TableCellData` is changed to include not only React nodes, but also arbitrary structured data
- A custom renderer can be provided to a column for displaying the structured data on screen
- Another renderer can be provided to a column for transforming the structured data into strings to include in an exported file
- A custom sorting function can be provided for a column when its content is cells with structured data. In absence of such sorting function, the default sorter will assume that the data in table cells is either strings or numbers, and sort the rows accordingly.

### Out of scope
This PR does not sort genomic locations by the rank of their regions in the karyotype. Instead, the sorting is done purely based on string comparison. It means that chromosomes 1, 2, and 10 will be arranged as 1, 10, and 2 in ascending order; or 2, 10, 1 in descending order.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1786

## Deployment URL(s)
http://blast-sort-by-gen-loc.review.ensembl.org


ens-st3 (approved):
Confirmed that sorting works as expected, ie just on chromosome, and that location data are in the table export

jyothishnt (approved):
Looks ok to me

PR #920, opened 2023-01-23
Update genome browser to version 0.6.0

azangru:
## Description
Version 0.6.0 brings new genome browser functionality, as well as breaking changes: 

- Zmenus have been generalised to "hotspots". Other "hotspots" include lozenges and the left corner of a track
- Added support for lozenge clicks for the focus track.
- Genome browser reports when user hovers over the left corner of a track; in response to which we are showing a tooltip
- We can now ask genome browser to put a grey bar to the left of a track, thus providing a way to mark them

### Important (breaking changes)
Notice that the focus gene track now behaves differently from other gene tracks. Specifically:
- _(breaking change)_ Previously, genome browser decided on its own how many transcripts to show in the focus track, based on whether the "several transcripts" switch was on or off. It doesn't do this any longer, and requires browser chrome to explicitly send it the transcripts to be displayed. Meanwhile, for regular gene tracks, genome browser makes the decision on its own.
- _(breaking change)_ Genome browser no longer changes the number of transcripts on the screen in response to the "several tracks" switch for the focus track. It does so, however, for the regular gene tracks.
- For regular gene tracks, genome browser changes the number of visible transcripts when the user clicks on a lozenge. For a focus track, genome browser does not to that, but instead reports to the browser chrome that a lozenge has been clicked, and expects the chrome to tell it how many transcripts to display.

## Related PRs
https://github.com/Ensembl/ensembl-genome-browser/pull/21

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1823

## Deployment URL(s)
http://gb-0-6-0.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.scss
+++ b/src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.scss
@@ -1,0 +1,29 @@
+.anchor {
+  position: absolute;
+  width: 0;
+  height: 0;
+}
+
+.message {
+  max-width: 250px;
+}
+
+.message ul {
+  padding-left: 2rem;
+}
+
+.message li {
+  padding-left: 1rem;
+}
+
+.message li:nth-child(1) {
+  list-style-type: 'G';
+}
+
+.message li:nth-child(2) {
+  list-style-type: 'V';
+}
+
+.message li:nth-child(3) {
+  list-style-type: 'R';
+}
--- a/src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
+++ b/src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
@@ -15,19 +15,19 @@
  */
 
 import React, { useEffect, memo } from 'react';
+import {
+  type HotspotAction,
+  type HotspotPayload,
+  type ZmenuCreatePayload,
+  IncomingActionType
+} from '@ensembl/ensembl-genome-browser';
 
 import useGenomeBrowser from 'src/content/app/genome-browser/hooks/useGenomeBrowser';
-
-import {
-  ZmenuCreateAction,
-  IncomingActionType,
-  ZmenuCreatePayload
-} from '@ensembl/ensembl-genome-browser';
 
 import Zmenu from './Zmenu';
 
 type Props = {
-  browserRef: React.RefObject<HTMLDivElement>;
+  containerRef: React.RefObject<HTMLDivElement>;
 };
 
 // when a zmenu is created, it’s assigned an id,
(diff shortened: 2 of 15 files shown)


src/server/middleware/proxyMiddleware.ts
```diff
@@ -53,15 +53,25 @@ const proxyMiddleware = [apiProxyMiddleware, docsProxyMiddleware];
...
+      changeOrigin: true,
+      secure: false
+    }
+  );
+
+  const browserProxyMiddleware = createHttpProxyMiddleware('/api/browser/**', {
+    target: 'http://52.56.215.72:3333',
+    pathRewrite: {
+      '^/api/browser': '/api' // rewrite path
+    },
```
azangru, line 69: **NOTE TO SELF:** remember to undo before merge

src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.scss
```diff
@@ -0,0 +1,33 @@
+.anchor {
+  position: absolute;
+}
+
+.actualAnchor {
+  position: relative;
+  width: 0;
+  height: 0;
+}
```
imransl, line 9: Is this styling used anywhere?

src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
```diff
@@ -71,12 +70,20 @@ const ZmenuController = (props: Props) => {
...
   ));
 
   return <>{zmenuElements}</>;
 };
 
+const isZmenuPayload = (
+  payload: HotspotPayload
+): payload is ZmenuCreatePayload => {
+  // NOTE: check for payload.variety.length should be unnecessary;
+  // but as of now, genome brwoser can't separate track's left corner hotspot from transcript zmenu hotspot
```
imransl, line 85: `browser` is misspelt in the comment above

src/content/app/genome-browser/hooks/useFocusTrack.ts
```diff
@@ -77,22 +85,58 @@ const useFocusGene = (params: Params) => {
   const focusObjectIdRef = useRef(focusObjectId);
   const geneIdRef = useRef(geneStableId);
   const visibleTranscriptIds = focusGene?.visibleTranscriptIds ?? null;
-  const trackSettingsForGenome = useAppSelector(getAllTrackSettings)
+  const focusGeneTrackSettings = useAppSelector(getAllTrackSettings)
     ?.settingsForIndividualTracks.focus as FocusGeneTrack | undefined;
 
+  const allSortedFocusGeneTranscriptsRef = useRef<string[]>([]);
+  const visibleTranscriptIdsRef = useRef(visibleTranscriptIds);
+  const showSeveralTranscriptsRef = useRef(
```
imransl, line 93: Curious why refs are used here in favour of state

azangru: - Setting a state causes a rerender
- All refs in this hook are intended to overcome stale closures (e.g. you subscribe to genome browser events early in the lifetime of the component; but you have to use up-to-date variables inside of the subscriber functions). Refs, being mutable, help with that. State, being immutable, doesn't.

imransl: That makes sense. Thanks.

src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.scss
azangru: Ah, good point. That was me struggling with the tooltip for the left corner of a track. Deleted now.

src/content/app/genome-browser/components/zmenu/ZmenuController.tsx
azangru: Updated.

But another **NOTE TO MYSELF:** this comment should be unnecessary after Dan fixes the hotspots code (and the line of code below the comment should be updated accordingly).

imransl (approved):

PR #929, opened 2023-02-09
Fix BLAST results table for cDNA

azangru:
### Bug reproduction steps
- Create a new BLAST submission agains a database of cDNAs or proteins
- Wait until jobs in this submission are completed, and open the results page
- Open the table with results of a single job
- Bug: the page explodes

### Cause
The bug must have been introduced in https://github.com/Ensembl/ensembl-client/pull/919, when we added table sorting by genomic location. This necessitated changing the shape of the data passed in the genomic location column. The same column is used for showing transcript ids and protein ids; but it has not been appropriately updated to handle these two cases.

This PR makes a minimum amount of changes needed to fix the bug. It isn't very clean or pretty.

## Deployment URL(s)
http://fix-blast-results-cdna-protein.review.ensembl.org


imransl:
The code looks okay but unfortunately can't test the review site since GitLab is down

imransl (approved):

jyothishnt (approved):

PR #938, opened 2023-02-23
Fix warning from incorrect SimpleSelect usage in tests of DataTable component

imransl:
## Description
This happened due to using `defaultValue` to specify the placeholder in `SimpleSelect` as React doesn't like this. This PR tries to come with a workaround for this by using the `value` to either specify the selected value or the placeholder (if no option is selected).

## Related JIRA Issue(s)
[ENSWBSITES-1828](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1828)

## Deployment URL(s)
http://fix-simple-select.review.ensembl.org

## Views affected
BLAST settings
Genome browser -> Navigation panel -> Go to new location

### DIFF
--- a/src/shared/components/simple-select/SimpleSelect.tsx
+++ b/src/shared/components/simple-select/SimpleSelect.tsx
@@ -74,7 +74,7 @@ const SimpleSelect = (
   props: SimpleSelectProps,
   ref: ForwardedRef<{ clear: () => void }>
 ) => {
-  const { className, placeholder, value, ...otherProps } = props;
+  const { className, placeholder, ...otherProps } = props;
   const selectRef = useRef<HTMLSelectElement | null>(null);
 
   useImperativeHandle(ref, () => ({
@@ -99,14 +99,6 @@ const SimpleSelect = (
   );
 
   let selectChildren: ReactNode;
-
-  const getSelectValue = () => {
-    if (value) {
-      return value;
-    }
-
-    return placeholder ? '' : undefined;
-  };
 
   if ('optionGroups' in otherProps) {
     selectChildren = otherProps.optionGroups.map(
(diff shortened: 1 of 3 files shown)


imransl:
There was another error that popped up due to `value` being specified without `onChange` in `SimpleSelect`. The fix for this seems a bit hacky, but couldn't think of anything better.

Screenshot 2023-02-23 at 20 57 49

src/shared/components/simple-select/SimpleSelect.tsx
```diff
@@ -127,7 +154,8 @@ const SimpleSelect = (
       <select
         ref={selectRef}
         className={styles.selectResetDefaults}
-        defaultValue={placeholder ? '' : undefined}
+        value={getSelectValue()}
+        onChange={selectProps.onChange ?? noop}
```
azangru, line 158: Why do you need this line (as well as the noop)?

src/shared/components/simple-select/SimpleSelect.tsx
```diff
@@ -40,6 +41,7 @@ type HTMLSelectProps = SelectHTMLAttributes<HTMLSelectElement>;
 export type Option = {
   value: string;
   label: string;
+  isSelected?: boolean;
```
azangru, line 44: Why is this needed? This adds confusion to the component's api, by introducing competition between `option.isSelected` and `props.value` (and by providing to limit to how many options can have the `isSelected` flag on them).

src/shared/components/simple-select/SimpleSelect.tsx
imransl: I've mentioned the reason in the above comment

azangru (approved):

PR #939, opened 2023-02-27
Default track panel tab

azangru:
## Description
When the focus object is a variant, switch to the variation tab of the track panel. When the focus object is any other type, switch to the genomic tab.

## Related JIRA Issue(s)
Part of https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1831

## Deployment URL(s)
http://default-track-panel-tab.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/state/track-panel/trackPanelSlice.ts
+++ b/src/content/app/genome-browser/state/track-panel/trackPanelSlice.ts
@@ -16,12 +16,14 @@
 
 import {
   createSlice,
+  createAsyncThunk,
   type PayloadAction,
   type ThunkAction,
   type Action
 } from '@reduxjs/toolkit';
 
 import browserStorageService from 'src/content/app/genome-browser/services/browserStorageService';
+import { parseFocusObjectId } from 'src/shared/helpers/focusObjectHelpers';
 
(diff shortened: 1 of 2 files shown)


src/content/app/genome-browser/state/track-panel/trackPanelSlice.ts
```diff
@@ -96,6 +103,25 @@ export const selectTrackPanelTab =
...
+export const updateTrackPanelTabForNewFocusObject = createAsyncThunk(
+  'track-panel-slice/updateTrackPanelTabWhenFocusObjectChanges',
+  (params: { genomeId: string; focusObjectId: string }) => {
+    const { genomeId, focusObjectId } = params;
+    const { type } = parseFocusObjectId(focusObjectId);
+    let trackPanelTab;
+    if (type === 'variant') {
+      trackPanelTab = TrackSet.VARIATION;
+    } else {
+      trackPanelTab = TrackSet.GENOMIC;
```
imransl, line 115: Shouldn't there be a check for genomic?

azangru: What we are doing here is one behaviour if focus object is a variant, and a different behaviour is focus object is anything else. No need to be more specific about focus object types.

imransl (approved):

PR #941, opened 2023-02-27
Change BLAST results download file from CSV to TSV

imransl:
## Description
Replace CSV with TSV since we recently decided that BLAST results files need to be in the latter format. 

## Related JIRA Issue(s)
[ENSWBSITES-1865](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1865)

## Deployment URL(s)
http://change-blast-download-format.review.ensembl.org

## Views affected
BLAST -> Unviewed jobs
BLAST -> Jobs list


src/shared/components/data-table/components/table-controls/components/table-actions/components/download-data/DownloadData.tsx
```diff
@@ -137,7 +137,7 @@ const DownloadData = () => {
   );
 };
 
-const formatCSV = (table: (string | number)[][]) => {
+const formatTSV = (table: (string | number)[][]) => {
   return table.map((row) => row.join(',')).join('\n');
```
azangru, line 141: see the separator

jyothishnt: After all the effort of renaming, it still gives back a CSV ;-) 
```suggestion
  return table.map((row) => row.join('\t')).join('\n');
```

imransl: Ah I missed this one. I made the same change here as well:
https://github.com/Ensembl/ensembl-client/pull/941/files#diff-e49005527206e2193fbbe69d64f07708ab87d2f9323a8596cbe1550b166e4867R193

Anyway why are we doing the same thing (function) in two separate files? Can't we do it in just one place?

azangru (approved):

PR #942, opened 2023-03-01
Focus variant track settings

azangru:
## Description
This PR connects the focus track settings panel both to redux and to IndexedDB.
- If you toggle track settings, close the settings panel, and open it again, you should see the toggles turned correctly on or off
- If you refresh the browser window, the settings should be preserved and applied to the focus variant track when it loads.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1876

## Deployment URL(s)
http://focus-variant-track-settings.review.ensembl.org/genome-browser/grch38?focus=variant:rs779377781

### DIFF
--- a/src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
+++ b/src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { createSelector } from '@reduxjs/toolkit';
+
 import { getBrowserActiveGenomeId } from '../browser-general/browserGeneralSelectors';
 
 import type {
@@ -56,3 +58,21 @@ export const getAllTrackSettingsForGenome = (
 ): TrackSettingsForGenome | null => {
   return state.browser.trackSettings[genomeId] ?? null;
 };
+
+export const getAllNonFocusTrackSettingsForGenome = createSelector(
+  getAllTrackSettingsForGenome,
+  (trackSettingsForGenome) => {
+    if (!trackSettingsForGenome) {
+      return null;
+    }
+    const settings: Record<string, TrackSettings> = {};
+    for (const [trackId, trackSettings] of Object.entries(
+      trackSettingsForGenome.settingsForIndividualTracks
+    )) {
+      if (trackId !== 'focus' && trackId !== 'focus-variant') {
+        settings[trackId] = trackSettings;
+      }
+    }
+    return settings;
+  }
+);
--- a/src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts
+++ b/src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts
@@ -36,7 +36,18 @@ export type GeneTrackSettings = {
   isVisible: boolean;
 };
 
+export type VariantTrackSettings = {
+  'label-snv-id': boolean;
+  'label-snv-alleles': boolean;
+  'label-other-id': boolean;
+  'label-other-alleles': boolean;
+  'show-extents': boolean;
+  name: boolean;
+  isVisible: boolean;
+};
+
 export type FocusGeneTrackSettings = Omit<GeneTrackSettings, 'isVisible'>;
+export type FocusVariantTrackSettings = Omit<VariantTrackSettings, 'isVisible'>;
 
 export type RegularTrackSettings = {
   showTrackName: boolean;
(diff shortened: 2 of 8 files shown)


imransl (approved):

src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts
```diff
@@ -247,6 +271,20 @@ const trackSettingsSlice = createSlice({
...
+      state,
+      action: PayloadAction<{
+        genomeId: string;
+        trackId: string;
+        trackSettings: TrackSettings;
+      }>
+    ) {
+      const { genomeId, trackId, trackSettings } = action.payload;
+      if (!state[genomeId]) {
+        state[genomeId] = { settingsForIndividualTracks: {} };
```
jyothishnt, line 284: Why do you need a nesting with `settingsForIndividualTracks` Can't we have all settings inside `state.browser.trackSettings`?

src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
```diff
@@ -56,3 +58,21 @@ export const getAllTrackSettingsForGenome = (
...
+  getAllTrackSettingsForGenome,
+  (trackSettingsForGenome) => {
+    if (!trackSettingsForGenome) {
+      return null;
+    }
+    const settings: Record<string, TrackSettings> = {};
+    for (const [trackId, trackSettings] of Object.entries(
+      trackSettingsForGenome.settingsForIndividualTracks
+    )) {
+      if (trackId !== 'focus' && trackId !== 'focus-variant') {
```
jyothishnt, line 72: can we do a match for `focus` to find all non-focus tracks?

src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts
azangru: I was asking myself the same question, trying to remember why we had it. I think this may be a remnant from the time when we had the all tracks/this track toggle, which we stored in redux, and which referred to all tracks in a genome. We aren't doing so now; so if we are feeling brave, we can remove this key. But in a subsequent PR, not in this one.

src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
azangru: you mean running `/focus/.test(trackId)`?

🤷‍♂️ Do we need to? The risk with the above is that other tracks (which we don't control) will need to be guaranteed to not have a string `f-o-c-u-s` anywhere in their ids.

src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts
jyothishnt: ok. Yeah better remove it if this is unnecessary. Please could you create a ticket? Thanks

src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
jyothishnt: I mean to use regex to match trackId. But yes its not good if there is another track that has matches `focus` but where did `focus-gene` go?

jyothishnt:
`Track name` toggle should turn on/off track names for all tracks. But when you do it via non focus track settings it doesn't turn on track name for focus.

https://user-images.githubusercontent.com/6347854/222422273-498a8c6c-7c96-47bc-ad3e-d5d95bf34709.mov

jyothishnt:
I also got an error below, but couldn't replicate due to the BE api failure. 
Screenshot 2023-03-02 at 11 52 52

src/content/app/genome-browser/state/track-settings/trackSettingsSelectors.ts
azangru: `focus-gene` is `focus` 😄

azangru:
> `Track name` toggle should turn on/off track names for all tracks. But when you do it via non focus track settings it doesn't turn on track name for focus.

We discussed this. I mentioned this at an earlier scrum; but forgot to write this down explicitly in the description of the PR. Older tracks are storing the name flag in the field called `showTrackName`. For the variant track, I decided to call all flags same way Dan calls the switches in the genome browser; which means, it got called `name`. In a later PR, we ought to refactor track settings code to make it more generic. In this PR, I am leaving the `track name` toggle unconnected with the names of all other tracks.

azangru:
> I also got an error below, but couldn't replicate due to the BE api failure.

I can't replicate this either.

jyothishnt:
Jotting down the bugs found
On beta:
1. Go to a [wrong location](https://beta.ensembl.org/genome-browser/iwgsc?focus=gene:TraesCS3D02G273600&location=3D:0-379541788) and click on `Continue` takes you to interstitial page. It should also have cleared this location entry from storage/redux, but it doesn't. Now go to Human and do a soft refresh and now back to Wheat, it will take you back to the above previous wrong location

2. Open browser console, go to human example gene in GB and now go to another species (in GB) that doesn't have variation data, you will see client requesting for variation tracks. Again bug on the client side.

On this review branch:
1. Variation only work for Human. Others fail due to some data bug - needs investigation.

PR #944, opened 2023-03-02
Add zmenu for focus variant track

azangru:
## Description
Show zmenu when clicking on a variant in the variant track (see https://github.com/Ensembl/ensembl-genome-browser/pull/24)

- Update genome browser package so as to include updated type definitions
- Extracted gene-and-one-transcript zmenu and a variant zmenu from the Zmenu component

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1874

## Deployment URL(s)
http://zmenu-variant.review.ensembl.org

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
-        "@ensembl/ensembl-genome-browser": "0.6.3",
+        "@ensembl/ensembl-genome-browser": "0.6.4",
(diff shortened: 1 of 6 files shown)


jyothishnt (approved):
Assuming it is very premitive, the code looks ok to me. I couldn't properly check the UI due to the BE bug. Also I couldn't bring up a proper variant region to check zmenu on Wheat.

imransl (approved):

PR #930, opened 2023-02-13
Implement new blast settings combination for gap-open/close and match/mismatch scores

imransl:
## Description
The default values and behaviour of some BLAST parameters such as gap penalties, match/mismatch and matrix scores have been updated to reflect the BLAST in the current site.

### TODO:
- [ ] Check and update the default values for all sensitivity presets (gap penalties are especially wrong; but don't trust anything else either)
- [x] Remove from sensitivity presets the settings that are irrelevant for a program (gap penalties, gap alignment)
- [x] Remove the `NONE` option for the scoring matrix, and update the tests
- [x] Update the UI to match the design
- [x] Disable the gap align checkbox

## Related JIRA Issue(s)
[ENSWBSITES-1838](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1838)

## Deployment URL(s)
http://blast-default-values.review.ensembl.org

## Views affected
BLAST

PR #946, opened 2023-03-06
Update blast settings parameters

imransl:
## Description
According to the BLAST manual gap align is only accepted for `blastn` program. So sending it for other programs is redundant and also displaying the option to the users (for programs other than `blastn`) is misleading. 

This PR updates to show the option only for `blastn` program. Also, the mock config has been updated to test this locally and BE will need to update it for the change to work in live.

There are also two more tasks that have been incorporated into this PR:

1. Show `Short sequences` sensitivity option only for `blastn` ([ENSWBSITES-1871](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1871))
2. Changing the default gap opening and extension penalties. The default settings for normal sensitivity of each program was updated based on the NCBI BLAST tool defaults: https://blast.ncbi.nlm.nih.gov/Blast.cgi. ([ENSWBSITES-1863](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1863))

## Related JIRA Issue(s)
[ENSWBSITES-1864](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1864) (commit: 34ef7d0)
[ENSWBSITES-1871](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1871) (commit: 8a026f3)
[ENSWBSITES-1863](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1863) (commit: 9040400)

## Deployment URL(s)
http://gap-align-checkbox.review.ensembl.org

## Views affected
BLAST form


azangru:
Looks good; but wouldn't you like to include ENSWBSITES-1863 in this PR as well? It also seems to be related to the updating of the config. Why not do both updates in one go?

tests/fixtures/blast/blastSettingsConfig.json
```diff
@@ -802,6 +802,13 @@
     "tblastn": ["matrix", "gapopen", "gapext"],
     "blastp": ["matrix", "gapopen", "gapext"]
   },
+  "valid_sensitivities_for_program": {
```
azangru, line 805: Why do we need this? Can't this be calculated from the keys of the `presets.settings` object?

imransl: I went by what we did with `valid_parameters_for_program` as it would be easier to access it, though now think it is more or less redundant. But this also means we do the same for ` valid_parameters_for_program` right?

imransl: Made the changes for this ✅

jyothishnt:
I still get Submission failed for some combinations. Try the below combination
Screenshot 2023-03-09 at 16 51 36

azangru (approved):
Looks good. Remember to undo the changes in blastApiSlice before merging.

jyothishnt:
Hi @imransl We found out that the labels and values for the gap penalties dropdown are different. This is the reason blast is failing. Please review the valid combinations again

azangru:
> I still get Submission failed for some combinations. Try the below combination

As a follow-up on this comment by Jyo. Here are the details, with steps to replicate the error:
- For a regular `blastn` job, choose the `Match/mismatch scores` of `5,-4`. The dropdown for gap penalties will change and show `25, 10`. So far, so good.
- Open the browser dev tools, and open the Network panel
- Submit the BLAST job
- Notice that the response from the tools api will contain errors inside of the `jobs` array:

```
{
    "submission_id": "phSCQehOQMVS2pDI5a1X3Q",
    "jobs": [
        {
            "sequence_id": 1,
            "genome_id": "a73351f7-93e7-11ec-a39d-005056b38ce3",
            "error": " Invalid parameters: Match/mismatch scores -> Gap existence and extension values of 5 and 2 not supported for substitution scores 5 and -4. Supported combinations are: [[25, 10], [10, 6], [8, 6], [-1, -1], [25, -1], [-1, 10]].",
            "status": 400
        }
    ]
}
```

- Notice that the error above says "Gap existence and extension values of 5 and 2 ..." If you then check the data that was submitted by the client, you will see that the `gapopen` and  `gapext` values are indeed `5` and `2`, rather than `25` and `10`, as one would expect from the dropdown:

```
{
    "genome_ids": [
        "a73351f7-93e7-11ec-a39d-005056b38ce3"
    ],
    "query_sequences": [
        {
            "id": 1,
            "value": "..."
        }
    ],
    "parameters": {
        "database": "dna_sm",
        "program": "blastn",
        "stype": "dna",
        "alignments": "100",
        "scores": "50",
        "hsps": "100",
        "dropoff": "0",
        "gapalign": "true",
        "filter": "T",
        "compstats": "F",
        "exp": "10",
        "gapopen": "5",
        "gapext": "2",
        "wordsize": "11",
        "match_scores": "5,-4"
    }
}
```

What this means is that the redux state **did not get updated appropriately** when the match/mismatch scores were updated. This is probably a bug that we missed back in https://github.com/Ensembl/ensembl-client/pull/930

imransl:
@jyothishnt @azangru Thanks for spotting this issue. I've fixed this with the latest commit: [9b06b9c](https://github.com/Ensembl/ensembl-client/pull/946/commits/9b06b9c751a722b666a532eded9c6187a7ba40ed).

src/content/app/tools/blast/components/blast-settings/BlastSettings.tsx
```diff
@@ -189,8 +185,8 @@ const BlastSettings = ({ config }: Props) => {
           </div>
           <div>
             <BlastSelect
-              options={getPresetsList(config).options}
-              label={getPresetsList(config).label}
+              options={getPresetsList(config, blastProgram).options}
+              label={getPresetsList(config, blastProgram).label}
```
azangru, line 189: let's create a `presetsList` variable before the return, instead of calling the `getPresetsList` function twice

imransl: ✅

azangru (approved):
I am happy with this PR.

Please remember to remove the use of `mockConfig` in `blastApiSlice` before merging.

PR #948, opened 2023-03-07
Fix broken restore defaults in data table

imransl:
## Description
The `Restore defaults` action in the `DataTable` was not properly restoring the hidden columns when selected. This was because the `TableContext` which is used to pass to the reducer for restoring, was being modified when hiding the columns. Changing the code to not to modify fixes the issue.

Steps to reproduce:

1. Run a BLAST job
2. Select the results when it has finished running
3. Click on the genome in a sequence it was run against to open the results table (far right)
4. Click on the `Actions` select field (far left) in the table header
5. Select `Show/hide columns` option and select any fields in the popup panel to hide
6. Now select `Restore defaults` to see if the hidden columns have been restored

## Related JIRA Issue(s)
[ENSWBSITES-1870](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1870)

## Deployment URL(s)
http://fix-data-table.review.ensembl.org

## Views affected
BLAST -> Results view -> Sequence -> Results table

### DIFF
--- a/src/shared/components/data-table/components/table-controls/components/table-actions/TableActions.test.tsx
+++ b/src/shared/components/data-table/components/table-controls/components/table-actions/TableActions.test.tsx
@@ -64,14 +64,14 @@ describe('<TableActions />', () => {
 
   it('respects the disabledActions list', () => {
     container = renderTableActions({
-      disabledActions: [TableAction.FILTERS, TableAction.DEFAULT]
+      disabledActions: [TableAction.FILTERS, TableAction.FIND_IN_TABLE]
     }).container;
 
     const filterOption = document.querySelector(
       `option[value="${TableAction.FILTERS}"`
     );
     const defaultOption = document.querySelector(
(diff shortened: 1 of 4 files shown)


azangru:
I've just realised there are tons of problems with the Actions section of the data table:

- The word "Actions" appears twice, and in fact "Actions" is a selectable option (this must already be in our main branch)

image

- More related to "restoring defaults", if you click on "Restore defaults" when you have hidden some table rows, the rows selection will be reset, but the actions UI will remain in the "select rows" mode. I doubt that this is correct. Don't know if this is something you want to include in this PR or not:

https://user-images.githubusercontent.com/6834224/223675355-a0028156-c394-4da3-aecc-b554105bddee.mov

imransl:
@azangru I've fixed both the issues in [ca944d4](https://github.com/Ensembl/ensembl-client/pull/948/commits/ca944d4107accdf7a48a42d195973b90c8842f61)

PR #949, opened 2023-03-08
Use an element inside the body as a container of the React app

azangru:
## Description
This is a follow-up on https://github.com/Ensembl/ensembl-client/pull/878, changing the strategy again.

The purpose of https://github.com/Ensembl/ensembl-client/pull/878 was to use React's new api for server-side rendering, and also to remove dependency on a couple of third-party libraries, such as react-helmet-async, and loadable-components.

One thing https://github.com/Ensembl/ensembl-client/pull/878 did was to assign the whole `document` element as the React root, making React responsible for the `html` element and everything inside. This is actually recommended in the React docs, which [say](https://beta.reactjs.org/reference/react-dom/server/renderToPipeableStream#rendering-a-react-tree-as-html-to-a-nodejs-stream): "Your root component should return the entire document including the root <html> tag".

However, it turned out that React there can be too much adversarial activity on a web page which can confuse React when it attempts to hydrate the server-rendered html. One common offender was React Dev Tools itself. Here is an example when React Dev Tools mess up a newly opened browser tab in development — notice that React does not recover from the hydration error, and we need to reload the page to get rid of the error:
 
https://user-images.githubusercontent.com/6834224/223750219-13f454af-ce6a-48a1-bde9-9bfd72423d78.mp4

and here is an example of the same error in a production build — notice that React seems to be able to recover from it; at least the page doesn't look broken

https://user-images.githubusercontent.com/6834224/223752567-aaf08513-6323-48c8-9014-e3571a84c131.mp4

The cause of the errors above was a script tag inserted by React Dev Tools chrome extension. Notice on the screenshot below how the chrome extension has injected a script tag immediately above the `head` tag, and right below the `body` tag:

image

I've informed the React team of this problem (see `https://github.com/facebook/react/issues/25924`); and it looks likely that they will fix this problem with React Dev Tools (see their PR: `https://github.com/facebook/react/pull/26233`); but there's no way to protect us against third-party scripts that we add to the page (sigh), or against arbitrary Chrome extensions that users might have in their browsers.

### Proposed solution
The solution explored in this ticket is to move back from giving React the whole html to control to giving it just a single DOM element inside the body. This way, we should not expect third-party agents to interfere with the DOM belonging to React.

#### Concerns
Why do the docs say that "Your root component **should** return the entire document including the root <html> tag". Is something going to break about React if the root component is not the entire document? I can't imagine that it would; but there's no way to be sure.

### Checklist
- Minimal html rendered on the server and hydrated on the client without errors
- Favicons work
- Page metadata is updated appropriately when navigating between pages (both server-side and client-side)
- Third-party scripts enabled when required

### TODO
- Confirm the piping is used correctly

## Related JIRA Issue(s)
None

## Deployment URL(s)
http://react-ssr-div.review.ensembl.org


azangru:
NOTE: see the long discussion in the React repo about hydration problems with various Chrome extensions; workarounds by different teams, and React team's position on hydrating anything other than the whole `document` (they don't like the idea):

`https://github.com/facebook/react/issues/24430`

Also, see [this repo](https://github.com/Xiphe/remix-island), which suggests that a fix on React side is hopefully expected in React 18.3

azangru:
Closing this PR because of the above comment.

PR #951, opened 2023-03-14
Use hotspot area coordinates to position tooltip

jyothishnt:
## Description
Use the "hotspot-area" block from the payload to position the tooltip, rather than the "x" and "y" fields of the payload.

There is a change needed in ensembl-genome-browser repo which is in [this PR](https://github.com/Ensembl/ensembl-genome-browser/pull/25)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1882

## Deployment URL(s)
http://track-tooltip-position.review.ensembl.org

## Views affected
GB

### DIFF
--- a/src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.tsx
+++ b/src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.tsx
@@ -59,8 +59,13 @@ const BrowserTrackLegend = (props: Props) => {
     if (!isTrackLegendHotspot(payload)) {
       return;
     }
+
+    const { bottom, top, right } = payload['hotspot-area'];
+    const toolTipX = right;
+    const toolTipY = top + (bottom - top) / 2 + 4;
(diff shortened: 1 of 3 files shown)


src/content/app/genome-browser/components/browser-track-legend/BrowserTrackLegend.tsx
```diff
@@ -59,8 +59,13 @@ const BrowserTrackLegend = (props: Props) => {
     if (!isTrackLegendHotspot(payload)) {
       return;
     }
+
+    const { bottom, top, right } = payload['hotspot-area'];
+    const toolTipX = right;
```
azangru, line 64: Isn't it a bit too far? Here's what it looks like:

image

image

Compare with Andrea's directions in the ticket:

image

image

azangru (approved):

imransl (approved):

PR #952, opened 2023-03-14
add spinner to drawer while fetching data

jyothishnt:
## Description
Add a spinner to the drawer while fetching data.
Added to gene and transcripts drawer as they are ones that fetches data as of now.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1842

## Deployment URL(s)
http://drawer-spinner.review.ensembl.org

## Views affected
GB

### DIFF
--- a/src/content/app/genome-browser/components/drawer/Drawer.tsx
+++ b/src/content/app/genome-browser/components/drawer/Drawer.tsx
@@ -22,17 +22,10 @@ import DrawerBookmarks from './drawer-views/DrawerBookmarks';
 import GeneSummary from './drawer-views/gene-summary/GeneSummary';
 import TranscriptSummary from './drawer-views/transcript-summary/TranscriptSummary';
 import VariantGroupLegend from './drawer-views/variant-group-legend/VariantGroupLegend';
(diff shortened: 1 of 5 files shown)


src/content/app/genome-browser/components/drawer/Drawer.tsx
```diff
@@ -22,11 +22,18 @@ import DrawerBookmarks from './drawer-views/DrawerBookmarks';
...
 
 import { getActiveDrawerView } from 'src/content/app/genome-browser/state/drawer/drawerSelectors';
 
 import styles from './Drawer.scss';
 
+export const Spinner = () => (
+  <div className={styles.spinnerContainer}>
+    <CircleLoader />
+  </div>
+);
```
azangru, line 35: I am a bit worried that `Drawer` imports components (`GeneSummary`, `TranscriptSummary`), which now depend on something that `Drawer` exports.

Is this a circular dependency? Maybe not quite; but it feels safer to put the spinner in its own `DrawerSpinner` file, next to `Drawer.tsx`

azangru (approved):

imransl (approved):

PR #507, opened 2021-06-03
Genome browser service

msmanoj:
TODOs before merging:

- [x] - Remove browserProxy from `src/ensembl/src/server/middleware/proxyMiddleware.ts`

## Type

- [ ] Bug fix
- [x] New feature
- [x] Improvement to existing feature

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1074

## Importance
Required for MAP

## Description

This PR is to migrate the Genome Browser to the [new version](https://github.com/Ensembl/ensembl-dauphin-style-compiler) which is made available through the dedicated [NPM package](https://github.com/Ensembl/ensembl-genome-browser).
Unlike the previous version, the new version does not require any assets to be stored in the codebase of the client app. To use the new version of the genome browser, we just need to add the above-mentioned package to the `package.json` file, which will expose the class `EnsemblGenomeBrowser` along with the required WASM and type definition files. 

## Deployment URL
http://genome-browser-service.review.ensembl.org/

PR #954, opened 2023-03-27
Genome browser page refactoring

azangru:
## Description
This PR changes the way we are consuming the genome browser in the following ways:
- It is proposed that https://github.com/Ensembl/ensembl-genome-browser repository be retired
- The part of the https://github.com/Ensembl/ensembl-genome-browser repository that is responsible for communicating with the genome browser is moved over in ensembl-client repository, thus undoing an earlier PR https://github.com/Ensembl/ensembl-client/pull/507. The reason for doing so is that the `ensembl-genome-browser` repository is too much entangled with `ensembl-client` anyway, and not generic enough to be consumed by any other client
- The part of the https://github.com/Ensembl/ensembl-genome-browser repository that is responsible for packaging and publishing the genome browser as an npm package will be moved to https://github.com/Ensembl/ensembl-dauphin-style-compiler — see https://github.com/Ensembl/ensembl-dauphin-style-compiler/pull/54 

## Significant changes
- Genome browser is imported directly as it is packaged by wasm-pack, rather than via an intermediary wrapper, which was the `ensembl-genome-browser` repo
- Instead of translating names of genome browser's switches into more explicit and conveniently written names (e.g. `showSeveralTranscripts` instead of just `several`; or `showTranscriptIds` instead of `transcript-label`, this PR reverts to using the original names of genome browser's switches. Mapping to clearer names isn't going to scale well.

## Related PRs
https://github.com/Ensembl/ensembl-dauphin-style-compiler/pull/54, which updates the process of packaging and publishing of the the genome browser

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1885

## Deployment URL(s)
http://gb-project-refactor.review.ensembl.org


src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  if (focusType === 'gene') {
+    setFocusGene(payload);
+  } else if (focusType === 'variant') {
+    setFocusVariant(payload);
+  } else if (focusType === 'location') {
+    setFocusLocation(payload);
+  }
+};
+
```
imransl, line 36: Isn't this a candidate for switch just in case more types get added?

src/content/app/genome-browser/services/genome-browser-service/types/genomeBrowserCongig.ts
imransl, line 1: `genomeBrowserCongig.ts` to `genomeBrowserConfig.ts`?

src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
azangru: I dunno; there are all these `break`s in a switch, which you mustn't forget. It's just about as ugly as a bunch of `else if`s.

imransl (approved):

src/content/app/genome-browser/components/track-settings-panel/track-settings-views/GeneTrackSettings.tsx
```diff
@@ -42,37 +42,31 @@ export const GeneTrackSettings = (
...
-    selectedTrackSettings.showSeveralTranscripts;
-  const shouldShowTranscriptIds = selectedTrackSettings.showTranscriptIds;
+  const shouldShowTrackName = selectedTrackSettings.name;
+  const shouldShowFeatureLabels = selectedTrackSettings.label;
+  const shouldShowSeveralTranscripts = selectedTrackSettings.several;
+  const shouldShowTranscriptIds = selectedTrackSettings['transcript-label'];
 
   const handleSeveralTranscriptsToggle = () => {
-    updateShowSeveralTranscripts(!shouldShowSeveralTranscripts);
+    updateTrackSetting('several', !shouldShowSeveralTranscripts);
```
jyothishnt, line 57: I think it would be good if there is a mention of transcripts instead of just `several` like `several-transcripts`

src/content/app/genome-browser/state/browser-nav/browserNavSlice.ts
```diff
@@ -25,10 +24,10 @@ export enum BrowserNavAction {
 }
 
 export const browserNavButtonActionMap = {
-  [BrowserNavAction.MOVE_LEFT]: OutgoingActionType.MOVE_LEFT,
-  [BrowserNavAction.MOVE_RIGHT]: OutgoingActionType.MOVE_RIGHT,
-  [BrowserNavAction.ZOOM_OUT]: OutgoingActionType.ZOOM_OUT,
-  [BrowserNavAction.ZOOM_IN]: OutgoingActionType.ZOOM_IN
+  // [BrowserNavAction.MOVE_LEFT]: OutgoingActionType.MOVE_LEFT,
```
jyothishnt, line 27: Do you still want to keep this? There is no more `OutgoingActionType`

src/content/app/genome-browser/components/track-settings-panel/track-settings-views/GeneTrackSettings.tsx
azangru: I don't disagree; but that should really be the name of the switch on the eard side. Using one set of names in ensembl-client, and a different set of names in eard, will lead to unnecessary confusion and unnecessary code for translating naming convention into another. Meahwhile, I've left [comments in trackSettingsSlice](https://github.com/Ensembl/ensembl-client/blob/f6bbe0321bca92cdefa612283b393ae63d970eb0/src/content/app/genome-browser/state/track-settings/trackSettingsSlice.ts#L32) explaining what these eard keywords mean.

src/content/app/genome-browser/state/browser-nav/browserNavSlice.ts
azangru: Good point. We aren't using `browserNavSlice` at all anymore. I deleted the slice file and the selector file. Should have deleted them back in https://github.com/Ensembl/ensembl-client/pull/914.

jyothishnt (approved):
Thanks

PR #956, opened 2023-03-27
Refactor Input component

imransl:
## Description
The styling of the `Input` component needs to be refactored, as it doesn't seem to be consistent across the site. Andrea created an [XD](https://xd.adobe.com/view/b008980c-aa53-4b5e-99ef-b7eae5553a1b-e55b/screen/e8283692-ac08-4291-b24f-491965a2c8e2) highlighting the styling that needs to be used for it in different parts of the site.

In this PR, the styling has been mostly consolidated into `Input` stylesheet. Also the right corner icon has been moved into it as well.

## Related JIRA Issue(s)
[ENSWBSITES-1822](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1822)

## Deployment URL(s)
http://update-input-styles.review.ensembl.org

## Views affected
- In Genome browser and Entity viewer: interstitial and sidebar search
- Genome browser -> Navigation modal
- Species selector page
- Contact form
- Blast form


src/shared/components/input/FlatInput.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+      inputRef = useRef<HTMLInputElement | null>(null);
+      return <FlatInput ref={inputRef} />;
+    };
+    render(<Wrapper />);
+
+    const inputElement = inputRef?.current;
+    expect(inputElement?.tagName).toBe('INPUT');
+  });
+
+  it('renders the input as disabled if prop passed as true', () => {
```
azangru, line 35: `'renders the input as disabled if prop passed as true'` -> `can be disabled`?

src/shared/components/input/FlatInput.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    let { container } = render(<FlatInput disabled={true} />);
+    expect(container.querySelector('.flatInputWrapper')?.classList).toContain(
+      'flatInputWrapperDisabled'
+    );
```
azangru, line 39: We should check here that the input is actually disabled. `expect(container.querySelector('.input').hasAttribute('disabled')).toBe(true)`, or something.

src/shared/components/input/FlatInput.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(
+      container.querySelector('.flatInputWrapper')?.classList
+    ).not.toContain('flatInputWrapperDisabled');
+  });
+
+  it('renders icon if passed as props', () => {
+    const MockIcon = () => <div className="mockIcon">Mock Icon</div>;
+    let { container } = render(<FlatInput rightCorner={<MockIcon />} />);
+    expect(container.querySelector('.rightCorner')).toBeTruthy();
+    expect(container.querySelector('.mockIcon')).toBeTruthy();
```
azangru, line 51: The icon should be inside of the right corner. Does it make the selector `'.rightCorner .mockIcon'`?

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,60 @@
...
+  &::placeholder {
+    color: $dark-grey;
+    font-weight: $light;
+  }
 
   &:focus,
   &:active {
     outline: none;
+    color: $black;
+    font-weight: $normal;
```
azangru, line 13: Is font-weight necessary? Isn't it the default?

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,60 @@
...
+.rightCorner {
+  display: flex;
+  align-items: center;
+  flex: 0 0 auto;
+}
+
+.inputWrapper {
+  background-color: var(--input-background-colour, $white);
+  display: inline-flex;
+  font-size: $global-font-size;
```
azangru, line 26: Why is this line needed?

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,60 @@
...
 }
 
-.shadedInput {
-  box-shadow: $form-field-shadow;
+.shadedInputWrapper {
   border: none;
-  padding: 0 15px;
-  height: 30px;
+  box-shadow: $form-field-shadow;
+  height: 25px;
```
azangru, line 41: If the `size` property of the shaded input is optional; it is probably worth setting the height via a css variable:

 ```
height: var(--shaded-input-height, 25px); // or whatever is the default; can be overridden from the parent

&Large {
    --shaded-input-height: 40px; // so that it cannot be overridden from the parent
  }

&Small {
    --shaded-input-height: 30px;
  }

```

File: src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx [hunk 7]
```diff
@@ -115,6 +115,7 @@ export const SpeciesSearchField = () => {
         onSelect={onMatchSelected}
         matchGroups={matchGroups}
         searchFieldClassName={styles.speciesSearchField}
+        size="large"
```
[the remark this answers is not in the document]
imransl: I put it in there since it is a shared component and uses the `SearchField` to which the size can be passed on. Will remove it since we don't use it anywhere else as you've mentioned.

src/shared/components/input/FlatInput.test.tsx
imransl: Yes that would be better than checking if `.rightCorner` exists separately

src/shared/components/question-button/QuestionButton.scss
```diff
@@ -17,8 +17,8 @@
 
 .in-input-field {
   padding: 0;
-  height: 14px;
-  width: 14px;
+  height: 16px;
+  width: 16px;
```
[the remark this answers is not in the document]
imransl: Perhaps we should use a CSS variable. We only have `16px` sized icons in use for now.

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.scss
```diff
@@ -3,9 +3,10 @@
 $rightCornerWidth: 40px;
 
 .speciesSearchLabel {
-  color: $dark-grey;
+  color: $black;
   display: block;
-  font-size: 12px;
+  font-size: 13px;
```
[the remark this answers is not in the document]
imransl: ✅

src/shared/components/input/FlatInput.test.tsx
imransl: ✅

src/shared/components/input/FlatInput.test.tsx
imransl: ✅

src/shared/components/input/Input.scss
imransl: Removed these ✅

src/shared/components/input/Input.scss
imransl: Removed it ✅

src/shared/components/input/Input.scss
imransl: ✅

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,60 @@
...
+    height: 40px;
+  }
+
+  &Small {
+    height: 30px;
+  }
 }
+
+.flatInputWrapper {
+  box-shadow: none;
```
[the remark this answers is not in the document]
imransl: Removed it ✅

src/shared/components/question-button/QuestionButton.scss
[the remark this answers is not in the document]
imransl: ✅

src/shared/components/in-app-search/InAppSearch.tsx
```diff
@@ -57,6 +57,7 @@ export type Props = {
   genomeId: string;
   genomeIdForUrl: string; // this should be a temporary measure; it should be returned by search api
   mode: InAppSearchMode;
+  size?: InputSize;
```
azangru, line 60: I don't understand your approach to InAppSearch:

- If it truly exists in only two sizes, then why is the `size` field optional?
- However, does it really need the `size` prop? Isn't the `mode` prop sufficient? Aren't all the interstitial instances of InAppSearch large and all sidebar instances small?

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,56 @@
...
 }
 
-.shadedInput {
-  box-shadow: $form-field-shadow;
+.shadedInputWrapper {
   border: none;
-  padding: 0 15px;
-  height: 30px;
+  box-shadow: $form-field-shadow;
+  height: var(--shaded-input-height, 40px);
```
azangru, line 38: Out of curiosity — which shaded input is going to be more common: 30px high or 40px high (depending on this, we will pick our default). What will Andrea say if we ask her?

src/shared/components/in-app-search/InAppSearch.tsx
imransl: I did lean towards using `mode` initially but then thought we might see the pattern breaking in future designs and went with the safe option.

azangru: But then we may find that the mode itself needs rethinking (what if Andrea puts the `InAppSearch` component that looks like the one on the interstitial page on some other, not interstitial, page?

For now, let's just rely on the mode to determine the height of the input. We can always change it if we come across surprises in the future. My suspicion is that in that case we will likely need to rethink the mode as well (or perhaps add another one).

src/shared/components/question-button/QuestionButton.scss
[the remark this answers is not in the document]
imransl: Hmm that doesn't look right. Going through the XD again there seems to be two sizes: 12px and 16px. I'll check with Andrea about this as well.

src/shared/components/in-app-search/InAppSearch.tsx
imransl: ✅

src/shared/components/input/Input.scss
azangru: image

Let's do 30 as a default.

imransl: ✅

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,60 @@
...
+  display: flex;
+  align-items: center;
+  flex: 0 0 auto;
+}
+
+.inputWrapper {
+  background-color: var(--input-background-colour, $white);
+  display: inline-flex;
+  font-size: $global-font-size;
+  padding: 0 20px;
```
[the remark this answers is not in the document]
imransl: It is going to be 20px so left it as if it is

stories/shared-components/input/Input.stories.tsx
```diff
@@ -68,11 +70,64 @@ export const ShadedInputStory = () => (
...
+      </div>
+    </div>
+    <div className={`${styles.stage} ${styles.greyStage}`}>
+      <p>Disabled with icon in grey background</p>
+      <div className={styles.wrapper}>
+        <FlatInput
+          placeholder="Placeholder for dev..."
+          disabled={true}
+          rightCorner={
+            <QuestionButton helpText="Shows some text when hovered over" />
```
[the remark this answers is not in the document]
imransl: Right corner should not be shown according to Andrea so it will be hidden with the latest change when disabled ✅

src/shared/components/question-button/QuestionButton.scss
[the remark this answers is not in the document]
imransl: 16px for big search boxes and 12px for smaller ones ✅

src/shared/components/input/Input.scss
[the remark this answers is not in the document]
imransl: Andrea changed her mind about this for flat inputs and small shaded ones as well. It is going to be 15px for them and 20px for the large ones. So I've changed the default to 15px.

src/shared/components/input/FlatInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const {
+    className: classNameFromProps,
+    disabled = false,
+    rightCorner,
+    ...otherProps
+  } = props;
+
+  const wrapperClasses = classNames(
+    styles.inputWrapper,
+    styles.flatInputWrapper,
```
azangru, line 43: Hmm... Why do we need both wrapper classes? What's inputWrapper class for?

src/shared/components/input/Input.scss
```diff
@@ -1,23 +1,63 @@
...
+  display: flex;
+  align-items: center;
+  flex: 0 0 auto;
+}
+
+.inputWrapper {
+  background-color: var(--input-background-colour, $white);
+  display: inline-flex;
+  padding: var(--input-padding, 0 15px);
+  width: var(--input-width, 100%);
```
azangru, line 25: Unsure about both the purpose of this class and the names of these variables. Something to discuss...

src/shared/components/input/Input.scss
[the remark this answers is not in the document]
imransl: ✅

src/content/app/genome-browser/components/browser-sidebar-modal/modal-views/navigate-modal/NavigateModal.scss
```diff
@@ -43,24 +41,11 @@
...
-  }
 }
 
 .segmentedInput {
   margin-top: 20px;
-
-  input {
-    width: 140px;
-  }
+  --input-width: 140px;
```
azangru, line 48: Here, and in a couple of places below: where is it used now?

azangru (approved):

PR #965, opened 2023-04-14
Update dependencies

azangru:
## Description
- Removing three dependencies that we aren't using and that were just left forgotten in package.json:
  - `@types/react-router-hash-link`
  - `what-input`
  - `html-webpack-plugin` (storybook still has it in its dependencies (`@storybook/builder-webpack5`); but the app is no longer relying on it
- Storybook is now out of release candidate. Updating it to 7.0.4

Also, using this opportunity to update other dependencies

**Controversial choice**
- Removed `image-webpack-loader`. It has been deprecated (see https://github.com/tcoopman/image-webpack-loader); and npm audit has been complaining about some outdated dependencies of this package that will never be updated. The author suggests [image-minimizer-webpack-plugin](https://github.com/webpack-contrib/image-minimizer-webpack-plugin) instead; but that plugin looks tedious to setup, and also depends on a bunch of stuff. I don't think we even need them, really. We have only very few non-svg images; and we compress them manually.

After removing `image-webpack-loader`, `npm audit` is happy for the first time in a long while:

**before removing `image-webpack-loader`:**

image

image

**after removing `image-webpack-loader`:**

image

## Deployment URL(s)

## Checks
- `npm test` — ✅ 
- `npm run check-types` — ✅
- development build — ✅
- production build — ✅
- storybook build — ✅


imransl:
> We have only very few non-svg images; and we compress them manually.

Is it just the favicons that are not SVG images?

imransl:
> One might conceivably argue for setting up a minimizer plugin for svg images that aren't imported as React components via svgr; but I don't know if we really need that.

If the size of those images are not that big then maybe we don't need to do that?

imransl (approved):

PR #969, opened 2023-04-21
Add breadcrumbs to About app

imransl:
## Description
The Help app has breadcrumbs, whereas the About app does not. This PR adds it to the latter, by reusing the breadcrumbs generation code in the former.

## Related JIRA Issue(s)
[ENSWBSITES-1815](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1815)

## Deployment URL(s)
http://about-app-breadcrumbs.review.ensembl.org

## Views affected
About app

### DIFF
--- a/src/content/app/about/About.scss
+++ b/src/content/app/about/About.scss
@@ -18,11 +18,6 @@ $article-right-padding: 1.5rem;
   }
 }
 
-.textArticle {
-  height: calc(100vh - 370px);
-  overflow-y: auto;
-}
-
 .spinnerContainer {
   display: flex;
   height: 100%;
--- a/src/content/app/about/About.scss
+++ b/src/content/app/about/About.scss
@@ -18,16 +18,16 @@ $article-right-padding: 1.5rem;
   }
 }
 
+.textArticle {
+  height: calc(100vh - 370px);
+  overflow-y: auto;
+}
+
 .spinnerContainer {
   display: flex;
   height: 100%;
   justify-content: center;
   align-items: center;
-}
-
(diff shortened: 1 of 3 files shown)


src/content/app/about/About.scss
```diff
@@ -18,18 +18,18 @@ $article-right-padding: 1.5rem;
   }
 }
 
+.textArticle {
+  height: calc(100vh - 370px);
```
azangru, line 22: We shouldn't do this.

imransl: I've added the height to the article area of the grid in `grid-template-rows` with the latest commit [4fef107](https://github.com/Ensembl/ensembl-client/pull/969/commits/4fef107518700c8b4f63d062511dca28c28df231). This should also fix the article height issue in the Help app.

src/content/app/help/Help.scss
```diff
@@ -30,7 +30,7 @@
   grid-template-areas:
     'breadcrumbs breadcrumbs navigation'
     '.  article  article';
-  grid-template-rows: 40px auto;
+  grid-template-rows: 40px calc(100% - 65px);
```
azangru, line 33: By `calc(100% - 65px)`, do you mean `1fr`?

imransl: It doesn't work with `1fr` unfortunately

azangru (approved):
Ok, let's get this merged, and I'll pick it up in https://github.com/Ensembl/ensembl-client/pull/973

PR #970, opened 2023-04-24
Show hide helptext icon based on helptext availability

jyothishnt:
## Description
Tasks done:
1. Hide question mark if no content is provided
2. Add biotype label before biotype on gene and transcript drawers

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1911

## Deployment URL(s)
http://tooltip-nodata-fix.review.ensembl.org

## Views affected
GB

### DIFF
--- a/src/content/app/genome-browser/components/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
+++ b/src/content/app/genome-browser/components/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
@@ -148,10 +148,13 @@ const TranscriptSummary = (props: Props) => {
 
             {metadata.biotype && (
               <div className={styles.featureDetail}>
+                <span className={styles.labelWithPadding}>Biotype</span>
                 <span>{metadata.biotype.label}</span>
-                <div className={styles.questionButton}>
-                  <QuestionButton helpText={metadata.biotype.definition} />
-                </div>
+                {metadata.biotype.definition && (
+                  <div className={styles.questionButton}>
+                    <QuestionButton helpText={metadata.biotype.definition} />
+                  </div>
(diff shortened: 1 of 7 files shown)


src/content/app/genome-browser/components/drawer/drawer-views/transcript-summary/TranscriptSummary.tsx
```diff
@@ -148,10 +148,13 @@ const TranscriptSummary = (props: Props) => {
 
             {metadata.biotype && (
               <div className={styles.featureDetail}>
+                <span className={styles.labelWithPadding}>Biotype</span>
                 <span>{metadata.biotype.label}</span>
-                <div className={styles.questionButton}>
-                  <QuestionButton helpText={metadata.biotype.definition} />
-                </div>
+                {metadata.biotype.definition && (
```
azangru, line 153: A couple of things.

`definition` is defined as a non-nullable field of `ValueSetMetadata` [here](https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/types/thoas/metadata.ts#L20-L24). For biotypes — for now — this is no longer the case; the definition field for biotype metadata has become nullable. I suggest we add a new type in the thoas/metadata file:

```
export type BiotypeMetadata = Omit<ValueSetMetadata, 'definition'> & {
  definition: string | null;
};
```

The reason I suggest a new type rather than changing the type of the definition field directly on the `ValueSetMetadata` type is because, I think, we are still hoping that the data coming from the value set api (when it finally becomes available) will  always have definitions.

azangru (approved):

imransl (approved):

PR #971, opened 2023-04-24
update variant legend group ids

jyothishnt:
## Description
This PR updates variant group ids based on its values in the data file.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1890

## Deployment URL(s)
http://update-var-group.review.ensembl.org

## Views affected
GB

### DIFF
--- a/src/content/app/genome-browser/components/drawer/drawer-views/variant-group-legend/VariantGroupLegend.scss
+++ b/src/content/app/genome-browser/components/drawer/drawer-views/variant-group-legend/VariantGroupLegend.scss
@@ -31,19 +31,19 @@
 }
 
 .variantColour1 {
-  background-color: $duckegg-blue;
+  background-color: $dark-pink;
 }
 .variantColour2 {
-  background-color: $teal;
+  background-color: $dark-yellow;
 }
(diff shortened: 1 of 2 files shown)


src/content/app/genome-browser/components/drawer/drawer-views/variant-group-legend/VariantGroupLegend.scss
```diff
@@ -31,19 +31,19 @@
 }
 
 .variantColour1 {
-  background-color: $duckegg-blue;
+  background-color: $dark-pink;
```
jyothishnt, line 34: git diff is bit weird. I only changed the class name and order of these :-)

azangru (approved):
Should merge together with updated genome browser backend.

jyothishnt:
> Should merge together with updated genome browser backend.

Why? This change doesn't have any effect as we dont use ids anywhere yet.

azangru:
> Why? This change doesn't have any effect as we dont use ids anywhere yet other than the linking to classnames.

True. Feel free to merge then.

PR #972, opened 2023-04-24
Update BLAST settings styles

azangru:
## Description
- Updated the width of select elements in the BLAST settings section from 70px to 90px
- Since something like BLOSUM62 won't fit in 90px, set the width of the matrix selector (as well as the select elements that can appear directly above it) to 110px
- Removed the set 140px width for labels, thus bringing the select elements closer together (as in design; consulted with Andrea)
- Updated other styles to fix behaviour at breakpoints

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1875

## Deployment URL(s)
http://blast-select-width.review.ensembl.org

## Views affected
BLAST form (specifically, the BLAST parameters section)

### DIFF
--- a/src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
+++ b/src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
@@ -31,7 +31,8 @@
 }
 
 .parametersColumn .select label {
-  grid-template-columns: 140px 70px;
+  grid-template-columns: max-content 90px;
+  justify-content: end;
   text-align: right;
 }
 
@@ -42,9 +43,10 @@
   justify-self: end;
 }
 
-.parametersColumn .matrixSetting .select label,
-.parametersColumn .gapPenalties .select label {
-  grid-template-columns: 140px 110px;
+.matrixSetting.select label,
+.wordSize.select label,
+.matchScores.select label {
+  grid-template-columns: max-content 110px;
 }
 
 .bottomLevel {
@@ -54,8 +56,8 @@
   padding-top: 20px;
   margin-left: 104px;
 
-  @media (max-width: 1610px) {
-    margin-left: -10px;
+  @media (max-width: 1686px) {
+    margin-left: 0px;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
```diff
@@ -54,8 +56,8 @@
   padding-top: 20px;
```
jyothishnt, line 56: Update `gap: 20px`?

jyothishnt:
Loosing the alignment at 1049px. May be not for this PR, but if it is an easy fix, then why not.
Screenshot 2023-04-26 at 11 22 15

azangru:
> if it is an easy fix, then why not

Updated. Removed the margin-left

There is a deeper problem, which is that at our smallest allowed width, the Submission name goes on a new line at all, when there is seemingly plenty of available space to the right; but that would be difficult to fix; so I am not touching that part.

https://user-images.githubusercontent.com/6834224/234708726-6f2d4515-66e7-401f-9b64-cdd5d65103da.mov

src/content/app/tools/blast/components/blast-settings/BlastSettings.scss
azangru: Not sure I understand what you mean here.

jyothishnt: gap as per XD is 20px. Its on line 55.

azangru: I've updated the gap.

There is one thing in which the result deviates from XD: the select box for gap penalties isn't perfectly aligned with the select boxes above it on a narrow screen:

https://user-images.githubusercontent.com/6834224/234836350-366d15f6-b7ea-4d21-b3fe-60771f6e5c65.mov

There isn't a clean solution. The best I could come up with was to set a certain width (190px) on the div containing the the gap penalties; but it is fragile. I asked Andrea if we can ignore this problem, and she agreed.

jyothishnt (approved):

PR #974, opened 2023-04-27
Bring interface of our button components in accordance with standard html buttons

azangru:
## Description
Button components (`PrimaryButton`, `SecondaryButton`...) were among the first we built for ensembl-client, and I made two wrong decisions then. I decided, first, that a button **must** have an `onClick` property (because how else do you interact with a button other than by clicking on it); and that `isDisabled` is a better property name for a button than the standard `disabled` attribute of the native html button. The first decision was wrong, because, inside of a form, a button, by default, submits the form, and does not need a click handler. Insisting on the `onClick` property has led to us passing dummy functions, such as lodash's `noop`, to the buttons, which clearly demonstrates that it was a wrong choice. As for the second decision, having to translate `isDisabled` to `disabled` inside our button component is just an unnecessary complexity.

Therefore, this PR aims to undo the damage done then. The programmatic interface of our button components is brought in accordance with the interface of native html button elements. Specifically:
- the `onClick` property is no longer required
- the `isDisabled` property is renamed to `disabled`

## Deployment URL(s)
http://update-button.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip.tsx
+++ b/src/content/app/entity-viewer/shared/components/entity-viewer-sidebar/entity-viewer-sidebar-toolstrip/EntityViewerSidebarToolstrip.tsx
@@ -15,7 +15,6 @@
  */
 
 import React from 'react';
-import noop from 'lodash/noop';
 
 import { useAppSelector, useAppDispatch } from 'src/store';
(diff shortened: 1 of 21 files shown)


imransl (approved):
Everything looks good when I tested except for some forms failing due to API endpoints returning 404

azangru:
> except for some forms failing due to API endpoints returning 404

Which forms were those?

jyothishnt (approved):

PR #975, opened 2023-05-01
Add species tabs slider

azangru:
## Description
This PR adds `SpeciesTabsSlider`, a component for displaying selected species lozenges in the top of the screen (we don't have the exact dev spec; but see [this XD](https://xd.adobe.com/view/6211d3e4-abf5-4649-a21e-89b84bb76b4e-3742/screen/ea6beea2-3c81-4eba-ba59-472270b2bee2?fullscreen) for an example).

### Behaviour
- Species tabs (lozenges) are always displayed in a single line; never wrap
- If there are too many species tabs to fit in the width of a screen, a slider UI is enabled
  - The slider can be navigated by:
    - pressing the left/right arrow buttons
    - dragging the slider area with a mouse (or a finger on a touch screen)
    - using a horizontal slide gesture on a touchpage
  -  If all species tabs can fit into the width of the screen, the left/right arrow buttons of the slider get hidden
- When user selects a species by clicking on a partially hidden species lozenge, that lozenge scrolls into view
- When user switches between apps or reloads the page, the slider always tries to scroll to a position where the currently active species lozenge is visible

### Notes
- Changed `SpeciesLozenge` component from a div into a button. They are now focusable via keyboard.
- Repurposed SpeciesTabsWrapper to be a simple container that positions the "Change" link next to the tabs slider.

### Not included in this PR
- The "Species in use" lozenge that, according to the [XD](https://xd.adobe.com/view/28612e7a-300e-46ca-a0db-c5f3d13593ad-7d32/screen/445af8ea-31ef-4e91-a241-3fbb59b17ea6/?fullscreen), will be positioned above the slider. Will do this in a separate PR.
- The order in which selected species get added (and, therefore, in which the lozenges are displayed). Andrea suggested that newly added species are always added to the beginning of the list. Currently, the code adds them to the end of the list. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2091

## Deployment URL(s)
http://species-tabs-slider.review.ensembl.org


src/shared/components/species-tabs-slider/useSliderGestures.ts
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+  // see https://greensock.com/forums/topic/32453-convert-vertical-scroll-to-horizontal-scroll-with-observer/
+  // this may be a bad idea in general
+  const handleWheel = (event: Event) => {
+    const wheelEvent = event as WheelEvent;
+    if (Math.abs(wheelEvent.deltaY) > Math.abs(wheelEvent.deltaX)) {
+      handleWheelVertical(wheelEvent);
+    }
+  };
```
azangru, line 103: Wonder if I should remove this...

imransl:
@azangru Don't have anything to comment about the code. Just one thing though about the `tabsSlider`? Can't it be its own component rather than being replicated?

src/shared/components/species-tabs-slider/useSliderGestures.ts
azangru: After some thought, removed this. No need to hijack native wheel behaviour.

azangru:
> Just one thing though about the `tabsSlider`? Can't it be its own component rather than being replicated?

I am not sure, @imransl; and I don't think there is any point. After the "Change" link gets removed, which will happen after we add the "Manage" page (see Andrea's XDs), the `tabsSlider` will turn from this:

```
const speciesTabsSlider = (
    <SpeciesTabsWrapper>
      <SpeciesTabsSlider>{speciesTabs}</SpeciesTabsSlider>
      <Link to={urlFor.speciesSelector()} onClick={trackSpeciesChange}>
        Change
      </Link>
    </SpeciesTabsWrapper>
  );
```

into just:

```
const speciesTabsSlider = <SpeciesTabsSlider>{speciesTabs}</SpeciesTabsSlider>;
```

Not much to extract any further here 🤷‍♂️

PR #977, opened 2023-05-02
Update VariantAllelesSequences component

azangru:
## Description
The `VariantAllelesSequences` component has been moved to shared components, and updated such that it can handle long or multiple sequences.

See the added Storybook stories for examples.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1908

## Deployment URL(s)
http://variant-allele-sequences.review.ensembl.org (the only purpose of this url is to check for regressions; the component should be examined in Storybook)


src/shared/components/variant-alleles-sequences/VariantAllelesSequences.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import ShowHide from 'src/shared/components/show-hide/ShowHide';
+
+import styles from './VariantAlleleSequences.scss';
+
+const COMPACT_MAX_LENGTH = 18;
+
+type Props = {
+  isExpandable?: boolean;
+  alleles: {
```
imransl, line 27: Isn't it just one allele's details that gets passed on?

azangru: There's a `[]` on line 30 🙂

imransl: Yes I missed that one. I need better glasses 😅

imransl (approved):

PR #978, opened 2023-05-04
Refactor thunk actions

imransl:
## Description
Previously the state passed to the thunk actions was typed as `any` when in fact it was `RootState`. This has been changed in this PR across all the reducers/slices. The type of the `getState` function now doesn't need to be explicitly declared as a result of this.

PS: I tried using `createAsyncThunk` as suggested in the JIRA ticket. However, it was causing the site to break and changing the state type from `any` to `RootState` seemed simpler as well.

## Related JIRA Issue(s)
[ENSWBSITES-1666](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1666)

## Deployment URL(s)
http://refactor-thunk-actions.review.ensembl.org

## Views affected
Genome browser, Entity viewer and Species selector

### DIFF
--- a/src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts
+++ b/src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts
@@ -17,10 +17,10 @@
 import {
   createAsyncThunk,
   createSlice,
+  type Action,
+  type ThunkAction,
   type PayloadAction
 } from '@reduxjs/toolkit';
-import type { Action } from 'redux';
-import type { ThunkAction } from 'redux-thunk';
 import cloneDeep from 'lodash/cloneDeep';
 
 import entityViewerStorageService from 'src/content/app/entity-viewer/services/entity-viewer-storage-service';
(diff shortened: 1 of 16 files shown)


src/content/app/entity-viewer/state/gene-view/transcripts/geneViewTranscriptsSlice.ts
```diff
@@ -14,9 +14,13 @@
...
-import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { Action } from 'redux';
-import { ThunkAction } from 'redux-thunk';
+import {
+  createAsyncThunk,
+  createSlice,
+  type PayloadAction
+} from '@reduxjs/toolkit';
+import type { Action } from 'redux';
+import type { ThunkAction } from 'redux-thunk';
```
azangru, line 23: Here and elsewhere: type Action and type ThunkAction are imported by redux-toolkit and then re-exported ([link1](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/index.ts#L2), [link2](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/index.ts#L19). Could you import them from `@reduxjs/toolkit`?

imransl: That's done with the latest commit [5a12a55](https://github.com/Ensembl/ensembl-client/pull/978/commits/5a12a55d6f6ed514bd216b55a348b8624087ad20). Also, fixed a few other type imports as well.

✅

azangru (approved):

PR #980, opened 2023-05-05
Fix GB breaking when clicked on navigation button before loading complete

imransl:
## Description
The site breaks when the navigation button in the sidebar is clicked before the Genome browser loading is completed. This is because `NavigationButtons` component relies on `browserLocation` data which is not available yet and the component breaks as a result.

I tried to fix this in `NavigationButtons` component itself, but couldn't figure out a way to do it properly. Hence, the fix in `RegionNavigation`.

## Related JIRA Issue(s)
[ENSWBSITES-1919](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1919)

## Deployment URL(s)
http://bug-pressing-nav-button.review.ensembl.org

## Views affected
Genome browser sidebar


imransl:
@azangru I had to add `actualChrLocations` to the mock test to fix this. Missed this one in the previous commit.

azangru (approved):

PR #985, opened 2023-05-24
More variant examples

azangru:
## Description
Variation team has updated variant examples. We now have a simple SNV (`rs699`), a more complex SNV (`rs202155613`), and an insertion (`rs71197234`).

Notice that `rs71197234` does not exist in the combined variation data file; so we cannot focus on it. I made it the default variant for display though, because it is sufficiently complex.

Noteworthy changes:
- Updated data types to include the latest changes in Variation data
- Reference allele is now included in the alleles array. It can be identified by its allele type being `biological_region`
- Fixed the code for preparing variant summary data. The previous version was overwriting an array of data produced by a previous cycle in every subsequent one

### TODO in a separate PR
Fix the behaviour of the VariantSummaryStrip component in the genome browser top grey bar (make it adapt to the available width of the grey bar)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1930

## Deployment URL(s)
http://more-variant-examples.review.ensembl.org


jyothishnt:
Example gene in Species home page takes me back to interstitial for some reason.

jyothishnt:
Thoas slow response which is mentioned above
https://github.com/Ensembl/ensembl-client/assets/6347854/620902ca-46e7-438f-a734-4848d11221d7

azangru:
Well, we haven't received any response from Thoas, and thus haven't validated the focus id, and so haven't recorded it anywhere in redux or browser storage. I would say the code behaves as expected. We may review this scenario at some later point

src/shared/helpers/variantHelpers.ts
```diff
@@ -14,16 +14,26 @@
...
-  size: number;
-  description: string;
-  type: ValueSetMetadata;
-  is_global: boolean;
-  is_from_genotypes: boolean;
-  display_group_name: string; // The name of the group of populations this population belongs to; serves for grouping purposes
-  super_population: VariantStudyPopulation | null;
-  sub_populations: VariantStudyPopulation[];
+  const referenceAllele = alleles[referenceAlleleIndex];
+  const alternativeAlleles = alleles.filter(
```
jyothishnt, line 31: Can you delete the reference allele from the alleles array in the step above to avoid this?

jyothishnt:
Where is the Copy/Copied button gone? Is it still in storybook and not added here?

azangru:
> I couldn't find Ancestral on the XDs.

https://xd.adobe.com/view/b485e6c9-546a-4627-a5a6-0edf77651f09-e1d6/screen/1004f946-a763-404b-bfc7-ac60e863121b/

> Btw, where is the Copy/Copied button?

https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/components/copy/Copy.tsx#L47-L50  <-- will hit this on http

src/shared/helpers/variantHelpers.ts
azangru: Why would that be any better?

jyothishnt:
Thanks for the changes.. this PR looks ok to me..

PR #989, opened 2023-06-08
Add grey background to code snippets

bilalebi:
## Description
Add a grey background to code snippets

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/EA-1081

## Deployment URL(s)
http://add-background-code-snippets.review.ensembl.org
> No changes will be visible since there are no code snippets in the docs yet

## Views affected
Help and Docs with code snippets

### DIFF
--- a/src/shared/components/help-article/HelpArticle.scss
+++ b/src/shared/components/help-article/HelpArticle.scss
@@ -70,6 +70,11 @@ $textColumnWidth: 450px;
       padding: 0 5px;
     }
   }
+
+  pre {
+    background-color: #F1F2F4; // Add a background color to code snippets
+    padding: 15px;
+  }
 }
 
 .indexArticle {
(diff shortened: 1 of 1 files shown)


src/shared/components/help-article/HelpArticle.scss
```diff
@@ -70,6 +70,11 @@ $textColumnWidth: 450px;
       padding: 0 5px;
     }
   }
+
+  pre {
+    background-color: #F1F2F4; // Add a background color to code snippets
```
azangru, line 75: We use special variables for colours to keep them consistent (see [link](https://github.com/Ensembl/ensembl-client/blob/dev/src/styles/_settings.scss#L37)). In this case, it should be `background-color: $light-grey;`

bilalebi: Done

PR #992, opened 2023-06-14
Replace clinical significance section in GB drawer for variants

imransl:
## Description
The Clinical Significance section is now obsolete as per the new [XD](https://xd.adobe.com/view/b485e6c9-546a-4627-a5a6-0edf77651f09-e1d6/screen/1004f946-a763-404b-bfc7-ac60e863121b/?fullscreen). It is now replaced with the text `has phenotype associations`, if there are any phenotypes in any of the variant alleles. This link can be used for testing: http://update-clinical-sig-field.review.ensembl.org/genome-browser/grch38?focus=variant:rs699&location=1:230710029-230710068

## Related JIRA Issue(s)
[ENSWBSITES-1959](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1959)

## Deployment URL(s)
http://update-clinical-sig-field.review.ensembl.org

## Views affected
Genome browser -> Sidebar -> Focus variant drawer


src/content/app/genome-browser/components/drawer/drawer-views/variant-summary/VariantSummary.scss
```diff
@@ -55,9 +55,18 @@ For the reference (from Andrea)
   max-width: 60ch;
 }
 
-.clinicalSignificance {
-  display: flex;
-  flex-direction: column;
+.phenotypeAssociations {
+  margin-top: 20px;
```
azangru, line 59: Is it intentional that the distance here is different from `.newRowGroup` class? If yes, could you confirm with Andrea that she wants the vertical space here to be different than, say, between "Ancestral" and "MAF", or between "GERP" and "Location."

src/content/app/genome-browser/components/drawer/drawer-views/variant-summary/prepareVariantSummaryData.ts
```diff
@@ -45,16 +41,25 @@ const prepareVariantSummaryData = (variant: Variant) => {
...
   for (const variantAllele of variant.alleles) {
     addVariantAllelePopulationFrequencyData(variantAllele, variantSummaryData); // iterates over population frequencies
-    addClinicalSignificance(variantAllele, variantSummaryData); // iterates over phenotype assertions
     addVariantAllelePredictions(variantAllele, variantSummaryData); // iterates over prediction results
   }
 
   return variantSummaryData as PreparedVariantSummaryData;
 };
 
+const updateHasPhenotypeAssociations = (
```
azangru, line 54: Why the word `update` in the name of the function?

azangru, line 54: `updateHas` -> just `add`?

azangru: Also, notice that there is a loop (lines 46-49 above) specifically for variant alleles, so that you don't have to iterate over same alleles over and over? Wouldn't it make more sense to transform this function such that it accepts a variant allele, alongside a store, and if `store.hasPhenotypeAssociations` if false, but allele has phenotype_assertions, then it sets `store.hasPhenotypeAssociations` to true?

imransl: ✅

src/content/app/genome-browser/components/drawer/drawer-views/variant-summary/VariantSummary.scss
imransl: @azangru So as Andrea mentioned on Slack, this is fine?

src/content/app/genome-browser/components/drawer/drawer-views/variant-summary/VariantSummary.scss
```diff
@@ -55,9 +55,18 @@ For the reference (from Andrea)
...
+
+  .value {
+    display: flex;
+    column-gap: 10px;
+  }
+
+  svg {
+    fill: $green;
+    width: 12px;
+  }
```
azangru, line 69: It works fine when there is a CADD line ([for rs699](http://update-clinical-sig-field.review.ensembl.org/genome-browser/grch38?focus=variant:rs699&location=1:230710029-230710068)) :

image

but not so great for variants that don't have a CADD ([example link](http://update-clinical-sig-field.review.ensembl.org/genome-browser/grch38?focus=variant:rs779377781&location=13:32357725-32357764)):

image


I suppose, a margin-bottom of 30px should help (it should collapse with the margin-top of CADD). Or perhaps both CADD and GERD will have to be wrapped in a container.

azangru (approved):
Looks good now.

PR #993, opened 2023-06-19
Add 404 screens for Help and About pages; and update their page meta

azangru:
## Description
- Added handling of not found articles to the Help and About apps
  - Server should respond with a `404` http status code
  - The page will show a "not found" screen
- Made page metadata for help pages dynamic
  - Dynamic `title` tag
  - Dynamic meta description
  - Fallbacks if metadata is unavailable

**WARNING:** do not merge until https://github.com/Ensembl/ensembl-client/pull/994 has been merged, and the updated config map applied to staging and production environments.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1808

## Deployment URL(s)
http://help-404.review.ensembl.org

### DIFF
--- a/src/content/app/help/state/api/helpApiSlice.ts
+++ b/src/content/app/help/state/api/helpApiSlice.ts
@@ -1,0 +1,70 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { FetchBaseQueryError } from '@reduxjs/toolkit/query';
+import type { SerializedError } from '@reduxjs/toolkit';
+
+import config from 'config';
+
+import restApiSlice from 'src/shared/state/api-slices/restSlice';
+
+import type {
+  TextArticleData,
+  VideoArticleData
+} from 'src/shared/types/help-and-docs/article';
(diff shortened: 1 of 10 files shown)


src/content/app/help/state/api/helpApiSlice.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { FetchBaseQueryError } from '@reduxjs/toolkit/query';
+import type { SerializedError } from '@reduxjs/toolkit';
+
```
imransl, line 19: Put all the type imports together?

src/content/app/help/state/api/helpApiSlice.ts
```diff
@@ -0,0 +1,70 @@
...
+export const isArticleNotFoundError = (
+  error?: SerializedError | FetchBaseQueryError
+): boolean => {
+  const hasErrorStatus = error && 'status' in error;
+  if (!hasErrorStatus) {
+    return false;
+  }
+
+  const errorStatus = error.status;
+  return typeof errorStatus === 'number' && errorStatus === 404;
```
imransl, line 69: Can't this just be `return errorStatus === 404;`?

src/content/app/help/state/api/helpApiSlice.ts
azangru: I don't think '@reduxjs/toolkit` exports `FetchBaseQueryError` as a top-level export

imransl: Ah yes there's that. I missed seeing it since these were the only imports were there and they both were 'type' imports 😀

src/content/app/help/state/api/helpApiSlice.ts
azangru: It totally can 😄 

Also, I renamed this function to `isMissingResourceError` and moved it to `restSlice`. It should be the same error for any service that responds with a 404 http status code.

imransl (approved):

PR #998, opened 2023-07-02
Species selector modal view

azangru:
## Description
- Add a modal view for species selector
- Add initial implementation of species search field (will need updating in subsequent PRs)
- Add type definitions for popular species and species search matches
- Add a redux api slice for data fetching

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2009
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2006

## Deployment URL(s)
http://species-selector-modal-view.review.ensembl.org


src/content/app/new-species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import React from 'react';
+
+import { useAppDispatch } from 'src/store';
+
+import { setModalView } from 'src/content/app/new-species-selector/state/species-selector-ui-slice/speciesSelectorUISlice';
+
+import ModalView from 'src/shared/components/modal-view/ModalView';
+
+const SpeciesSelectorSelectionModalView = () => {
```
imransl, line 25: Can't the same name be used for both the component and the file name?

src/content/app/new-species-selector/types/speciesSearchMatch.ts
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+export type SpeciesSearchMatch = {
+  genome_id: string;
+  common_name: string | null;
+  scientific_name: string;
+  type: {
+    kind: string; // e.g. "population"
+    value: string; // e.g. "European"
```
imransl, line 23: Do we have predefined values for these two properties rather than random ones?

src/content/app/new-species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
azangru: Sure. This is the result of me first naming the file one way, and then the other, while neglecting to rename the component inside of it.

✅

src/content/app/new-species-selector/types/speciesSearchMatch.ts
azangru: The list of "kinds" must be small (but I do not have it); the list of values, large and open. I think, `string` is an appropriate representation for both.

imransl: Okay so let's change the type of `kind`, when we know what the values for it, later and ignore `value`.

src/root/rootReducer.ts
```diff
@@ -39,6 +40,7 @@ const createRootReducer = () =>
     genome,
     global,
     speciesSelector,
+    newSpeciesSelector, // TODO: rename to just species slector when we retire the old one
```
imransl, line 43: Shouldn't this be going into the server side reducer as well?

azangru: We'll investigate this in ENSWBSITES-2019

imransl (approved):

PR #1004, opened 2023-07-13
Use new variant id schema

azangru:
## Description
As agreed with the Variation team, the ids in the url that are able to uniquely identify a variant need to contain region name, variant start coordinate, variant name (e.g. an rsID), and the sequence of the reference allele, in the following format:

`<region_name>:<start>:<variant_name>`

This PR switches to this convention. The variant string is read from the url, and parsed into significant parts, which are used to direct the genome browser to bring the variant into view. The viewport is calculated as start±25bp. This will allow us to retire the jump file for variants. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2016

## Deployment URL(s)
http://new-variant-id.review.ensembl.org

**Examples to try:**
- http://new-variant-id.review.ensembl.org/genome-browser/grch38?focus=variant:1:230710048:rs699
- http://new-variant-id.review.ensembl.org/genome-browser/grch38?focus=variant:13:57932509:rs71197234
- http://new-variant-id.review.ensembl.org/genome-browser/grch38?focus=variant:13:32379902:rs202155613

### DIFF
--- a/src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
+++ b/src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
@@ -224,23 +224,23 @@ const checkFocusLocation = async (params: CheckFocusObjectParams) => {
 
 const checkFocusVariant = async (params: CheckFocusObjectParams) => {
   // NOTE: so far, only checking that focus variant id conforms to the following format
-  // <region_name>:<start_coordinate>:<variant_name>:<reference_sequence>
+  // <region_name>:<start_coordinate>:<variant_name>
   const { parsedFocusObjectId } = params;
   const { objectId } = parsedFocusObjectId;
 
   const variantIdParts = objectId.split(':');
 
-  if (variantIdParts.length !== 4) {
-    // expect four parts in a focus variant id
+  if (variantIdParts.length !== 3) {
+    // expect three parts in a focus variant id
     return {
       isMissingFocusObject: true
     };
   }
 
-  const [regionName, start, variantName, referenceSequence] = variantIdParts; // eslint-disable-line
+  const [regionName, start, variantName] = variantIdParts; // eslint-disable-line
 
-  // we know that at least start coordinate must be a number, and reference sequence consists only of letters
-  if (/\D/.test(start) || /[^a-z]/i.test(referenceSequence)) {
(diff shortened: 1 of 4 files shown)


src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
```diff
@@ -220,6 +222,35 @@ const checkFocusLocation = async (params: CheckFocusObjectParams) => {
...
+  // <region_name>:<start_coordinate>:<variant_name>
+  const { parsedFocusObjectId } = params;
+  const { objectId } = parsedFocusObjectId;
+
+  const variantIdParts = objectId.split(':');
+
+  if (variantIdParts.length !== 3) {
+    // expect three parts in a focus variant id
+    return {
+      isMissingFocusObject: true
```
imransl, line 236: Isn't this incorrect focus object rather than missing focus object?

azangru: Technically, yes; but in the future we are going to send the variant id to the variation api, and it will tell us that the variant with the supplied id does not exist.  I'd rather keep the error as missing focus object.

imransl (approved):

jyothishnt:
Hi Andrey, This may not be for this PR, but I was playing with the UI and found when there is no variant ID in the url, GB still works (which is good) but Previously Viewed shows empty. Check screenshot
Screenshot 2023-07-20 at 09 49 46

src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
jyothishnt: If there is a `:` and no variant ID this condition becomes false which I guess is not correct. It should still capture the missing variant id

jyothishnt: > If there is a `:` and no variant ID this condition becomes false which I guess is not correct. It should still capture the missing variant id `focus=variant:1:230710048:`

Sorry this one is marked as resolved and you may miss it.

jyothishnt (approved):

PR #1006, opened 2023-07-18
Use Intl NumberFormat for number formatting

azangru:
## Description
Replace the old implementation of the function that we had for formatting numbers to comma-separated strings with a proper Intl-based solution.

Not only is this the right thing to do, but it also will fix the problem discovered when reviewing species stats:

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1971

## Deployment URL(s)
http://number-formatting.review.ensembl.org

### DIFF
--- a/src/shared/helpers/formatters/numberFormatter.ts
+++ b/src/shared/helpers/formatters/numberFormatter.ts
@@ -18,7 +18,7 @@
     Formats the input number to comma separated representation
     eg: 10000 -> 10,000
 */
-export const getCommaSeparatedNumber = (
+export const formatNumber = (
   input: number,
   options?: Intl.NumberFormatOptions
 ) => {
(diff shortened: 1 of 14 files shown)


src/shared/helpers/formatters/numberFormatter.ts
```diff
@@ -18,8 +18,12 @@
     Formats the input number to comma separated representation
     eg: 10000 -> 10,000
 */
-export const getCommaSeparatedNumber = (input: number) => {
-  return input.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
+export const getCommaSeparatedNumber = (
```
azangru, line 21: I guess a question for me is whether to keep calling this function `getCommaSeparatedNumber` or change its name to `getFormattedNumber` or `formatNumber`.

imransl: But how many types of formatted numbers will we have?

azangru: 🤷‍♂️

PR #1010, opened 2023-08-04
Update all deployments in gitlab-ci.yaml for the new k8s cluster

azangru:
Updated all deployments in gitlab-ci.yaml for the new k8s cluster.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -205,7 +205,7 @@ Test_N_Build:
     ENVIRONMENT: production
     API_HOST: ""
   rules:
-    - if: $CI_COMMIT_BRANCH == "master" || $CI_COMMIT_BRANCH == "dev"
+    - if: $CI_COMMIT_BRANCH == "master" || $CI_COMMIT_BRANCH == "new-cluster-deployments"
 
 # Job to build static asset for internal environment
 # master branch -> Internal
@@ -234,7 +234,7 @@ Test_N_Build:review:
 Node:Staging:
   extends: .build-node
   rules:
-    - if: $CI_COMMIT_BRANCH == "dev"
+    - if: $CI_COMMIT_BRANCH == "new-cluster-deployments"
   needs:
     - Test_N_Build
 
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -209,6 +209,7 @@ Test_N_Build:
   only:
   - dev
   - master
+  - new-cluster-deployments
```
kamaldodiya, line 212: Can you change this to based on rules 

https://github.com/Ensembl/ensembl-web-metadata-api/blob/main/.gitlab-ci.yml#L55-L57
```
rules:
    - if: '$CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "main"'
      when: on_success
```

.gitlab-ci.yml
```diff
@@ -138,7 +138,7 @@ variables:
 # Template for deployment to the new kubernetes cluster
 # For live deployment, it does not need to deploy a static assets container
 # Will need updating after the migration is completed (the name of the template, name of the ensembl-k8s-manifests branch)
-.deploy-wp40:
+.deploy-newcluster:
   stage: deploy
   image: dockerhub.ebi.ac.uk/ensembl-web/deploy-tools:0.6
```
kamaldodiya, line 143: Use latest deploy-tools image
`dockerhub.ebi.ac.uk/ensembl-web/deploy-tools:latest`

.gitlab-ci.yml
```diff
@@ -222,6 +223,7 @@ Test_N_Build:internal:
 
   only:
   - master
+  - new-cluster-deployments
```
kamaldodiya, line 226: See above to use rules

kamaldodiya (changes requested):

.gitlab-ci.yml
azangru: But we've never used deploy freeze for client. I remember you introduced it for complex cases like Thoas, where we were sharing the repository with another team, and wanted a fine-grained control over the deployment; but with ensembl-client, we are in full control. Do we really need this?

And if we really need this level of control for deployments to production, why would we need it for deployments to any other environment? Here, you left the comment under `Test_N_Build:internal`

kamaldodiya: That's valid point but only/except are not being actively developed by gitlab and prefer rules over only/except 
https://archives.docs.gitlab.com/15.11/ee/ci/yaml/#only--except

kamaldodiya (approved):

PR #1011, opened 2023-08-06
Update search field behaviour in new species selector

azangru:
## Description
- Add a distinction between a "Find" and an "Add" button
- The "Find" button is disabled if the search field content is less than 3 characters
- On pre-selection of one or more genomes in the search results table, the "Find" button changes to "Add"
- Changing the text in the search field removes the table of search results as well as any pre-selected genomes
- Clicking the "Add" button adds selected species
- Added placeholder text and help icon to the search field 
- Added dedicated style for the selected species rows in the search results table
- Added a dedicated DisabledExternalLink component to use in the search results table for selected species

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2049

## Deployment URL(s)
http://species-search-field.review.ensembl.org/new-species-selector

### DIFF
--- a/src/content/app/new-species-selector/components/species-search-field/SpeciesSearchField.scss
+++ b/src/content/app/new-species-selector/components/species-search-field/SpeciesSearchField.scss
@@ -1,3 +1,5 @@
+@import 'src/styles/common';
+
 .speciesSearchField {
   display: grid;
   grid-template-areas:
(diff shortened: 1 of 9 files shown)


imransl:
The search results only ever returns `human` results even if I search for `mouse` or `cat`. Is this  a known issue?

azangru:
> The search results only ever returns human results even if I search for mouse or cat. Is this a known issue?

We don't have a real search api, and are using mock data, which happens to be human,  to populate the table.

PR #1012, opened 2023-08-09
Add circular and linear fraction indicators

azangru:
## Description
This PR adds components to draw simple diagrams that make their appearance in the Homology table; but Andrea says they may appear outside of the table too.
- The diagrams are, in effect, a linear and a circular progress indicators. But I didn't want to call them a ProgressIndicator, because that would be confusing.
- The final decition was to have:
  - `LinearProportionIndicator` — just the bar diagram
  - `LinearPercentageIndicator` — bar diagram with label; the input is between 0 and 100
  - `LinearFractionIndicator` — bar diagram with label; the input is between 0 and 1
  - `CircularProportionIndicator`  — just the circle diagram
  - `CircularPercentageIndicator`– circle diagram with label; the input is between 0 and 100
  - `CircularFractionIndicator` – circle diagram with label; the input is between 0 and 1
- Added stories in Storybook showing these components in action

[Link to XD](https://xd.adobe.com/view/e8af9441-569c-4226-9b9f-9c2ac298b4c3-bdfc/screen/e17c8d90-26f9-4566-9386-69b838f1c49f?fullscreen)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2133

## Deployment URL(s)
None. Please check Storybook


src/shared/components/fraction-indicator/LinearProportionIndicator.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <span className={styles.label}>{formattedValue}</span>
+    </div>
+  );
+};
+
+/**
+ * This component renders LinearProportionIndicator with a value next to it.
+ * It is meant to be used with values between 0 and 1
+ */
+export const LinearFractionIndicator = (props: Props) => {
```
imransl, line 84: I wonder if we need to have some kind of check in case values bigger than `1` here and `100` above are provided?

azangru: I don't know how we can prevent it. If this happens, the component will show a fully coloured bar or circle. This doesn't seem worse than returning `null`, and is arguably better than throwing an error. I would wait and see what we discover when this component starts being used.

imransl (approved):

PR #1019, opened 2023-09-04
Use the trigger field from track api payload to pass messages to genome browser

azangru:
## Description
After enabling the expansion node for variation tracks in the genome browser (https://github.com/Ensembl/ensembl-dauphin-style-compiler/pull/86), the genome browser, sadly, will use different triggers for different tracks — e.g. `["track", track_id]` for tracks declared in the boot file; but `["track", expansion_node, track_id]` for tracks registered via the expansion node. It thus became necessary the additional information to browser chrome about how any given track needs to be called (see PRs https://github.com/Ensembl/ensembl-web-track-api/pull/19 and https://github.com/Ensembl/ensembl-web-track-api/pull/20).

The updates in this PR:
- use the new `trigger` field in the track api payload to build messages for the genome browser
- use the new `on_by_default` field in the track api payload to determine whether to enable a track by default (notice that 1000genomes track is hidden by default, because the track api suggests that it should not be turned on by default)

**NOTE:** It may be possible that the genome browser server will be broken at the time you review this. If this is the case, please run the branch locally.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2136

## Deployment URL(s)
http://use-track-api-updates.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
+++ b/src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
@@ -59,26 +59,26 @@ export const setBrowserLocation = (payload: {
 
 export const toggleTrack = (payload: {
   genomeBrowser: EnsemblGenomeBrowser;
-  trackId: string;
+  trackPath: string[];
   isEnabled: boolean;
 }) => {
-  const { genomeBrowser, trackId, isEnabled } = payload;
-  genomeBrowser.switch(['track', trackId], isEnabled);
+  const { genomeBrowser, trackPath, isEnabled } = payload;
+  genomeBrowser.switch(trackPath, isEnabled);
 };
 
 // NOTE: this method can only handle boolean flags
 export const toggleTrackSetting = (payload: {
   genomeBrowser: EnsemblGenomeBrowser;
-  trackId: string;
+  trackPath: string[];
   setting: string;
   isEnabled: boolean;
 }) => {
-  const { genomeBrowser, trackId, setting, isEnabled } = payload;
+  const { genomeBrowser, trackPath, setting, isEnabled } = payload;
 
-  if (trackId === 'focus-variant') {
+  if (trackPath.includes('focus-variant')) {
     toggleFocusVariantTrackSetting(payload);
   } else {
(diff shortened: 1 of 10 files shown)


src/server/middleware/proxyMiddleware.ts
```diff
@@ -55,15 +55,25 @@ const serverConfig = getConfigForServer();
...
+      changeOrigin: true,
+      secure: false
+    }
+  );
+
+  const browserProxyMiddleware = createHttpProxyMiddleware('/api/tracks/**', {
+    target: 'http://update.review.ensembl.org',
+    // pathRewrite: {
+    //   '^/api/browser': '/api' // rewrite path
+    // },
```
azangru, line 71: **NOTE TO SELF:** remember to undo this before merging.

src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
```diff
@@ -59,26 +59,26 @@ export const setBrowserLocation = (payload: {
...
+  trackPath: string[];
   isEnabled: boolean;
 }) => {
-  const { genomeBrowser, trackId, isEnabled } = payload;
-  genomeBrowser.switch(['track', trackId], isEnabled);
+  const { genomeBrowser, trackPath, isEnabled } = payload;
+  genomeBrowser.switch(trackPath, isEnabled);
 };
 
 // NOTE: this method can only handle boolean flags
```
imransl, line 69: This isn't related to this PR but is the comment still relevant?

imransl:
Again don't think this is unique to the PR but spotted some warnings in the console while testing.

Screenshot 2023-09-05 at 10 52 28

azangru:
> Again don't think this is unique to the PR but spotted some warnings in the console while testing.

Yes, this is not something for this PR; but if you can find the conditions under which this warning can be reproduced, that would be useful.

src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
azangru: Yes. This method still only handles boolean flags.

imransl (approved):

PR #1023, opened 2023-09-05
Update rules for focusing on variants via zmenu

azangru:
## Description
As defined in the XD, when variant length is greater than the viewport, genome browser is expected to zoom out, such that the variant occupies 70% of the viewport.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2140

## Deployment URL(s)
http://variant-refocus.review.ensembl.org

### Example of how to test
Navigate to http://variant-refocus.review.ensembl.org/genome-browser/grch38?focus=variant:1:230710246:rs1471899965&location=1:230710230-230710262, and then click on the deletion immediately to the left of the SNV

image

### DIFF
--- a/src/content/app/genome-browser/components/zmenu/zmenus/VariantZmenu.tsx
+++ b/src/content/app/genome-browser/components/zmenu/zmenus/VariantZmenu.tsx
@@ -92,11 +92,12 @@ const VariantZmenu = (props: Props) => {
 
 /**
  * Consider that the genome browser shows the detailed zoomed-in view for variants
- * at scales 2**1 to 2**6 base pairs. 2**6 is 64; and when the genome browser zooms out to this point,
+ * at scales between 2**1 and 2**6 base pairs. 2**6 is 64; and when the genome browser zooms out to this point,
  * it switches to the zoomed-out view.
  */
-const MAX_NUCLEOTIDES_PER_VIEWPORT = 64 - 2;
-const MIN_NUCLEOTIDES_PER_VIEWPORT = 32;
+const VARIANT_DETAILS_MAX_NUCLEOTIDES = 2 ** 6 - 1; // variant details program of the genome browser shows up to 63 nucleotides
+const MAX_NUCLEOTIDES_PER_VIEWPORT = VARIANT_DETAILS_MAX_NUCLEOTIDES - 2; // to leave at least one nucleotide on either side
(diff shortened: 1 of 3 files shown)


src/content/app/genome-browser/services/genome-browser-service/genomeBrowserCommands.ts
```diff
@@ -53,8 +56,6 @@ export const setBrowserLocation = (payload: {
       focusType: focus.type
     });
   }
-
-  genomeBrowser.goto(start, end);
```
azangru, line 57: Moved the goto instruction before the setting the focus, because otherwise the genome browser does this:

https://github.com/Ensembl/ensembl-client/assets/6834224/48b85e02-4f7b-4cd8-aa9d-53529d1becb5

I have no idea why this should have made a difference. Bloody genome browser.

src/content/app/genome-browser/components/zmenu/zmenus/VariantZmenu.tsx
```diff
@@ -82,4 +90,39 @@ const VariantZmenu = (props: Props) => {
   );
 };
 
+/**
+ * Consider that the genome browser shows the detailed zoomed-in view for variants
+ * at scales 2**1 to 2**6 base pairs. 2**6 is 64; and when the genome browser zooms out to this point,
+ * it switches to the zoomed-out view.
+ */
+const MAX_NUCLEOTIDES_PER_VIEWPORT = 64 - 2;
```
imransl, line 98: Can't the value of this just be `62`?

azangru: I was trying to make it explicit that 64 is 2**6. I should probably make it even more explicit still. Like:


```js
const VARIANT_DETAILS_TRACK_NUCLEOTIDES_LIMIT = 2 ** 6 - 1; // 63 nucleotides
const MAX_NUCLEOTIDES_PER_VIEWPORT = VARIANT_DETAILS_TRACK_NUCLEOTIDES_LIMIT - 1; // to make it an even number
```

imransl: This would actually be easier to understand

imransl (approved):

ens-st3 (approved):
spacing of 70:30 is not always accuarate, but this works well overall

PR #1032, opened 2023-09-29
Disable species search field

azangru:
## Description
Andrea has updated the design of the species search results view. In it she disables the species search field. See [XD](https://xd.adobe.com/view/b8d67194-34d7-40de-bcd9-a3413ea5d727-ed85/screen/d06762a7-6b1e-43e8-baba-1ef4b7996d0a)

While I think disabling the input field is hostile to users, because they cannot easily correct their query, and that this is a mistake that we will later have to correct, it is easy to set up. In fact, this decision simplifies our code. This is what this PR is about.

Also, added a view for empty search results (see [XD](https://xd.adobe.com/view/b8d67194-34d7-40de-bcd9-a3413ea5d727-ed85/screen/bd236467-8165-4187-8384-b6b34f366630))

Also in the PR:
- Added the `disabled` behaviour to the ShadedInput component
- Added a `TextButton` component, which uses native html button and thus is better for accessibility than spans or divs

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2197

## Deployment URL(s)
http://disable-species-search-field.review.ensembl.org

### DIFF
--- a/src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
+++ b/src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
@@ -27,10 +27,11 @@ export type Props = {
   query: string;
   canAdd: boolean;
   onAdd: () => void;
+  onCancel: () => void;
 };
 
 const AddFoundSpecies = (props: Props) => {
-  const { query, canAdd, onAdd } = props;
+  const { query, canAdd, onAdd, onCancel } = props;
 
   return (
     <div className={styles.grid}>
@@ -45,7 +46,7 @@ const AddFoundSpecies = (props: Props) => {
         <PrimaryButton disabled={!canAdd} onClick={onAdd}>
           Add
         </PrimaryButton>
-        <TextButton>Cancel</TextButton>
+        <TextButton onClick={onCancel}>Cancel</TextButton>
       </div>
     </div>
   );
--- a/src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
+++ b/src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
@@ -1,0 +1,54 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
```diff
@@ -0,0 +1,55 @@
...
+import styles from './SpeciesSearchField.scss';
+
+export type Props = {
+  query: string;
+  canAdd: boolean;
+  onAdd: () => void;
+  onCancel: () => void;
+};
+
+const AddFoundSpecies = (props: Props) => {
```
imransl, line 33: Won't `AddSelectedSpecies` be more self descriptive?

src/content/app/new-species-selector/components/species-search-results-summary/SpeciesSearchResultsSummary.scss
```diff
@@ -9,3 +9,21 @@
...
+
+.searchHelp p {
+  margin: 0;
+}
+
+.searchHelp ul {
+  padding-inline-start: 20px;
+}
+
+.searchHelp li {
```
imransl, line 27: This overrides the `list-style-type` in `ul` (Normalize.css) so how about moving it to `searchHelp ul`?

azangru: I don't think it matters. `list-style-type` property applies to `li` element anyway. Setting it on `ul` in normalize was using the css cascade to pass that rule down to the `li`s. Setting it here directly on `li` feels more natural.

See MDN:

> Only a few elements (<li>) and <summary>) have a default value of display: list-item. However, the list-style-type property may be applied to any element whose display value is set to list-item. Moreover, because this property is inherited, it can be set on a parent element (commonly <ol> or <ul>) to make it apply to all list items.

src/content/app/new-species-selector/components/species-search-field/AddFoundSpecies.tsx
azangru: It's that word again — selected — I am never quite sure what it means in the context of our site.

imransl: So how about `AddStagedSpecies` instead?

azangru: I suppose I can change it to `AddSpecies` (or `AddGenomes`).

azangru: Renamed `AddFoundSpecies`  to `AddSpecies`

imransl (approved):

PR #1037, opened 2023-10-16
Add sorting for the species table

azangru:
## Description
- Add the `ColumnHead` component to use with the `Table` component. The `ColumnHead` component can show the sort arrows if it allows sorting.
-  Added the sorting logic for genomes results table
- Enabled sticky header for the genomes table
- Updated CSS of ExternalLink component to fix its positioning along z-axis when overlaid by the sticky header 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2205

## Deployment URL(s)
http://species-table-sorting.review.ensembl.org/new-species-selector

### DIFF
--- a/src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
+++ b/src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
@@ -1,0 +1,198 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useState, useCallback } from 'react';
+import type { SelectableGenome } from './useSelectableGenomesTable';
+
+type SortableColumn =
+  | 'common_name'
+  | 'scientific_name'
+  | 'type'
+  | 'assembly_name'
+  | 'assembly_accession_id'
+  | 'coding_genes_count'
+  | 'contig_n50';
+
+type SortOrder = 'asc' | 'desc';
+
+export type SortRule = {
+  column: SortableColumn;
+  sortOrder: 'asc' | 'desc';
+};
+
+export type ChangeSortRule = (
+  column: SortableColumn,
+  sortOrder: SortOrder | 'none'
+) => void;
+
+const useOrderedGenomes = (genomes: SelectableGenome[]) => {
+  const [sortRule, setSortRule] = useState<SortRule | null>(null);
+
+  let orderedGenomes = genomes;
+
+  const changeSortRule: ChangeSortRule = useCallback((column, sortOrder) => {
+    if (sortOrder === 'none') {
+      setSortRule(null);
+    } else {
+      setSortRule({ column, sortOrder });
+    }
+  }, []);
+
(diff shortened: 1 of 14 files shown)


src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
```diff
@@ -0,0 +1,198 @@
...
+  column: SortableColumn,
+  sortRule: SortRule | null
+) => {
+  if (sortRule?.column === column) {
+    return sortRule.sortOrder;
+  } else {
+    return 'none';
+  }
+};
+
```
azangru, line 157: The four functions below look generic enough to be moved to the helpers directory.

src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
```diff
@@ -0,0 +1,198 @@
...
+type SortableColumn =
+  | 'common_name'
+  | 'scientific_name'
+  | 'type'
+  | 'assembly_name'
+  | 'assembly_accession_id'
+  | 'coding_genes_count'
+  | 'contig_n50';
+
+type SortOrder = 'asc' | 'desc';
```
imransl, line 29: Shouldn't `SortOrder` also have `none` as a value?

azangru: No.

I've been struggling with the way how to represent "no sort order". For the ColumnHeader component, I was debating between `'asc' | 'desc' | 'none'` and `'asc' | 'desc' | null`; and managed to convince myself that if `ColumnHeader` were an html element, then the string 'none', which can be passed as an attribute, would be preferable to `null`, which only has to be passed as a property. I am still not sure whether the interface I chose for `ColumnHeader` is correct.

But for the hook, there is no need for the `none` string — it is responsible for determining the sort order by listening for signals from all columns; and there may only be either a sorting rule for one of the columns (ascending or descending), or no sorting rule at all (i.e. null).

src/shared/components/table/ColumnHead.scss
```diff
@@ -0,0 +1,47 @@
...
+.columnHeadSortingApplied {
+  font-weight: $bold;
+}
+
+.sortArrow {
+  width: var(--_sort-arrow-width);
+  transform: rotate(180deg);
+  fill: $grey;
+  // display: inline-block;
+  // margin-right: var( --_sort-arrow-right-margin);
```
imransl, line 38: Remove these comments?

src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
imransl: On the other hand, `SortOrder` type is defined in quite a few places and seems to be almost of the same type (except for `none` not used in a couple of places). Can it be reused rather than being defined everywhere?

src/shared/components/table/Table.scss
```diff
@@ -19,9 +19,9 @@
   background-clip: padding-box;
 }
 
-.table th,
+// .table th,
```
imransl, line 22: Remove comment?

src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
azangru: ✅

src/content/app/new-species-selector/components/selectable-genomes-table/useOrderedGenomes.ts
azangru: Ok, I created two types, `SortOrder` and `SortOrderWithNone`, and moved them into a dedicated file.

src/shared/components/table/ColumnHead.scss
azangru: ✅

src/shared/components/table/Table.scss
azangru: ✅

imransl (approved):

PR #1041, opened 2023-10-20
Fix page metadata for Entity Viewer and Help page

azangru:
## Description
**Bug:** Switching on the client side to Entity Viewer was not updating the title of the browser tab; although you could see the correct text if you refreshed the page.

https://github.com/Ensembl/ensembl-client/assets/6834224/802ae04f-00b1-4a99-b45f-6d447f60fd57

**Cause:** The code that was responsible for updating page metadata was not wrapped in the right context provider. (Although it's weird that it didn't cause a runtime error, as I would expect from [these lines](https://github.com/Ensembl/ensembl-client/blob/dev/src/content/app/entity-viewer/hooks/useEntityViewerIds.ts#L24-L28).)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2254

## Deployment URL(s)
http://fix-ev-page-meta.review.ensembl.org

## Views affected
- Entity Viewer pages (if you switch to Entity Viewer from any other app)
- Index page of the Help & Docs app

### DIFF
--- a/src/content/app/entity-viewer/EntityViewer.tsx
+++ b/src/content/app/entity-viewer/EntityViewer.tsx
@@ -39,7 +39,6 @@ import {
   initializeSidebar
 } from 'src/content/app/entity-viewer/state/sidebar/entityViewerSidebarSlice';
 
-import EntityViewerIdsContextProvider from 'src/content/app/entity-viewer/contexts/entity-viewer-ids-context/EntityViewerIdsContextProvider';
 import { StandardAppLayout } from 'src/shared/components/layout';
 import EntityViewerAppBar from './shared/components/entity-viewer-app-bar/EntityViewerAppBar';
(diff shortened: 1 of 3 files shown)


imransl:
@azangru I am seeing `Gene: BRCA2` when viewing `BRCA2` rather than Entity viewer. Is this as expected?

Screenshot 2023-10-20 at 17 26 46

azangru:
>  I am seeing Gene: BRCA2 when viewing BRCA2 rather than Entity viewer. Is this as expected?

Yep, completely expected. This is the title I intended EntityViewer to show when there is an entity to display — it is what it is showing on staging/beta if you refresh the page. We may be told to change this title in the future (this requires more in-depth discussions that will consider how the titles of our pages might appear on Google); but the real bug so far was that Entity Viewer pages were not  registering their title during client-side rendering, i.e. when you switched from a different app to Entity Viewer

imransl (approved):

PR #1048, opened 2023-11-03
Replace old species selector with the new one

azangru:
## Description
This PR replaces the old Species Selector with the new one.

### Notes
- The single SpeciesSelectorSlice is now split into several smaller slices with their own responsibility. This PR adds `species-selector-general-slice`, whose purpose is to just keep track of the species that were added via the SpeciesSelector.
  - As a consequence, the path to the file with redux selectors that are responsible for reading selected species from redux has changed, and that required updates in of multiple imports
  - The `CommittedItem` type has been moved to its own file, which also meant a change of file path and changes to imports in many other files
  - The shape of the species selector state has changed, which required updates to a bunch of tests
- Because the new SpeciesSelector does not have an autocomplete; and the list of popular species is fetched via redux toolkit, this allowed us to remove the code responsible for search autocomplete and for popular species that previously resided in species selector slice and species selector epics 
- Updated the `CommittedItem` type to include additional fields, such as `species_taxonomy_id`, `type`, and `is_reference`; and also made the value of the `assembly` field a nested object. 
- Marking popular species buttons as selected by comparing the `species_taxonomy_id` field with the one in committed items.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2260

## Deployment URL(s)
http://replace-species-selector.review.ensembl.org


src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
```diff
@@ -34,7 +34,7 @@ import type { BlastJobResultResponse } from 'src/content/app/tools/blast/types/b
 import type { SubmittedSequence } from 'src/content/app/tools/blast/types/blastSequence';
 
 export type BlastSubmissionPayload = {
-  species: Species[];
+  species: Partial<Species>[];
```
azangru, line 37: This looks suspiciously permissive.

src/content/app/species-selector/state/speciesSelectorEpics.ts
```diff
@@ -64,10 +64,16 @@ export const ensureCommittedSpeciesEpic: Epic<Action, Action, RootState> = (
       const { genomeInfo } = action.payload;
```
azangru, line 64: If I understand correctly, at the moment, in this branch, the payload from `fetchGenomeInfo` will not contain the fields expected below. It is still using the old endpoint from the genome search api, whose payload can be seen e.g. [here](https://beta.ensembl.org/api/genomesearch/genome/info?genome_id=a7335667-93e7-11ec-a39d-005056b38ce3).

The endpoint is being replaced in https://github.com/Ensembl/ensembl-client/pull/1047, after which we should have access to these fields.

src/content/app/tools/blast/state/blast-api/blastApiSlice.ts
jyothishnt: It is a test, so does it really matter?

azangru: What do you mean a test? This is not a test file.

jyothishnt: What I mean is this is `createBlastSubmissionPayload` is used in blastEpics.test

jyothishnt: Happy to update it, but the question is, is it necessary.

jyothishnt: Sorry my bad. It is not just used on test. Will push my changes once you resolve the conflict

src/content/app/species-selector/SpeciesSelector.tsx
```diff
@@ -14,47 +14,55 @@
...
-    ) : (
-      <div>
-        <SpeciesSearchPanel />
-        <PopularSpeciesPanel />
-      </div>
-    );
+  const modalView = useAppSelector(getSpeciesSelectorModalView);
+  const dispatch = useAppDispatch();
+
+  const isSpeciesSelectorResultsView = [
```
imransl, line 36: This variable name is confusing. How about `shouldShowResultsView` instead?

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -14,110 +14,77 @@
...
-  }, []);
+export type Props = {
+  query: string;
+  onSearchSubmit: () => void;
+  canSubmit?: boolean;
+  onInput?: ((event: FormEvent<HTMLInputElement>) => void) | (() => void);
+};
 
-  const { trackAutocompleteSpeciesSelect } = useSpeciesSelectorAnalytics();
+export const SpeciesSearchField = (props: Props) => {
```
imransl, line 37: Since we still haven't got a better name for this component was wondering how e can rename it. Would `SimpleSpeciesSearchField` make it less confusing/better for imports?

src/content/app/species-selector/services/species-selector-storage-service.ts
```diff
@@ -18,7 +18,7 @@ import storageService, {
   StorageServiceInterface
 } from 'src/services/storage-service';
 
-import { CommittedItem } from 'src/content/app/species-selector/types/species-search';
+import { CommittedItem } from 'src/content/app/species-selector/types/committedItem';
```
imransl, line 21: Another type import?

src/content/app/species-selector/speciesSelectorHelper.ts
```diff
@@ -14,17 +14,10 @@
  * limitations under the License.
  */
 
-import {
-  CommittedItem,
-  PopularSpecies
-} from 'src/content/app/species-selector/types/species-search';
-import { CurrentItem } from 'src/content/app/species-selector/state/speciesSelectorSlice';
-import { SearchMatch } from 'src/content/app/species-selector/types/species-search';
+import { CommittedItem } from 'src/content/app/species-selector/types/committedItem';
```
imransl, line 17: Type import?

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
azangru: It's fine; we have a default export vs a named export.

src/content/app/species-selector/SpeciesSelector.tsx
azangru: ✅

imransl (approved):

PR #1050, opened 2023-11-08
Bugfixes for the species page

azangru:
### Bug 1
If the example features metadata endpoint does not return example objects; the "Example gene" or "Example location" links were still visible; but because they did not have any associated example object data, the "View in app" tooltip was empty, and looked broken:

image

### Bug 2
Open the species page, then click on the chevron closing the sidebar. This produces an error because state[genome_id] is undefined at that point.

Screenshot from Sentry:

image

## Deployment URL(s)
http://species-page-fixes.review.ensembl.org

### DIFF
--- a/src/content/app/species/state/general/speciesGeneralHelper.ts
+++ b/src/content/app/species/state/general/speciesGeneralHelper.ts
@@ -661,51 +661,50 @@ const getExampleLinks = (props: {
       (object) => object.type === 'gene'
     );
 
-    const focusId = geneExample?.id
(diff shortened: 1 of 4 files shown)


imransl:
@azangru I now need to click on the chevron twice to close the sidebar in the species page

azangru:
@imransl thank you for noticing. When I tested the sidebar on my side before submitting the PR, I must have dismissed the silent first click thinking that I must have missed the chevron.

I've pushed a fix.

imransl (approved):

PR #1051, opened 2023-11-09
Show dynamic legends for regulation

jyothishnt:
## Description
Use metadata api genome regulation stats to dynamically show regulation legends on track panel.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-1976

## Deployment URL(s)
http://custom-legend.review.ensembl.org

## Views affected
GB


src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -36,12 +36,22 @@ import { getBrowserActiveGenomeId } from 'src/content/app/genome-browser/state/b
 import styles from '../TrackPanelList.scss';
 import trackPanelItemStyles from './TrackPanelItem.scss';
 import regulationStyles from 'src/content/app/genome-browser/components/drawer/drawer-views/regulation-legend/RegulationLegend.scss';
+import { useGetRegulationStatsQuery } from 'src/content/app/genome-browser/state/api/genomeBrowserApiSlice';
```
azangru, line 39: Nitpick: could you move this import up, to other imports from redux slices.

src/content/app/genome-browser/state/api/genomeBrowserApiSlice.ts
```diff
@@ -37,6 +38,11 @@ import {
 } from 'src/content/app/genome-browser/state/api/queries/variantQuery';
 import { regionQuery, type RegionQueryResult } from './queries/regionQuery';
 
+import {
+  RegulationStatistics,
+  SpeciesStatsResponse
+} from 'src/content/app/species/state/api/speciesApiTypes';
```
imransl, line 44: Type imports?

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -36,12 +36,22 @@ import { getBrowserActiveGenomeId } from 'src/content/app/genome-browser/state/b
...
   const activeGenomeId = useAppSelector(getBrowserActiveGenomeId) as string;
   const dispatch = useDispatch();
   const { trackDrawerOpened, reportTrackPanelSectionToggled } =
     useGenomeBrowserAnalytics();
+  const { data: regulationStats } = useGetRegulationStatsQuery(activeGenomeId);
+
+  if (!regulationStats) {
+    return;
+  }
+  const availableRegulationLegends = regulationLegends.filter((legend) => {
```
azangru, line 51: Mmm, there is something about the names that we missed last time.

A legend, if applied to an explanatory note next to an image, is the note as a whole. It is not just one colour/label combination; it is all of them. Thus, when this PR is using the word "legend" to refer to just a single colour square with text next to it, and calls them collectively by the plural form "legends", it is using this word incorrectly.

Notice that the component itself is called `...Legend`.

Could we possibly rename "legends" (the constants file and its contents) to "legend"; and change the variable name here to something different. `availableRegulatoryFeatures` would be nice.

src/content/app/genome-browser/constants/regulationLegends.ts
```diff
@@ -17,30 +17,35 @@
 const regulationLegends = [
   {
     id: 1,
+    stats_api_key: 'promoters',
```
azangru, line 20: I have a slight reservation for this name of the field — "api key" is usually something we use to authenticate with an api. Wouldn't it be clearer if we turned these strings into the value of the `id` field; and moved the current value of the id field to a field called something like `colour_code`:

```
{
    id: 'promoters',
    colour_code: 1,
    ....
}
```

src/content/app/species/state/api/speciesApiTypes.ts
```diff
@@ -93,9 +93,12 @@ type VariationStatistics = {
   structural_variants_with_phenotype_assertions: number;
 };
 
-type RegulationStatistics = {
+export type RegulationStatistics = {
   enhancers: number;
   promoters: number;
+  ctcf_count: number;
+  tfbs_count: number;
+  open_chromatin_count: number;
```
azangru, line 101: Notice the type defined in the api spec. These three added fields are nullable; therefore their type is `number | null`

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -36,12 +36,22 @@ import { getBrowserActiveGenomeId } from 'src/content/app/genome-browser/state/b
...
 import trackPanelItemStyles from './TrackPanelItem.scss';
 import regulationStyles from 'src/content/app/genome-browser/components/drawer/drawer-views/regulation-legend/RegulationLegend.scss';
+import { useGetRegulationStatsQuery } from 'src/content/app/genome-browser/state/api/genomeBrowserApiSlice';
 
 const TrackPanelRegulationLegend = (props: { disabled: boolean }) => {
   const activeGenomeId = useAppSelector(getBrowserActiveGenomeId) as string;
   const dispatch = useDispatch();
   const { trackDrawerOpened, reportTrackPanelSectionToggled } =
     useGenomeBrowserAnalytics();
+  const { data: regulationStats } = useGetRegulationStatsQuery(activeGenomeId);
```
azangru, line 46: Why the need for this separate query in the first place? Why not just:

```
import { useGetSpeciesStatisticsQuery } from 'src/content/app/species/state/api/speciesApiSlice';
...

const { currentData: genomeStats } = useGetSpeciesStatisticsQuery({ genomeId: activeGenomeId });
const regulatoryStats = genomeStats?.regulation

if (!regulatoryStats) {

...and so on
```

src/content/app/genome-browser/components/drawer/drawer-views/regulation-legend/RegulationLegend.tsx
```diff
@@ -16,7 +16,7 @@
 
 import React from 'react';
 
-import regulationLegends from 'src/content/app/genome-browser/constants/regulationLegends';
+import regulationLegend from 'src/content/app/genome-browser/constants/regulationLegend';
```
azangru, line 19: image

Should rename the file as well

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -42,6 +43,22 @@ const TrackPanelRegulationLegend = (props: { disabled: boolean }) => {
   const dispatch = useDispatch();
   const { trackDrawerOpened, reportTrackPanelSectionToggled } =
     useGenomeBrowserAnalytics();
+  const { data: regulationStats } = useGetSpeciesStatisticsQuery(
```
azangru, line 46: It is not just the regulation stats that you will be getting here. You will get the whole species stats object, which has `regulation_stats` as a key inside of it.

By treating the whole species stats object as regulation stats, you are currently getting the following in the sidebar.

image

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -42,6 +43,22 @@ const TrackPanelRegulationLegend = (props: { disabled: boolean }) => {
...
+    {
+      skip: !activeGenomeId
+    }
+  );
+
+  if (!regulationStats) {
+    return;
+  }
+  const availableRegulatoryFeatures = regulationLegend.filter((legend) => {
+    return Object.keys(regulationStats).includes(legend.id);
```
azangru, line 59: No need to repeatedly create arrays with the same values (`Object.keys`) inside of the loop. Create the array once outside of the loop.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
jyothishnt: Oops! sorry my bad! I think my browser wasn't responding to this for some reason. Will update it

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -42,6 +43,26 @@ const TrackPanelRegulationLegend = (props: { disabled: boolean }) => {
...
+  const regulationStats = data?.genome_stats.regulation_stats;
+
+  if (!regulationStats) {
+    return;
+  }
+
+  const regulationStatKeys = Object.keys(regulationStats);
+  const availableRegulatoryFeatures = regulationLegend.filter((legend) => {
+    return regulationStatKeys.includes(legend.id);
+  });
```
azangru, line 64: The check above is incorrect. The statistics object coming from the backend will always include all keys; but the values of some of those keys will be null

See Chicken. All regulatory statistics fields are there; but some are marked as null. Notice that the legend in the sidebar is listing all features:

image

Change

```
regulationLegend.filter((legend) => {
  return regulationStatKeys.includes(legend.id);
});
```
to

```
regulationLegend.filter((item) => {
  return !!regulationStatKeys[item];
});
```

jyothishnt: this doesn't work
Screenshot 2023-11-20 at 15 06 54

azangru: Sorry, not `!!regulationStatKeys[item]`, but `!!regulationStatKeys[item.id]`. An `item` is an object, and, of course, cannot be used as a key. My mistake.

src/content/app/genome-browser/components/track-panel/components/track-panel-list/track-panel-items/TrackPanelRegulationLegend.tsx
```diff
@@ -42,6 +43,26 @@ const TrackPanelRegulationLegend = (props: { disabled: boolean }) => {
...
+    }
+  );
+
+  const regulationStats = data?.genome_stats.regulation_stats;
+
+  if (!regulationStats) {
+    return;
+  }
+
+  // const regulationStatKeys = Object.keys(regulationStats);
```
azangru, line 61: ✂️

azangru (approved):

PR #1053, opened 2023-11-12
Refactor species stats types and add link to example variant

azangru:
## Description
- Add link to example variant to the species page
- Refactored types related to species statistics in speciesGeneralHelper.ts
- Simplified `speciesApiSlice` — moved the step for munging the data into the step required by the Species page into the component responsible for displaying the stats accordion

### Why change the speciesApiSlice?
Species statistics data are now going to be used not only on the species page, but also on the genome browser page for the regulatory features legend. Moving the code that transforms the backend response out of the RTK endpoint makes the endpoint reusable.

### Why refactor the types?
The types were heavily based on enums. Not only enums are a rather bad idea in typescript; but they were also incorrectly written in this case, which the linter is only now picking up:

Screenshot 2023-11-11 at 15 49 17

See the squigglies under most of the options of this enum? This is because the enum is using differently named members set to the same string value, for example NON_CODING_AVERAGE_GENOMIC_SPAN and PSEUDOGENES_AVERAGE_GENOMIC_SPAN are both set to the value "average_genomic_span" that is a key in the payload sent by the api. Below, you can see the eslint error that produces the squigglies

Screenshot 2023-11-11 at 15 49 34

I took this opportunity to move away from enums in this file.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2283

## Deployment URL(s)
http://add-example-variant-link.review.ensembl.org

### DIFF
--- a/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
+++ b/src/content/app/species/components/species-main-view/SpeciesMainViewStats.tsx
@@ -19,11 +19,7 @@ import classNames from 'classnames';
 
 import { useAppDispatch, useAppSelector } from 'src/store';
 import useSpeciesAnalytics from '../../hooks/useSpeciesAnalytics';
-
-import ViewInAppPopup from 'src/shared/components/view-in-app-popup/ViewInAppPopup';
-import SpeciesStats from 'src/content/app/species/components/species-stats/SpeciesStats';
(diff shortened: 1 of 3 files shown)


jyothishnt:
I have checked the species_stats part of it and it looks good :-)  I will check the remaining later

imransl (approved):

PR #1067, opened 2023-12-18
Fix ternary operator issue

jyothishnt:
## Description
Removes the extra 0 displayed when a species search in blast returns no results

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2370

## Deployment URL(s)
http://zero-search.review.ensembl.org

## Views affected
Blast species search


azangru:
Or `!!currentData?.matches.length &&`. Whichever is the easiest.

PR #1069, opened 2023-12-19
Update Home & About styles migration

jyothishnt:
## Description
Updated Home and About pages

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2075

## Deployment URL(s)
http://home-scss.review.ensembl.org


src/content/app/about/About.module.css
```diff
@@ -34,6 +29,6 @@ $article-right-padding: 1.5rem;
 }
 
 .asideTitle {
-  font-weight: $light;
+  font-weight: var(--color-light);
```
azangru, line 32: No! Font-weight is not a color!

src/content/home/Home.module.css
```diff
@@ -0,0 +1,215 @@
...
+.mainTopMiddle .logotype {
+  height: 36px;
+  fill: var(--color-white);
+  justify-self: center;
+}
+
+.mainTopMiddle .text {
+  text-align: center;
+  font-size: 18px;
+  font-weight: var(--color-light);
```
azangru, line 32: Font-weight is not a color!

src/content/home/Home.module.css
```diff
@@ -0,0 +1,215 @@
...
+  position: relative;
+  width: 221px;
+  height: 40px;
+  background-color: var(--color-blue);
+  color: var(--color-white);
+  border-radius: 6px;
+  text-align: center;
+  padding: 5px;
+  font-size: 18px;
+  font-weight: var(--color-bold);
```
azangru, line 58: Font-weight is not a color!

src/content/home/Home.module.css
```diff
@@ -0,0 +1,215 @@
...
+}
+
+.appListItem .button span {
+  margin-left: -20px;
+}
+
+.appListItem .button .appIconWrapper {
```
azangru, line 65: When you unwrap the nested classes, check if that specificity was really needed. For example, would just the `.appIconWrapper` be sufficient?

src/content/home/Home.module.css
```diff
@@ -0,0 +1,215 @@
...
+.homeTopRight {
+  position: absolute;
+  top: 2px;
+  right: var(--standard-gutter);
+  display: grid;
+  grid-template-columns: repeat(2, auto);
+  align-items: start;
+}
+
+.homeTopRight .helpLabel {
```
azangru, line 91: For classes from here and below, check whether the increased specificity produced by combining the two class names was really necessary. For example, won't `.helpLabel` be a sufficient selector on its own?

src/content/home/Home.module.css
```diff
@@ -0,0 +1,215 @@
...
+    max-height: 60%;
+    transform: translateX(calc(-1 * (100% - 100vw) / 2));
+  }
+
+  .speciesContainer .platform {
+    height: 6%;
+  }
+}
+
+footer {
```
azangru, line 152: Erm. This is a tag name; a tag name cannot be scoped to a module; so its CSS rules will infect any other footer, if we ever have one. Should change the selector (could be `.home footer`, for example)

src/content/app/about/About.module.css
jyothishnt, line 32:
```suggestion
  font-weight: var(--font-weight-light);
```

src/content/home/Home.module.css
jyothishnt, line 32:
```suggestion
  font-weight: var(--font-weight-light);
```

src/content/home/Home.module.css
jyothishnt, line 58:
```suggestion
  font-weight: var(--font-weight-bold);
```

src/content/home/Home.module.css
```diff
@@ -0,0 +1,216 @@
...
+  grid-template-columns: repeat(2, auto);
+  align-items: start;
+}
+
+.helpLabel {
+  color: var(--color-blue);
+  font-weight: var(--font-weight-bold);
+}
+
+.howToUse {
```
azangru, line 96: I don't think this class is used anywhere?

azangru (approved):

PR #1084, opened 2024-02-05
Refactor ExternalLink and ExternalReference components

azangru:
## Description
Both the `ExternalLink` component and the `ExternalReference` component use our old and error-prone way of styling through the `classNames` prop. This PR refactors them so that:
- Instead of passing the link content through the `linkText` property, it is passed as a regular component child
- Relevant styleable properties are exposed through CSS custom properties.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2408

## Deployment URL(s)
http://ext-link-refactor.review.ensembl.org


imransl (approved):
The external links look much better now in the review site :)

PR #1085, opened 2024-02-06
move Copy link above the vcf string

jyothishnt:
## Description

Copy link when the vcf strings are long go beyond the screen below and become hidden at the bottom.
Moved it above the vcf string as in the updated xd
I updated the VariantVCF component itself so that this change affects everywhere we show vcf strings (Confirmed with Andrea)

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2401

## Deployment URL(s)
http://move-copy-link.review.ensembl.org

## Views affected
Variant EV, GB drawer

### DIFF
--- a/src/shared/components/variant-vcf/VariantVCF.tsx
+++ b/src/shared/components/variant-vcf/VariantVCF.tsx
@@ -52,6 +52,7 @@ const VariantVCF = (props: Props) => {
 
   return (
     <div className={componentClasses}>
+      {props.withCopy && <Copy value={vcfSequenceParts.vcfString} />}
       <span className={styles.vcfString}>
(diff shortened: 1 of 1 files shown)


azangru (approved):
This placement of the Copy button on top of the VCF string feels awkward and wrong to me. I think it was a rash decision, and I like @kamaldodiya 's [suggestion](https://genomes-ebi.slack.com/archives/C67KM6TD0/p1707385839685209?thread_ts=1707311463.487799&cid=C67KM6TD0) about the change of the copy button itself, which then becomes much easier to position, far better. But, from the technical perspective, there is no reason why this PR cannot be merged.

PR #1093, opened 2024-02-22
Update the Copy button and the VariantVCF component

azangru:
## Description
- Add a new property to the Copy button, so that it could align the text saying "Copy" to the left, middle, or right of the reserved space, and not push the surrounding elements to the side when the button is pressed (all because when we click the "Copy" button, a "Copied" lozenge appears, which is wider than the initial label 🤦 )
- Following Andrea's instructions, made the VariantVCF component always start a new line after the "Copy" button. Don't know what impact this is going to have for us later on.
- Added the Copy button to Storybook

https://github.com/Ensembl/ensembl-client/assets/6834224/30504737-c9d7-4dee-9fad-a30d3a3e8744

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2424

## Deployment URL(s)
http://copy-button.review.ensembl.org

_(NOTE: You won't see the copy button on review deployments because the copy functionality does not work over http. Please check locally.)_

### DIFF
--- a/src/content/app/entity-viewer/variant-view/variant-view-sidebar/overview/VariantOverview.module.css
+++ b/src/content/app/entity-viewer/variant-view/variant-view-sidebar/overview/VariantOverview.module.css
@@ -23,6 +23,7 @@
 
 .labelValueWrapper {
   display: flex;
+  align-items: baseline;
   margin-top: 8px;
 }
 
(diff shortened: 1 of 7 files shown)


jyothishnt (approved):
Looks good to me.

imransl (approved):

PR #1095, opened 2024-02-27
Update download title

jyothishnt:
## Description
Make Download title consistent one apps

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2440

## Deployment URL(s)
http://update-title.review.ensembl.org

## Views affected
Species selector, GB, Zmenu, EV


src/content/app/genome-browser/state/browser-sidebar-modal/browserSidebarModalSlice.ts
```diff
@@ -29,7 +29,7 @@ export enum BrowserSidebarModalView {
   SEARCH = 'search',
   BOOKMARKS = 'Previously viewed',
   SHARE = 'Share',
-  DOWNLOADS = 'Downloads',
+  DOWNLOADS = 'Download',
```
azangru, line 32: If you are also changing the values in enums, then I wonder whether we should also change the name of the enum key as well (`DOWNLOAD` instead of `DOWNLOADS`). If yes, then we would do so in all apps that have such enums (genome browser, entity viewer, species page). If we make this change, it would be purely for our (developers') benefit; so it depends on how much more annoying the form "downloads" is than "download".

azangru (approved):

PR #1096, opened 2024-02-29
Refactor feature sequence downloading, and add downloading of exons

azangru:
## Description
The code responsible for the downloading of genomic sequences got refactored in the following way.

**Reason for refactoring:** We used to have a [single function](https://github.com/Ensembl/ensembl-client/blob/dev/src/shared/workers/sequenceFetcher.worker.ts#L30) in a web worker, which received a checksum and a refget url, and fetched a sequence and transformed it into a FASTA-formatted string, which it passed back to the caller. The problem with this approach was that if, for example, the user chose to download genomic sequences for all transcripts in a gene, we would send multiple requests to refget, one per transcript, which for some genes may mean hundreds of unnecessary requests. Moreover, now that we are adding the capability of downloading exons, the number of network requests for all exons in a gene would get much greater. All these requests are completely unnecessary, because it is sufficient to just fetch the genomic sequence of a gene, and then to subslice transcripts and exons out of it in the client, which is what this PR does.

**Summary of the approach:**
- The code for fetching the sequences and generating the FASTA strings still runs in a web worker (it probably doesn't have to; but there shouldn't be any harm in doing so).
- The logic for how to deal with different kinds of sequences has been moved from the main thread (files that used to be called `fetchForGene`, `fetchForTranscript`, etc.) into the worker. Now, the functions responsible for generating FASTA strings, also contain the logic about how these strings should be built.

**New functionality:** It is now possible to download exons, either for a single transcript, or for all transcripts in a gene.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2391

## Deployment URL(s)
http://exons-download.review.ensembl.org


imransl (approved):
Looks good to me

src/shared/components/instant-download/instant-download-gene/InstantDownloadGene.tsx
```diff
@@ -215,12 +243,35 @@ const TranscriptSection = (props: TranscriptSectionProps) => {
...
+      import.meta.url
+    )
   );
+
+  try {
+    const service = wrap<WorkerApi>(worker);
+    const sequences = await service.downloadSequencesForGene({
+      genomeId: params.genomeId,
+      geneId: params.geneId,
+      geneSequences: params.options.gene,
```
jyo-nt, line 265: Are you planning to rename these two keys?

jyothishnt, line 265: Are you planning to rename these keys?

azangru: Renamed them to `...sequenceTypes` (`geneSequenceTypes`, `transcriptSequenceTypes`).

jyothishnt (approved):

PR #1097, opened 2024-02-29
Move species search results views to dedicated urls

imransl:
## Description
This PR improves the browser back button behaviour of the Species selector page, by adding the species selectors results view into the browser history. Additionally, it has made preparation for navigation to work with popular species, find gene and to create the link assemblies in the Species page. Perhaps these changes can be done in this PR itself.

**TODO**

- Fix broken gene search (need to create a route for it)
- Remove UI redux slice for species selector 

## Related JIRA Issue(s)
[ENSWBSITES-2271](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2271)
[ENSWBSITES-2400](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2400)

## Deployment URL(s)
http://species-results-back-button.review.ensembl.org

## Views affected
Species selector page


src/content/app/species-selector/SpeciesSelector.tsx
```diff
@@ -30,13 +31,14 @@ import SpeciesSelectorGeneSearchView from './views/species-selector-gene-search-
...
+  const [searchParams] = useSearchParams();
   const modalView = useAppSelector(getSpeciesSelectorModalView);
   const dispatch = useAppDispatch();
 
-  const shouldShowResultsView = [
-    'species-search',
-    'popular-species-genomes'
-  ].includes(modalView || '');
+  const shouldShowResultsView =
+    pathname.includes('/search') &&
```
azangru, line 40: Shouldn't this (the `/search` part) be done via react-router's react components? See EntityViewer as an example.

src/content/app/species-selector/SpeciesSelector.tsx
```diff
@@ -30,13 +31,14 @@ import SpeciesSelectorGeneSearchView from './views/species-selector-gene-search-
...
+    (searchParams.has('query') || searchParams.has('species_taxonomy_id'));
```
azangru, line 41: This should go into the component that matches the `/search` route.

src/content/app/species-selector/components/popular-species-list/PopularSpeciesList.tsx
```diff
@@ -43,8 +44,10 @@ const PopularSpeciesList = () => {
 
   const onPopularSpeciesButtonClick = (species: PopularSpecies) => {
     dispatch(setPopularSpecies(species));
-    dispatch(setModalView('popular-species-genomes'));
```
azangru, line 46: Now that we are getting rid of the modal view mechanism, we should remove the code that relied on it (from redux slice, and from the components that used the selectors that were reading that view).

src/content/app/species-selector/components/popular-species-list/PopularSpeciesList.tsx
```diff
@@ -43,8 +44,10 @@ const PopularSpeciesList = () => {
...
     trackPopularSpeciesClick(species);
+    navigate(
+      `/species-selector/search?species_taxonomy_id=${species.species_taxonomy_id}`
```
azangru, line 49: We probably want to move the generation of these urls into the urlHelper file.

src/content/app/species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
```diff
@@ -73,19 +70,21 @@ const Content = (props: { onClose: () => void }) => {
     for (const addedGenome of genomes) {
       trackAddedGenome(addedGenome);
     }
+
+    navigate('/species-selector');
```
azangru, line 74: I am very unsure and confused about how the navigation should work when user presses the "Add" button. Last I heard, Andrea was saying that the user should be taken to the screen they came from (which, if they pressed on the "Assemblies" link on the species page, will not be "/species-selector"). I do not know how much of this is still true. You can discuss this with Andrea.

src/content/app/species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
```diff
@@ -73,19 +70,21 @@ const Content = (props: { onClose: () => void }) => {
...
-  ) : selectedPopularSpecies ? (
+  ) : searchParams.has('species_taxonomy_id') && selectedPopularSpecies ? (
     <GenomeSelectorBySpeciesTaxonomyId
       onSpeciesAdd={onSpeciesAdd}
-      speciesTaxonomyId={selectedPopularSpecies.species_taxonomy_id}
-      speciesImageUrl={selectedPopularSpecies.image}
+      speciesTaxonomyId={selectedPopularSpecies?.species_taxonomy_id}
+      speciesImageUrl={selectedPopularSpecies?.image}
     />
   ) : null; // this last option should never happen
```
azangru, line 89: We should probably be a bit smarter with the handling of the query parameters. Do we really want to show a mostly empty screen if the query parameter is neither `query` nor `species_taxonomy_id`; or do we want to rewrite the url such that we push the user to the main species selector screen?

src/content/app/species-selector/SpeciesSelector.tsx
imransl: ✅

src/content/app/species-selector/SpeciesSelector.tsx
imransl: ✅

src/content/app/species-selector/components/popular-species-list/PopularSpeciesList.tsx
imransl: ✅

src/content/app/species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
imransl: For now I am redirecting to `/species-selector` but we might need to check if this is alright with Andrea?

imransl: ✅

imransl:
@azangru I have pushed the changes to remove the `popular_species` from Species selector search slice and link the assembly count in Species selector to the list of assemblies in Species search results view.

src/content/app/species-selector/SpeciesSelector.tsx
```diff
@@ -30,33 +26,20 @@ import SpeciesSelectorGeneSearchView from './views/species-selector-gene-search-
...
-    <SpeciesSelectorResultsView />
-  ) : modalView === 'gene-search' ? (
-    <SpeciesSelectorGeneSearchView />
-  ) : (
-    <SpeciesSelectorMainView />
+  const body = (
+    <Routes>
+      <Route index element={<SpeciesSelectorMainView />} />
+      <Route path="/search" element={<SpeciesSelectorResultsView />} />
+      <Route path="/search/gene" element={<SpeciesSelectorGeneSearchView />} />
```
azangru, line 41: One thing to verify is that we expect the gene search view to be accessible only from the species selector. Technically, there is nothing species selector-ish in gene search view. Ideally, I would have expected it to go inside of `/search` app.

src/shared/helpers/urlHelper.ts
```diff
@@ -36,12 +35,53 @@ type SpeciesPageUrlParams = {
...
+  speciesTaxonomyId?: string | number;
+};
+
 export const speciesPage = (params: SpeciesPageUrlParams) => {
   const speciesPageRootPath = '/species';
 
   return `${speciesPageRootPath}/${params.genomeId}`;
 };
 
+export const speciesSelector = (params?: SpeciesSelectorParams) => {
```
azangru, line 52: This should probably be split into three:
- `speciesSelector`: just returns `/species-selector`
- `speciesSelectorSearch`: returns `/species-selector/search?query...` or `/species-selector/search?species_taxonomy_id: ...`
- `geneSearch` returns whatever url we decided to use for gene search.

src/content/app/species-selector/state/species-selector-search-slice/speciesSelectorSearchSlice.ts
```diff
@@ -25,21 +25,17 @@ import { saveMultipleSelectedSpecies } from 'src/content/app/species-selector/se
...
 
 type State = {
   query: string;
-  popularSpecies: PopularSpecies | null;
 };
 
 const initialState: State = {
-  query: '',
-  popularSpecies: null
+  query: ''
```
azangru, line 38: I wonder if we still need to keep track of the query in redux. I can't think of what it might be needed for now.

src/content/app/species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
```diff
@@ -54,15 +52,26 @@ const SpeciesSelectorResultslView = () => {
...
   const dispatch = useAppDispatch();
   const { trackAddedGenome, trackTotalSelectedGenomesCount } =
     useSpeciesSelectorAnalytics();
 
+  useEffect(() => {
+    if (
+      !(
+        (searchParams.has('query') && searchParams.get('query')) ||
+        (searchParams.has('species_taxonomy_id') &&
+          searchParams.get('species_taxonomy_id'))
```
azangru, line 68: Do you actually need the `has` here? What happens if you just `get`?

src/content/app/species/components/species-title-area/SpeciesTitleArea.module.css
```diff
@@ -41,12 +41,14 @@
 }
 
 .assemblyCountWrapper {
+  cursor: pointer;
```
azangru, line 44: - is the whole wrapper supposed to be clickable, or just the blue "Assemblies"?
- you wouldn't need this line if you turned your span (or your div) into an html button element

src/content/app/species/components/species-title-area/SpeciesTitleArea.tsx
```diff
@@ -122,7 +130,10 @@ const SpeciesTitleArea = () => {
           <h1 className={styles.speciesName}>{getDisplayName(species)}</h1>
           <span className={styles.assemblyName}>{species.assembly.name}</span>
         </div>
-        <div className={styles.assemblyCountWrapper}>
+        <div
+          className={styles.assemblyCountWrapper}
+          onClick={navigateToAssemblies}
+        >
```
azangru, line 136: Either this `div`, or the Assemblies `span` (depending on which element is expected to be clickable) should instead become a `button`.

_(Personally, I would only expect the `Assemblies` to be clickable; but of course if Andrea said otherwise, stick with that.)_

src/content/app/species/components/species-title-area/SpeciesTitleArea.module.css
azangru, line 44: Now that you changed the div to the button, cursor: pointer is unnecessary.

src/content/app/species/components/species-title-area/SpeciesTitleArea.tsx
azangru: Now the InfoPill has collapsed again...

image

I think we should replace the default padding of InfoPill with `2px 8px` instead of `0 8px`.

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -75,16 +71,19 @@ If no results are shown, please try a different spelling or attribute
 // Species search field, but wrapped in redux.
```
azangru, line 71: This comment is no longer correct (the search field is no longer using redux).

Rather, it should say, `Species search field, but wrapped in a component that reads a query parameter from the url`.

_(The reason this wrapped component exists is because in BLAST, the search field will behave differently)._

src/content/app/species-selector/views/species-selector-results-view/SpeciesSelectorResultsView.tsx
```diff
@@ -73,19 +74,20 @@ const Content = (props: { onClose: () => void }) => {
...
       onSpeciesAdd={onSpeciesAdd}
       onClose={props.onClose}
     />
-  ) : selectedPopularSpecies ? (
+  ) : searchParams.has('species_taxonomy_id') ? (
     <GenomeSelectorBySpeciesTaxonomyId
       onSpeciesAdd={onSpeciesAdd}
-      speciesTaxonomyId={selectedPopularSpecies.species_taxonomy_id}
-      speciesImageUrl={selectedPopularSpecies.image}
+      speciesTaxonomyId={searchParams.get('species_taxonomy_id') || ''}
```
azangru, line 90: ` || ''` -> `as string` (you have checked that `searchParams.get('species_taxonomy_id')` exists above)

src/content/app/species/components/species-title-area/SpeciesTitleArea.module.css
```diff
@@ -47,6 +47,7 @@
 }
```
azangru, line 47: currently:

image

to fix, add:
`align-items: center;` to `.assemblyCountWrapper`

imransl: I don't see it like this on Chrome. Which browser does it appear like this in? Also, when I add centre it the padding reverts to what it was before.
Screenshot 2024-03-18 at 17 18 49

imransl: I checked this in the wrong branch so please ignore the above comment

azangru (approved):

PR #1099, opened 2024-03-13
Add CDS diagram to transcript consequences view

azangru:
## Description
This PR add CDS diagram to the transcript consequences view.

**NOTE:** Because the diagram requires the display of exons within a CDS, this PR attempts to calculates exon locations within the CDS using relative locations both of spliced exons to transcript and of CDS to transcript. Notice that both these locations are relative to a single transcript, which means that if we ever get examples of true trans-splicing (in which exons from different genes can combine together to produce a product), then they will not show up properly. We should be getting exon locations relative to cDNA and to CDS from an api; but currently aren't. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2432

## Deployment URL(s)
http://tr-cons-cds.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/variant-view/transcript-consequences/transcript-variant-cds/TranscriptVariantCDS.tsx
+++ b/src/content/app/entity-viewer/variant-view/transcript-consequences/transcript-variant-cds/TranscriptVariantCDS.tsx
@@ -1,0 +1,340 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


src/content/app/entity-viewer/variant-view/transcript-consequences/transcript-variant-cds/TranscriptVariantCDS.tsx
```diff
@@ -0,0 +1,323 @@
...
+  });
+
+  return <g>{exonBlocks}</g>;
+};
+
+const ExonBlock = (props: { x: number; width: number }) => {
+  const { x, width } = props;
+  const initialChevronWidth = 32; // <-- from the viewBox attribute of the svg image, in which the chevron is pointing down
+  const initialChevronHeight = 20; // <-- from the viewBox attribute of the svg image, in which the chevron is pointing down
+  const targetChevronWidth = 9; // <-- before rotating the chevron 90°; after the rotation this will become its height
```
imransl, line 122: Can't these values be turned into constants like above?

azangru: They are local to a single function, and to a rather minor concern of chevron dimensions. I'm not sure it's worth  there'll be any benefit

imransl (approved):

PR #1100, opened 2024-03-13
Update header bar styling for GB and EV

imransl:
## Description
Update the header bar styling in Genome browser and Entity viewer

## Related JIRA Issue(s)
[ENSWBSITES-2445](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2445)

## Deployment URL(s)
http://update-headerbar-styling.review.ensembl.org

## Views affected
Genome browser and Entity viewer

### DIFF
--- a/src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.module.css
+++ b/src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.module.css
@@ -22,7 +22,7 @@
   margin-right: 10px;
 }
 
-.emphasized {
+.variantName {
   font-weight: var(--font-weight-bold);
   margin-right: 12px;
 }
--- a/src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.tsx
+++ b/src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.tsx
@@ -89,7 +89,7 @@ export const VariantSummaryStrip = (props: VariantSummaryStripProps) => {
     <div className={componentClasses}>
       <div className={styles.section}>
         <span className={styles.label}>Variant</span>
-        <span className={styles.emphasized}>{variant.name}</span>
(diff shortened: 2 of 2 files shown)


azangru:
This is from the review deployment. The main bug seems to be still present.

image

src/shared/components/feature-summary-strip/VariantSummaryStrip.tsx
```diff
@@ -70,28 +77,36 @@ const VariantSummaryStrip = (
...
-        <span className={styles.featureNameEmphasized}>{variant.name}</span>
+      <div className={featureStripStyles.section}>
+        <span className={featureStripStyles.featureSummaryStripLabel}>
+          Variant
+        </span>
+        <span className={featureStripStyles.featureNameEmphasized}>
+          {variant.name}
+        </span>
+        <span className={styles.variantType}>
+          {referenceAllele?.allele_type.value}
```
azangru, line 95: If you do this, then the label will always read "biological_region" (this is what the type of the reference allele always is).

What you should use instead is `variant->allele_type->value`

src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.module.css
```diff
@@ -24,10 +24,10 @@
 
 .emphasized {
   font-weight: var(--font-weight-bold);
+  margin-right: 12px;
```
azangru, line 27: This contradicts the name of the class. And also, you no longer seem to be using this class anywhere; you are using `featureStripStyles.featureNameEmphasized` in `VariantSummaryStrip` component

imransl: It's probably the other `VariantSummaryStrip` component that you are talking about. This class is used in `src/content/app/entity-viewer/variant-view/variant-summary-strip/VariantSummaryStrip.tsx`.

azangru (approved):

PR #1102, opened 2024-03-18
Use gene stable ids from transcript consequences to get lists of transcripts

azangru:
## Description
Instead of reading transcript's stable id from the first predicted molecular consequence and assuming that all other transcripts are from the same gene, in this PR, we read gene stable ids, and, if there are different genes in the list of predicted molecular consequences, then we fetch all these genes.

Also, added the  `TranscriptQualityLabel` ("Ensembl canonical", etc.) to show to the left of the transcripts.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2466

## Deployment URL(s)
http://tr-cons-cds.review.ensembl.org

Examples with transcript consequences for more than one gene:
- http://tr-cons-cds.review.ensembl.org/entity-viewer/grch38/variant:1:227732439:rs1660720802?allele=0&view=transcript-consequences
- http://tr-cons-cds.review.ensembl.org/entity-viewer/grch38/variant:1:12167:rs112125468?allele=0&view=transcript-consequences

### DIFF
--- a/src/content/app/entity-viewer/state/api/entityViewerThoasSlice.ts
+++ b/src/content/app/entity-viewer/state/api/entityViewerThoasSlice.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { request } from 'graphql-request';
+
 import graphqlApiSlice from 'src/shared/state/api-slices/graphqlApiSlice';
 
 import config from 'config';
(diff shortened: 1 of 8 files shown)


src/content/app/entity-viewer/state/api/entityViewerThoasSlice.ts
```diff
@@ -235,15 +237,46 @@ const entityViewerThoasSlice = graphqlApiSlice.injectEndpoints({
...
+              }
+            }
+          };
+        }
+      }
+      // query: (params) => ({
+      //   url: config.coreApiUrl,
+      //   body: geneForVariantTranscriptConsequencesQuery,
+      //   variables: params
+      // })
```
imransl, line 279: Are these still needed?

azangru: Nope. Deleted.

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
```diff
@@ -145,6 +128,88 @@ const PanelHeader = (props: {
...
+              <span className={styles.label}>Gene</span>
+              {gene.symbol && (
+                <span className={styles.geneSymbol}>{gene.symbol}</span>
+              )}
+              <span className={styles.geneStableId}>{gene.stable_id}</span>
+            </div>
+          </div>
+        </div>
+        <div className={classnames(styles.right, styles.transcriptsCount)}>
+          <span>{`${transcriptConsCount} `}</span>
```
imransl, line 195: Can't this be just `{transcriptConsCount}`?

azangru: Yes; I initially had the content of this and the next span as a single string, until it turned out different parts of it needed different styling...

Removed the string interpolation.

imransl (approved):

PR #1108, opened 2024-03-31
Ev var protein seq

azangru:
## Description
Display a slice of the protein that is affected by the variant:
- contains the reference sequence directly affected by the variant in the middle, and flanking sequences
- the total length of the displayed slice is 41 amino acids
- if the protein is longer, to either side, an ellipsis is displayed
- the protein impact element, displayed under the reference sequence, can show the altered sequence itself (when it is known and not long), the number of changed amino acids (when the change is long, or when the variant is a deletion), or the word "uncertain" (when the allele sequence sent by the api contains the letter "X")

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2454

## Deployment URL(s)
http://ev-var-protein-seq.review.ensembl.org


src/shared/state/api-slices/refgetSlice.ts
```diff
@@ -46,6 +46,8 @@ const refgetApiSlice = restApiSlice.injectEndpoints({
           ? `?${queryParameters.toString()}`
           : '';
 
+        // FIXME:
+        // Use urlFor.refget helper. !!!Remember that urlFor.refget helper accepts start in Ensembl coordinates
```
azangru, line 50: This will be done in a separate PR.

src/content/app/entity-viewer/variant-view/transcript-consequences/transcript-variant-protein/TranscriptVariantProtein.tsx
```diff
@@ -0,0 +1,360 @@
...
+      {modificationTypeLabel && (
+        <span className={styles.modificationTypeLabel}>
+          {modificationTypeLabel}
+        </span>
+      )}
+    </div>
+  );
+};
+
+// FIXME: extract in a reusable component
```
azangru, line 328: The appropriate time to do this is in https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2494

PR #1111, opened 2024-04-02
Fix review deployment/cleanup

veidenberg:
## Description
Update review deployment and cleanup jobs to fix any remaining bugs.

## Deployment URL(s)
http://test-review-job.review.ensembl.org
The review app uses both ensembl-client and genome browser from review branch/namespace.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -154,17 +154,17 @@ variables:
     - kubectl config set-context --current --namespace=${DEV_NAMESPACE}
   when: manual
   script:
-    - kubectl delete ingress compara-server-ingress-${NAMESPACE} || true
-    - kubectl delete ingress ensembl-help-docs-ingress-${NAMESPACE} || true
-    - kubectl delete ingress ensembl-in-app-search-api-ingress-${NAMESPACE} || true
-    - kubectl delete ingress ensembl-thoas-ingress-${NAMESPACE} || true
-    - kubectl delete ingress ensembl-tools-api-ingress-${NAMESPACE} || true
-    - kubectl delete ingress ensembl-track-api-ingress-${NAMESPACE} || true
-    - kubectl delete ingress genome-browser-server-${NAMESPACE}-hi || true
-    - kubectl delete ingress genome-browser-server-${NAMESPACE}-lo || true
-    - kubectl delete ingress metadata-api-ingress-${NAMESPACE} || true
-    - kubectl delete ingress refget-proxy-ingress-${NAMESPACE} || true
-    - kubectl delete ingress variation-graphql-ingress-${NAMESPACE} || true
+    - kubectl delete ingress compara-server-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress ensembl-help-docs-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress ensembl-in-app-search-api-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress ensembl-thoas-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress ensembl-tools-api-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress ensembl-track-api-ingress-${CI_COMMIT_REF_SLUG} || true
+    - kubectl delete ingress genome-browser-server-${CI_COMMIT_REF_SLUG}-hi || true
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -463,11 +461,13 @@ CleanUpReview:
...
-      namespace: ${CI_COMMIT_REF_SLUG}
+  variables:
+    AGENT: ${DEV_AGENT}
+    DEV_NAMESPACE: ${DEV_NS}
+    NAMESPACE: ${CI_COMMIT_REF_SLUG}
   except:
-  - dev
-  - master
+    - dev
+    - main
```
azangru, line 470: ensembl-client doesn't have a `main` branch; it still has a `master`

veidenberg: Fixed

kamaldodiya:
> any remaining bugs.

What remaining bus are there ?

veidenberg:
> > any remaining bugs.
> 
> What remaining bus are there ?
Genome browser ingress was deployed to wrong namespace ([relevant slack thread](https://genomes-ebi.slack.com/archives/CC2RKR525/p1712049183688939)), review cleanup was not working.

kamaldodiya:
I think there is some misunderstanding here. 
Considering we are only focusing on default backend. 
1. Only ingress should go into the ensembl-dev namespace ( which is not happening in the PR) it is creating ing in the dedicated namespace ( This is my major concern)
2. I  kept CI_COMMIT_REF_SLUG so that it would be explicit that the review apps are for branches. no doubt namespace, which is assigned from CI_COMMIT_REF_SLUG technically does the same thing but having CI_COMMIT_REF_SLUG is more readable in my opinion.

veidenberg:
> I think there is some misunderstanding here. Considering we are only focusing on default backend.
> 
> 1. Only ingress should go into the ensembl-dev namespace ( which is not happening in the PR) it is creating ing in the dedicated namespace ( This is my major concern)
I agree. The [previous update](https://github.com/Ensembl/ensembl-client/pull/1110/files) moved some ingresses to dev ns, but left GB ingresses to dedicated ns. This PR intends to also move GB ingresses to dev ns (why do you say it's not happening in this PR?)
> 2. I  kept CI_COMMIT_REF_SLUG so that it would be explicit that the review apps are for branches. no doubt namespace, which is assigned from CI_COMMIT_REF_SLUG technically does the same thing but having CI_COMMIT_REF_SLUG is more readable in my opinion.
Why not to just use $CI_COMMIT_REF_SLUG then (do we need $NAMESPACE variable in this job)?

kamaldodiya:
>  why do you say it's not happening in this PR?

```
 nkhdev -n  test-review-job get ing
NAME                               CLASS   HOSTS                                ADDRESS          PORTS   AGE
ensembl-client-nginx-ingress       nginx   test-review-job.review.ensembl.org   10.252.164.146   80      26h
ensembl-client-node-ingress        nginx   test-review-job.review.ensembl.org   10.252.164.146   80      26h
genome-browser-server-ingress-hi   nginx   test-review-job.review.ensembl.org   10.252.164.146   80      26h
genome-browser-server-ingress-lo   nginx   test-review-job.review.ensembl.org   10.252.164.146   80      26h
```

> Why not to just use $CI_COMMIT_REF_SLUG then (why do we need $NAMESPACE variable in this job)?

$CI_COMMIT_REF_SLUG indicates  when deploying which branch this particular k8s resource belongs to ( explicit to the reader )
$NAMESPACE is for in which namespace the resource are going to be created. 
In this case it is same. It's just while reading the code it will make it explicit that k8s resource is for particular branch.  where as it is still deployed in DEV_NAMESPACE. 
To me personally it doesn't make much difference but generally I feel it is more readable/understandable if we keep it explicit

veidenberg:
> ```
> $ kubectl -n  test-review-job get ing
> NAME                               CLASS   HOSTS                                   PORTS   AGE
> ensembl-client-nginx-ingress       nginx   test-review-job.review.ensembl.org      80      26h
> ensembl-client-node-ingress        nginx   test-review-job.review.ensembl.org      80      26h
> genome-browser-server-ingress-hi   nginx   test-review-job.review.ensembl.org      80      26h
> genome-browser-server-ingress-lo   nginx   test-review-job.review.ensembl.org      80      26h
> ```

The above are ingresses are not for default backens but for the deployments inside the same ns (ens-client deploy+ing was created by ens-client repo ci/cd pipeline and gb resources by gb repo pipeline). The default ingresses (including gb) are created in dev ns, you can see that in the code changes.

kamaldodiya (approved):

PR #1112, opened 2024-04-03
Replace the classNames prop in PointerBox with className and set zmenu max-width

azangru:
## Description
- Minor refactoring of the `PointerBox` component and the components that use it, to replace the `classNames` property on `PointerBox` (an object) with a `className` property (a string).
- Used this `className` property after the refactoring to set a `max-width` of 400px on Zmenu on genome browser page.

**NOTE:**
PointerBox uses `width: max-content;` and `box-sizing: content-box;` (the latter due to Safari's behaviour); so the max-width of 400px on Zmenu does not include the padding.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2458

## Deployment URL(s)
http://pointerbox-classes-refactor.review.ensembl.org

### DIFF
--- a/src/shared/components/pointer-box/PointerBox.module.css
+++ b/src/shared/components/pointer-box/PointerBox.module.css
@@ -5,7 +5,6 @@
   cursor: default; /* to override cursor: pointer if e.g. parent is a button */
   position: absolute;
   width: max-content;
-  box-sizing: content-box; /* Safari doesn't seem to respect "width: max-content" if box-sizing is border-box */
   padding: var(--pointer-box-padding, 6px);
   z-index: 1000;
 }
--- a/src/shared/components/pointer-box/PointerBox.module.css
+++ b/src/shared/components/pointer-box/PointerBox.module.css
@@ -1,9 +1,13 @@
 .pointerBox {
+  --_pointer-box-color: var(--pointer-box-color, var(--color-black));
+  color: var(--pointer-box-text-color, var(--color-white));
+  background-color: var(--_pointer-box-color);
+  cursor: default; /* to override cursor: pointer if e.g. parent is a button */
   position: absolute;
-  z-index: 1000;
   width: max-content;
   box-sizing: content-box; /* Safari doesn't seem to respect "width: max-content" if box-sizing is border-box */
-  padding: 6px;
+  padding: var(--pointer-box-padding, 6px);
+  z-index: 1000;
 }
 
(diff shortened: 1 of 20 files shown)


src/shared/components/pointer-box/PointerBox.module.css
```diff
@@ -1,9 +1,12 @@
 .pointerBox {
+  --_pointer-box-color: var(--pointer-box-color, var(--color-black));
+  color: var(--pointer-box-text-color, var(--color-white));
+  background-color: var(--_pointer-box-color);
+  cursor: default; /* to override cursor: pointer if e.g. parent is a button */
   position: absolute;
-  z-index: 1000;
   width: max-content;
-  box-sizing: content-box; /* Safari doesn't seem to respect "width: max-content" if box-sizing is border-box */
```
azangru, line 5: This doesn't seem to be a concern any longer.

src/shared/components/pointer-box/PointerBox.module.css
```diff
@@ -18,6 +21,15 @@ hence it is placed in top left corner of the screen with position: fixed */
 }
 
 .pointer {
+  fill: var(--_pointer-box-color);
```
imransl, line 24: Curious why `pointer` have an underscore prepended

azangru: I've declared this variable above (line 2) as:

```css
 --_pointer-box-color: var(--pointer-box-color, var(--color-black));
```

the underscore tries to communicate that this CSS variable is sort of private (of course nothing in CSS is private; but this is the intention). It is used only within this component (you will notice that this component uses it in two places). The variable that the component exposes outside and allows its parents to modify is a prettier looking `--pointer-box-color`. So, with one variable (`--_pointer-box-color`), we either capture the value set by the parent (`--pointer-box-color`), or the default value.

The alternative would be to repeat in multiple (in this component, two) places, the following line:

```
background-color: var(--pointer-box-color, var(--color-black));

...

fill: var(--pointer-box-color, var(--color-black));
```


This is the approach popularized by Lea Verou, e.g. [here](https://lea.verou.me/blog/2021/10/custom-properties-with-defaults/):

> My preferred solution is what I call pseudo-private custom properties. You use a different property internally than the one you expose, which is set to the one you expose plus the fallback:

imransl (approved):

PR #1115, opened 2024-04-10
Add table of selected genomes to species manager

azangru:
## Description
- Added a table that lists selected genomes (in the same order species tabs are displayed)
- Each row of the table has a delete button; and the table goes into the genome removal mode when this button is pressed
- Each row of the table has a use / do not use toggle
- Added reusable components for displaying various parts of species name
- Added disabled state for the SlideToggle component

### NOTE
This PR introduces two sets of very similarly-looking components: the generic components to display parts of a species name, and more specific components to display parts of a species name in a table. I separated them out, because it seems that there is a recurring pattern in tables that does not necessarily repeat in other contexts — for example, if a particular part of a name is missing, the table would normally display a "-", whereas in other contexts, we would just skip the display of that part entirely. 

### NOT INCLUDED IN THIS PR
- Handling of the empty screen (e.g. when you deleted all species, or when after you deleted all species you go back in the browser history)
- Showing/hiding columns in the table

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2513

## Deployment URL(s)
http://species-manager-table.review.ensembl.org


src/shared/components/species-name-parts-for-table/AssemblyAccessionIdForTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const SpeciesAssemblyIdForTable = (props: Props) => {
+  const { className: classNameFromProps } = props;
+
+  const componentClasses = classNames(
+    styles.assemblyAccessionId,
+    classNameFromProps
+  );
+
+  return <AssemblyAccessionId {...props} className={componentClasses} />;
```
jyothishnt, line 38: why a different file and not in this component?

src/shared/components/species-name-parts-for-table/SpeciesScientificNameForTable.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Nothing special for the scientific name in the table.
+ * Just re-export it for convenience.
+ */
+import { ScientificName } from '../species-name-parts';
```
jyothishnt, line 21: Scientic_name is always italic. Should we add the class here?

src/shared/components/species-name-parts-for-table/SpeciesTypeForTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * In a table, species type column typically combines the type information (strain/cultivar/etc)
+ * with whether this assembly is the reference.
+ * If no type information is available and the assembly is not the reference one,
+ * we typically display a "-".
+ */
+
+const SpeciesTypeForTable = (props: Props) => {
+  const { is_reference, type: speciesType } = props;
+
```
jyothishnt, line 33: Should we do the same as above here?
const { fallback = '-' } = props;

src/content/app/species-selector/views/species-manager/selected-genomes-table/SelectedGenomesTable.tsx
```diff
@@ -14,10 +14,346 @@
...
+        <button className={styles.cancelGenomeRemoval} onClick={props.onCancel}>
+          Do not remove
+        </button>
+      </td>
+    </tr>
+  );
+};
+
+const removalWarning =
+  'Any configuration of views for this species will be lost if you remove it - do you wish to continue?';
```
jyothishnt, line 357: Any or All?

src/shared/components/species-name-parts-for-table/SpeciesScientificNameForTable.tsx
azangru: The imported `ScientificName` component has an italic font style.

src/shared/components/species-name-parts-for-table/AssemblyAccessionIdForTable.tsx
azangru: Because in tables, assembly accession id is usually rendered in a light font; but outside of tables who knows.

src/content/app/species-selector/views/species-manager/selected-genomes-table/SelectedGenomesTable.tsx
azangru: 🤷  The [XD](https://xd.adobe.com/view/2ab31a29-bcf3-4c8f-b50c-906a3949cfe2-7fe0/screen/2b49466b-a058-4484-8aea-0da2b01db7ca) says any. It's a question for Andrea, really.

src/shared/components/species-name-parts-for-table/AssemblyAccessionIdForTable.tsx
jyothishnt: No, I was saying something different I think
We could have moved `<AssemblyAccessionId/>`   as an exported component inside `AssemblyAccessionIdForTable.tsx`

azangru: I'm still a bit lost.

The plain `AssemblyAccessionId` component just extracts the accession id from the appropriate field in the data, wraps the result in a span and returns the result.

The `AssemblyAccessionIdForTable` adds a little bit of its own opinion to this, adding a CSS class with styles to be applied by default. It wants the assembly accession id to render with a lighter font by default. I've seen this used consistently in tables, so it feels appropriate to set this style by default; but I was unsure if this is an appropriate default for other contexts. Which is why I only made `AssemblyAccessionIdForTable` add this font-weight style.

Here's a diagram illustrating the intent:

image

src/content/app/species-selector/views/species-manager/selected-genomes-table/SelectedGenomesTable.tsx
```diff
@@ -14,10 +14,346 @@
...
+
+type DispatchedTableAction =
+  | EnterDeletionModeAction
+  | ExitDeletionModeAction
+  | UpdateDeletionListAction;
+
+// The function accepts id of the genome in the table row where the delete button was clicked
+const getNewDeleteModeSettings = (genomeId: string): DeletionModeSettings => {
+  return {
+    rowDeletionInitiated: genomeId,
```
imransl, line 93: This key name seems like it wants a boolean so is a bit confusing as to what it is doing

azangru: It's not a boolean though. It is a string that keeps track of which row in the table contains the Delete button that was clicked to toggle the genome removal mode on.

Are you suggesting that this field should be named differently? Could you suggest a name?

azangru: The message I wanted to convey is "genome id associated with the row that triggered the deletion mode". In `genomeRowForDeletion`, I am reading "the genome row that is marked for deletion".

imransl: Okay so how about `genomeTriggeringDeletion`?

azangru: `initialGenomeId`?

imransl: That is okay too

PR #1120, opened 2024-04-23
Allow updates to search that returns empty results

azangru:
## Description
**Problem:** Currently, if user submits a search that returns no result, he cannot submit a new search — although we are showing an editable input, the new query does not get submitted.

This PR updates the `GenomeSelectorBySearchQuery` component to always read the query from the url string, and to change the url string when the user submits a new query.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2527

## Deployment URL(s)
http://allow-search-updates.review.ensembl.org


imransl (approved):

bethflint (approved):
New funcitonality works as expected.

Searched for Bombus - no results. Search for human - see results.

PR #1123, opened 2024-04-25
Expand first transcript consequence by default

jyothishnt:
## Description
Expand first transcript consequence by default for every gene.
State stored per genome->variantId->alleleId means it will preserve the state per genome.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2507

## Deployment URL(s)
http://expand-variant-transcript.review.ensembl.org
http://expand-variant-transcript.review.ensembl.org/entity-viewer/grch38/variant:1:227732439:rs1660720802?allele=0&view=transcript-consequences

### More test variants
- `1:230709436:rs34924695` (has multiple alleles; can demonstrate that the code works when switching between alleles)

## Views affected
EV transcript consequence


src/content/app/entity-viewer/state/variant-view/transcriptConsequenceSlice.ts
```diff
@@ -0,0 +1,125 @@
...
+    };
+  }
+};
+
+export const toggleTranscriptConsequenceIds =
+  (
+    genomeId: string,
+    variantId: string,
+    alleleId: string,
+    expandIds: string[]
```
azangru, line 61: Naming. Here, and elsewhere: `expandIds` -> `expandedTranscriptIds`. `expandIds` doesn't tell us what these are the ids of. Similarly, `expandedTranscriptConseqeuenceIds` -> `expandedTranscriptIds` (a consequence itself does not have an id).

src/content/app/entity-viewer/variant-view/hooks/useExpandedTranscriptConsequence.ts
```diff
@@ -0,0 +1,46 @@
...
+    getExpandedTranscriptConseqeuenceIds(state, genomeId, variantId, alleleId)
+  );
+
+  useEffect(() => {
+    if (expandedIds === null) {
+      dispatch(
+        toggleTranscriptConsequenceIds(genomeId, variantId, alleleId, expandIds)
+      );
+    }
+  }, []);
```
azangru, line 43: don't we want same check when we switch between genomes, variants, and alleles (i.e. array of dependencies)

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
```diff
@@ -23,13 +23,16 @@ import { getReverseComplement } from 'src/shared/helpers/sequenceHelpers';
...
+import { getExpandedTranscriptConseqeuenceIds } from 'src/content/app/entity-viewer/state/variant-view/transcriptConsequenceSelectors';
+import { toggleTranscriptConsequenceIds } from 'src/content/app/entity-viewer/state/variant-view/transcriptConsequenceSlice';
 import { formatAlleleSequence } from '../variant-view-sidebar/overview/MainAccordion';
 
 import useTranscriptConsequencesData, {
   type TranscriptConsequencesData
 } from './useTranscriptConsequencesData';
+import useExpandedTranscriptConsequence from '../hooks/useExpandedTranscriptConsequence';
+
+import { type GeneInResponse } from '../../state/api/queries/variantTranscriptConsequencesQueries';
```
azangru, line 35: Style.

- If we are only importing a type, and no values, from a module, then prefer `import type { SomeType }` over `import {type SomeType}`.
- Could you move this import down, below components but above css? This is where we usually place type imports.

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
```diff
@@ -44,6 +47,12 @@ type Props = {
   activeAlleleId: string;
 };
 
+type transcriptConsequencesWithTranscriptType = Array<
```
azangru, line 50: Style.
- Begin type names with capital letter
- Don't need the word `type` in the end of the name (unless there is a clash with the name of a variable)
- When you declare types, describe the type of a single member of an array (i.e. type of a transcript consequence with transcript, not the type of the whole array). Then, when you use this type, you can just mark that you are using it as an array (i.e. instead of `const foo = (things: Things) => { ... }`, prefer `const foo = (things: Thing[]) => { ... }`).

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
```diff
@@ -52,6 +61,39 @@ const TranscriptConsequences = (props: Props) => {
...
+      dispatch(
+        toggleTranscriptConsequenceIds(
+          genomeId,
+          variantId,
+          activeAlleleId,
+          transcriptConseqeuencesToExpandByDefault
+        )
+      );
+    }
+  }, [transcriptConseqeuencesToExpandByDefault, expandedTranscriptIds]);
```
azangru, line 96: Why do you have both this hook, and the `useExpandedTranscriptConsequence` hook? Is the `useExpandedTranscriptConsequence` hook needed anymore?

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
```diff
@@ -88,24 +130,64 @@ const TranscriptConsequences = (props: Props) => {
     );
   }
 
+  type geneDataWithTranscriptConsequencesType = GeneInResponse & {
+    transcriptConsequences: transcriptConsequencesWithTranscriptType;
+  };
```
azangru, line 135: Style:
- same comments as for the type above
- plus, please don't declare types inside of functions

src/content/app/entity-viewer/variant-view/hooks/useExpandedTranscriptConsequence.ts
jyothishnt: Oh yeah missed it. But it was working when I tested.

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
jyothishnt: There is another word `transcriptConsequencesWithTranscript` so adding the suffix is better.

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
jyothishnt: When the `useExpandedTranscriptConsequence` is called first, it initialises the state. And later when we get the transcripts to expand after sorting and ordering, this hook gets called.

src/content/app/entity-viewer/variant-view/transcript-consequences/TranscriptConsequences.tsx
jyothishnt: same thing as before. I am leaving the `Type` suffix as it is.

azangru (approved):

imransl (approved):

PR #1127, opened 2024-05-02
Species manager updates

azangru:
## Description
- Two right columns of the selected genomes table in species manager stick to the right edge of the table when the table scrolls horizontally on small screens
- Removal of ticks from all checkboxes in the genome deletion mode makes the table exit the deletion mode
- Added a text field to filter the species shown in the table; as well as an "Add a species" button that takes user to the species selector home page. 
- Fixed the orange icon of external references in genome search matches table

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2510
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2535

## Deployment URL(s)
http://species-manager-updates.review.ensembl.org

### DIFF
--- a/src/content/app/species-selector/components/species-lozenge-display-selector/SpeciesLozengeDisplaySelector.module.css
+++ b/src/content/app/species-selector/components/species-lozenge-display-selector/SpeciesLozengeDisplaySelector.module.css
@@ -1,3 +1,4 @@
 .label {
+  font-weight: var(--font-weight-light);
   padding-right: 10px;
 }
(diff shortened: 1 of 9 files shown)


imransl:
@azangru Noticed while playing around with the changes that the `Add species` button and the close button seem to quite close to each other (orange arrow):

Screenshot 2024-05-08 at 10 58 17

Also, when searching for a species you now need to type twice to enter a letter from the second letter onwards. It is the same when deleting the third letter too.

azangru:
@imransl 

I've updated the styles for the "Add species" button. Andrea's directions were:

> Add icon to left edge of blue Close icon - never less than 45px.

Here is the result:

https://github.com/Ensembl/ensembl-client/assets/6834224/69113fb3-badd-48e8-bb15-f85537bc9ae5

As for the bug with the input field, I created a dedicated ticket (ENSWBSITES-2557) to address it.

imransl (approved):

PR #1129, opened 2024-05-07
Make species selector button clickable

jyothishnt:
## Description
Species selector button needs to be clickable (always) from different pages - manage, search and homepage.
Leave it active (black), but clickable.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2531

## Deployment URL(s)
http://enable-species-selector-button.review.ensembl.org

## Views affected
Species selector

### DIFF
--- a/src/header/launchbar/LaunchbarButton.tsx
+++ b/src/header/launchbar/LaunchbarButton.tsx
@@ -33,7 +33,7 @@ export type LaunchbarButtonProps = {
   icon: FunctionComponent<unknown>;
   enabled?: boolean;
   isActive?: boolean;
-  isClickable?: boolean | false;
+  isClickableWhenActive?: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


src/header/launchbar/LaunchbarButton.tsx
```diff
@@ -33,6 +33,7 @@ export type LaunchbarButtonProps = {
   icon: FunctionComponent<unknown>;
   enabled?: boolean;
   isActive?: boolean;
+  isClickable?: boolean | false;
```
azangru, line 36: - All buttons are clickable... `isClickableWhenActive`?
- `false` is a narrower type than a boolean; meaning that the type `boolean | false` is indistinguishable from type `boolean`

azangru (approved):

PR #1132, opened 2024-05-09
Refactor previously viewed entities in Entity Viewer

azangru:
## Description
- Migrated from the localStorage-based store for previously viewed objects/entities to an indexedDB-based one
- Updated the shape of previously viewed entity in entity viewer, such that it is common for both genes and variants
- Added code for recording previously viewed variants in EntityViewer

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2506

## Deployment URL(s)
http://previously-viewed-vars.review.ensembl.org


imransl:
I noticed that the previously viewed objects don't get deleted immediately if I get back to entity viewer after adding them again in species selector:


https://github.com/Ensembl/ensembl-client/assets/32512909/83229e0f-acff-472d-ac18-b47b85bec9ac

imransl (approved):

PR #1134, opened 2024-05-10
Memoize the input field to prevent bugs due to re-rendering

azangru:
## Description
**Problem:** Search input field in in-app search component was not registering the first typed letter.

**Cause:** We add a `type="search"` attribute to the search input field, in order for it to show the clear button. This transition from question button to clear button in the `ShadedInput` component caused a re-rendering of the input field, and as a result, it cleared user's input before sending it to click handler.

**Solution:** This PR wraps the search field in React's `memo` function; such that the `Input` component does not re-render when its parent (`ShadedInput`) does.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2557

## Deployment URL(s)
http://memoize-input-field.review.ensembl.org

### DIFF
--- a/src/shared/components/input/Input.tsx
+++ b/src/shared/components/input/Input.tsx
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-import { forwardRef, type InputHTMLAttributes, type ForwardedRef } from 'react';
+import {
+  forwardRef,
+  memo,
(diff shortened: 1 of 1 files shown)


imransl (approved):
Works nicely now

PR #1143, opened 2024-06-04
Show colour values in colour cards in Storybook

azangru:
## Description
During migration from SCSS to CSS, I removed the colour values (the actual hexadecimal colour codes) from the colour cards in Storybook, because, when we still used SCSS, the colour values were imported into javascript via a non-standard SCSS export; and there were no analogs for it in CSS.

It turns out Andrea looks at our Storybook for the list of colours used on the website and wants the colour values back.

This PR reads the values of the custom properties from the root DOM element on which they are registered, and displays them in the cards.

**Before:**

image

**After:**

image

### DIFF
--- a/stories/design-primitives/colours/ColourCard.tsx
+++ b/stories/design-primitives/colours/ColourCard.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { FunctionComponent } from 'react';
+import { useState, useEffect } from 'react';
 
 import styles from './ColourCard.module.css';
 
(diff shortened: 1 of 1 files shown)


stories/design-primitives/colours/ColourCard.tsx
```diff
@@ -23,7 +23,21 @@ type Props = {
   variableName: string;
 };
 
-const ColourCard: FunctionComponent<Props> = (props) => {
+const ColourCard = (props: Props) => {
```
imransl, line 26: Shouldn't we just stick with `color`?

azangru: Ha! I was hesitant to go broader with renaming 😄 

I pushed another commit, where I renamed all variables and files in this directory (and renamed the directory too). I left the 'u" in a couple of labels that appear on the screen, because this is the only bit that might matter to the viewer.

imransl (approved):

PR #1145, opened 2024-06-14
Add the variants section to the VEP input form

azangru:
## Description
- Added a VepFormVariantsSection component, which encapsulates the logic that allows user to paste variants or append a file with variants
- The section starts as collapsed and disabled. It needs a selected species to become enabled.
- The "Add variants" button and the plus button of the section can toggle its expanded/collapsed state
- If text is pasted in the textared, the file drop zone is hidden. If a file is appended, the textarea gets disabled.
- If either text is entered into textarea or a file is appended, the Add button is enabled, and the Clear button appears under it

### NOTE
The design of the VEP form has introduced a different shape of the upload button — you can see that it is both shorter than what we used previously, and has a label that looks differently (single line; the icon on the same line):

Upload button

To achieve this, I split our Upload component into two parts — the `FileDropZone` component that does the heavy work, and the `Upload` component that retains the previous shape of the button for default use cases.

### TODO (in separate PRs)
- Some (or all) aspects of the state of this section, which are currently stored in component's own state will need to be moved to redux
- Show an outline on the FileDropZone component when user focuses on it using keyboard
- Fetch example variant for a genome, and put its vcf representation in the textarea when user clicks on the vcf example button

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2593

## Deployment URL(s)
http://vep-user-input.review.ensembl.org

### DIFF
--- a/src/shared/components/upload/FileDropZoneOutline.tsx
+++ b/src/shared/components/upload/FileDropZoneOutline.tsx
@@ -1,0 +1,38 @@
+/**
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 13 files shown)


src/shared/components/upload/FileDropZoneOutline.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+import classNames from 'classnames';
+import type { ReactNode } from 'react';
+
+import styles from './FileDropZoneOutline.module.css';
+
+/**
+ * This component has a certain visual resemplance to the FileDropZone,
```
imransl, line 23: typo -> resemblance?

imransl (approved):

PR #1147, opened 2024-06-19
Vep results table

azangru:
## Description
This is an initial attempt at building the VEP results table using example data from a json dump created in [ENSWBSITES-2589](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2589).

### Also done in this PR
- Moved the `VariantColor` component from genome browser directory into the shared directory (and also renamed it to remove `u` from colour)
- Created a shared `VariantConsequence` component that combines the string label of a variant consequence with a corresponding colour.
- Updated the `ViewInAppPopup` component such that: a) it uses a button element instead of a span element to wrap around the anchor; b) it sets blue colour to children's text

### Not included in this PR
- The list of consequences only includes transcript consequences. We will have intergenic consequences as well; but currently they are filtered out.
- The transcript list does not expand or collapse. Currently, all transcripts are shown. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2571

## Deployment URL(s)
http://vep-results-table.review.ensembl.org

the table can be accessed at the `/vep/submissions/:id` path. Since there are no real submission ids yet, any string will do. For example: http://vep-results-table.review.ensembl.org/vep/submissions/1


src/content/app/tools/vep/views/vep-submission-results/VepSubmissionResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+ * - Consider pagination (should it be part of url?)
+ * - Make the table scrollable
+ */
+
+const VepSubmissionResults = () => {
+  const { currentData: vepResults } = useVepResultsQuery();
+
+  return (
+    <div className={styles.container}>
+      <div>Vep analysis</div>
```
imransl, line 46: Shouldn't VEP be all capitalised?

src/content/app/tools/vep/views/vep-submission-results/VepSubmissionResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+      </tbody>
+    </Table>
+  );
+};
+
+const VariantRow = (props: {
+  variant: VEPResultsResponse['variants'][number];
+}) => {
+  // group transcript consequences by gene
+  // const variant = updateVariant(props.variant);
```
imransl, line 87: Remove comments?

src/content/app/tools/vep/views/vep-submission-results/VepSubmissionResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+          <ColumnHead>Ref</ColumnHead>
+          <ColumnHead>Location</ColumnHead>
+          <ColumnHead>Alt allele</ColumnHead>
+          <ColumnHead>Genes</ColumnHead>
+          <ColumnHead>Transcripts</ColumnHead>
+          <ColumnHead>Consequences</ColumnHead>
+        </tr>
+      </thead>
+      <tbody>
+        {/* Use something more reliable for key than index */}
```
imransl, line 74: Can a rsID be used for this?

azangru: No; the variant name is defined (or omitted) by the user. It is not guaranteed to be unique (in a VCF input, if user does not define it, it is represented with a dot).

src/content/app/tools/vep/views/vep-submission-results/VepSubmissionResults.tsx
azangru: I think this was me just adding a placeholder string; but I'll capitalise it.

azangru: ✅

src/content/app/tools/vep/views/vep-submission-results/VepSubmissionResults.tsx
azangru: ✅

imransl (approved):

PR #1154, opened 2024-07-16
Use indexedDB to store VEP submissions, and add an initial display of VEP submissions list

azangru:
## Description
- Added vepStorageService to persist VEP submission data in indexedDB
  - Moved VEP input file from redux into indexedDB
- Added a view for listed VEP submissions
  - It displays a list of VEP submissions (without separating them into viewed and unviewed)
  - It is not yet styled appropriately
- Upon pressing the VEP form submit ("Run") button, the user is taken to the submissions list view

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2650

## Deployment URL(s)
http://vep-indexeddb.review.ensembl.org


src/content/app/tools/vep/services/vepStorageServiceConstants.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {
+  ONE_HOUR_IN_MILLISECONDS,
+  ONE_DAY_IN_MILLISECONDS
+} from 'src/shared/constants/timeConstants';
+
+export const VEP_SUBMISSIONS_STORE_NAME = 'vep-submissions';
+
+export const VEP_RESULTS_AVAILABILITY_DURATION =
+  ONE_DAY_IN_MILLISECONDS * 7 - ONE_HOUR_IN_MILLISECONDS;
```
imransl, line 25: Why is it one hour less rather than a full 7 day period?

azangru: I copied this from BLAST, where we couldn't be sure how long they kept the results (they promised up to seven days, but we didn't know when the expiration date actually happens). We can update this for VEP when we are clear on how long exactly we can / want to store results on our (server) side..

src/content/app/tools/vep/types/vepSubmission.ts
```diff
@@ -18,25 +18,55 @@ import type { CommittedItem } from 'src/content/app/species-selector/types/commi
 
 export type VepSelectedSpecies = Omit<CommittedItem, 'isEnabled'>;
 
+type ClientSideSubmissionStatus =
+  | 'NOT_SUBMITTED' // initial status of a submission while it is being prepared by the user
+  | 'SUBMITTING' // during the time between user clicking the submit button, and the submission being accepted by the backend
+  | 'FAILED_ON_SUBMISSION' // some unspecified (most likely network timeout) error happened during submission
```
imransl, line 24: How about `SUBMISSION_FAILED` instead?

src/content/app/tools/vep/views/vep-form/vep-form-variants-section/VepFormVariantsSection.tsx
```diff
@@ -60,13 +61,14 @@ const VepFormVariantsSection = () => {
     setIsExpanded(false);
   }, [selectedSpecies]);
 
+  // TODO: create a useEffect for component unmount, which will update stored VEP form
```
imransl, line 64: Should this still be done?

src/content/app/tools/vep/types/vepSubmission.ts
azangru: Couldn't this be misinterpreted as a problem for the submission to successfully run to completion (what is currently represented by the `FAILED` server-side submission status). I wanted to express that here the tools api did not accept the submission for some reason (a network failure, etc.). `SUBMISSION_REJECTED`?

azangru: `UNSUCCESSFUL_SUBMISSION`, or `SUBMISSION_UNSUCCESSFUL` was another suggested option, which can work.

PR #1155, opened 2024-07-18
add gene symbol to search results table

jyothishnt:
## Description
Add gene symbol to gene search results  table

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2604

## Deployment URL(s)
http://add-gene-symbol.review.ensembl.org

## Views affected
Gene search in species selector

### DIFF
--- a/src/shared/components/gene-search-panel/GeneSearchPanel.module.css
+++ b/src/shared/components/gene-search-panel/GeneSearchPanel.module.css
@@ -96,7 +96,7 @@
 
 .geneSymbol {
   font-weight: var(--font-weight-bold);
-  padding-right: 8px;
+  margin-right: 8px;
 }
 
 .closeButton {
--- a/src/shared/components/gene-search-panel/GeneSearchPanel.module.css
+++ b/src/shared/components/gene-search-panel/GeneSearchPanel.module.css
@@ -94,6 +94,12 @@
   cursor: pointer;
 }
 
+.geneSymbol {
+  font-weight: var(--font-weight-bold);
+  text-transform: uppercase;
+  padding-right: 8px;
+}
+
 .closeButton {
   position: absolute;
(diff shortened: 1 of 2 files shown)


src/shared/components/gene-search-panel/GeneSearchPanel.module.css
```diff
@@ -94,6 +94,12 @@
   cursor: pointer;
 }
 
+.geneSymbol {
+  font-weight: var(--font-weight-bold);
+  text-transform: uppercase;
```
azangru, line 99: Umm, we shouldn't modify the text of the symbol

src/shared/components/gene-search-panel/GeneSearchPanel.module.css
```diff
@@ -94,6 +94,12 @@
...
+  padding-right: 8px;
```
azangru, line 100: Curious: do we need to customise the space to the right of the gene symbol? It is not just a regular space between words?

jyothishnt: XD has more gap than a regular space

src/shared/components/gene-search-panel/GeneSearchPanel.module.css
azangru: > XD shows all caps though

Some symbols will have all capital letters; some will have lower-case letters (e.g. genes of C. elegans). This cannot be a design decision; it is the way appropriate scientific communities decide to name a gene.

src/shared/components/gene-search-panel/GeneSearchPanel.module.css
azangru: Ok. Let's do margin instead of the padding then.

PR #1156, opened 2024-07-19
Update listed VEP submissions; add SpeciesName component

azangru:
## Description
**VEP-related updates**
- Updated the the VEP top bar (the grey bar above the form):
  - Split it into two components (`VepFormTopBar` and `VepGenericTopBar`) — one for the VEP form page, and the other for all the rest of VEP pages. Their content is different.
  - Enabled navigation buttons, such that it is now possible to navigate between the list of unviewed submissions, and the VEP form 
- Updated the `ListedVepSubmission` component to show a spinner when the form data is being submitted to the server.
- Added some content to show inside of the box of the `ListedVepSubmission` component, and above that box (the `VepSubmissionHeader` component).

**General changes**
- Created a `SpeciesName` component to display the species name formatted according to the display option chosen in the species manager
  - This component is connected to redux
  - This component is reused in the `SpeciesLozenge`
  - What this means is that the `SpeciesLozenge` component has now also become connected to redux. Thus, I had to update the Storybook stories that use this component.

Here's what it looks like currently:
localhost_8080_vep_submissions

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2676

## Deployment URL(s)
http://vep-submissions-list.review.ensembl.org

If you want to update the "SUBMITTING" status to "SUBMITTED" to see what it looks like, you'll need to run a script in the console:

```js
vepSubmissionId = <id of your submission goes here>

dbRequest = indexedDB.open("ensembl-website");
dbRequest.onsuccess = (event) => {
  const db = event.target.result;
  const transaction = db.transaction("vep-submissions", 'readwrite')
  const objectStore = transaction.objectStore("vep-submissions")

  objectStore.get(vepSubmissionId).onsuccess = (event) => {
    updateSubmissionStatus(event.target.result, objectStore);
  };
};

updateSubmissionStatus = (data, objectStore) => {
  data.status = 'SUBMITTED';
  const requestUpdate = objectStore.put(data, vepSubmissionId);
  requestUpdate.onsuccess = (event) => {
    console.log('vep submission date updated')
  };
}
```


src/shared/components/species-name/SpeciesName.tsx
```diff
@@ -0,0 +1,160 @@
...
+const Content = (props: {
+  species: Props['species'];
+  displayOption: SpeciesNameDisplayOption;
+}) => {
+  const { species, displayOption } = props;
+  const scientificNameClasses = classNames(styles.prominent, styles.italic);
+
+  const scientificNameElement = (
+    <ScientificName
+      scientific_name={species.scientific_name}
```
imransl, line 73: Shouldn't this prop also be in camel case?

src/shared/components/species-name/SpeciesName.tsx
```diff
@@ -0,0 +1,160 @@
...
+      className={scientificNameClasses}
+    />
+  );
+
+  if (displayOption === 'common-name_assembly-name') {
+    return (
+      <>
+        <CommonName
+          common_name={species.common_name}
```
imransl, line 82: This prop as well?

src/shared/components/species-name/SpeciesName.tsx
```diff
@@ -0,0 +1,160 @@
...
+    );
+  } else if (displayOption === 'common-name_type_assembly-name') {
+    return (
+      <>
+        <CommonName
+          common_name={species.common_name}
+          fallback={scientificNameElement}
+          className={styles.prominent}
+        />
+        {!!(species.type || species.is_reference) && (
```
imransl, line 97: Does this mean that `SpeciesType` or `SpeciesReference` can be rendered even if they are not available?

stories/shared-components/species-lozenge/SpeciesLozenge.stories.tsx
```diff
@@ -15,9 +15,9 @@
  */
 
 import SpeciesLozenge from 'src/shared/components/selected-species/SpeciesLozenge';
+import WrapInRedux from './SpeciesLozengeReduxWrapper';
```
imransl, line 18: Curious about whether this a better alternative to react hooks?

src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+    submission: { status }
+  } = props;
+
+  const canGetResults = status === 'SUCCEEDED';
+
+  return (
+    <div className={styles.controls}>
+      <DeleteButton />
+      <DownloadButton onClick={noop} disabled={!canGetResults} />
+      <ButtonLink isDisabled={!canGetResults} to="/">
```
imransl, line 77: Not related to this PR but wondering about the consistency of the prop `disabled` also used as `isDisabled`. Shouldn't we just stick to one?

src/shared/components/species-name/SpeciesName.tsx
azangru: No. It comes from the selected species; selected species data comes from the backend; on the backend it is written with an underscore. Using the same case for the property name instead of changing the case allows developer to write `<ScientificName  {...species} />` rather than always having to write `<ScientificName scientificName={species.scientific_name} />`

src/shared/components/species-name/SpeciesName.tsx
azangru: Each of them should return `null` if it isn't available.

stories/shared-components/species-lozenge/SpeciesLozenge.stories.tsx
azangru: I am not sure I understand... What alternative do you have in mind?

src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader.tsx
azangru: I wondered the same when I used this component. Previously, I preferred `isDisabled`; but now I am firmly in the `disabled` camp, because this is the name of attribute of native html buttons and inputs. But, like you say, this is something for a different PR.

imransl (approved):
Discussed the PR comments with @azangru during the code review, and happy to approve.

PR #1157, opened 2024-07-29
Add polling for VEP submission status

azangru:
## Description
The client needs a way to learn about the status of a VEP submission, which it currently can only do by polling a dedicated server endpoint at a certain interval. This PR enables that. Given the flow of actions through the redux store over time:

image

this PR adds the capability to react to specific actions, such as the successful submission of the VEP form, by starting a process that would periodically query the status of the submission:

image

The solution should be capable of the following:

- Recognise that the VEP form has been submitted successfully, and start querying the status of that submission
- The querying should be done via long-polling, at a regular interval (say, once every 15 seconds)
- Picking up a second submission while the first one is still being queried should put the second submission in the global queue, such that requests are sent as follows:

```
submission1 ---delay---> submission2 ---delay---> submission1 ---delay---> submission2
```

- When a submission is deleted on the client, the polling for the status of this submission should stop
- When the app is restarted (e.g. browser refresh), all submissions saved on the client should be checked for whether any of them are not yet in a final status; and if such submissions are found, they should all be polled in a quick succession.

### Notable considerations
This PR introduces a proposal to use the action listener middleware that is part of redux-toolkit. So far, we have been using the `redux-observable` package for similar purposes; but its support has not been great. The creators of the `redux-observable` package no longer maintain it; and the current maintainer seems to be busy with other work. This adds a risk that if/when the major version of `redux` is updated, there will be an indeterminate lag before `redux-observable` is updated as well. For example, when redux was updated from version 4 to version 5, `redux-observable` had to bump their major version as well, released an rc (`3.0.0-rc.2`), and stayed on it ever since

### Alternatives
Alternative approaches are explored in [this gist](https://gist.github.com/azangru/25b4798255f5ad46311a8c1f053ee1df).

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2682

## Deployment URL(s)
http://vep-status-polling-listener.review.ensembl.org


src/content/app/tools/vep/state/vep-action-listeners/vepSubmissionStatusPolling.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    await this.pause();
+
+    const submissionId = this.queue.pop() as string;
+    const url = `${config.toolsApiBaseUrl}/vep/submissions/${submissionId}/status`;
+    const response = await fetch(url);
+    await this.handleResponse({ submissionId, response });
+
+    // this makes it a recursive function; but by avoiding the return statement, we shouldn't be leaking memory
+    this.processQueueWithDelay();
```
imransl, line 73: So `return` happens only when the queue goes down to zero?

azangru: No, the return happens implicitly right after line 73 executes. It doesn't wait for this internal `this.processQueueWithDelay();` to finish.

src/content/app/tools/vep/state/vep-action-listeners/vepSubmissionStatusPolling.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    if (status !== this.submissions[submissionId].status) {
+      this.submissions[submissionId].status = status;
+      this.reportSubmissionStatus({
+        submissionId,
+        status
+      });
+    } else {
+      // put submission id back in the queue to query its status again later
+      this.queue.unshift(submissionId);
```
imransl, line 135: Sorry I am trying get my head around this. Does this assume the queue will be FIFO?

azangru: Yes. Items enter on the left, and move towards the right.

PR #1161, opened 2024-08-08
Enable download of VEP submissions

azangru:
## Description
A VEP submission file is downloaded by directly sending a request to a back-end endpoint `/api/tools/vep/submissions/:submission_id/download`. Thus instead of a button with javascript logic, it should be sufficient to just use an html anchor tag.

Changes made in this PR:
- A `DownloadLink` component is added. Visually, the component looks identical to the DownloadButton, and therefore reuses its styles. I put the `DownloadLink.tsx` file inside of the `DownloadButton` directory. Don't know if that was a good idea or not.
- Added the `DownloadLink` component to each submission component of the submissions list.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2686

## Deployment URL(s)
http://vep-submission-download.review.ensembl.org

### DIFF
--- a/src/server/middleware/proxyMiddleware.ts
+++ b/src/server/middleware/proxyMiddleware.ts
@@ -56,7 +56,7 @@ const serverConfig = getConfigForServer();
 const createApiProxyMiddleware = () => {
   const apiProxyMiddleware = createHttpProxyMiddleware({
     pathFilter: '/api',
-    target: 'https://staging-2020.ensembl.org',
+    target: 'https://dev-2020.ensembl.org',
     changeOrigin: true,
     secure: false
   });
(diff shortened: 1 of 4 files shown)


src/server/middleware/proxyMiddleware.ts
```diff
@@ -56,7 +56,7 @@ const serverConfig = getConfigForServer();
 const createApiProxyMiddleware = () => {
   const apiProxyMiddleware = createHttpProxyMiddleware({
     pathFilter: '/api',
-    target: 'https://staging-2020.ensembl.org',
+    target: 'https://dev-2020.ensembl.org',
```
azangru, line 59: I'll revert this change before merging.

imransl (approved):

PR #1162, opened 2024-08-21
Update VEP results page

azangru:
## Description
- Added VEP submissions header to the results page
- Set up links to viewed submissions vs unviewed submissions.
- When user visits the results page for a given VEP submission, mark that submission as viewed
- Make sure the expansion/collapsing of transcripts is scoped within a gene and within an alternative allele; i.e. pressing the expand/collapse button does not affect transcripts in other rows.

https://github.com/user-attachments/assets/50587fb5-8678-45af-9d60-ac298d67d583

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2705

## Deployment URL(s)
http://vep-results-page.review.ensembl.org

### DIFF
--- a/src/content/app/tools/vep/VepPageContent.tsx
+++ b/src/content/app/tools/vep/VepPageContent.tsx
@@ -60,10 +60,10 @@ const Main = () => {
       <Route index={true} element={<VepForm />} />
       <Route
         path="unviewed-submissions"
-        element={<div>List of unviewed submissions</div>}
+        element={<VepSubmissions unviewed={true} />}
       />
       <Route path="submissions">
-        <Route index={true} element={<VepSubmissions />} />
+        <Route index={true} element={<VepSubmissions unviewed={false} />} />
         <Route path=":submissionId" element={<VepSubmissionResults />} />
       </Route>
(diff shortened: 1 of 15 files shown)


azangru:
> The VEP jobs I submit keep failing with 500 errors. Am I doing something wrong? Also noticed the edit button is broken.

The edit/rerun button still hasn't been made to work (there's a ticket in the backlog). As for the 500 errors on submission, I bet it is the tower agent being down again.

See this Slack thread for instructions on how to bring it back.

imransl:
When I view the VEP result and try to go to the region or gene, I get get an error. This is a python error though.


https://github.com/user-attachments/assets/db196f97-038f-4dab-9822-1fadfc52f8f6

azangru:
> When I view the VEP result and try to go to the region or gene, I get get an error. This is a python error though.

Is the url formed correctly? This is the only thing we should be truly concerned with in this case. From the video recording, we are seeing a problem with the track api in review deployment, which is quite a separate issue. You will see that the track api is broken in dev-2020 as well.

imransl:
@azangru Yes it seems to be broken regardless. Doesn't seem to be affected by this PR.

PR #1163, opened 2024-09-01
Add notifications support to CommunicationPanel

azangru:
## Description
Up until now, the CommunicationPanel (the panel that slides from the right upon click on the conversation icon) only showed the "Contact Us" form. This PR adds the capability to show notifications ("Messages").

The proper, long-term architecture for notifications is unclear. For now, the requirement is only to show one notification ("welcome to beta, etc."), which automatically opens the communication panel if the user hasn't seen it. If the user simply closes the panel or clicks on the "use beta" button, the notification remains accessible (shows up as a green dot next to the conversation icon). If the user chooses to "clear the message", it disappears completely.

https://github.com/user-attachments/assets/c053d40c-618f-45e2-992f-90b1ee36d548

I wonder if the `is_important: true` field should instead be something like `behavior: "noisy"`. 

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2690

## Deployment URL(s)
http://announce-rapid.review.ensembl.org

### DIFF
--- a/src/content/app/about/About.tsx
+++ b/src/content/app/about/About.tsx
@@ -41,7 +41,7 @@ import {
 import HelpMenu from 'src/content/app/help/components/help-menu/HelpMenu';
 import Breadcrumbs from 'src/shared/components/breadcrumbs/Breadcrumbs';
 import { CircleLoader } from 'src/shared/components/loader';
-import ConversationIcon from 'src/shared/components/communication-framework/ConversationIcon';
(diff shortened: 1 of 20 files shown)


imransl (approved):

azangru:
**NOTE:** this shouldn't be merged and released until roughly beta3.

bethflint:
I believe that the behaviour is as intended.

PR #1165, opened 2024-09-08
Vep results pagination

azangru:
## Description
- Added the Pagination component to the VEP results view, so that the user can navigate through a long list of results
- Added a dropdown to select the number of variants per page of VEP results (no idea why this is even needed)
- Updated the Pagination component so that it triggers a change event when user presses the Enter key rather than when he just types the number
- Updated styles of the grey bar right above the results (ToolsTopBar, VepTopBar)
- Updated the CircleLoader component by wrapping its rotating element into a static container element. This makes it possible to apply CSS transforms to CircleLoader without interfering with the rotation transform that makes the rotating element spin.
- Removed the 'Change' link to the right of SpeciesTabsSlider, by Andrea's request. As a consequence, deleted the `SpeciesTabsWrapper` component whose purpose it was to temporarily display the "Change" link to the right of species tabs until species manager was ready.

https://github.com/user-attachments/assets/975e2724-d706-4492-8995-7981335bfd28

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2623

## Deployment URL(s)
http://vep-results-pagination.review.ensembl.org


src/content/app/tools/vep/components/vep-input-summary/VepInputSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+const VepInputSummary = (props: Props) => {
+  const { submission } = props;
+
+  const componentClasses = props.className;
+
+  if (submission.inputText) {
+    return (
+      <span className={classNames(componentClasses, styles.smallLight)}>
+        Pasted data
+      </span>
```
imransl, line 44: Won't the `inputText` be shown here?

src/shared/components/pagination/Pagination.tsx
```diff
@@ -24,46 +30,75 @@ export type PaginationProps = {
...
+          disabled={lastPageNumber === 1}
+        />
+      </form>
       of {lastPageNumber}
       <ChevronButton
         direction="right"
         className={styles.showHide}
-        disabled={currentPageNumber === lastPageNumber}
-        onClick={() => props.onChange(currentPageNumber + 1)}
+        disabled={pageNumberFromProps === lastPageNumber}
```
imransl, line 100: Should this be `pageNumberFromProps >= lastPageNumber`?

azangru:
> As discussed with @azangru, the pagination breaks when switching from lower number per page to a higher number per page (e.g. 10/50 per page to 100 per page). This happens when the selected page is higher than the range of the newly selected per page option.

I tried to address this comment in a new commit

src/shared/components/pagination/Pagination.tsx
azangru: Updated.

imransl (approved):
It is working like a charm now :)

PR #1170, opened 2024-09-25
Use vep config endpoint to populate form parameters

jyothishnt:
## Description
Use new /form_config endpoint from tools api to populate client vep form

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2758

## Deployment URL(s)
http://use-vep-config-endpoint.review.ensembl.org

## Views affected
VEP


azangru:
Question not about the client, but about the backend — why is wheat showing `None`?

image

azangru (approved):

PR #1172, opened 2024-09-30
Use dedicated components for species name parts in species search results table

azangru:
## Description
Dedicated components for species name parts will make sure that the species scientific name in the species search results table is italicized.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2689

## Deployment URL(s)
http://species-name-styles.review.ensembl.org


jyothishnt (approved):
Looks good to me

PR #1177, opened 2024-10-10
Design updates for VEP submission list and results page

jyobhai:
## Description
Applying design updates from Andrea

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2773

## Deployment URL(s)
http://vep-style.review.ensembl.org

## Views affected
VEP submission list and results

### DIFF
--- a/src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader.module.css
+++ b/src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader.module.css
@@ -29,7 +29,7 @@ for ListedVepSubmission, and for the VEP results
 }
 
 .timezone {
-  font-size: 11px;
+  font-size: 10px;
   font-weight: var(--font-weight-light);
 }
 
--- a/src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
+++ b/src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
@@ -13,7 +13,8 @@
   border: 1px solid var(--color-grey);
   padding-left: var(--standard-gutter);
   padding-right: var(--standard-gutter);
-  line-height: 60px;
+  padding-top: 20px;
+  padding-bottom: 20px;
 }
 
 .bodyLoading {
@@ -69,16 +70,4 @@
 .statusRunning,
 .statusFailed {
   color: var(--color-red);
-}
-
-.speciesName [data-part="common-name"] {
-  font-size: 14px;
-}
-
-.speciesName [data-part="scientific-name"] {
-  font-size: 14px;
-}
-
-.speciesName [data-part="assembly-name"] {
-  font-size: 11px;
 }
--- a/src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
+++ b/src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
@@ -13,8 +13,8 @@
   border: 1px solid var(--color-grey);
   padding-left: var(--standard-gutter);
   padding-right: var(--standard-gutter);
-  padding-top: 20px;
-  padding-bottom: 20px;
+  padding-top: 15px;
+  padding-bottom: 15px;
 }
 
 .bodyLoading {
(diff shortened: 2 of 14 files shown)


src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
```diff
@@ -13,8 +13,7 @@
   border: 1px solid var(--color-grey);
   padding-left: var(--standard-gutter);
   padding-right: var(--standard-gutter);
-  padding-top: 24px;
-  padding-bottom: 24px;
+  line-height: 60px;
```
azangru, line 16: Spinner off-center now.

image

But line-height is not the right approach anyway. It will lead to this:

image

src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
```diff
@@ -71,3 +70,15 @@
...
+  font-size: 14px;
+}
+
+.speciesName [data-part="scientific-name"] {
+  font-size: 14px;
+}
+
+.speciesName [data-part="assembly-name"] {
+  font-size: 11px;
+}
```
azangru, line 84: I don't like the repetition of these styles in two files.

- Should the species name in the VEP submission form (i.e. right after you've selected a species) also have the same styles?
- We are probably looking at a new component here. At least one specific to VEP (and to be placed in the vep/components directory). Perhaps even one that should be common across all tools (both VEP and BLAST); but we can start with VEP.

The repetition of these styles in two files suggests that we may be dealing with a

src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader.module.css
```diff
@@ -29,7 +29,7 @@ for ListedVepSubmission, and for the VEP results
 }
 
 .timezone {
-  font-size: 11px;
+  font-size: 10px;
```
azangru, line 32: Does this have implications for styling in BLAST?

andrea-ebi:
vep looking much better
re Blast and the need to restyle - no changes needed at the moment
(note:  neither box is actually 60px in height, but close enough !)

jyobhai:
Updated with new component to display species name for VEP. Also applied the same style to vep form

src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.tsx
```diff
@@ -19,7 +19,7 @@ import classNames from 'classnames';
 import { isFailedVepSubmission } from 'src/content/app/tools/vep/utils/vepResultsAvailability';
 
 import VepSubmissionHeader from 'src/content/app/tools/vep/components/vep-submission-header/VepSubmissionHeader';
-import SpeciesName from 'src/shared/components/species-name/SpeciesName';
+import { VepSpeciesName } from '../../../components/vep-species-name/VepSpeciesName';
```
azangru, line 22: `../../../` -> `src/blah`

src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
```diff
@@ -28,6 +28,7 @@
   grid-template-columns: [species-name] 300px [input-summary] 1fr [status] auto;
   column-gap: var(--standard-gutter);
   align-items: center;
+  line-height: 30px;
```
azangru, line 31: Is this for setting the height of the submission box? `line-height` is such a non-obvious way of achieving this; and it will have undesirable consequences if text wraps into several lines. Wasn't min-height capable of achieving the desired effect?

azangru:
image

We probably want to align this label as well?

src/content/app/tools/vep/views/vep-submissions/listed-vep-submission/ListedVepSubmission.module.css
azangru: I can see that you've updated this line; but before your latest commit had a chance to deploy, I saw a good example demonstrating why using line-height would be bad.

https://github.com/user-attachments/assets/e966ffda-76e5-4e0b-96a0-975c2af0c66c

jyothishnt:
> We probably want to align this button as well?

I dont see any issue here after the recent code update :-)

azangru (approved):

PR #1184, opened 2024-11-05
Update social media links

azangru:
## Description
- Extract a group of social media links into a dedicated component
- Use the SocialMediaLinks component in the footer on the Home page; on the About page; and in the ContactUs tab of the communications panel
- Added a base CSS layer, so that the CSS rules that are declared as part of the components layer do not get overridden by the regular (non-layered) styles.

Also
- Updated a label on the VEP page (`Variant effect predictor` -> `Variant Effect Predictor`)

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2809
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2559

## Deployment URL(s)
http://social-media-links.review.ensembl.org


src/styles/base.css
azangru, line 1: Notice: This file contains the styles that were extracted from main.css, such that main.css can be used for importing styles and assigning them to CSS layers.

imransl (approved):

PR #1191, opened 2024-12-18
Update cleanup job

veidenberg:
## Description
Update review app cleanup job

## Deployment URL(s)
http://ens2465.review.ensembl.org

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -113,7 +113,6 @@ variables:
 # Template for deployment to "stable" (non-review) environments that have static assets on a mounted drive
 # (therefore, does not need to deploy a static assets container)
 .deploy:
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -263,6 +261,7 @@ LiveFallback:
 # Publish static assets
 Pub::LiveFallback:
   extends: .publish_assets
+  stage: deploy-live
```
azangru, line 264: Both `Pub::Live` and `Pub::LiveFallback` extend the `.publish_assets` template, which has `stage: deploy-live` defined in it.

If you want to move the stage from the template to the jobs, as you did for Live and LiveFallback, then you should probably delete the stage from the `.publish_assets` template and add it to `Pub::Live`. Or leave `deploy-live` stage in `.publish_assets` and remove it from here 🤷

PR #1193, opened 2025-01-05
Update icons

azangru:
## Description
- Change the species home icon in the selected genomes table in species manager from the home icon to the species selector icon
- Update the VEP icon in the launchbar

**Before:**

image

**After:**

image

### Also
Added a new component (called it `VerticallyCenteredCellContent`) to help with the vertical centring of `inline-block` or `inline-flex` items in table cells.

Here is an image demonstrating the principle. Table cells are capable of running the `vertical-align: middle` CSS rule; but it does not align the inline elements vertically (see the first and the third row in the screenshot of a table below). However, `vertical-align: middle` would align a block element vertically. Thus, if I wrap an inline-block element in a block element, and change the `display` property of that wrapper to `flex`, I can properly align its contents vertically. See the second and the fourth row, as well as the bottom row with the buttons on the screenshot below): 

image

Detailed view. Without wrapping (content box of table's `td` shown in green):

image

With wrapping in a div with `display: flex` and `align-items: center`:

image

## Related JIRA Issue(s)
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2711
- https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2841

## Deployment URL(s)
http://update-icons.review.ensembl.org


shashankbrgowda:
LGTM

- Change home icon to species selector icon in species selector manage page
- Vep icon has been updated
- Improvements to vertically align elements inside table cell

PR #1195, opened 2025-01-20
Add epigenome activity visualisation, and switch from mock data to data fetched from backend apis

azangru:
## Description
- Added visualisation of epigenome activity

image

- Switched from using mock data to fetching the data from the backend apis
  - Deleted the mock data files that are no longer in use
- Refactoring
  - Moved the logic for dealing with ids (genome id, assembly accession id, etc.) into a dedicated context. In the future, the provider of this context will also be responsible for checking whether the genome with given uuid exists, and fetching it if it does.
  - Moved the logic for dealing with epigenomes (fetching the list of base genomes, fetching epigenome metadata, filtering epigenomes based on the selected metadata) into a dedicated context.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2859

## Deployment URL(s)


src/content/app/regulatory-activity-viewer/components/region-activity-section/RegionActivitySection.tsx
```diff
@@ -91,15 +60,31 @@ const RegionActivitySection = (props: Props) => {
...
+            <>
+              <RegionActivitySectionImage
+                width={width}
+                regionOverviewData={preparedData.regionOverviewData}
+                featureTracks={preparedData.featureTracksData}
+                start={preparedData.location.start}
+                end={preparedData.location.end}
+              />
+              {/* A temporary vertical separator component below */}
+              <div style={{ margin: '1rem 0' }} />
```
imransl, line 74: Shouldn't this div be closed rather than be self-enclosing?

azangru: It's a temporary div; it was easier for me to just add the vertical spacing inline rather than add a css class to something. I expect it to go away when the dust settles. It doesn't have any content, which is why I made it self-closing. Most of self-closing tags aren't valid html, but are valid jsx; react just expands it to `<div style="something"></div>`.

src/content/app/regulatory-activity-viewer/contexts/ActivityViewerIdContextProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const contextValue = {
+    activeGenomeId,
+    assemblyAccessionId: assemblyAccessionId ?? null,
+    assemblyName: assemblyName ?? null,
+    // Below is a test location. Later on, we will read it from the url or from an input element
+    location: {
+      regionName: '17',
+      start: 58190566,
+      end: 58290566 // <-- 100kB slice
+      // end: 59190566 // <-- 1MB slice
```
imransl, line 61: Is this for future reference or can it be removed?

src/content/app/regulatory-activity-viewer/components/activity-viewer-app-bar/ActivityViewerAppBar.tsx
shashankbrgowda, line 1: If I disable species in species-selector manage page, tracks are still rendered. Is this expected ?

src/content/app/regulatory-activity-viewer/helpers/filter-epigenomes/filterEpigenomes.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  if (!selectedMetadataDimensions.length) {
+    return [];
+  }
+
+  // Iterate over the groups of enabled filters (the selected metadata dimensions),
+  // and gradually narrow down the list of epigenomes by applying filters from each subsequent group
+  // to the results of the previous pass
+
+  let finalFilteredEpigenomes: typeof epigenomes = epigenomes;
```
shashankbrgowda, line 76: define type `Epigenome[]` ?

src/content/app/regulatory-activity-viewer/contexts/ActivityViewerIdContextProvider.tsx
azangru: For future reference, to easily switch to a 1MB slice when the backend api becomes more performant.

src/content/app/regulatory-activity-viewer/contexts/ActivityViewerEpigenomesContext.ts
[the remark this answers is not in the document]
azangru: You are right that we do use redux (modernised and improved via redux-toolkit) for global state management. Since we use redux, we also went for redux-toolkit query for data fetching. Ideally, I should be able to grab all the data I need from the redux state using appropriate selectors. In practice though, there doesn't seem to be (or I couldn't find in the documentation) a way to create a custom selector to freely access different pieces of data loaded via redux-toolkit query.

Consider what is happening to get various data related to epigenomes (in `ActivityViewerEpigenomesContextProvider`). I need access to `baseEpigenomes`, and I also need access to `epigenomeMetadataDimensions`, each of which is fetched from a separate endpoint. Only then I can apply user-selected filters (`epigenomeSelectionCriteria`, stored in redux) and `epigenomeCombiningDimensions` (also stored in redux) to finally create a list of epigenomes both to show in the table and to fetch the data for in `EpigenomesActivityImage`. It would have been great if I could do all that in a custom selector; but I couldn't find a way to achieve this. So instead, not to duplicate the computations involved in the combination of the different sources of data and the creation of the final list of epigenomes, I do that in a single place, and then inject the results through the context into the components that use it.

src/content/app/regulatory-activity-viewer/state/api/activityViewerApiSlice.ts
```diff
@@ -16,53 +16,109 @@
...
+      })
+    }),
+    epigenomesActivity: builder.query<
+      EpigenomeActivityResponse,
+      EpigenomesActivityRequestParams
+    >({
+      queryFn: async (params, _, __, baseQuery) => {
+        const { assemblyAccessionId, epigenomeIds, locations, regionName } =
+          params;
+        const url = `${config.regulationApiBaseUrl}/epigenomes/v0.1/activity_viewer/assembly/${assemblyAccessionId}`;
```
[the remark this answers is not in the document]
azangru: Yes; this api endpoint is currently very slow; but there isn't anything I can do about it other than show a spinner. It is being written in rust by the regulation team, who promise to improve its performance (it is unacceptably slow for them either). It is doing something CPU-intensive (could be just reading bigwig files, or could be munging the data read from them). Regulation team says that its slowness has something in particular to do with the async runtime; but I did not inquire further.

src/content/app/regulatory-activity-viewer/components/activity-viewer-app-bar/ActivityViewerAppBar.tsx
azangru: So far, yes.

src/content/app/regulatory-activity-viewer/helpers/filter-epigenomes/filterEpigenomes.ts
azangru: Updated.

src/content/app/regulatory-activity-viewer/state/api/activityViewerApiSlice.ts
```diff
@@ -16,53 +16,109 @@
...
-          'tests/fixtures/activity-viewer/mockRegionOverviewDense'
-        );
-        const data = module.default;
-
-        return { data };
+    regionOverview: builder.query<OverviewRegion, RegionOverviewRequestParams>({
+      query: (params) => {
+        const { assemblyName, location } = params;
+        return {
+          url: `${config.regulationApiBaseUrl}/region-of-interest/v0.1.0/${assemblyName}?location=${location}`
```
imransl, line 61: Shouldn't the version (`v0.1.01) be a constant or an environment variable?

src/content/app/regulatory-activity-viewer/hooks/useEpigenomes.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { ActivityViewerEpigenomesContext } from 'src/content/app/regulatory-activity-viewer/contexts/ActivityViewerEpigenomesContext';
+
+const useEpigenomes = () => {
+  const context = useContext(ActivityViewerEpigenomesContext);
+
+  if (!context) {
+    throw new Error(
+      'useEpigenomes must be used within ActivityViewerEpigenomesContext'
+    );
```
imransl, line 27: Is this error caught anywhere or supposed to completely break things?

src/content/app/regulatory-activity-viewer/state/api/activityViewerApiSlice.ts
azangru: Mmm, it isn't our constant/variable; it's regulation's. And we don't reuse it anywhere other than in this url string. I think it's too early to worry about those; we will see how regulation develop their apis and how they will agree with Platforms to expose their endpoints.

src/content/app/regulatory-activity-viewer/hooks/useEpigenomes.ts
azangru: It is supposed to break things. It can only occur if the developer tries to use the `useEpigenomes` hook somewhere other than in the activity viewer. It is supposed to explode in that case.

imransl (approved):

PR #1196, opened 2025-01-20
Swap add species and close button in species home page

shashankbrgowda:
## Description
The "Add a species" button and the "Close" button have been swapped to improve the user experience in species home page.

image

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2703

## Deployment URL(s)
http://species-page-style-updates.review.ensembl.org

### DIFF
--- a/src/content/app/species/SpeciesPage.module.css
+++ b/src/content/app/species/SpeciesPage.module.css
@@ -4,7 +4,7 @@
   height: 100%;
 }
 
-.topbar {
+.topbarContents {
   display: grid;
   grid-template-columns: 1fr auto;
   align-items: center;
@@ -13,7 +13,7 @@
   max-width: 1380px;
 }
 
-.topbar_withoutSidebarNavigation {
+.topbarWithSidebarClosed {
   flex-grow: 1;
   margin-right: 39px;
 }
--- a/src/content/app/species/SpeciesPage.module.css
+++ b/src/content/app/species/SpeciesPage.module.css
@@ -11,11 +11,14 @@
   justify-items: start;
   line-height: 1;
   max-width: 1380px;
-  padding-left: 30px;
 }
 
 .addSpeciesButton {
   --add-button-icon-color: var(--color-green);
   --add-button-label-color: var(--color-black);
-  padding-right: 35px;
+  padding-left: var(--double-standard-gutter);
+}
+
+.closeButton {
+  padding-right: var(--double-standard-gutter);
 }
--- a/src/content/app/species/SpeciesPage.module.css
+++ b/src/content/app/species/SpeciesPage.module.css
@@ -14,9 +14,7 @@
 }
 
 .topbar_withoutSidebarNavigation {
-  display: grid;
-  align-items: center;
-  width: 100%;
+  flex-grow: 1;
   margin-right: 39px;
 }
 
--- a/src/content/app/species/SpeciesPage.module.css
+++ b/src/content/app/species/SpeciesPage.module.css
@@ -20,5 +20,5 @@
 }
 
 .closeButton {
-  padding-right: var(--double-standard-gutter);
+  padding-right: 55px;
 }
--- a/src/content/app/species/SpeciesPage.module.css
+++ b/src/content/app/species/SpeciesPage.module.css
@@ -13,6 +13,13 @@
   max-width: 1380px;
 }
 
+.topbar_withoutSidebarNavigation {
+  display: grid;
+  align-items: center;
+  width: 100%;
+  margin-right: 39px;
+}
+
 .addSpeciesButton {
   --add-button-icon-color: var(--color-green);
   --add-button-label-color: var(--color-black);
--- a/src/content/app/species/SpeciesPageContent.tsx
+++ b/src/content/app/species/SpeciesPageContent.tsx
@@ -48,6 +48,7 @@ import { BreakpointWidth } from 'src/global/globalConfig';
 import type { CommittedItem } from 'src/content/app/species-selector/types/committedItem';
 
 import styles from './SpeciesPage.module.css';
+import classNames from 'classnames';
 
 type SpeciesPageParams = {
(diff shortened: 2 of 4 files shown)


azangru:
A couple of things I noticed:

- The close button changes its wobbles a bit horizontally when the width of the window is changed
- At one extreme, the close button is shifted slightly to the left of the trashcan icon; at the other extreme, slightly to the right

https://github.com/user-attachments/assets/a79cbe05-52eb-4844-8c33-b723c764f32f

Looking at the image in the [ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2703), I think the intention is that the close button is right-aligned with the trashcan icon below it (we can ask Andrea to clarify). I wonder if this is achievable.

src/shared/components/layout/StandardAppLayout.module.css
```diff
@@ -41,7 +41,7 @@
 
 .main {
   height: 100%;
-  overflow: auto;
```
azangru, line 44: We still need `overflow:auto` though! There won't be a scrollbar without it.

azangru:
Now we have two problems, because a stable scrollbar gutter will always take space

1. On the species selector page, the content of the page shifted to the left, and 55px is no longer the right padding for the close button:

image

2. There is now an extra padding on the genome browser page:

image

For problem 1, we could either shift main content to the right or the close button to the left at the distance of scroll gutter (e.g. `padding-right: calc(20px - var(--scrollbar-width));`).

Problem 2 is harder to address. Either the StandardAppLayout needs a prop that would tell it whether it should set a stable scrollbar gutter in the main area, or the `browserImageWrapper` div needs a negative right margin. Let's discuss what is more appropriate.

src/content/app/species/SpeciesPage.module.css
```diff
@@ -11,11 +11,21 @@
   justify-items: start;
   line-height: 1;
   max-width: 1380px;
-  padding-left: 30px;
+}
+
+.topbar_withoutSidebarNavigation {
+  display: grid;
+  align-items: center;
```
azangru, line 18: I don't think you are using any grid features for this element; so the two CSS rules above are probably unnecessary.

src/content/app/species/SpeciesPageContent.tsx
```diff
@@ -136,25 +137,31 @@ const SpeciesPageContent = () => {
...
+const TopBar = (props: { isSidebarOpen: boolean }) => {
   const navigate = useNavigate();
 
   const returnToSpeciesSelector = () => {
     navigate(urlFor.speciesSelector());
   };
 
+  const topbarClassnames = classNames({
+    [styles.topbar_withoutSidebarNavigation]: !props.isSidebarOpen
+  });
```
azangru, line 149: Here is how the component is rendered into the DOM:

image

Notice:
- There is an empty `class` attribute on the top element of your component. While this isn't bad, it is unnecessary. The reason the empty attribute appears is because the `topbarClassnames` equals to an empty string when `props.isSidebarOpen` is `true`. You only seem to be using a single class name; so you don't even need the `classNames` function for help; it could have just been:

```
const topbarClassname = !props.isSidebarOpen ? styles.someClass : undefined
```

- Since you added a wrapper div, this div, logically, now represents our top bar. So the `.topbar` class on its child (line 153) should probably be renamed to something like `.topbarContents`
- Regarding the `styles.topbar_withoutSidebarNavigation` class on line 148:
  - Our convention in this codebase is to use camelCase for classNames. So, no underscore. The reason there is an underscore in CSS class names of the `StandardAppLayout` component is because that component is very old, created before we decided to standardise on camelCase. We should probably update its class names too; but for new CSS that we write, let's stick to camel case.
  - As for the name of that class, since species pages knows nothing about sidebar navigation, perhaps `topbarFullWidth` or `topbarWithSidebarClosed` would be more appropriate?

src/content/app/species/SpeciesPage.module.css
```diff
@@ -11,11 +11,19 @@
   justify-items: start;
   line-height: 1;
   max-width: 1380px;
-  padding-left: 30px;
+}
+
+.topbar_withoutSidebarNavigation {
+  flex-grow: 1;
+  margin-right: 39px;
```
azangru, line 18: I've been puzzled by the padding on this element:

image

until I realised that it is defined by the parent component:

https://github.com/Ensembl/ensembl-client/blob/main/src/shared/components/layout/StandardAppLayout.module.css#L33-L35

Whatever I was thinking when I wrote those styles for `StandardAppLayout`, that was a bad idea — a parent component shouldn't unceremoniously alter styles of its children. Could you please move [this padding-right](https://github.com/Ensembl/ensembl-client/blob/main/src/shared/components/layout/StandardAppLayout.module.css#L34) from `.topbar_withoutSidebarNavigation > div` to `.topbar_withoutSidebarNavigation` class in `StandardAppLayout.module.css`?

src/content/app/species/SpeciesPageContent.tsx
shashankbrgowda: - Wrapper div with empty class was intentional, I used classNames to keep it consistent with codebase as it does't cause any harm.
- Regarding naming convention, I've followed what's already in StandardAppLayout

 Thanks for letting me know about this, will update accordingly

src/content/app/species/SpeciesPage.module.css
shashankbrgowda: I didn't understand why that padding was added in StandardAppLayout so didn't wanted to change it as it could affect other components

azangru (approved):

veidenberg (approved):
Looks good to me.

Btw in Safari the assembly counter pushes the bin icon away (leaving all right-side icons misaligned) but that's already live and outside the scope of this PR (cc @azangru).
Screenshot

imransl (approved):

PR #1198, opened 2025-01-29
Add flag to toggle species selector modal in VEP tools

shashankbrgowda:
## Description
Currently VEP tools uses react router to display the species selector modal. When VEP species selector modal is open and if we come back to VEP tools from any other page (ex: species selector page), clicking close button of VEP species selector modal navigates to previous page in history instead of closing the modal.

Change this by adding a flag in redux which toggles the VEP species selector modal.

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2891

## Deployment URL(s)
http://vep-species-selector-toggle.review.ensembl.org


shashankbrgowda:
While this pr fixes vep species selector close bug, it doesn't address below scenario
Help page -> vep tools page -> click on vep species selector -> click on browser back button
then, It shouldn't take us to previous page in history (help page) but, It should take us back to vep tools page.

So, vep species selector should be a separate page instead of modal.

This issue exists in blast species selector as well so should be fixed.

PR #1201, opened 2025-02-11
Read the `location` url parameter to display regulatory activity data

azangru:
## Description
- Read the `location` url parameter to display regulatory activity data.
- Added an interstitial screen — very primitive so far; just shows example locations for reference human and mouse.

https://github.com/user-attachments/assets/56692c3e-9f4f-4d1f-a5ca-20924e95d87a

### Other changes
- Reshuffled the buttons ("Configure") now moved to the right, and changed text in one of them (`"Epigenomes"` replaced with `"Table"`)
- Added the "Configure" button to the sidebar (in the "Region activity" view)
- If user selects a location in RegionOverview while the bottom of the page shows the epigenomes table, the view automatically changes to regulatory activity visualisation

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2901

## Deployment URL(s)


veidenberg (approved):
Everything works.
Note: the `<int> epigenomes` text at the top-right table corner often gets cut off when not full-screen (but there is a scrollbar at the bottom to hint that so should to be ok).

PR #1202, opened 2025-02-19
Add gene expression level indicators with mock data

azangru:
## Description
- Added gene expression indicators (to the right of epigenome activity visualisation), currently displaying mock data. In the future, this should only show up if a gene has been selected.
-  Added the `genome` namespace to url paths (e.g. `/activity-viewer/genome/grch38`). This does not quite follow url convention used for other pages, where genome id follows immediately after the name of the 'app'; but Garth feels strongly that regulatory data should be accessible not just by providing a genome id; and for now, this seems to be a decent compromise
- If no filters are selected, regulatory activity viewer now shows the full list of epigenomes instead of showing none
- When the "Configure" button in the main section of the page is pressed, the sidebar now changes its contents to 'region activity' view. This might come as a surprise to users; but at the same time, is a way to show the applied filters. We shall see if this is the right behaviour.

https://github.com/user-attachments/assets/d0758736-41b1-46ae-b5c6-838c69962870

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2902

## Deployment URL(s)


veidenberg (approved):
Looks good to me.

PR #1206, opened 2025-04-14
Update epigenome labelling in regulatory activity viewer

azangru:
## Description
Some exploratory changes to the regulatory activity viewer. Andrea is not directing the changes at the moment; so they have been suggested by Garth. The understanding is that this is still an early-stage prototype that will be used in user testing and to inform design decisions.

The changes include:
1) Added text labels for tracks to the left of colour labels
2) Added a table overlay that expands over epigenome activity tracks to provide annotation for those tracks in-place (without having to switch to the view to the table of epigenomes)
  - the default view of the table contains only the three columns for the dimensions that are used to sort the epigenomes
  - the table can be expanded further to show the remaining dimensions

**Before:**

staging-2020 ensembl org_activity-viewer_genome_grch38_location=17_58190566-58290566

**After:**

https://github.com/user-attachments/assets/bd275dc4-37f1-49d8-a336-d815e4ae0d9a

## Deployment URL(s)

### DIFF
--- a/src/content/app/regulatory-activity-viewer/RegulatoryActivityViewer.module.css
+++ b/src/content/app/regulatory-activity-viewer/RegulatoryActivityViewer.module.css
@@ -4,3 +4,8 @@
   align-items: start;
   grid-template-rows: auto 1fr;
 }
+
+.mainContentContainer {
+  container-name: main-content-container;
+  container-type: size;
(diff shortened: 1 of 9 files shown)


veidenberg (approved):
Looks good to me.

PR #1210, opened 2025-05-09
Make release a mandatory part of selected species; migrate species stored in indexeddb

azangru:
## Description
Added an indexedDB migration to add release information to the genomes that the user has already selected.

During the migration, the script will query the `explain` endpoint of the metadata api for every saved genome, and update the stored genomes with the fetched release data to ensure that from now on selected species always have information about their release. 

### Steps for testing
The easiest way to test this PR is to run the code locally. Here are the steps that should confirm that the code is working:

1. Currently, species matches retrieved from the search api already contain release information. So, in order to simulate life before releases, strip release information off the selected species. For example, since we are testing indexedDB, you can do so before saving data to the db. One way of doing so is the following:

In `speciesSelectorStorageService`, look up function `saveMultipleSelectedSpecies`, and change the code inside the `for (const species of speciesList)` block:

```ts
    for (const species of speciesList) {
      // strip release information away
      const x = [...Object.entries(species)].reduce((acc, [key, val]) => {
        if (key !== 'release') {
          acc[key] = val;
        }
        return acc
      }, {})
      await transaction.store.put(x, species.genome_id);
    }
```

2. Open `indexeddb-service.ts`. Make sure `DB_VERSION` is set to 6.
3. Start the site, delete indexedDB (Chrome -> Dev tools -> Application -> Storage -> Clear site data)
4. Reload the page, and select several species
5. Take a look at the saved species in indexedDB (Chrome -> Dev tools -> Application -> IndexedDB -> ensembl-website -> selected-species). Notice that they do not have any release information.

image

6. In `indexeddb-service.ts`, change `DB_VERSION` to 7.
7. Refresh the browser. Take another look at the saved species in IndexedDB. Your previously saved species should now have release information 

image

## Related JIRA Issue(s)
This is part of support for integrated releases.

## Deployment URL(s)
http://update-indexeddb-species.review.ensembl.org — **BUT** there is nothing new that could be seen on that deployment

### DIFF
--- a/src/content/app/species-selector/state/species-selector-general-slice/speciesSelectorGeneralSlice.ts
+++ b/src/content/app/species-selector/state/species-selector-general-slice/speciesSelectorGeneralSlice.ts
@@ -122,6 +122,7 @@ const prepareSelectedSpeciesForCommit = (
       accession_id: species.assembly.accession_id,
       name: species.assembly.name
     },
+    release: species.release,
     is_reference: species.is_reference,
     type: species.type,
     isEnabled: true
--- a/src/content/app/species-selector/state/speciesSelectorEpics.ts
+++ b/src/content/app/species-selector/state/speciesSelectorEpics.ts
@@ -144,6 +144,7 @@ const buildCommittedItemFromBriefGenomeSummary = (
       accession_id: genome.assembly.accession_id,
       name: genome.assembly.name
     },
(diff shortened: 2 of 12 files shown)


veidenberg (approved):
Looks good to me (didn't do manual testing but went through the code and confirmed the automatic migration tests passed).

PR #1212, opened 2025-05-13
Enable navigation outside of the initial location in regulatory activity viewer

azangru:
## Description
This PR adds an ability to navigate beyond the boundaries of the initially rendered slice in the region overview image. For comparison, with the code that is currently in the main branch (see http://activity-viewer.review.ensembl.org), it is possible only to zoom in into the originally rendered slice. Notice, too, that the location url parameter in the main branch does not change during the zooming or the dragging, as opposed to this PR.

**Before**

https://github.com/user-attachments/assets/f36b4eca-0325-4fa0-bb6c-5c6f3a3747a6

**After**

https://github.com/user-attachments/assets/9436915f-48cb-4eae-846e-874945aa8e7b

## Overview of the approach

1. Region overview image is rendered as an svg
  - Is this a good idea? Not sure. There are definitely too many DOM elements that the svg generates; and it becomes expecially concerning at a far-enough zoom-out.

2. Since the image is rendered as an svg, we cannot afford recreating the DOM at every animation frame when viewport contents are dragged left/right. Therefore, the image renders data for three viewports:
  - One visible viewport in the middle
  - One viewport worth of svg elements to the left of the svg border
  - One viewport worth of svg elements to the right of the svg border

image

3. The data for the region overview image is retrieved over the course of two subsequent requests:
  - First, data is requested just for the viewport (plus a full viewport worth of data to the right, and a full viewport worth of data to the left). This is what is needed for the immediate rendering of the image.
  - Second, a request is sent to fetch annotation data for the full region (i.e. the full chromosome). The response is several megabytes of gzipped json.
  - Is this a good idea? I don't know. On the one hand, requesting several megabytes worth of data is wasteful; but on the other, this makes it possible to easily zoom out.

4. Location in the url now changes whenever the contents of the region overview image are dragged or zoomed in/out.

## Deployment URL(s)
http://activity-viewer-region-nav.review.ensembl.org


veidenberg (approved):
Works as described, also with throttled network.
One small glitch is that one can zoom/pan past the end of chromosome (but not past the start of chromosome), but not a blocker for merging.

PR #1221, opened 2025-07-03
Fix the display of the 404 screen for missing gene

azangru:
## Description

**Problem:** When the Entity Viewer loads a page for a non-existing genome, it shows a mostly blank page, without the expected error screen:

image

**Causes:**

Although code was written previously to detect and display the 'gene not found' errors, it must have rotted over time and react version updates.

- The most proximate cause was that the function for getting the data based on which it could check the absence of the gene was not called, because it was skipped if the url hasn't changed ([link](https://github.com/Ensembl/ensembl-client/blob/c9b929985f70fd3cb117a927d0359137e9374f9f/src/content/app/entity-viewer/hooks/useEntityViewerUrlChecks.ts#L58)).
  - **CAUTION!** This check was introduced to "Fix infinite requests to Thoas on broken network". As it turns out, without it, the requests will indeed be infinite.
- **Next problem:** When the error screen appeared, the "Continue" button didn't work due to a race condition. Specifically, the "Continue" button should take user to the interstitial screen; and in order to do so, it needs to remove the active entity id from the state ([link](https://github.com/Ensembl/ensembl-client/blob/dac57dac6134f6ce83f0053d4c4bd14c94c3e073/src/content/app/entity-viewer/EntityViewerForGene.tsx#L63-L66)). However, there is another hook with a function that immediately restores the active entity id from the url ([link](https://github.com/Ensembl/ensembl-client/blob/dac57dac6134f6ce83f0053d4c4bd14c94c3e073/src/content/app/entity-viewer/EntityViewer.tsx#L148-L153)). This is a race condition, which probably used to work out fine before, but likely became apparent after an update to React.
- There is a very old file entity-viewer-storage-service, which stores some EntityViewer data in localStorage. Since we should be storing data in indexedDB anyway, this service should be rewritten later; but in the meantime, it became obvious that it cannot properly delete data from nested objects, i.e. if it has saved something like below:

```
{
  foo: {
    bar: 'a',
    baz: 'b'
  }
}
```

and I want to remove the `baz` field from the stored object, the service will fail to do so.

## Deployment URL(s)
http://gene-404-page.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/hooks/useEntityViewerUrlChecks.ts
+++ b/src/content/app/entity-viewer/hooks/useEntityViewerUrlChecks.ts
@@ -16,7 +16,6 @@
 
 import useEntityViewerIds from 'src/content/app/entity-viewer/hooks/useEntityViewerIds';
 import { useGeneSummaryQuery } from 'src/content/app/entity-viewer/state/api/entityViewerThoasSlice';
-import usePrevious from 'src/shared/hooks/usePrevious';
 
 /**
  * A hook for validating individual parts of the url.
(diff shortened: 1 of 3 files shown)


andrea-ebi:
Suggest updating the help text (black)
Use Find a gene or the example gene link

PR #1290, opened 2026-04-10
Enable pagination for species search

azangru:
## Description
- [x] Add pagination component to species search results table shown on species search page, on BLAST page and on VEP page
- [x] Replace client-side logic for sorting of species search results with appropriate call to the server
- [x] Add per-page dropdown (do we really have to?)
- [x] Figure out styles for how to position pagination components 

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/ENSWBSITES-2992

## Deployment URL(s)
http://paginated-species-search.review.ensembl.org


src/content/app/species-selector/components/genome-selector-by-search-query/GenomeSelectorBySearchQuery.tsx
```diff
@@ -49,35 +65,39 @@ type Props = {
...
-    sortRule,
-    changeSortRule
-  } = useSelectableGenomesTable({
-    genomes: currentData?.matches ?? [],
-    selectedGenomes: committedSpecies,
-    filterQuery
+  let pageNumber = parseInt(pageString);
+  if (!pageNumber) {
+    pageNumber = 1;
+  }
```
shashankbrgowda: can be one liner  (nitpick)

```
const pageNumber = parseInt(searchParams.get('page') || '1');
```

src/content/app/species-selector/components/genome-selector-by-search-query/GenomeSelectorBySearchQuery.tsx
```diff
@@ -96,32 +116,80 @@ const GenomeSelectorBySearchQuery = (props: Props) => {
...
+    const newPageParam = new URLSearchParams(searchParams);
+    newPageParam.set('page', `${page}`);
+    setSearchParams(newPageParam, { replace: true });
+  };
+
+  const onResultsPerPageChange = (perPage: number) => {
+    const newSearchParams = new URLSearchParams(searchParams);
+    newSearchParams.set('per_page', `${perPage}`);
+    newSearchParams.set('page', `${pageNumber}`);
+    setSearchParams(newSearchParams, { replace: true });
```
shashankbrgowda, line 125: Should we maintain page history through browser nav?

src/content/app/species-selector/helpers/genomeSearchHelpers.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import {
+  defaultPerPageOptions,
+  defaultPerPageValue
+} from 'src/shared/components/pagination/PerPage';
+
+import type { SortOrder } from 'src/shared/types/sort-order';
+import type { SortProps } from 'src/content/app/species-selector/components/species-search-results-table/SpeciesSearchResultsTable';
+
+export const DEFAULT_NUM_RESULTS_PER_PAGE = defaultPerPageValue;
```
shashankbrgowda, line 25: Why create again ?

src/content/app/species-selector/state/species-selector-api-slice/speciesSelectorApiSlice.ts
```diff
@@ -51,23 +59,61 @@ const speciesSelectorApiSlice = restApiSlice.injectEndpoints({
...
     })
   })
 });
 
-export const {
-  useGetPopularSpeciesQuery,
-  useLazyGetSpeciesSearchResultsQuery,
-  useLazyGetGenomesBySpeciesTaxonomyIdQuery
-} = speciesSelectorApiSlice;
+const prepareGenomeSearchParams = (
```
shashankbrgowda, line 77: should this function and getSpeciesSearchLastPageNumber be moved to some helper from api slice file ?

shashankbrgowda (approved):
Added some minor comments otherwise looks good.

Also, download button is not present in vep and blast species search page. Not sure if its expected

src/content/app/species-selector/components/genome-selector-by-search-query/GenomeSelectorBySearchQuery.tsx
azangru: I don't think so. Also, if we did that, I think that would break the "close" button, which works as one step back in history.

src/content/app/species-selector/helpers/genomeSearchHelpers.ts
azangru: The pagination dropdown is weird. I don't know where the defaults for the numbers of items per page should go, and whether they should be the same across all Ensembl pages, or tailored specifically to some views.

In this case, I didn't want genome/species search components to pull the constant directly from a shared `PerPage` component; so I moved it to a file with reusable logic, which happened to be this helpers file.

I am open to suggestions for what a better place for this constant would be.

src/content/app/species-selector/state/species-selector-api-slice/speciesSelectorApiSlice.ts
azangru: I don't think `prepareGenomeSearchParams` should move to another file; it is not used anywhere outside of this module.

As for `getSpeciesSearchLastPageNumber`, I do not have a strong preference one way or the other. I put it in this slice file, because of how closely it's tied to the shape of the data that comes back from the api (the  `SpeciesSearchResponse`). But I also wouldn't mind moving it to the `genomeSearchHelpers.ts` file. Do you have a particular preference either way?

shashankbrgowda: We have url search param related code in urlHelpers.. Is this one different ?

shashankbrgowda:
Don't we need download button in vep and blast species search page ?

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
```diff
@@ -100,40 +116,99 @@ const BlastSpeciesSelector = (
...
+      sortBy,
+      sortOrder
+    });
+    setSearchResultsPage(initialSearchPage);
+    setSearchResultsPerPage(perPage);
+  };
+
+  const sortRule = getSortRule(sortBy, sortOrder);
+
+  const onSortRuleChange = useCallback(
```
veidenberg, line 137: Picked up by agent and confirmed in review app: column sorting in species results table does not work as expected in BLAST and VEP pages. `onSortRuleChange` (also in [VEPSpeciesSelector](https://github.com/Ensembl/ensembl-client/pull/1290/changes#diff-539b2e0c186b253c1db7d52456fed61d444c58856a95f12c60b27cf40b3104fdR146)) only updates local sorting state, while server-side sorting requests happen in `onSearchSubmit`/`onPageNumberChange`/`onResultsPerPageChange` handlers. So the sorting icon changes, but the row order itself will not update until clicking on the table page navigation button.

veidenberg (changes requested):

azangru:
> Don't we need download button in vep and blast species search page ?

Regarding the download button, I think that downloading a list of genomes makes sense when user is on the species selector page, and is exploring our list of species. When user is in the context of a tool (BLAST or VEP), I think he will be focused on running the tool rather than on exploring what genomes we've got; and the download button in this context feels unnecessary.

I've asked Beth what she thinks. She agrees that she wouldn't expect a download button in the context of BLAST or VEP.

I'm going to omit it from the tools pages, and see if anyone complains.

src/content/app/tools/blast/components/blast-species-selector/BlastSpeciesSelector.tsx
azangru: Good agent 😃 

I've refactored the code for BLAST and VEP species selectors to move from lazy queries to eager queries. Locally, I am not seeing the bug anymore. See if this has improved matters.

veidenberg (approved):

PR #1294, opened 2026-04-21
EV for transcript: add routing, and store active view in redux

azangru:
## Description
- Connect the view within the main content area to the url (default view vs transcript view)
- Store the selected view in redux such that last visited view can be restored after navigation between genomes or between apps

## Related JIRA Issue(s)
Part of https://embl.atlassian.net/browse/ENSWBSITES-2993

## Deployment URL(s)
http://ev-transcript-as-object.review.ensembl.org
Specific example: http://ev-transcript-as-object.review.ensembl.org/entity-viewer/grch38/transcript:ENST00000380152

### DIFF
--- a/src/content/app/entity-viewer/state/entityViewerReducer.ts
+++ b/src/content/app/entity-viewer/state/entityViewerReducer.ts
@@ -19,6 +19,7 @@ import { combineReducers } from 'redux';
 import entityViewerGeneralReducer from './general/entityViewerGeneralSlice';
 import entityViewerSidebarReducer from './sidebar/entityViewerSidebarSlice';
 import entityViewerGeneViewReducer from './gene-view/entityViewerGeneViewReducer';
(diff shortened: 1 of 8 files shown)


veidenberg (approved):

shashankbrgowda (approved):
why do we need to maintain state of transcript function view while navigating between apps,  can't we open transcript view by default ? 

Otherwise looks good

azangru:
@shashankbrgowda 

> why do we need to maintain state of transcript function view while navigating between apps

I am not sure about specifically maintaining state while navigating between apps (although this is what the Entity Viewer does for genes); but it probably is a good idea to maintain the view when switching from one genome to another within Entity Viewer. And once you start remembering the view when switching from one genome to another, remembering the view when switching from one app to another comes for free.

I agree that this makes the code more complex and confusing, and would be super happy to remove it; but I doubt we'll be allowed to. I'll keep this functionality if only for consistency with the behaviour of Entity Viewer for gene.

PR #1295, opened 2026-04-24
Add protein diagram to entity viewer for transcript

azangru:
## Description
Add content for the "Transcript function" tab of entity viewer for transcript.

The content of the tab has been lifted from the gene view, where there are many transcripts and thus, potentially, many proteins. This is why the imports have names such as `ProteinListItem` or `ProteinListItemInfo`.

## Related JIRA Issue(s)
Part of https://embl.atlassian.net/browse/ENSWBSITES-2993

## Deployment URL(s)
http://ev-transcript-as-object.review.ensembl.org

Example: http://ev-transcript-as-object.review.ensembl.org/entity-viewer/grch38/transcript:ENST00000380152?view=protein


shashankbrgowda (approved):
Looks good

veidenberg (approved):

PR #1300, opened 2026-05-13
Add genome group categories view in home page

shashankbrgowda:
## Description
This PR adds genome group categories tabs along with 42 popular species in the home page.

Enabled in non production environments only

## Deployment URL(s)
http://genome-groups-ui.review.ensembl.org


src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.tsx
```diff
@@ -46,12 +70,44 @@ const SpeciesSelectorMainView = () => {
...
+        {activeTab === popularSpeciesTab && <PopularSpeciesList />}
+        {activeGenomeGroupCategory && (
+          <GenomeGroupsList category={activeGenomeGroupCategory} />
+        )}
       </div>
     </div>
   );
 };
 
+const SpeciesSelectorTabs = (props: {
```
azangru, line 89: It isn't the first time I see a tabs component being recreated. We need to review (and possibly rebuild), and reuse our `Tabs` component. Not in this PR; but we really ought to do this.

src/content/app/species-selector/components/genome-groups/GenomeGroups.tsx
azangru, line 1: `Groups` is already plural; I wonder if we need a `List` in the name as well.

src/content/app/species-selector/components/genome-groups-list/GenomeGroupsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+type Props = {
+  category: GenomeGroupCategory;
+};
+
+const GenomeGroupsList = (props: Props) => {
+  const { category } = props;
+
+  if (!category.groups.length) {
+    return null;
+  }
```
azangru: Do we need this block at all?

src/content/app/species-selector/components/genome-groups-list/GenomeGroupsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <div className={styles.container}>
+      {category.groups.map((group) => (
+        <GenomeGroupItem key={group.group_id} group={group} />
+      ))}
+    </div>
+  );
+};
+
+const GenomeGroupItem = (props: { group: GenomeGroup }) => {
```
azangru: Just `GenomeGroup`? If it clashes with the type name, you could rename the type on import (`import type { GenomeGroup as GenomeGroupType}`

src/content/app/species-selector/components/genome-groups-list/GenomeGroupsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { group } = props;
+
+  return (
+    <section className={styles.group}>
+      <h2 className={styles.groupTitle}>{group.title}</h2>
```
azangru: Why are we doing an h2 here?

Our headings situation on this site is so cursed. Do we even have an h1?

src/content/app/species-selector/components/genome-groups-list/GenomeGroupsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+      ))}
+    </div>
+  );
+};
+
+const GenomeGroupItem = (props: { group: GenomeGroup }) => {
+  const { group } = props;
+
+  return (
+    <section className={styles.group}>
```
azangru: Why is it a `section`?

src/content/app/species-selector/components/genome-groups/GenomeGroups.tsx
shashankbrgowda: Named it similar to Popular species list for consistency. Shall I change it to GenomeGroups ?

azangru: Yeah; I thought that was the reason :-) I would change it to GenomeGroups. The reason there is the word List in popular species is because the word species is the same in both singular and plural.

src/content/app/species-selector/components/genome-groups/GenomeGroups.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <div className={styles.container}>
+      {category.groups.map((group) => (
+        <GenomeGroupItem key={group.group_id} group={group} />
+      ))}
+    </div>
+  );
+};
+
+const GenomeGroupItem = (props: { group: GenomeGroupType }) => {
```
azangru: `GenomeGroupItem ` -> `GenomeGroup`

azangru (approved):
Ok.

We should remember to do something about the tabs that keep reappearing in various screens.

PR #1303, opened 2026-05-21
Implement transcript search in species selector, Interstitial and sidebar

shashankbrgowda:
## Description
- Add Find a gene, transcript and variant links in species selector home page
- Add transcript search tab in species selector feature search page
- Add transcript search tab in Interstitial pages
- Add transcript search panel in species selector, genome browser, entity viewer sidebar modals
- Links to genome browser transcript focus view and transcript entity viewer page

## Related JIRA Issue(s)
[ENSWBSITES-2980](https://embl.atlassian.net/browse/ENSWBSITES-2980)

## Deployment URL(s)
http://transcript-search-ui.review.ensembl.org


src/shared/components/search-match/TranscriptSearchMatches.tsx
```diff
@@ -0,0 +1,196 @@
...
+    setShouldShowTooltip(!shouldShowTooltip);
+
+    if (app === 'entityViewer') {
+      analyticsTracking.trackEvent({
+        category: `${app}_${mode}_search`,
+        action: 'select_link',
+        label: match.stable_id,
+        value: position
+      });
+    }
```
azangru: Let's maybe not do this?

These analytics events are added very inconsistently. There was a time, years ago, when Lucy wanted to collect analytics for various parts of the UI; but there was never a clear purpose stated (like, let's run an experiment for a while to decide whether a particular UI pattern is working or whether it is better than an alternative).

Our attempt at adding the analytics code was rather ugly. Since this is a shared component, unless anyone tells us anything about analytics, let's not add it.

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -50,5 +134,51 @@ const searchApiSlice = restApiSlice.injectEndpoints({
   })
 });
 
+const transcriptSearchApiSlice = graphqlApiSlice.injectEndpoints({
```
azangru: `transcriptSearchApiSlice` -> `graphqlSearchApiSlice`

You may want to expand it in the future to search for other features (e.g. search genes via Thoas instead of the search hub / solr)

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.module.css
```diff
@@ -56,3 +66,27 @@
...
+  row-gap: 10px;
+}
+
+.searchLinks {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  row-gap: 8px;
+  column-gap: 30px;
+  margin-left: 20px;
```
azangru, line 81: add `line-heigh: 1` to avoid the vertical shift when navigating from the main screen to the feature search screen

Currently, the height of this element is different.

Main screen:

Image

Feature search screen:

Image

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.module.css
```diff
@@ -56,3 +66,28 @@
...
+.searchFieldWithLinks {
+  display: flex;
+  flex-direction: column;
+  row-gap: 10px;
+}
+
+.searchLinks {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
```
azangru: are there scenarios in which we wrap the links?

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.module.css
```diff
@@ -56,3 +66,28 @@
...
+  line-height: 1;
+  row-gap: 8px;
```
azangru: are there scenarios in which the row-gap is needed?

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.module.css
```diff
@@ -56,3 +66,28 @@
...
+  column-gap: 30px;
+  margin-left: 20px;
+}
+
+.activeSearchLink {
```
azangru: Why is this class needed? It looks like you are applying it to the element that already has a black colour.

src/content/app/species-selector/views/species-selector-search-view/SpeciesSelectorSearchView.tsx
```diff
@@ -84,24 +90,31 @@ const SpeciesSelectorSearchView = () => {
...
   const isGeneSearchMode = activeSearchMode === 'gene';
+  const isTranscriptSearchMode = activeSearchMode === 'transcript';
   const isVariantSearchMode = activeSearchMode === 'variant';
 
   useEffect(() => {
-    if (isGeneSearchMode) {
-      dispatch(setGeneQuery(queryFromParams));
-    } else if (isVariantSearchMode) {
-      dispatch(setVariantQuery(queryFromParams));
+    if (queryFromParams !== null) {
```
azangru, line 100: Technically, the contents of this block should only run if query from params exists and is different from query from redux, right?

src/content/app/species-selector/components/species-selector-app-bar/SpeciesSelectorAppBar.tsx
```diff
@@ -44,7 +42,9 @@ export const PlaceholderMessage = () => (
 
 const isInSearchMode = (pathname: string) => {
```
azangru: I wonder why we've been doing this rather than passing a prop during [route matching](https://github.com/Ensembl/ensembl-client/blob/b149d90773f4eb864737ee42ca5d6e74fcf11b8c/src/content/app/species-selector/SpeciesSelector.tsx#L34-L36)

src/shared/components/feature-search-results/FeatureSearchResults.module.css
```diff
@@ -79,11 +79,21 @@
...
+
 .geneSymbol {
   font-weight: var(--font-weight-bold);
   margin-right: 8px;
 }
 
+.transcriptSymbol {
+  font-weight: var(--font-weight-bold);
+  margin-right: 8px;
+}
```
azangru, line 94: I wonder if we even need the transcript symbol. The user only searches by transcript id; and there will only at most be a single match; and transcript symbols are non-standard.

src/content/app/species-selector/views/species-selector-search-view/SpeciesSelectorSearchView.tsx
shashankbrgowda: It should only check for query from params

src/content/app/species-selector/views/species-selector-search-view/SpeciesSelectorSearchView.tsx
```diff
@@ -166,8 +191,9 @@ const SpeciesSelectorSearchView = () => {
         {isFeatureSearchMode(activeSearchMode) && (
           <>
             <FeatureSearchForm
+              key={activeSearchMode}
```
azangru, line 194: Why is the `key` necessary here?

src/shared/components/feature-search-results/FeatureSearchResults.tsx
```diff
@@ -159,22 +159,105 @@ const FeatureSearchTableRows = (props: {
...
+      url: urlForEntityViewer
+    }
+  };
+
+  return (
+    <>
+      <span
+        className={styles.transcriptMatch}
+        ref={setAnchorElement}
+        onClick={handleClick}
```
azangru, line 238: This suggests that this `span` should be a `button`

src/shared/components/feature-search-results/FeatureSearchResults.tsx
```diff
@@ -287,13 +374,17 @@ const GeneRecord = (props: { match: SearchMatch; species: CommittedItem }) => {
 
   return (
     <>
-      <span className={styles.geneMatch} ref={anchorRef} onClick={handleClick}>
+      <span
+        className={styles.geneMatch}
+        ref={setAnchorElement}
+        onClick={handleClick}
```
azangru, line 381: Same as above.

src/shared/components/search-match/TranscriptSearchMatches.tsx
```diff
@@ -0,0 +1,172 @@
...
+    null
+  );
+
+  const onMatchClick = () => {
+    setShouldShowTooltip(!shouldShowTooltip);
+  };
+
+  const onAppClick = () => {
+    if (app === 'genomeBrowser') {
+      dispatch(changeHighlightedTrackId(''));
```
azangru: I forget — does this do anything?

src/shared/components/search-match/TranscriptSearchMatches.tsx
```diff
@@ -0,0 +1,172 @@
...
+  };
+
+  const searchMatchAnchorClass = classNames(
+    styles.searchMatchAnchor,
+    styles[`${mode}SearchMatchAnchor`]
+  );
+
+  return (
+    <>
+      <div className={styles.searchMatch} onClick={onMatchClick}>
```
azangru: `onclick` suggests that this should be a button

wait... there already is a button inside

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -27,6 +29,88 @@ type SearchParams = {
...
+  const transcriptSearch = response.transcript_search;
+
+  const matches = (transcriptSearch?.matches ?? []).flatMap((match) => {
+    const { stable_id: stableId, unversioned_stable_id: unversionedStableId } =
+      match;
+    const genomeId =
+      match.genome_id ??
+      (params.genome_ids.length === 1 ? params.genome_ids[0] : undefined);
+
+    if (!stableId || !unversionedStableId || !genomeId) {
```
azangru: Why would any of this happen? Why wouldn't a transcript match above have a genome id?

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -50,5 +134,51 @@ const searchApiSlice = restApiSlice.injectEndpoints({
...
+        }
+
+        return {
+          data: parseTranscriptSearchResults(
+            data as TranscriptSearchQueryResponse,
+            params
+          )
+        };
+      }
+    })
```
azangru, line 69: I am a little concerned here about the return type – normally, I think, it would have both the data and the error; but here, you are separating one from the other. I cannot remember if this is ok or not quite.

Plus the empty checking. Should it really be responsibility of this endpoint to specifically handle this edge case? Shouldn't it be the responsibility of the function that calls the endpoint to make sure there is always something in the genome ids and in the query? And what happens if an empty query like this is sent to Thoas?

I would simplify this to:

```
    {
      query: (params) => ({
        url: config.coreApiUrl,
        body: transcriptSearchQuery,
        variables: params
      }),
      transformResponse(response: AppropriateResponseType) {
        return parseTranscriptSearchResults(response)
      }
    ),
```

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -27,6 +29,88 @@ type SearchParams = {
   per_page: number;
 };
 
+type TranscriptSearchQueryResponse = {
+  transcript_search?: {
+    meta?: SearchResults['meta'];
```
azangru: Why is this optional?

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -27,6 +29,88 @@ type SearchParams = {
...
+    matches?: TranscriptSearchMatchResponse[];
```
azangru: Why is this optional?

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -27,6 +29,88 @@ type SearchParams = {
   per_page: number;
 };
 
+type TranscriptSearchQueryResponse = {
+  transcript_search?: {
```
azangru: Why is this optional? If there is an error with the query, the field would be null; but it won't be undefined.

src/shared/state/api-slices/searchApiSlice.ts
shashankbrgowda: It returns either error or data. I can simplify the endpoint with query
https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#queryfn-function-return-value

> normally, I think, it would have both the data and the error;

src/shared/components/feature-search-results/FeatureSearchResults.tsx
shashankbrgowda: This approach has been followed in existing variant and gene search as well, which was taken from existing code

src/shared/components/feature-search-results/FeatureSearchResults.tsx
shashankbrgowda: This approach has been followed in existing variant and gene search as well, which was taken from existing code

src/shared/components/search-match/TranscriptSearchMatches.tsx
shashankbrgowda: Taken from gene search match. I think it clears any highlighted track before navigating to gb focus track

src/shared/components/feature-search-results/FeatureSearchResults.module.css
shashankbrgowda: I was checking Jbrowse and gene/transcript name/symbol seems to be standard. 

It shouldn't cause any issues because undefined values are handled.

src/content/app/species-selector/views/species-selector-search-view/SpeciesSelectorSearchView.tsx
shashankbrgowda: Yes, because there is an issue when I clear the input (ShadedInput component)

src/shared/components/feature-search-results/FeatureSearchResults.tsx
azangru: Yes; the existing code is wrong here. We were not very particular in the early years about which html elements we used. Let's fix this here and in the existing code.

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -74,39 +74,20 @@ const transcriptSearchQuery = gql`
...
-      }
-    ];
-  });
+  const matches = transcriptSearch.matches.map((match) => ({
+    type: 'Transcript' as const,
+    stable_id: match.stable_id,
+    unversioned_stable_id: match.unversioned_stable_id,
+    symbol: match.symbol,
+    genome_id: match.genome_id
+  }));
```
azangru: What's happening in this map? Are you just adding the `type` field and keeping the same values of the rest of the fields?

This could be expressed as 

```
{
  ...match,
  type: 'Transcript'
}
```

But! The schema of `Transcript` in Thoas already includes the `type` field, which, for transcripts, is exactly the string `Transcript`. We might include the `type` field in the query. This would simplify the code on the client side. Or, if you feel that it is unwise to send the string "Transcript" over the wire, use the spread option above.

src/shared/state/api-slices/searchApiSlice.ts
```diff
@@ -30,32 +30,32 @@ type SearchParams = {
...
 type TranscriptSearchMatchResponse = {
-  stable_id?: string | null;
-  unversioned_stable_id?: string | null;
-  symbol?: string | null;
-  genome_id?: string | null;
+  stable_id: string;
+  unversioned_stable_id: string;
+  symbol: string | null;
+  genome_id: string;
 };
```
azangru: Since we are dealing with thoas, we should derive this type from the core api types (see [here](https://github.com/Ensembl/ensembl-client/blob/main/src/shared/types/core-api/transcript.ts)). It might also make sense to move the query, along with the type of the response, in its own file. See [examples](https://github.com/Ensembl/ensembl-client/blob/main/src/content/app/entity-viewer/state/api/queries/defaultGeneQuery.ts) in EntityViewer

src/shared/state/api-slices/searchApiSlice.ts
shashankbrgowda: Simplified with spread operator

src/shared/components/feature-search-results/FeatureSearchResults.module.css
azangru: It isn't the handling of undefined that I was concerned about; but the display of the transcript symbol at all, and the prominence that this symbol now has in a search result, being the first part of a label. We have never before shown transcript symbols anywhere on the site — not in the genome browser (including the drawer or the zmenus), not in the entity viewer. If you consider the old site, then on the [gene page](https://www.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000139618;r=13:32315086-32400268;t=ENST00000380152), the most prominent element is the gene symbol; whereas on the [transcript page](https://www.ensembl.org/Homo_sapiens/Transcript/Summary?db=core;g=ENSG00000139618;r=13:32315086-32400268;t=ENST00000380152), the most prominent element is the transcript id. Besides, while we can search gene by gene symbol, and can get multiple matches in search results, in which case it becomes useful to show additional information, such as gene stable id — transcripts can only be searched by exact match against the transcript id. So, there will be only one match in transcript search results, and the user will already have to know this id in order to find the transcript.

I would just remove the transcript symbol for consistency with the rest of the site. But if we are so indecisive and hesitant that we can't just agree to remove it here, then at the very least, it should be displayed after the transcript id, not before it.

src/shared/components/feature-search-results/FeatureSearchResults.tsx
```diff
@@ -237,10 +237,10 @@ const TranscriptRecord = (props: {
         ref={setAnchorElement}
         onClick={handleClick}
       >
+        <span className={styles.stableId}>{match.stable_id}</span>
```
azangru, line 240: And same in the sidebar:

Image

azangru (approved):
All right; let's go.

Please check the `changeHighlightedTrackId` after you merge; and if it isn't used anywhere, please remove that part of redux state.

PR #1305, opened 2026-05-27
Fix focus object during species change in Genome Browser

azangru:
### Bug description
- Have several species selected in the genome browser
- In one of the species, view a gene, then, by modifying the url, view a transcript (this will cause the page to reload); then, open the list of previously viewed features, and go to a previously viewed gene. The point is to keep more than one focus objects for a genome in memory
- Now, switch to a different species, then back to the initial one. You will notice that the species that you started with will sometimes focus on a gene, and other times on a transcript, in what seems to be a random order

https://github.com/user-attachments/assets/f359f406-d27f-460c-ab5e-4ab1e77943bc

### What this PR does
My hypothesis is that something is wrong in the `useBrowserRouting` hook. It is awfully written; and hopefully, we will get to refactoring it in the future. Meanwhile, I have updated the dependency array of a `useEffect` to include all variables that it captures from the outer scope. At first, this caused an infinite re-render of the component, because every time the hook runs, it updates the redux state, which causes a re-render, which triggers the hook. So I added a check before updating the redux state. 

## Deployment URL(s)
http://fix-species-switch-in-gb.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
+++ b/src/content/app/genome-browser/contexts/genome-browser-ids-context/useGenomeBrowserUrlValidator.ts
@@ -178,18 +178,21 @@ const useGenomeBrowserUrlValidator = (params: Params) => {
     stateRef.current = currentState;
   }, [currentState]);
 
-  const runAfterValidation = (fn: () => unknown) => {
-    if (
-      !hasFocusObjectToValidate ||
-      (stateRef.current.doneValidating &&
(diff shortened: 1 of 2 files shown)


shashankbrgowda (commented):
I couldn't reproduce this issue while switching between focus gene track of two genomes in live site.

The issue I reported was when switching between gene and transcript focus tracks.

shashankbrgowda (approved):
Switching between gene and transcript focus tracks is working fine as expected

PR #1307, opened 2026-05-28
Fix bugs with messaging to the genome browser

azangru:
## Description
This PR attempts to fix several bugs that are currently visible on staging.

1. The "dancing genome browser" bug, which is a regression introduced in [this PR](https://github.com/Ensembl/ensembl-client/pull/1305). It results from a desynchronisation between the location that the the client instructs the genome browser to go to, and the browser's confirmation that it has reached that location. It manifests as follows: user moves the genome browser to the side, or zooms in/out; the genome browser sends back a message to update the url; and this starts an infinite cycle of the client sending to the genome browser an older url location, the genome browser navigating to it, and sending back a message that updates the url, and so on, and so forth.

https://github.com/user-attachments/assets/95a24770-0594-49e1-a252-08936b2203d0

The fix for this bug is a slight modification of the `useBrowserRouting` hook. The regression was caused by adding all useEffect's dependencies to the dependency array, which increased the rate at which that hook was called, and broke the conditional logic somewhat. Note that the `useGenomeBrowserPosition` hook has also been updated slightly (one function wrapped in a useCallback and moved above a useEffect) to satisfy the linter.

2. Transcripts within gene become hidden after navigation from different app to genome browser.

This is a fairly old bug, which might have gotten worse recently. It presents as follows: you visit the genome browser for the first time (e.g. refresh the page) and confirm that you can see the gene with an enabled transcript. Then you navigate to a different page; and then you return to the genome browser — and you notice that the transcript in the gene has become disabled (visibly, only a faintly drawn canonical transcript).

https://github.com/user-attachments/assets/3a3e7b0f-72b5-452a-b30c-7b93253a010c

The cause of the bug was a race condition. On a return visit to the genome browser page, the focus object information is already present in redux, but the genome browser module has not loaded yet; and the useFocusTrack hooks sends a message prematurely, which the genome browser misses. See line 259 in the `useFocusTrack` file in the diff.

ALSO:
The PR updates the genome browser launchbar button, so that when user returns to the genome browser by pressing it, the generated url is not just the `/genome-browser`; but `/genome-browser/:genome?focus=focus-object&location=location`.  

## Deployment URL(s)
http://fix-gb-position.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/hooks/useBrowserRouting.ts
+++ b/src/content/app/genome-browser/hooks/useBrowserRouting.ts
@@ -34,6 +34,7 @@ import {
   parseFocusIdFromUrl
 } from 'src/shared/helpers/focusObjectHelpers';
 
+import { setActiveGenomeId } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
 import { setDataFromUrlAndSave } from 'src/content/app/genome-browser/state/browser-general/browserGeneralSlice';
 import { fetchFocusObject } from 'src/content/app/genome-browser/state/focus-object/focusObjectSlice';
 
@@ -113,6 +114,7 @@ const useBrowserRouting = () => {
         location: chrLocation ? getChrLocationStr(chrLocation) : null
       };
 
+      dispatch(setActiveGenomeId(genomeId));
       // Consider it as first render when we change genome
       firstRenderRef.current = true;
 
@@ -132,6 +134,7 @@ const useBrowserRouting = () => {
       allActiveFocusObjectIds,
       allChrLocations,
       allCommittedSpecies,
+      dispatch,
       navigate
     ]
   );
@@ -152,8 +155,8 @@ const useBrowserRouting = () => {
     previousChrLocation: ChrLocation | null | undefined;
   }) => {
     return (
-      genomeId === previousGenomeId ||
-      focusObjectIdInUrl === previousFocusObjectIdInUrl ||
+      genomeId === previousGenomeId &&
+      focusObjectIdInUrl === previousFocusObjectIdInUrl &&
       JSON.stringify(chrLocation) === JSON.stringify(previousChrLocation)
     );
   };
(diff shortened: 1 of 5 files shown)


shashankbrgowda:
@azangru Genome browser auto pan issue is fixed but there is still an issue with the state when we toggle between the species lozenges  :(


https://github.com/user-attachments/assets/b1e317f9-b4d3-49be-991a-382337213558

image

shashankbrgowda:
It's hard to reproduce as the issue disappears when I refresh the page. But you can notice state issue when we zoom in one species and toggle between using species lozenge, the new zoomed location is not updated properly.

azangru:
@shashankbrgowda I think you are showing a broken genome browser (as in, the rust/webassembly bit), not the part that ensembl-client is responsible for. Which is why it is so hard to reproduce.

azangru:
> But you can notice state issue when we zoom in one species and toggle between using species lozenge, the new zoomed location is not updated properly.

You are right about this one. Argh!

It's different from what you show in the screen capture though.

Actually, wait; this is genome browser's shenanigans as well.

src/content/app/genome-browser/hooks/useBrowserRouting.ts
```diff
@@ -131,11 +132,32 @@ const useBrowserRouting = () => {
...
+  }: {
+    genomeId: string | undefined;
+    previousGenomeId: string | undefined;
+    focusObjectIdInUrl: string | null;
+    previousFocusObjectIdInUrl: string | null;
+    chrLocation: ChrLocation | null | undefined;
+    previousChrLocation: ChrLocation | null | undefined;
+  }) => {
+    return (
+      genomeId === previousGenomeId ||
```
shashankbrgowda: Shouldn't it be && instead of || ?

and comparing values of ChrLocation makes it clearer instead of using JSON.stringify(...)

veidenberg (approved):

shashankbrgowda (approved):

PR #1308, opened 2026-06-03
Link genome groups to genome results table

shashankbrgowda:
## Description
This PR provides navigation from genome groups (External projects and Ensembl genome collection) to sepcies selector results view. 

## Deployment URL(s)
http://extended-search-results.review.ensembl.org

### DIFF
--- a/src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
+++ b/src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
@@ -11,7 +11,10 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
+  align-self: flex-start;
   row-gap: 6px;
+  color: inherit;
+  text-decoration: none;
 }
 
 .groupTitle {
--- a/src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
+++ b/src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
@@ -11,10 +11,7 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
-  align-self: flex-start;
   row-gap: 6px;
-  color: inherit;
-  text-decoration: none;
 }
 
 .groupTitle {
@@ -35,4 +32,5 @@
   align-items: center;
   column-gap: 8px;
   color: var(--color-blue);
+  text-decoration: none;
 }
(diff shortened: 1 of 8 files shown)


src/content/app/species-selector/components/genome-groups/GenomeGroups.tsx
```diff
@@ -52,7 +61,7 @@ const GenomeGroup = (props: { group: GenomeGroupType }) => {
         <Pill>{group.genomes_count}</Pill>
         <span>genomes</span>
       </div>
-    </div>
+    </Link>
```
azangru: I can see in Andrea's XD that the whole project element is clickable; but so far, people have confirmed that the clickable area should be (2)

Image

src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
```diff
@@ -11,7 +11,10 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
+  align-self: flex-start;
   row-gap: 6px;
+  color: inherit;
+  text-decoration: none;
```
azangru: we won't need these if the group stops being a link, right?

src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
```diff
@@ -11,7 +11,10 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
+  align-self: flex-start;
```
azangru: I tried disabling this in dev tools, but couldn't see any difference. What was this for?

shashankbrgowda: Because pointer appeared outside content of project element (height matched with largest element in row). If we make just  count element clickable we can remove this.

src/content/app/species-selector/components/genome-groups/GenomeGroups.module.css
```diff
@@ -32,4 +32,5 @@
   align-items: center;
   column-gap: 8px;
   color: var(--color-blue);
+  text-decoration: none;
```
azangru: Image

I am not seeing any consequence to removing this rule (it is picked up from the global styles). Are you?

azangru (approved):

PR #1309, opened 2026-06-03
Combine all feature searches into one box

bethflint:
## Description
- Updated the search on species selector to have two distinct search boxes: one for species and one for features.  
- Feature search box searches through all available feature search (gene, transcript and variant)
- New results views showing results of all features
- Enabled search page (top left launch) to also contain species and feature search boxes.

## Deployment URL(s)
http://search-experiment.review.ensembl.org

## Views affected
https://search-experiment.review.ensembl.org/species-selector
https://search-experiment.review.ensembl.org/search
https://search-experiment.review.ensembl.org/search/results?query=brca2 (when species are added)
https://search-experiment.review.ensembl.org/species-selector/search/results?query=brca (when species are added)
all views - launchbar search icon enabled.


src/content/app/entity-viewer/transcript-view/components/transcript-details/TranscriptDetails.module.css
```diff
@@ -6,7 +6,6 @@
 
 .header {
   border-bottom: 1px solid var(--color-orange);
-  padding: 0;
```
azangru, line 9: I don't believe this falls within the scope of the PR? It is changing something for Entity Viewer.

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
```diff
@@ -87,9 +94,12 @@ If no results are shown, please try a different spelling or attribute
...
   const [searchParams] = useSearchParams();
-  const [searchInput, setSearchInput] = useState(
-    searchParams.get('query') || ''
-  );
+  const queryFromParams = searchParams.get('query') || '';
+  const [searchInput, setSearchInput] = useState(queryFromParams);
+
+  useEffect(() => {
+    setSearchInput(queryFromParams);
+  }, [queryFromParams]);
```
azangru: I don't understand how this is happening; but the linter is not complaining about this code block here. Meanwhile, it seems to be in violation of [this rule](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect).

For example, eslint would complain about the code below:

```
const Test = () => {
  const [foo, setFoo] = useState('');

  useEffect(() => {
    setFoo('bar');
  }, [foo]);
};
```

But it did not pick up the usage I am commenting on — due to some shortcomings in static analysis.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
```diff
@@ -26,26 +27,51 @@ import SpeciesSearchField, { type Props } from './SpeciesSearchField';
...
+    showFeatureSearchLinks?: boolean;
+  }
+) => {
+  const {
+    title = 'Find a species',
+    titleIcon,
+    isFeatureSearch = false,
+    showFeatureSearchLinks = true,
+    ...searchFieldProps
+  } = props;
```
azangru: Isn't `SpeciesSearchFieldWithLinks` now becoming redundant? Haven't we removed the links (i.e. the tabs) associated with the species search field in this PR? Compare to the [PR](https://github.com/Ensembl/ensembl-client/pull/1303) in which this component was introduced.

**UPDATED:** Ok; I see that you are using this component below. We should think whether this component is still necessary (might want to discuss with Shashank); but if it stays, then at the very least, it should be renamed to avoid confusion. There are no "links" in this component anymore.

src/content/app/entity-viewer/transcript-view/components/transcript-details/TranscriptDetails.module.css
bethflint: This is removing the duplicate padding which was ignored by the code and triggered a warning.

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.module.css
```diff
@@ -10,12 +10,20 @@
   grid-template-columns: 600px 1fr;
   grid-template-rows: 1fr auto;
   align-items: start;
-  height: 270px;
+  height: 300px;
```
azangru: We probably want to be more flexible here to avoid the below:

Image

perhaps a min-height of 300px will do the trick? Although we'll need a padding to avoid the button bumping into the tabs too.

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    const featureSearchForm = featureSearchInput.closest('form');
+    await userEvent.click(
+      within(featureSearchForm as HTMLFormElement).getByRole('button', {
+        name: 'Find'
+      })
+    );
+
+    expect(routerInfo.location?.pathname).toBe(
+      '/species-selector/search/results'
```
azangru: I don't like this url.

Compare with our species selector urls when we are searching for species: `https://staging-2020.ensembl.org/species-selector/search?species_taxonomy_id=10090` or `https://staging-2020.ensembl.org/species-selector/search?query=human`.

What if, instead, users are directed to `/search/...`, similarly to how users from the `/search` page are directed to `/species-selector` when they search for species?

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.tsx
```diff
@@ -18,23 +18,37 @@ import { useState } from 'react';
...
+import {
+  FeatureSearchIcon,
+  SpeciesSelectorIcon
+} from 'src/shared/components/app-icon';
 import GenomeCounts from 'src/shared/components/genome-counts/GenomeCounts';
 import PopularSpeciesList from 'src/content/app/species-selector/components/popular-species-list/PopularSpeciesList';
 import GenomeGroups from 'src/content/app/species-selector/components/genome-groups/GenomeGroups';
 import { useGenomeGroupCategoriesQuery } from 'src/content/app/species-selector/state/species-selector-api-slice/speciesSelectorApiSlice';
 import TextButton from 'src/shared/components/text-button/TextButton';
+import { getCommittedSpecies } from 'src/content/app/species-selector/state/species-selector-general-slice/speciesSelectorGeneralSelectors';
```
azangru: Let's move this line, as well as the `useGenomeGroupCategoriesQuery` import on line 34 above, further up, in a section of its own, above line 26.

The style rule for imports I am trying to follow is:
- imports from native modules and third-party libraries first (including type imports)
- then helpers, utility functions, constants
- then selectors of redux state, if present
- then imports from redux slices, if present
- then imports of components
- then imports of types
- finally, imports of styles

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.tsx
```diff
@@ -63,11 +77,42 @@ const SpeciesSelectorMainView = () => {
...
     <div className={styles.main}>
       <div className={styles.searchPanel}>
-        <div className={styles.speciesSearchFieldWrapper}>
-          <SpeciesSearchFieldWithLinks onSearchSubmit={onSearchSubmit} />
+        <div className={styles.searchFields}>
+          <SpeciesSearchFieldWithLinks
+            titleIcon={
+              <span role="img" aria-label="Find a species">
+                <SpeciesSelectorIcon />
+              </span>
```
azangru: Why does this icon need a span wrapper?

As for the accessibility attributes (role and aria-label), this image, as well as the image for the field below, is purely decorative; so there is no point in having them.

Codex's attempt at accessibility here is a good reminder that it is worth checking what the screenreader says when you focus on any of the input fields. We may want to turn the text elements above the input fields into proper html label elements. Or, if not, then add the `aria-label` attribute to the search fields.

src/content/app/species-selector/views/species-selector-search-results-view/SpeciesSelectorSearchResultsView.module.css
azangru, line 1: The name of the module is now very misleading.

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
```diff
@@ -0,0 +1,50 @@
+.main {
+  height: 100%;
+  overflow: hidden;
```
azangru: is this style necessary?

src/content/search/views/search-results-view/SearchResultsView.module.css
```diff
@@ -0,0 +1,50 @@
...
+  min-height: 0;
+  overflow-y: auto;
+}
+
+.title {
+  margin: 0;
+  margin-bottom: 24px;
+  font-size: 20px;
+  font-weight: var(--font-weight-normal);
+}
```
azangru, line 4: I am not seeing this class applied anywhere on the page

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
```diff
@@ -0,0 +1,50 @@
...
+
+.section {
+  display: flex;
+  flex-direction: column;
+  row-gap: 12px;
+  margin-bottom: 32px;
+}
+
+.sectionTitle {
+  margin: 0;
```
azangru: why is this style necessary?

src/content/app/search/components/combined-feature-search-results/CombinedFeatureSearchResults.module.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.sectionTitle {
+  margin: 0;
+  font-size: 14px;
+  font-weight: var(--font-weight-normal);
+}
+
+.warning {
+  margin: 0;
```
azangru, line 22: why is this style necessary?

src/content/app/search/views/search-results-view/SearchResultsView.tsx
azangru, line 1: We are clearly using this component and the `SpeciesSelectorSearchResults` component for the same purpose. They should be unified.

src/content/app/search/views/search-results-view/SearchResultsView.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+  }
+}));
+
+vi.mock('src/shared/state/api-slices/searchApiSlice', () => ({
+  useLazySearchGenesQuery: () => useLazySearchGenesQuery(),
+  useLazySearchTranscriptsQuery: () => useLazySearchTranscriptsQuery(),
+  useLazySearchVariantsQuery: () => useLazySearchVariantsQuery()
+}));
+
+const committedSpecies: CommittedItem[] = [
```
azangru: We have a function for generating these test objects: https://github.com/Ensembl/ensembl-client/blob/main/tests/fixtures/selected-species.ts#L21

src/content/search/views/search-results-view/SearchResultsView.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+const ResultsSearchField = () => {
+  const navigate = useNavigate();
+  const [searchParams, setSearchParams] = useSearchParams();
+  const queryFromParams = searchParams.get('query') ?? '';
+  const [searchInput, setSearchInput] = useState(queryFromParams);
+
+  useEffect(() => {
+    setSearchInput(queryFromParams);
+  }, [queryFromParams]);
```
azangru: same as earlier comment

src/content/search/views/search-results-view/SearchResultsView.tsx
```diff
@@ -0,0 +1,305 @@
...
+      featureSearchMode: 'variant' as const,
+      searchResults: currentVariantSearchResults,
+      variantSearchError
+    }
+  ];
+
+  if (!hasLoadedStoredSpecies || !committedSpecies.length) {
+    return null;
+  }
+
```
azangru, line 91: Here is what the loading state currently looks like. We should be showing a spinner instead

Image

src/content/search/SearchPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    dispatch(
+      updatePageMeta({
+        title: pageTitle,
+        description: pageDescription
+      })
+    );
+    dispatch(loadStoredSpecies());
```
azangru: We shouldn't need this here, nor the line below. The responsibility for loading data stored in browser storage should not belong to the search page.

src/content/search/SearchPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+import SearchAppBar from './components/search-app-bar/SearchAppBar';
+import SearchMainView from './views/search-main-view/SearchMainView';
+import SearchResultsView from './views/search-results-view/SearchResultsView';
+
+import styles from './SearchPage.module.css';
+
+import type { ServerFetch } from 'src/routes/routesConfig';
+
+const pageTitle = 'Search';
+const pageDescription = 'Search';
```
azangru: These two lines define the content of the `<title>` tag and of the `<meta name="description" >` tag in the page header. We may want to have them more descriptive than this. For example, on the species selector page, the title tag says: `"Species selector — Ensembl"` and the description metatag says `"Select one or more species to start using Ensembl"`

src/header/launchbar/Launchbar.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    renderComponent();
+
+    const siteSearchLink = screen.getByRole('link', { name: 'Site search' });
+    expect(siteSearchLink.getAttribute('href')).toBe('/search');
+
+    await userEvent.click(siteSearchLink);
+    expect(trackLaunchbarAppChange).toHaveBeenCalledWith('Site search');
+  });
+});
```
azangru: These are pointless tests.

src/shared/components/app-icon/SearchIcon.tsx
azangru, line 1: Since this is one of app icon components, we should probably call it using the name that we plan to give this app.

src/shared/components/table/Table.module.css
```diff
@@ -4,11 +4,12 @@
...
 
-.tableWithStickyHeader thead tr {
+.tableWithStickyHeader thead th {
   position: sticky;
   top: 0;
   background-color: var(--table-head-background, var(--_table-background));
-  z-index: 1;
+  z-index: 2;
+  background-clip: padding-box;
 }
```
azangru, line 12: What problem is it trying to solve with changes in this file?

src/shared/components/view-in-app/ViewInApp.module.css
```diff
@@ -17,6 +17,14 @@
   color: var(--color-white);
 }
 
+.viewInAppCompact .label {
+  display: none;
+}
+
+.viewInAppCompact {
+  justify-content: center;
+}
```
azangru: Compare styles (e.g. icon size), as well as approach, with the species manage table: https://search-experiment.review.ensembl.org/species-selector/manage

Do we want a `compact` version of the `ViewInApp` component (thus complicating the component), which shows the tooltip if you hover over the icon, even though the icon is in a column titled "Genome browser"/"Entity viewer"? Or do we just want a simple app icon in a link? I am not sure.

static/icons/icon_feature_search.svg
azangru, line 1: We already have this icon: https://github.com/Ensembl/ensembl-client/blob/main/static/icons/icon_search.svg

azangru:
I believe this PR makes [this redux slice](https://github.com/Ensembl/ensembl-client/blob/834025b89290abb8cf19b1e8a7aa22f6b2bfffa6/src/content/app/species-selector/state/species-selector-feature-search-slice/speciesSelectorFeatureSearchSlice.ts) (and therefore, [this whole directory](https://github.com/Ensembl/ensembl-client/tree/834025b89290abb8cf19b1e8a7aa22f6b2bfffa6/src/content/app/species-selector/state/species-selector-feature-search-slice)) redundant. I see some remnants of its usage in `SpeciesSelectorSearchView.tsx`; but this is probably because of some redundancy. The purpose of `speciesSelectorFeatureSearchSlice` was to keep track of search queries of different types of features (genes/transcripts/variants) as the tabs above the search field were switched. Since there are no longer any tabs, there shouldn't be any need of this redux state.

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
bethflint: Should we derive directly from the URL?
If the input value should always match ?query=..., drop useState and useEffect:

```
const WrappedSpeciesSearchField = (props: Omit<Props, 'query'>) => {
  const [searchParams] = useSearchParams();
  const queryFromParams = searchParams.get('query') || '';

  return (
    <SpeciesSearchField
      {...props}
      query={queryFromParams}
      onInput={props.onInput}
    />
  );
};
```
That removes the lint issue because there is no state sync in an effect.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
```diff
@@ -26,26 +27,51 @@ import SpeciesSearchField, { type Props } from './SpeciesSearchField';
...
+            return (
+              <Link
+                key={mode}
+                className={styles.featureSearchLink}
+                to={urlFor.speciesSelectorFeatureSearch(mode)}
+              >
+                {label}
+              </Link>
+            );
+          })}
```
azangru: Is this even used anywhere anymore? With this PR, we don't have feature search modes any longer — all features are being searched at once.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
```diff
@@ -26,26 +27,51 @@ import SpeciesSearchField, { type Props } from './SpeciesSearchField';
...
+    title = 'Find a species',
+    titleIcon,
+    isFeatureSearch = false,
+    showFeatureSearchLinks = true,
+    ...searchFieldProps
+  } = props;
+
   return (
     <div className={styles.searchFieldWithLinks}>
       <div className={styles.searchLinks}>
```
azangru: There aren't any links anymore. The names of these CSS classes are a lie.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
```diff
@@ -26,26 +27,51 @@ import SpeciesSearchField, { type Props } from './SpeciesSearchField';
...
+        <span
+          className={[
+            styles.searchFieldTitle,
+            isFeatureSearch ? styles.featureSearchTitle : '',
+            searchFieldProps.disabled ? styles.disabledSearchFieldTitle : ''
+          ].join(' ')}
+        >
+          {titleIcon}
+          {title}
+        </span>
```
azangru: There is a vertical shift in the position of the input field as you navigate from the species selector page (or the general search page) to the search results view. Not horrible, but noticeable. I would try to avoid that by checking what defines the height of the span here, and maybe even assigning position: absolute to the icon so that it doesn't impact the layout.

https://github.com/user-attachments/assets/bbf9248b-d605-4e8a-8ae3-20557bfdb943

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.module.css
bethflint: I adjusted the species selector panel spacing so the search controls no longer crowd the tabs on narrower screens.

Replaced the fixed height: 300px with min-height: 300px
Reduced the grey panel bottom padding slightly
Added a vertical row-gap between the search area and the tabs
This keeps the layout flexible while giving the Find button more breathing room before the popular species / external project tabs

Will push changes after reviewing the other comments.

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.tsx
bethflint: I reordered the imports in SpeciesSelectorMainView.tsx to match the project’s import grouping convention, without changing any runtime behavior.

Specifically:

moved the Redux selector and RTK query hook higher in the import block
kept native/third-party imports first
left component and style imports lower in the file
This is just a cleanup to align with the local style rule.  I reordered the imports in SpeciesSelectorMainView.tsx to match the project’s import grouping convention, without changing any runtime behavior.

Specifically:

moved the Redux selector and RTK query hook higher in the import block
kept native/third-party imports first
left component and style imports lower in the file
This is just a cleanup to align with the local style rule.

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.tsx
bethflint: I removed the span wrappers and decorative role="img" / aria-label attributes from the icons in the species selector search fields.

To keep the inputs accessible, I also passed the visible field titles through as aria-label values on the underlying search inputs, so screen readers have a proper label without relying on the icon wrappers.

I also verified the affected tests still pass.

src/content/app/species-selector/views/species-selector-search-results-view/SpeciesSelectorSearchResultsView.module.css
bethflint: Do you have a preferred name?

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
bethflint: The intention is:
height: 100% lets the view fill the content area
overflow: hidden prevents the page itself from scrolling
display: flex and flex-direction: column are needed so the results area can take the remaining space

src/content/search/views/search-results-view/SearchResultsView.module.css
bethflint: This class is only used in the missing-resource error branch for variant search results, so it’s not applied on the default path.

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
bethflint: The style provides the vertical stack and spacing for each result block:

```
display: flex
flex-direction: column
row-gap: 12px
margin-bottom: 32px
```

Do you think the spacing should live elsewhere?

src/content/app/search/components/combined-feature-search-results/CombinedFeatureSearchResults.module.css
bethflint: This class is used in SearchResultsView.tsx (lines 289-290):
`isMissingResourceError(variantSearchError) ? <p className={styles.warning}>...`
So .warning is what makes that error message red and removes default paragraph margin:
```
margin: 0
color: var(--color-red)
```

Do you think it could be handled differently?

src/content/app/search/views/search-results-view/SearchResultsView.tsx
bethflint: I unified the duplicate feature search results implementations by extracting the shared search/fetch/render logic into a reusable FeatureSearchResultsView component.

The global search results page and the species-selector search results page now use that shared component while keeping their existing route-specific redirects and close-button behavior. I also removed the two old duplicated CSS modules and moved the shared layout styles into the new shared component.

Targeted tests, ESLint, and stylelint all pass.

src/content/app/search/views/search-results-view/SearchResultsView.test.tsx
bethflint: I replaced the hand-written CommittedItem objects in the search results tests with the existing createSelectedSpecies fixture helper.

I kept the explicit species values that the tests depend on, such as genome_id, names, taxonomy ID, assembly, and release, but let the fixture provide the standard default fields.

The affected tests still pass.

src/content/search/views/search-results-view/SearchResultsView.tsx
bethflint: Updated the loading state to show an existing CircleLoader spinner under each feature result heading, rather than showing a single generic loading message.

Each section now reflects its own loading state, so gene, transcript, and variant results can show their loaders independently. Targeted tests, ESLint, and stylelint pass.

src/content/search/SearchPage.tsx
bethflint: Removed the duplicate stored-species restoration from SearchPage.

The search page now only handles its own page metadata. Loading species data and the species-name display option remains the responsibility of the root-level useRestoredReduxState hook, which already runs when the app mounts.

Relevant search tests and targeted ESLint pass.

src/content/search/SearchPage.tsx
bethflint: Updated the search page metadata to be more descriptive.

The page title is now Search — Ensembl, and the meta description explains that the page is for searching genes, transcripts, or variants in Ensembl.

Targeted ESLint passes.

src/header/launchbar/Launchbar.test.tsx
bethflint: Removed the duplicate launchbar test cases that were checking the same site-search link behavior under different committed-species states.

Since the species-dependent launchbar buttons are mocked in this test, the committed-species state was not affecting the behavior under test. The remaining test still verifies the /search link and analytics tracking.

Targeted test and ESLint pass.

src/shared/components/table/Table.module.css
bethflint: That change was meant to fix sticky table headers for the new feature search results table after switching it to the shared Table component.

What it does:

changes sticky positioning from thead tr to individual thead th
raises the header stacking order from z-index: 1 to z-index: 2
adds background-clip: padding-box

Is there a preferred approach?

src/content/search/views/search-results-view/SearchResultsView.tsx
azangru: >  spinner under each feature result heading,

Woah, are we loading them and updating the UI independently? Can't we have just one spinner, wait for all search requests for all feature types to resolve, and then replace the spinner with the results?

src/shared/components/view-in-app/ViewInApp.module.css
bethflint: Lets go with how it is for now and see if there's any feedback

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
azangru: > Should we derive directly from the URL?

Yes.

static/icons/icon_feature_search.svg
bethflint: switched to use the icon we already have and removed the one that was added.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
bethflint: Agreed, this looks like leftover behavior from the old mode-specific feature-search flow. Both current callers now pass `showFeatureSearchLinks={false}`, and the links point to the older `/species-selector/search/:mode` routes.

I think this should be cleaned up together with the redundant `speciesSelectorFeatureSearchSlice` / mode-specific route cleanup, once we confirm those old routes are no longer needed.

Shall we clean it up?

azangru: We should.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
bethflint: Removed the now-unused feature-mode links from SpeciesSearchFieldWithLinks.

Since the current flow searches all feature types together, the old gene/transcript/variant links were no longer used. I also renamed the related CSS classes so they describe the current title/header layout rather than referring to links.

Focused tests, ESLint, and stylelint pass.

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
bethflint: I tried to fix the alignment, but couldn't quite make it work.

src/content/app/species-selector/components/species-search-field/SpeciesSearchField.tsx
bethflint: Updated WrappedSpeciesSearchField so it derives its value directly from the URL query param instead of mirroring it into local state with an effect.

For editable search fields, the draft input state now lives in the titled search-field wrapper, where it is actually needed. Results pages pass the URL query as the initial value and remount the wrapper when the query changes, so the input still resets correctly without a state-sync effect.

Targeted ESLint and focused tests pass.

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.test.tsx
bethflint: Should be fixed now.

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
bethflint: overflow: hidden removed

src/content/search/views/search-results-view/SearchResultsView.module.css
bethflint: Removed

src/shared/components/feature-search-results-view/FeatureSearchResultsView.module.css
bethflint: Removed margin: 0

src/content/app/search/components/combined-feature-search-results/CombinedFeatureSearchResults.module.css
bethflint: this rule removes margin from paragraph, so it can stay.

src/content/app/search/views/search-results-view/SearchResultsView.tsx
bethflint: Keep, but maybe review in the future.

src/content/search/views/search-results-view/SearchResultsView.tsx
bethflint: Should be resolved now.

src/content/search/views/search-results-view/SearchResultsView.tsx
bethflint: Fixed

src/header/launchbar/Launchbar.test.tsx
bethflint: Removed test file

src/shared/components/app-icon/SearchIcon.tsx
bethflint: Fixed

src/shared/components/table/Table.module.css
bethflint: reverted to how it was

src/content/app/species-selector/components/species-search-field/SpeciesSearchFieldWithLinks.tsx
bethflint: done

src/content/app/species-selector/views/species-selector-main-view/SpeciesSelectorMainView.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+);
+
+vi.mock(
+  'src/content/app/species-selector/components/genome-groups/GenomeGroups',
+  () => ({
+    default: () => <div>Genome groups</div>
+  })
+);
+
+const committedSpecies: CommittedItem[] = [
```
azangru: We still have a test object that should have been created via a helper function in `fixtures/selected-species.ts`

PR #1313, opened 2026-06-17
Entity Viewer for gene: fix gene svg width in Safari

azangru:
## Description
**In Safari**, the svg image of a gene at the top of the Entity Viewer page for the gene renders larger than expected, and extends beyond the allocated space, overlapping other elements:

image

After the fix:

image

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/EBD-123

## Deployment URL(s)
http://fix-gene-svg-width.review.ensembl.org


shashankbrgowda (approved):
One small thing, we have scrollbar just for view in button for the gene

Image

PR #1316, opened 2026-06-19
Fix genome browser navigation

azangru:
## Description
This is a follow-up on https://github.com/Ensembl/ensembl-client/pull/1307, which turned out to be insufficient to prevent unnecessary location change messages to the genome browser causing surprising jumps, or even infinite animation loops.

The idea behind this PR is to avoid sending location change message to the genome browser if the reason the url got updated was because of a message from the genome browser. There is no need to ask genome browser to navigate to a location where it presumably already is anyway.

To tell url changes that were caused by the genome browser as opposed to other reasons, the client performs the following checks:
- Check whether browser history state contains key/value `updateSource: 'genome-browser'`. This state is set when the url is updated due to a message from the genome browser.  
- Check whether the navigation change is of a type "REPLACE". This excludes back/forward navigation (e.g. user navigating web browser history)

ALSO:
This PR filters out target_position messages with invalid locations that sometimes are emitted by the genome browser — specifically locations where start coordinate is a negative number. 

The recording below demonstrates both kinds of bugs: first, you may notice genome browser's erratic behaviour on zoom out; and then, continuous zoom in/out near the start of the region triggers the genome browser to send an invalid location to the client , which fails url validation and causes an error screen

https://github.com/user-attachments/assets/ebc483ed-13bc-41d7-a244-71429f23ad70

## Deployment URL(s)
https://fix-gb-navigation.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/components/browser-app-bar/BrowserAppBar.tsx
+++ b/src/content/app/genome-browser/components/browser-app-bar/BrowserAppBar.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { memo, useMemo } from 'react';
+import { memo, useCallback, useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import isEqual from 'lodash/isEqual';
 
@@ -38,16 +38,20 @@ type BrowserAppBarProps = {
 };
 
 const BrowserAppBar = (props: BrowserAppBarProps) => {
+  const { onSpeciesSelect: speciesSelectHandlerFromProps } = props;
   const enabledCommittedSpecies = useSelector(getEnabledCommittedSpecies);
   const activeGenomeId = useSelector(getBrowserActiveGenomeId);
 
   const { trackGenomeChanged } = useGenomeBrowserAnalytics();
 
-  const onSpeciesSelect = (species: CommittedItem) => {
-    props.onSpeciesSelect(species.genome_id);
+  const onSpeciesSelect = useCallback(
+    (species: CommittedItem) => {
+      speciesSelectHandlerFromProps(species.genome_id);
 
-    trackGenomeChanged();
-  };
+      trackGenomeChanged();
+    },
+    [speciesSelectHandlerFromProps, trackGenomeChanged]
+  );
 
   const speciesTabs = useMemo(() => {
     return enabledCommittedSpecies.map((species, index) => (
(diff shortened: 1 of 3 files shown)


src/content/app/genome-browser/components/browser-app-bar/BrowserAppBar.tsx
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { memo, useMemo } from 'react';
+import { memo, useCallback, useMemo } from 'react';
```
azangru, line 17: NOTE: changes to this file are irrelevant to the genome browser bug; but are fixing a linter warning.

shashankbrgowda (commented):
GB navigation is smooth now without any jumps or loops and there is no error while navigating around the start of the region. 

But there is still some issue in switching between lozenges

https://github.com/user-attachments/assets/1328b787-439f-4ca2-ac53-422c4aa6e48b

azangru:
@shashankbrgowda 

I tried to look into switching between genomes, and discovered a couple of problems in the code (specifically, there was an early return in the useEffect hook that prevented updates to redux). But I am not sure whether my changes have fixed the problem that you were describing, because I didn't quite understand the specific description of the problem.

Could you please check whether my latest changes have improved the switching between genomes?

shashankbrgowda:
@azangru The issue still persists :( In the attached video, I've selected four genomes and when I switch between the genome lozenges, you can notice content is changed in the right hand sidebar and even In the bar above GB but, track data in GB is not updated. It was easy to reproduce - we have to select multiple genomes and load example focus gene for all of them and try switching between the lozenges.


https://github.com/user-attachments/assets/11445d22-a498-4922-94ff-0f78a48dd485

azangru:
@shashankbrgowda 

One of the zebrafishes that you've selected (the `ASM3317019v3` one) has server-side problems — you can see errors coming both from the track api and from the genome browser server.  

image

When the genome browser receives errors, back from the server, it repeats its request several time; and apparently ignores your commands to switch to a different genome. There's nothing I can do about this on the client side.

Pick another genome instead of zebrafish `ASM3317019v3`. For example, human `grch37` should do (note that the partial grch38 is also broken in dev). I didn't see any problems with switching between genomes on my end:


https://github.com/user-attachments/assets/05c78b40-e375-4295-b65b-7342fe0936a0

shashankbrgowda (approved):
GB data is updated properly when I selected different genomes and switched between genome lozenges. Looks good

PR #1318, opened 2026-06-24
Update elements for genome browser navigation

azangru:
## Description
This PR makes the following updates to the genome browser page:
- The browser location indicator in the top bar becomes an editable input field through which user can change genome browser location
- To the left of the input field, there are navigation buttons that can be used to navigate the genome browser
- Entering an invalid location into the input field brings up a tooltip with a warning
- In the sidebar, the component responsible for changing the location has been simplified and moved out of the "sidebar modal" to the genomic view of the track panel. 
- Switched from `/karyotype` to `/top-regions` endpoint of the metadata api, in order to get lists of top-level regions that are not assembled to chromosome level.

image

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/ENSWBSITES-3015

## Deployment URL(s)
http://gb-navigation.review.ensembl.org


src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.tsx
```diff
@@ -14,155 +14,146 @@
...
+
 import styles from './BrowserLocationIndicator.module.css';
 
 type Props = {
   className?: string;
-  containerRef?: RefObject<HTMLElement | null>;
-  nonOverlapElementRef?: RefObject<HTMLElement | null>;
 };
 
 export const BrowserLocationIndicator = (props: Props) => {
```
shashankbrgowda, line 43: Functionality of this component is chnaged, It's no longer just a Indicator. Can we rename this ?

src/shared/components/region-navigation-buttons/RegionNavigationButtons.tsx
```diff
@@ -0,0 +1,118 @@
...
+  viewportStart: number;
+  viewportEnd: number;
+  regionLength: number;
+  className?: string;
+};
+
+const PANNING_STEP = 0.1;
+const ZOOMING_STEP = 0.5;
+
+const RegionNavigationButtons = (props: Props) => {
```
shashankbrgowda, line 48: When I navigate to start or end of the region and try to pan left/right, it adjusts the focus location to previous location before navigation

https://github.com/user-attachments/assets/fcccc44d-5ee7-40a2-ab7b-e9822df99dd8

src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.tsx
azangru: Do you have any suggestions for new name?

src/shared/components/region-navigation-buttons/RegionNavigationButtons.tsx
azangru: Thank you for finding the bug :-) 

The reason I tried to separate the buttons from the updates coming from from the outside is because the updates from the outside might interfere with user's quick presses on the navigation buttons, such that sometimes it feels as is a press has not been responded to.

I'll try to add a timeout to debounce updates from the outside; although I am not sure this is good idea either.

azangru: Updated.

src/content/app/genome-browser/components/browser-location-indicator/BrowserLocationIndicator.tsx
shashankbrgowda: I'm not sure. Maybe BrowserLocationNavigationInput ?

shashankbrgowda (approved):
Looks good

shashankbrgowda (approved):

PR #1319, opened 2026-06-24
Delete genomes from species manager

luanbroo:
## Description
Make it easy to delete genomes from the genome manager
- Table has checkboxes in the left column to select genomes. There is a master checkbox for selecting all genomes at once.
- There is a "Remove genomes" button in top right that is activated when checkboxes for one or more genomes are ticked off, and deletes these genomes.

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/ENSWBSITES-3007

## Deployment URL(s)
http://manage-species-select-delete-all.review.ensembl.org


src/content/app/species-selector/views/species-manager/SpeciesManager.module.css
```diff
@@ -1,52 +1,81 @@
...
+.removeGenomesButton {
+  --primary-button-color: var(--color-red);
+  --button-border: 1px solid var(--color-red);
+}
+
+@media (width < 1200px) {
+  .mainContentTop {
+    grid-template-areas:
+      "genome-label-selector search-field add-genome-button"
+      ".                     .            delete-button";
```
shashankbrgowda: could we remove unnecessary whitespaces ?

shashankbrgowda (approved):

src/content/app/species-selector/views/species-manager/SpeciesManager.module.css
azangru: Hmm; I was trying to position cells belonging to same columns under one another in the two rows :-)

Yes; I can remove the whitespaces.

PR #1322, opened 2026-07-03
Do not use lodash to update species general redux slice

azangru:
## Description
A bug with the same cause as https://github.com/Ensembl/ensembl-client/pull/1321, but now affecting species general slice.

Steps to reproduce: try opening any of the stats accordions on the species page.

image

## Deployment URL(s)
http://fix-species-expandable-sections.review.ensembl.org

### DIFF
--- a/src/content/app/species/state/general/speciesGeneralSlice.ts
+++ b/src/content/app/species/state/general/speciesGeneralSlice.ts
@@ -88,15 +88,13 @@ const speciesGeneralSlice = createSlice({
         expandedSections: SpeciesStatsSection[];
       }>
     ) {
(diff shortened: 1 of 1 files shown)


src/content/app/species/state/general/speciesGeneralSlice.ts
shashankbrgowda, line 61: Do we need to update this as well ?

azangru: It isn't necessary for this PR. This other function is called in a thunk, and the state there (the return value of `getState()`) is a plain object rather than an immer-wrapped proxy.

shashankbrgowda (approved):

PR #1325, opened 2026-07-09
Update gene track colour markers

veidenberg:
## Description
This PR ~changes~ removes all gene track colour markers in the right-hand panel, to match the [updates in Genome browser](https://github.com/Ensembl/ensembl-dauphin-style-compiler/pull/149).

## Related JIRA Issue(s)
[ENSWBSITES-2973](https://embl.atlassian.net/browse/ENSWBSITES-2973)

## Deployment URL(s)
http://blue-genes.review.ensembl.org

## Views affected
Genome Browser view

### DIFF
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.module.css
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.module.css
@@ -56,20 +56,6 @@
   padding: 10px 0 20px;
 }
 
-.trackPanelSectionLegend {
-  display: flex;
-  align-items: center;
-  column-gap: 12px;
-  padding: 4px 10px;
-}
-
-.trackPanelSectionLegendMarker {
-  width: 10px;
-  height: 10px;
-  background-color: var(--color-blue);
-  flex-shrink: 0;
-}
-
 .modalLinksWrapper {
   margin: 0 0 20px 10px;
   display: flex;
--- a/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.module.css
+++ b/src/content/app/genome-browser/components/track-panel/components/track-panel-list/TrackPanelList.module.css
@@ -56,6 +56,20 @@
   padding: 10px 0 20px;
 }
 
+.trackPanelSectionLegend {
+  display: flex;
+  align-items: center;
+  column-gap: 12px;
+  padding: 4px 10px;
+}
+
+.trackPanelSectionLegendMarker {
+  width: 10px;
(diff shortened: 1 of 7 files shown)


azangru:
image


Argh! These blue squares are now a completely useless noise (what's their purpose? why do we have them next to gene tracks, but not the others?). I vote we remove them altogether. Of course my vote counts for almost nothing, and we should run this past decision-makers

veidenberg:
I agree, I first started with removal but then reverted to match the ticket. I'll verify the design.

veidenberg:
@azangru we now have less blue squares.

azangru:
> we now have less blue squares.

Fewer squares is a win.

But why on earth do we need to add that final blue square? Compared to all the tracks that are already in the sidebar, and all the multicoloured thingies in the genome browser, how is it this particular square helpful?

And the added label text repeats the text in the accordion button 😭 

Argh!

image

veidenberg:
The marker indicates that the tracks with blue elements show genes and transcripts (see the comments in the [ticket](https://embl.atlassian.net/browse/ENSWBSITES-2973)). I agree that it needs more iteration (e.g. move the marker to section header?)

azangru:
But the gene tracks show things that already look very recognisably like genes and transcripts. And, if anyone has any doubts, there is also track names, which are now enabled by default, and which say that the things in these tracks are genes and transcripts. I am finding it hard to imagine that the blue colour is going to trip people up in any way. Compare with the repeat elements track, which shows rectangles of some colour, but we are not adding a coloured square anywhere to say that this is the colour of the repeat elements. Compare also with the [regulation subsite](https://regulation.ensembl.org/genome/homo_sapiens?location=1:100728560-100762343), where they added a colour legend for regulatory features, but not for the genes, which are all blue. Similarly, in the [structural variants viewer](https://staging-2020.ensembl.org/structural-variants?ref-genome-id=a7335667-93e7-11ec-a39d-005056b38ce3&ref-location=11:50233524-50302020&alt-genome-id=27be510b-c431-434c-a6f5-158d8c138507&alt-location=11:50352547-50421043), we don't explain the uniformly blue colour of the genes.

veidenberg:
As discussed, the colour markers are now gone from the track list panel.

azangru:
So the client is no longer relying on the `colour` field coming from the track api?

Could you remove the [`colour` property](https://github.com/Ensembl/ensembl-client/blob/main/src/content/app/genome-browser/state/types/tracks.ts#L47) from the `GenomicTrack` type, run `npm run check-types`, and see if anything complains?

azangru (approved):
Looks good!

PR #1330, opened 2026-07-15
Add an indeterminate state for Checkbox component

azangru:
## Description
It seems that we are starting to use a pattern in which one checkbox can control multiple subordinate checkboxes to select/deselect all. Thus, such a checkbox can be in one of three states:
- All subordinate checkboxes are checked --> the "select all" checkbox is also checked
- All subordinate checkboxes are unchecked --> the "select all" checkbox is also unchecked
- Some subordinate checkboxes are checked, others are not --> what should the state of the "select all" checkbox be?

It is a common practice to represent this third state as an "indeterminate" state. This is so common that the indeterminate state is part of the native DOM input element.

This PR adds an indeterminate state to the checkbox. It has **not** been deliberately designed and officially blessed by Andrea.

https://github.com/user-attachments/assets/6c543ecc-b416-453f-8839-de696fd3492c

### Docs and design systems:
- MDN [docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate) about the indeterminate state of the input element
- [Web Awesome's checkbox component](https://webawesome.com/docs/components/checkbox/#indeterminate)
- [Material UI's checkbox component](https://mui.com/material-ui/react-checkbox/#indeterminate)

### Prior art
Github notifications

image

Gitlab packages

image

### Notable exceptions

NCBI doesn't seem to use this pattern. [Example](https://www.ncbi.nlm.nih.gov/datasets/gene/GCF_000001405.40/)

## Deployment URL(s)
Not relevant. The changes made are in Storybook

### DIFF
--- a/src/shared/components/checkbox/Checkbox.module.css
+++ b/src/shared/components/checkbox/Checkbox.module.css
@@ -6,14 +6,31 @@
     cursor: pointer;
     height: var(--checkbox-side-size, 13px);
     width: var(--checkbox-side-size, 13px);
-    transition: background-color 0.1s ease-in-out;
+    /* transition: background-color 0.1s ease-in-out; */
   }
   
   .checkbox:checked:not(:disabled) {
     background-color: var(--checkbox-checked-background-color, var(--color-green));
   }
+
+  .checkbox:indeterminate:not(:disabled) {
+    --_bg-color: var(--checkbox-background-color, var(--color-white));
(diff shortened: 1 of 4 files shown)


azangru:
Closing this PR as there doesn't seem to be an appetite for a checkbox with indeterminate state. Will reopen if interest returns.

PR #1331, opened 2026-07-19
Genome browser fix: switch off non-focus tracks when switching between genomes

azangru:
## Description
**Problem:** when switching from one genome to another, non-focus tracks from one genome are carried over to another genome.

For example, in screen capture below, human repeatmasker track and dbSNP short variants track are still visible in the genome browser after switching to zebrafish. You can also see that these human tracks do not have corresponding cogs in zebrafish view

https://github.com/user-attachments/assets/6dad45d0-bb60-4240-8c3e-5c44ccb4b467

**Cause:** The client should instruct the genome browser to turn off previous genome's tracks after switching to a new genome, but is not currently doing so.

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/EBD-121

## Deployment URL(s)
http://fix-gb-non-focus-tracks.review.ensembl.org

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
@@ -13,7 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { useContext, useRef, useEffect, useCallback, useMemo } from 'react';
+import {
+  useState,
+  useContext,
+  useRef,
+  useEffect,
+  useCallback,
+  useMemo
+} from 'react';
 
 import config from 'config';
 import { useAppSelector } from 'src/store';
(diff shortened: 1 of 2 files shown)


src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -255,4 +278,14 @@ const getTrackIdToTrackPathMap = (trackCategories: GenomeTrackCategory[]) => {
...
 
+// const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+// const getTrackPath = (trackId: string) => {
+//   if (uuidRegex.test(trackId)) {
+//     return ['track', 'expand', trackId];
+//   } else {
+//     return ['track', trackId];
+//   }
+// };
```
azangru: **For discussion:** I am not sure which option to take.

At the moment, the client is trying to construct the path for a genome browser track using the `trigger` field in track api response. Trouble is, when user switches from one genome to another, the track api response for the previous genome is forgotten before the client has time to tell the genome browser to switch off the tracks for the previous genome.
 
So, in this PR, I have complicated the logic further by remembering previous genome's track api response.

A much simpler option would be to disregard the `trigger` field in the track api response entirely, and to check whether a track id looks like a uuid (in which case the path for the track is ['track', 'expand', trackId]), or not (in which case, the path for the track is ['track', trackId]). This is what the commented out code block does.

While much simpler, this option requires the client to know about how the path to a genome browser track is constructed; and that doesn't feel right.

So, I am left undecided between the two options: one feels more proper, but is more complex; the other is simple, but feels like a hack.

veidenberg: I'm using [uuid check in GB sandbox](https://github.com/Ensembl/ensembl-dauphin-style-compiler/blob/master/peregrine-generic/index.html#L208), because the expand/non-expand path discrepancy (and the `trigger` field in Track API) itself is a "temporary" hack that we should get rid of.
So I would vote for uuid check (maybe with a "remove this in the bright future" note).

azangru: > I'm using uuid check in GB sandbox

Yes; that's where I got the idea 😄 

It's certainly ok for genome browser development; but I've been debating with myself whether it's appropriate to put this into ensembl-client. But ultimately, it's so much easier than storing track api response for the current and the previous genome and generating an id-to-trigger map from them that I am agreeing with you in favour of the hack.

Code updated.

veidenberg (approved):

shashankbrgowda (approved):

PR #1332, opened 2026-07-21
Change urls for tools (BLAST and VEP)

azangru:
## Description
Changing urls for BLAST and tools
- For BLAST: from `/blast/*` to `/tools/blast/*`
- For VEP: from `/vep/*` to `/tools/vep/*`

The primary reason for the change is that `/vep` is a page on the old site that has links to documentation and source code, and needs to stay as is.

## Deployment URL(s)
http://change-tools-routes.review.ensembl.org


shashankbrgowda (approved):
Looks good

PR #1337, opened 2026-07-28
VEP V2: integrate standalone-web-vep

Alexey-ebi:
This PR is a major update to VEP. It goes together with https://github.com/Ensembl/ensembl-web-tools-api/pull/61

The VEP interface has been updated to:
- Display a dynamic list of VEP options and plugins, tailored to individual genomes.
- Display the VEP annotations in a dedicated expandable section of the results table
- Enable filtering of VEP results

Both the list of VEP options and the annotations in the table of results are rendered according to instructions parsed from json configs sent from the backend api. The client thus becomes agnostic to the specifics of the data that it is displaying.

This code in this PR has been initially generated by Claude, and then reviewed manually.

Deployment url: https://vep2.review.ensembl.org


Alexey-ebi:
Review of this branch: `review/client-review.md`. 16 findings, no blockers.

**The merge is sound.** Every post-fork hunk on the VEP directory survives except
the `genome-selector` route, which V2 deliberately replaces with a modal. No file
was lost that shouldn't have been, and no conflict resolution silently dropped a
change from `main`.

**Five worth fixing before launch:**

1. The transcript and gene-ID results filters validate against `/^ENST\d{11}/` and
   `/^ENSG\d{11}/`, so every valid mouse id is rejected and silently dropped from
   the filter — and mouse is one of the four quick-select presets.
   `resultsFilterFields.ts:91,112`
2. The options panel lets you untick the "All" ancestry under a gnomAD / All of Us
   source. That job comes back with no frequencies at all, and the results view
   renders it as the Allele frequencies section simply not being there. The filter
   on that column still works, because it reads raw CSQ — so you can filter on a
   frequency the page won't display. Real fix is API-side.
3. Alleles are looked up by `allele_sequence` string
   (`VepSubmissionResults.tsx:653-662`), so a multi-allelic SV whose alleles share
   `<DEL>` shows the first allele's rows with the second allele's annotations.
4. The server-supplied display spec is executed unguarded — `num(value as number)`,
   `joinList(value as string[])`. A format/type mismatch throws inside render and
   the nearest boundary is the app root, so the whole page goes. No live mismatch
   in the shipped spec — blast radius, not a bug today.
5. The allele-frequency filter offers ≥ and =, but the backend keeps any allele
   with no AF data for every operator — correct for ≤, inverted for ≥.

Eleven MINORs, mostly around the filter layer, which has no tests at all. Every
finding in the file cites `path:line` and carries a proposed diff.

**Needs a decision, not a fix:** whether `VepAppBar` stays (V2 deleted it; if it
goes, the genome-removal button goes with it), whether `EnsemblMark` is now
rendering the brand mark twice, and whether `/tools/vep/genome-selector` gets a
redirect.

PR #1338, opened 2026-07-29
Refactor the panel component and update no homologies messages 

azangru:
## Description
The main change in this PR was for the `Panel` component. It used to have an old api, which 1) passed the head content as a react node via a property, and 2) accepted a `classNames` property, which was an object with optional additional class names for different parts of the panel:

```
export type PanelProps = {
  header?: ReactNode;
  children: ReactNode;
  classNames?: {
    panel?: string;
    header?: string;
    body?: string;
    closeButton?: string;
  };
  onClose?: () => void;
};
```

After refactoring, the `Panel` component was split into its constituents: `Panel`, `PanelHead`, and `PanelBody`, which are expected to be used like so:

```
 <Panel>
  <PanelHead>
    Content for the head
  </PanelHead>
  <PanelBody>
        Content for the body
  </PanelBody>
</Panel>
```

Each of those components can be styled independently. The close button has, for now, been removed. In Andrea's early designs, the close button was in the top right corner of the panel body; and in her later designs it was moved to the right corner of the panel head; but we have never yet had a need for it.

### ALSO:
1. Grey tabs that don't do anything and were sitting in the panel head as an advertisement of future functionality have been removed. The future has come, but the advertised functionality hasn't.
2. I have removed the down-pointing arrows from the tabs. You can see in newer Andrea's designs that she has stopped using them. 
3. There was a minor annoyance related to react's double rendering in dev mode that prevented the first transcript from being opened by default in ~entity viewer~ feature explorer.  This is now fixed.
4. This PR also updates the messaging to distinguish between the two cases of missing gene homology data: 1) the homology pipeline wasn't run for this genome vs 2) the pipeline was run, but no homologues were found

### BUG FIXES:
This PR also fixes the bug reported months earlier in helpdesk about the unnecessary whitespace in the bottom of the panel in gene homologies view, which gave a misleading impression that there weren't any more homologies than what was being visible in the table (people didn't notice the scrollbar).

image

## Related JIRA Issue(s)
- https://embl.atlassian.net/browse/ENSWBSITES-3020 — to distinguish between the two cases of no homologies

## Deployment URL(s)
http://refactor-panel.review.ensembl.org

## Views affected
- Feature explorer -> gene -> gene function
- Feature explorer -> gene -> gene relationships
- Feature explorer -> transcript -> exons
- Feature explorer -> transcript -> transcript function
- Feature explorer -> variant -> allele frequency ([link](https://refactor-panel.review.ensembl.org/feature-explorer/GCA_000001405.29/variant:1:230710048:rs699?allele=1&view=allele-frequencies))
- Feature explorer -> variant -> transcript consequences ([link](https://refactor-panel.review.ensembl.org/feature-explorer/GCA_000001405.29/variant:1:230710048:rs699?allele=1&view=transcript-consequences))


src/content/app/entity-viewer/gene-view/components/gene-homology/GeneHomology.tsx
```diff
@@ -58,13 +58,32 @@ const GeneHomology = () => {
...
+      <span className={styles.noData}>
+        An error occurred while fetching the data
+      </span>
+    );
+  }
+
+  // The following two cases distinguish between the homology pipeline not having been run at all,
+  // and the pipeline having been run, but no homologues found
+
+  if (!currentData.homologies) {
```
veidenberg, line 70: Would `currentData.homologies === null` be a better check, since API explicitly returns `null` (and `homologies` cannot be `undefined` since it's a required prop)?

src/content/app/entity-viewer/gene-view/components/gene-homology/GeneHomology.tsx
veidenberg: If we explicitly check for `null` value above (and `homologies` is not optional), do we still need `GeneHomologyType[]` casting?

src/content/app/entity-viewer/gene-view/components/gene-homology/GeneHomology.tsx
azangru: I'm on the fence here. I don't think that between `!currentData.homologies` and `currentData.homologies === null` one has a clear advantage over the other; but for some reason, I feel a tiny bit more comfortable with `!currentData.homologies`.

src/content/app/entity-viewer/gene-view/components/gene-homology/GeneHomology.tsx
azangru: No, no need. Good point.

shashankbrgowda (approved):
Looks good

PR #1349, opened 2026-08-24
Add Ensembl UI development skill

Alexey-ebi:
## Description
Adds two complementary repository-specific Codex skills:

  - An implementation-context skill that pulls relevant coding and UI conventions from the current checkout, including shared React components, design tokens, Storybook guidance, routing, SSR, Redux, APIs, and testing.
  - An explicit standards-curator skill that maintains the repository’s evidence-backed CODING_STANDARDS.md.

  The existing standalone Ensembl UI skill is incorporated into the broader implementation-context skill. The shared standards document can also be discovered and applied by code-review workflows.

## Related JIRA Issue(s)
[ENSWBSITES-3029](https://embl.atlassian.net/browse/ENSWBSITES-3029)

## Deployment URL(s)
N/A — this change only adds development guidance and does not affect the deployed application.

## Views affected
None. This change does not alter any website views or runtime behaviour.


shashankbrgowda (approved):

azangru:
I have several additions:

1. For React:
```
- Check the version of react in package.json, and use version-appropriate api and coding conventions. For example when we use ref callbacks for DOM nodes,  we return a cleanup function from the callback, and therefore, expect that react will never pass a null to the ref callback.
- Use react best practices regarding the useEffect hook. For example, do not synchronously update component's state from the useEffect callback.
```

2. For coding style, following Shashank's recommendation that we formalise this:
```
- Import order
Try grouping imported modules by their kind. Observe the following order of imports: third-party dependencies imported from node_modules go first; then various helpers, services, redux selectors or slices; then components; then type imports; then images; and finally, styles. If you need to import a type from a module that has already been imported, then extend the same import statement; no need to create a new one.
```

3. Google Chrome web rels, as well as a bunch of other experts, are codifying modern frontend web development practices into skills in https://github.com/GoogleChrome/modern-web-guidance-src

PR #1350, opened 2026-08-26
Add pre-populated lists of genomes that can be selected for VEP

azangru:
## Description
Add a pre-populated list of genomes to VEP genome selector. The list consists of the genomes suggested by the tools api (this can be changed in the future to some call to the metadata api) and the common list of selected genomes. 

Also:
- Enable genome lozenges such that clicking them will add that genome to VEP form
- A genome lozenge whose associated genome is now selected in the form is marked as active

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/ENSWBSITES-3057

## Deployment URL(s)
https://vep-prepopulated-genomes.review.ensembl.org

### DIFF
--- a/src/content/app/tools/vep/components/vep-app-bar/VepAppBar.tsx
+++ b/src/content/app/tools/vep/components/vep-app-bar/VepAppBar.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


veidenberg (approved):
Nice! 👍

PR #1351, opened 2026-08-27
Refactor(vep): take option help from the API

jamie-m-a:
> **Do not merge until Ensembl/ensembl-web-tools-api#66 is merged and deployed.**
> On its own this is a regression. It deletes the local help map and the version resolution, so against a backend without #66 every option tooltip disappears — or, once #65 alone is deployed, shows a literal `{version}` and lists both assemblies' links.
> (#65, the first half, is already merged.)

Option help comes from `form_config`, authored in the annotation spec alongside the label and panel that decide where the option appears. The frontend renders what it is given.

### Two commits

**Take option help from the API.** The local `OPTION_HELP` map goes; `getOptionHelp` reads `option.help`.

**Read help straight off the option.** #66 resolves `{version}` and a link's `majorVersion` server-side, against the option's own label. That was the only thing `getOptionHelp` still did, so it collapsed to `option.help` and both call sites — the form panel, and the results annotation detail, which resolves help off the pinned form panels — now do that directly. `optionHelp.ts` and its tests go, and `majorVersion` leaves `OptionHelpLink` since the API no longer sends it.

Net: **+2 / −644**, with no change in what a reader sees.

### What is left on the frontend

`OptionHelpText` only, which renders a description and its links, and turns `*span*` into emphasis — a small markdown subset that keeps the text a plain string in the data layer rather than markup the backend has to escape.

### Testing

`vitest src/content/app/tools/vep` — 266 passed across 19 files; `tsc --noEmit` clean.

The version cases moved to the backend with the logic they test (`app/tests/test_option_help.py` in #66), where they cover what a single client could not: that no option in either assembly serves an unresolved placeholder, and that GRCh38 and GRCh37 draw different text and different links from one authored entry. The content assertions moved with them — that every allele-frequency option notes the source naming, and that gnomAD CNV reads "sample frequencies" where SV reads "allele frequencies", a deliberate distinction since gnomAD reports CNVs as the fraction of samples carrying the call rather than as an allele count.

### A deliberate absence

The client can no longer render a description containing `{version}`, so an unresolved template would reach a user as a literal `{version}` in a tooltip. There is no guard against that here on purpose: a guard would hide a backend bug rather than surface it. `test_no_option_serves_a_placeholder` in #66 covers every option in both assemblies instead. The two do need to stay deployed in step.

### One thing left behind

`OPTION_HELP` carried an entry for `clinvar` that was already unreachable: there is no `clinvar` form option — only `clinvar_short`, which declares no control, and `clinvar_sv` — and `helpFor` needs a form option to exist. It is carried in the spec but never attached. Whether ClinVar should offer help at all is a separate decision, so nothing here tries to make it.


src/content/app/tools/vep/views/vep-form/vep-form-options-section/vep-form-options-panel/optionHelp.test.ts
```diff
@@ -17,169 +17,116 @@
...
+      expect(links).toEqual([GNOMAD_SV_LINKS[0]]);
     });
 
-    it('gives gnomAD SV v2.1 only the v2 paper', () => {
-      const links = getOptionHelp(option('gnomad_sv', 'gnomAD SV v2.1'))?.links;
-      expect(links).toHaveLength(1);
-      expect(links?.[0].href).toBe(
-        'https://europepmc.org/article/MED/32461652'
-      );
+    it('gives v2.1 only the v2 paper', () => {
```
azangru: Where is this knowledge coming from? Why does the client choose some particular link instead of the backend sending only the appropriate link(s)?

veidenberg:
[Backend PR](https://github.com/Ensembl/ensembl-web-tools-api/pull/65) merged.

src/content/app/tools/vep/views/vep-form/vep-form-options-section/vep-form-options-panel/optionHelp.test.ts
jamie-m-a: Probably these tests were overkill in the first place. I can move them to backend as well or you can just remove them entirely @azangru - your choice.

veidenberg:
Backend [#66](https://github.com/Ensembl/ensembl-web-tools-api/pull/66) now merged.

PR #1353, opened 2026-08-28
DO NOT MERGE until Ensembl/ensembl-web-tools-api#68 lands — refactor(vep): take the results filter catalogue from the API

jamie-m-a:
> **Do not merge until Ensembl/ensembl-web-tools-api#68 is merged and deployed.** It deletes the catalogue the query builder is built from, so against a backend without #68 the filter panel offers no fields at all.

Which fields the query builder offers, their labels, editors, placeholders and option sets now arrive on the results response and are read from there. `resultsFilterFields.ts` goes from 354 lines to 155: what remains is per-row bookkeeping — which field a row may take, given what the others have — which is interface state rather than data.

`isHumanGRCh38` goes with it. Transcript groups come from the response, where they are decided from the columns the output has rather than from the species; the FIXME on that check said it belonged on the server. `species` is no longer a prop of the filter panel.

VEP no longer reads the genome browser's `variantGroups` constant for its consequence terms. That constant stays for its four other readers and the tools API carries its own copy for now — one source of truth for the grouping is worth its own change.

The tests keep the row-level cases and take a fixture catalogue rather than the constants they used to assert against. What those constants said — the score categories, each score's range hint, that every offered field is one the backend can compile — is asserted in #68, where the catalogue now lives.

`vitest` — 265 passed; `tsc --noEmit` clean. Also clears two comments describing the `pattern` validation removed in 626192e6.

### DIFF
--- a/src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
+++ b/src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
@@ -15,7 +15,6 @@
  */
 
 import { useState, memo, type InputEvent } from 'react';
-import classNames from 'classnames';
 
 import type { FilterField } from 'src/content/app/tools/vep/types/vepResultsFilters';
 
@@ -24,7 +23,7 @@ import styles from './VepResultsFilters.module.css';
 type Props = {
   values: string[];
   onChange: (values: string[]) => void;
-  config: Pick<FilterField, 'placeholder' | 'mono'>;
+  config: Pick<FilterField, 'placeholder'>;
 };
 
 // Split free text (comma / whitespace / newline separated) into unique tokens,
@@ -65,9 +64,7 @@ const TokenListInput = (props: Props) => {
     <div className={styles.tokenField}>
       <input
         type="text"
-        className={classNames(styles.tokenInput, {
-          [styles.tokenInputMono]: config.mono
-        })}
+        className={styles.tokenInput}
         value={text}
         onInput={onInput}
         placeholder={config.placeholder}
--- a/src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
+++ b/src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
@@ -17,14 +17,14 @@
 import { useState, memo, type InputEvent } from 'react';
 import classNames from 'classnames';
 
-import type { TextInputConfig } from './resultsFilterFields';
+import type { FilterField } from 'src/content/app/tools/vep/types/vepResultsFilters';
 
 import styles from './VepResultsFilters.module.css';
 
 type Props = {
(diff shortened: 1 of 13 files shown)


src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/resultsFilterFields.ts
```diff
@@ -14,246 +14,48 @@
...
 // track which rows have been applied). Allele frequency starts on the requested
 // defaults: match any, <= (le), threshold 0.05.
 let conditionCounter = 0;
 export const createCondition = (
-  field: ResultsFilterField
+  field: ResultsFilterField,
+  fields: FilterField[]
 ): ResultsFilterCondition => {
   const id = `condition-${++conditionCounter}`;
   if (field === 'allele_frequency') {
```
veidenberg: This looks hacky (special case)

src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/resultsFilterFields.ts
veidenberg: shouldn't these values (e.g what operator to use) be part of the filter spec coming from BE?

src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
```diff
@@ -17,14 +17,14 @@
...
-import type { TextInputConfig } from './resultsFilterFields';
+import type { FilterField } from 'src/content/app/tools/vep/types/vepResultsFilters';
 
 import styles from './VepResultsFilters.module.css';
 
 type Props = {
   values: string[];
   onChange: (values: string[]) => void;
-  config: TextInputConfig;
+  config: Pick<FilterField, 'placeholder' | 'mono'>;
```
veidenberg: Hmm, I don't like the idea of special styling (monospace) coming from backend, it should be generic (a filter textbox) and live in client CSS. Currently, a filter textbox for gene name uses regular font, but the same textbox for gene ID is monospace (because of BE spec override). Is there a good reason for this or can we just drop `mono` from the filter spec?

azangru: Let's drop `mono`.

src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/resultsFilterFields.ts
veidenberg: Moved to BE spec.

src/content/app/tools/vep/views/vep-submission-results/components/vep-results-filters/TokenListInput.tsx
veidenberg: Dropped.

veidenberg (approved):

PR #1362, opened 2026-09-07
Add healthcheck endpoint and opentelemetry endpoint for collecting metrics

azangru:
## Description
- Add healthcheck endpoint to use as readiness and liveness probe
- Add opentelemetry metrics endpoint for prometheus

## Deployment URL(s)
http://try-prometheus.review.ensembl.org


shashankbrgowda (approved):
Nice. I didn't know we could do this for client

PR #1365, opened 2026-09-08
Remove track trigger dependencies

veidenberg:
## Description
This PR removes the dependency on track "trigger" values, instead using track IDs for identifying all tracks.
Depends on the backend update in [this PR](https://github.com/Ensembl/ensembl-dauphin-style-compiler/pull/154).

## Related JIRA Issue(s)
[ENSWBSITES-2404](https://embl.atlassian.net/browse/ENSWBSITES-2404)

## Deployment URL(s)
http://migrate-legacy-tracks.review.ensembl.org

## Views affected
All genome browser tracks

### DIFF
--- a/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
@@ -207,20 +207,7 @@ const useGenomeBrowser = () => {
   };
 };
 
-/**
- * To toggle a genome browser track or its settings, the client has to generate
- * a "track path" — an array of strings that genome browser will use to identify the track.
- *
- * There are currently two types of genome browser tracks:
- *   - Older tracks with human-readable ids (e.g. protein-coding genes on the forward strand)
- *   - Newer tracks identified by uuids. They are registered by the genome browser
- *     using a mechanism called 'expansion'.
- *
- * Ideally, the client shouldn't know any of this. Ideally, tracks would be identified
- * Track API tracks are all registered through the common expansion node. Focus
- * controls are boot-only and retain their explicit path.
- */
-
+// Focus tracks have a hard-coded identifier path in genome browser.
 const getTrackPath = (trackId: string) => {
   if (trackId === 'focus') {
     return ['track', trackId];
--- a/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
+++ b/src/content/app/genome-browser/hooks/useGenomeBrowser.ts
@@ -217,23 +217,15 @@ const useGenomeBrowser = () => {
  *     using a mechanism called 'expansion'.
  *
  * Ideally, the client shouldn't know any of this. Ideally, tracks would be identified
- * just by their ids. But, while this is not the case, the client will use a hack
- * as an easy option of generating a track path.
- *
- * The hack is: if track id is uuid-shaped, then it is an "expansion" track
- * that uses one path pattern; and in other cases, it is a "non-expansion" track,
- * which uses a different path pattern.
+ * Track API tracks are all registered through the common expansion node. Focus
+ * controls are boot-only and retain their explicit path.
  */
 
-const uuidRegex =
-  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-
 const getTrackPath = (trackId: string) => {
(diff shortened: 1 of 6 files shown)


src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -217,23 +217,15 @@ const useGenomeBrowser = () => {
  *     using a mechanism called 'expansion'.
  *
  * Ideally, the client shouldn't know any of this. Ideally, tracks would be identified
```
azangru: Should this comment line also go now? Or be somehow modified?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -217,23 +217,15 @@ const useGenomeBrowser = () => {
...
- * just by their ids. But, while this is not the case, the client will use a hack
- * as an easy option of generating a track path.
- *
- * The hack is: if track id is uuid-shaped, then it is an "expansion" track
- * that uses one path pattern; and in other cases, it is a "non-expansion" track,
- * which uses a different path pattern.
+ * Track API tracks are all registered through the common expansion node. Focus
+ * controls are boot-only and retain their explicit path.
```
azangru: Is this AI-speak? "Focus controls are boot-only"? Could you translate this into human?

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
```diff
@@ -217,23 +217,15 @@ const useGenomeBrowser = () => {
...
  */
 
-const uuidRegex =
-  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-
 const getTrackPath = (trackId: string) => {
-  if (uuidRegex.test(trackId)) {
-    return ['track', 'expand', trackId];
-  } else {
+  if (trackId === 'focus') {
```
azangru, line 212: I forget: what is passed to this function as track id for focus variants and transcripts? Or are focus tracks handled in a separate function?

src/content/app/genome-browser/state/api/genomeBrowserApiSlice.ts
```diff
@@ -134,25 +134,9 @@ const genomeBrowserRestApiSlice = restApiSlice.injectEndpoints({
...
-         */
-        return response.track_categories.map((category) => ({
-          ...category,
-          track_list: category.track_list.map((track) => ({
-            ...track,
-            track_id: (track.track_id = track.trigger.at(-1) as string)
-          }))
-        }));
+        // Track API UUIDs are the browser and persistence identity for every
+        // visual track. The backend translates the UUID into an expansion.
```
azangru: Sounds LLMish? I don't think this we need this comment here at all.

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
veidenberg: The focus track has a track ID `focus`, switch path `['track', 'focus']`, and is already enabled at boot time.
The focus track type and focus ID is set with `['track', 'focus', 'item', ...]` switch (see `setFocusObject()` above).

src/content/app/genome-browser/state/api/genomeBrowserApiSlice.ts
veidenberg: Comment removed. Not sure if we need this transformation function since triggers are not used anymore.

src/content/app/genome-browser/hooks/useGenomeBrowser.ts
veidenberg: Left a short comment to remind us that focus track is a special case, but feel free to remove if not needed.

src/content/app/genome-browser/state/api/genomeBrowserApiSlice.ts
azangru: It seems that what comes out of that query function (because of the transformation) is an array of track categories rather than an object with a `{ track_categories }` field in it. So we can't just remove the transformer without updating code on the consuming side. Which is annoying; and the transformer becomes silly; but probably not worth changing in this PR.

azangru (approved):

PR #1366, opened 2026-09-10
Add a tab for transcripts table to FeatureExplorer gene view

azangru:
## Description
Add a transcripts table to Feature explorer's gene view

- A new tab opens up the view with the table (it was decided that the transcripts diagram is still the default view)
- The table displays the data in a similar way to the old Ensembl site
- Transcript ids, when clicked, show a "view in app" popup to let the transcript be viewed in either the genome browser or feature explorer
- The table can be downloaded as a .csv file
- The transcripts in the table reuse the same filtering and sorting rules as in the default transcripts view (the one with diagrams)
- The table shows all transcripts of a gene, rather than just the first 100. 

## Related JIRA Issue(s)
https://embl.atlassian.net/browse/ENSWBSITES-3065

## Deployment URL(s)
http://transcripts-table.review.ensembl.org

### DIFF
--- a/src/content/app/entity-viewer/gene-view/GeneView.tsx
+++ b/src/content/app/entity-viewer/gene-view/GeneView.tsx
@@ -50,6 +50,7 @@ import GeneOverviewImage from './components/gene-overview-image/GeneOverviewImag
 import DefaultTranscriptsList from './components/default-transcripts-list/DefaultTranscriptsList';
 import GeneViewTabs from './components/gene-view-tabs/GeneViewTabs';
 import TranscriptsFilter from 'src/content/app/entity-viewer/gene-view/components/transcripts-filter/TranscriptsFilter';
+import GeneTranscriptsTable from './components/gene-transcripts-table/GeneTranscriptsTable';
 import GeneFunction from 'src/content/app/entity-viewer/gene-view/components/gene-function/GeneFunction';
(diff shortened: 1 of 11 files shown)


shashankbrgowda:
Should it cover entire table body width similar to gene relationships? 
image

Also, I feel a toggle to hide visualization could be useful or fit screen icon in the table section header which toggles RHP and visualization ?

azangru:
@shashankbrgowda 

> Should it cover entire table body width similar to gene relationships?

I believe this table stops growing horizontally when it fits all its content inside. I don't know if there is any benefit in its growing to fill all the available space.

> Also, I feel a toggle to hide visualization could be useful or fit screen icon in the table section header which toggles RHP and visualization ?

Yes; we discussed this in the UX review meeting yesterday. The decision was that yes, we should only keep the top (gene) diagram in the view where we are showing all the transcript diagrams; but that it is more important to release the table now than to revamp feature explorer. But yes, this is something that we should do (but later on).

shashankbrgowda (approved):

azangru:
>> Should it cover entire table body width similar to gene relationships?

> I believe this table stops growing horizontally when it fits all its content inside. I don't know if there is any benefit in its growing to fill all the available space.

I have updated the styles to stretch the table 😃

PR #1370, opened 2026-09-14
Improve right hand panel feature search

shashankbrgowda:
## Description
This pr updates feature search in species selector , genome browser and feature explorer right hand panel. Instead of panels with respective search fields, there will be single search input field which accepts search query input for gene, transcript and variant search. It ranks results in descending order of result count and pagination is provided only for gene results.

## Related JIRA Issue(s)
[ENSWBSITES-3067](https://embl.atlassian.net/browse/ENSWBSITES-3067)

## Deployment URL(s)
http://rhp-search.review.ensembl.org


azangru:
@shashankbrgowda does variant search return different responses for genomes that don't have a variant index at all vs genomes that have a variant index? If yes, could you hide the variants section from search results for genomes that don't have a variant index?

azangru:
When the OS shows scrollbars,  you can see that the button is uncomfortably close to the scrollbar.



I do not have a clear solution. We can leave it as is. We can add a bit of padding on the right, but that would make the input field narrower. We can make the input field stretch across the full available space, and move the submit button below it. My personal preference is this last option.

src/shared/components/sidebar-search/SidebarSearch.tsx
```diff
@@ -14,36 +14,25 @@
...
 import { formatNumber } from 'src/shared/helpers/formatters/numberFormatter';
-import {
-  getFeatureSearchLabelsByMode,
-  getFeatureSearchModes,
-  type FeatureSearchMode,
-  type FeatureSearchAppName
+import type {
+  FeatureSearchMode,
+  FeatureSearchAppName
 } from 'src/shared/helpers/featureSearchHelpers';
```
azangru: could you move the type import to type imports?

src/shared/components/sidebar-search/SidebarSearch.tsx
```diff
@@ -66,309 +55,213 @@ type Props = {
...
-    const newQuery = event.currentTarget.value;
-    setSearchInputs((currentSearchInputs) => ({
-      ...currentSearchInputs,
-      [searchMode]: newQuery
-    }));
-    if (!newQuery.trim()) {
-      submitSearch(searchMode, '', 1, 20);
+  const searchResults = [
+    {
+      searchMode: 'gene' as const,
```
azangru: is the 'as const' needed in these three objects? A string is immutable, so is a const anyway. If typescript complains that a string here is too broad, perhaps it needs `as FeatureSearchMode`? Or the whole objects in the array can be marked as const.

shashankbrgowda: searchMode needs to be of type FeatureSearchMode. as const is used to avoid type error

azangru: But wouldn't `as FeatureSearchMode` make more sense?

azangru (approved):

